Issue #9, opened 2015-02-28
NREL SPA license problems


wholmgren:
It appears to me that we may be in violation of NREL's SPA [license](https://github.com/pvlib/pvlib-python/blob/master/pvlib/spa_c_files/SPA_NOTICE.md)

"The Software is being provided for internal, noncommercial purposes only and shall not be re-distributed."

A couple of potential problems:
* We are redistributing their code rather than just making binaries with it.
* We (or at least I) don't want to limit pvlib-python to non-commerical purposes. Everything else in pvlib-python is, to my knowledge, able to be used in commercial applications.

The license goes on to say that you must contact them about using it in commercial applications.

Am I reading this wrong? @Calama-Consulting did you have a different interpretation of this when you added the SPA code to pvlib?

Calama-Consulting:
My inclusion of it was based on it's original inclusion in the MATLAB version of the code, which carries the following liscence:

>Copyright (c) 2013, Sandia Labs
>All rights reserved.

>Redistribution and use in source and binary forms, with or without modification,
are permitted provided that the following conditions are met:

>  Redistributions of source code must retain the above copyright notice, this
  list of conditions and the following disclaimer.

>  Redistributions in binary form must reproduce the above copyright notice, this
  list of conditions and the following disclaimer in the documentation and/or
  other materials provided with the distribution.

>  Neither the name of the {organization} nor the names of its
  contributors may be used to endorse or promote products derived from
  this software without specific prior written permission.

>THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
[...]

PR #14, opened 2015-03-03
Add commercial licence warning for NREL SPA algorithm

Calama-Consulting:
Add commercial limitation warning for NREL SPA code, until licensing is determined from NREL


wholmgren:
Wondering if it would be better to use [``warnings.warn``](https://docs.python.org/2/library/warnings.html#warnings.warn). This way you only get the message once instead of every time that you use the function.

Calama-Consulting:
I think the word ing on this needs to be changed anyways to be consistent with #9, so i'll ditch this one. `warnings.warn` would be better.

Issue #11, opened 2015-03-01
Versioning and labels


bmu:
I think we should use github milestones. Also it is not clear to me on which version we are at the moment (`pvlib.version` says 0.3, but we are on something like `0.4dev` I think).

I would also like to use github labels for the issues.

wholmgren:
Agreed on both. I was wishing for milestone labels yesterday. 

@Calama-Consulting did you have a plan for what version to call the first PyPI release? 1.0 would make some sense for a first release, but I also feel like there's still a lot of loose ends for that label. So, I have a slight preference for 0.4.

Calama-Consulting:
I agree with @wholmgren , the initial release will probably not warrant the 1.0 label, but I'd like to start at a fresh versioning with the new package. Let's reset  and release initially under 0.1? If there's agreement I'll set up the tags appropriately

wholmgren:
Sure, I'm ok with calling it 0.1. One slight hiccup is that anybody that ``pip`` installed the package will need to specify ``-force`` instead of ``-upgrade``. That's probably only a concern for 3 people though.

bmu:
I created milestones and some labels. Maybe you should have a look.
If they seem reasonable, we can close this issue.

wholmgren:
Nice, they look good.

Let's leave this issue open until we fix the version (it's currently 0.3, which is a hold-over from the UARENForecasting project)

wholmgren:
I changed the version to ``0.1.0`` in a28e45d2. I suggest that we keep it this way until after the release, at which point we change it to ``0.1.1dev``. I think that there are some scripts for fancy versioning, but I don't think we need them now. Not sure if you want to close this issue or not.

dacoex:
What about adding a label such as "Good as first PR" / "easy catch"

https://github.com/pydata/pandas/labels/Good%20as%20first%20PR

For contibutors with litlle time or who are just starting to understand this library or do not use it everyday.

PR #18, opened 2015-03-08
changed unix file modes, so that the files are not executable

bmu:
For some of the original files (I created a new clone to double check this), the unix file modes were set to 755, which means that they are executable (see #11).
I think, it should be save to merge this (however I am not 100% sure).
Maybe the best is to wait for travis and to test this on another linux machine. 
Any others, that work on linux?


bmu:
Ok, it fails, but the reason is a failure in `test_tools`. I'll create another pull request to fix this.

bmu:
Something is going wrong here: `test_tools.py` seems to be outdated and is not tested on the other pull requests, but is tested on the one, that I submitted! 

Any ideas?

wholmgren:
Good catches. I apparently missed updating ``test_tools.py`` on UARENForecasting/PVLIB_Python and I never noticed because nose didn't run the tests automatically due to the permissions issue. 

Looks like everything in ``test_tools.py`` is for the old vars/expect pattern, so it will just be a stub when you remove all of that. This is fine with me for ``0.1``, but we should probably add some tests for ``0.2``.

bmu:
Ok, so I think I should remove the file. Ok?

wholmgren:
Yes, we can add it back later if tools expands.

bmu:
this should be ready to merge from my side.

wholmgren:
Great, merged.

PR #19, opened 2015-03-08
spa license fixes

wholmgren:
As discussed in #9, this PR:

* Removes NRELS's ``spa.c``, ``spa.h``, and ``spatester.c`` files.
* Updates ``setup.py`` so that it works with or without the spa files. I tested this, but see below.
* Updates ``test_solarposition.py`` to skip the spa tests if the files are missing. @bmu probably won't like the copy/paste test code, but I think that it can be fixed in ``0.2``.

(Now resolved) Major issue: NREL made a few minor changes to the SPA code that will require modifying the cython code. From the most recent spa.c file:

> // Revised 10-JUL-2014 Andreas
> //         Change second in spa_data structure from an integer to double to allow fractional second
> // Revised 08-SEP-2014 Andreas
> //         Corrected description of azm_rotation in header file
> //         Limited azimuth180 to range of 0 to 360 deg (instead of -180 to 180) for tech report consistency
> //         Changed all variables names from azimuth180 to azimuth_astro
> //         Renamed 2 "utility" function names for consistency

~~This PR currently works with the 01-APR-2013 version of the spa code. It should probably be updated to work with the latest version. I'm guessing that it would be a hassle to make it work with both versions.~~

This PR currently works with the 08-SEP-2014 version of the spa code. It does *not* work with the older code.


wholmgren:
I think this is ready, but it should definitely be tested in full by somebody else.

.gitignore
```diff
@@ -30,6 +30,11 @@ build/
 # spa_c_files builds
 pvlib/spa_c_files/build*
 
+# spa_c_files source
+pvlib/spa_c_files/spa.c
+pvlib/spa_c_files/spa.h
+pvlib/spa_c_files/spa_tester.c
+
```
bmu: I think `pvlib/spa_c_files/spa_py.so` should be removed from the repo and also be added to the ignore file (also because the SPA tests are passing before SPA is installed!)

README.md
```diff
@@ -32,6 +32,14 @@ pip install .
 
 Add ``-e`` to install in [develop mode](http://pip.readthedocs.org/en/latest/reference/pip_install.html#editable-installs).
 
+``pvlib-python`` comes with Python bindings for NREL's solar position C code,
+however you must manually download the C code directly from [NREL](http://www.nrel.gov/midc/spa/).
+To use the NREL SPA algorithm,
+
+1. Download the SPA files from [NREL](http://www.nrel.gov/midc/spa/)
+2. Copy the SPA files into ``pvlib/spa_c_files``
+3. From this ``pvlib-python`` directory, run ``pip install .``
```
bmu: For step 3 I would prefer `python setup.py build_ext --inplace` as this is the way described in `pvlib/spa_c_files/README.md` (however I am not sure what happens if you install pvlib afterwards with pip. Is this pip install here needed for pvlib to work with spa, when it is installed?)

bmu:
I checked out your branch locally and tested it. Everything seems to be ok (the only difference to your install instructions was the use of `python setup.py install --inplace` instead of `pip install .` (see my comment above). All tests were ok.

However I'm not sure about a Windows install, which could be a problem in gereneral because of the compilation of the c code (but this is not a problem introduced by this pull request).

.gitignore
wholmgren: Agreed, I'll fix this.

README.md
wholmgren: On further testing, I don't actually think that my actually instructions work. 

I had to ``pip uninstall pvlib; pip install .`` to get the ``.so`` file into my ``site-packages`` directory. ``pip install . -I --no-deps`` also works. I think that the tests only passed because they were being run in the root package directory.  

I couldn't figure out the syntax to get ``python setup.py`` to do copy that file over after the package has already been installed.

bmu: Maybe it is a good idea to move the spa to a separate package, if this is easier. I think you already discussed this with @Calama-Consulting. 

This could also have the advantage, that users, that are not interessted in PV, but want to calculate solar position can install only this package.

wholmgren: Well @Calama-Consulting, after seeing this implementation do you want to stick to your proposed solution in #9? 

For the 0.1 release I'm equally ok with accepting this PR (with minor revisions) or removing the spa wrapper entirely. If we remove the spa wrappers for 0.1, then I would support somebody (not me) creating a separate spa wrapper package and reimplementing its usage in pvlib in a future release. 

Maybe we should take the conversation back over to that issue and tag some of the other users and contributors to ask for their input?

.gitignore
wholmgren: Hm, on second thought I already removed the file and line 10 of ``.gitignore`` is ``*.so``. You may be seeing the file because of some git oddity when switching between branches.

bmu: You are right, I see. Sorry that I was causing this confusion. The problem that the `so` will stay in the file structure exists, because it is ignored and will stay at its place, if you switch branches.

Calama-Consulting:
I did testing on @wholmgren's branch. I think the confusion is coming from running the install from the local pvlib-python folder vs site-packages. When doing a pip install, the spa_c_files directory isn't created in site-packages when installing, so it isn't possible to go back and do a `python setup.py install --inplace` and have it apply globally. That will only work on the pvlib-python folder (the one that isn't in the site-packages directory). 

So, I think the only way to propagate this into the site-packages directory properly is using pip, and so I'd reccomend updating the README.md with this:

wholmgren:
I updated the readme with @Calama-Consulting's suggestion and rebased on the numerous master changes.

Issue #17, opened 2015-03-06
meta, location, and systemdef philosophy


jforbess:
I don't fully understand the differences of meta, location, and systemdef, in terms of how they should be used. 

systemdef expects TMY meta to be used in its first parameter. But what if I want to use the lat/long for my plant, not the TMY station? What if I am looking at operating data (which I am), and have no need to pull in a TMY file?

Why doesn't systemdef use a location instead of a TMY meta as its first parameter? (Why isn't a TMY meta a Location object?)

wholmgren:
These are all good points/questions. This is definitely a place where we need to improve the package. 

I propose that we:
* Make TMY importers return ``Location`` objects instead of meta dicts.
* Create a ``System`` class that inherits from ``Location`` and replaces ``systemdef``. ``System`` objects could be instantiated directly or accept ``Location`` objects plus the current ``systemdef`` arguments.
* Make use of ``System`` objects where appropriate throughout the library.
* These improvements should be a ``0.2`` milestone.

For ``0.1`` (#15), I think that we should just change the ``tmy_meta`` argument name to be ``meta`` and clarify the docs (#16) to specify that it should just be a dict that has these particular keys. There's nothing special about a dict that comes from the TMY importers vs. a dict that is user generated.

jforbess:
Great, that is what I was thinking, but I wasn't sure there was something I was missing. Both the 0.1 and 0.2 solutions make sense to me.

wholmgren:
@jforbess would you like to make a pull request with the 0.1 fixes? 

Should probably add a ``test_systemdef_dict`` function to [``test_pvsystem``](https://github.com/pvlib/pvlib-python/blob/master/pvlib/test/test_pvsystem.py) to prove that it works as expected.

jforbess:
@wholmgren sure, I'll do that, but not until Sunday or Monday.

wholmgren:
@bmu and @Calama-Consulting do you support the proposal above or do you have other ideas for how to resolve this? I think is a high priority for 0.2 since it could impact the API in fairly significant ways.

@jforbess, sorry, just now realizing that I did not respond to your question regarding contributing. Does the [Development information wiki](https://github.com/pvlib/pvlib-python/wiki/Development-information) help?

bmu:
[...]

Issue #15, opened 2015-03-03
Initial release to PyPi


Calama-Consulting:
At what point should we push the initial release to PyPi? I'm leaning towards pushing after #6 and #13 are resolved with the 0.1 tag. 

Thoughts?

wholmgren:
Sounds good for the code. 

What should we do about the tutorials? My suggestion is to just promise that one good tutorial works at this point. I can fix up, or even remove, the module level ones later. Another idea is to make a new repo under the pvlib org with the tutorials (at least the detailed ones). Not sure if that would be confusing to people that are new to git/github/python. 

We should also double check that there are no big issues that we want to fix in the docs.

Calama-Consulting:
The tutorials on the dev branch of the sandia site are pretty good, and fixed for the new package layout, I'll port those over

wholmgren:
@Calama-Consulting do you have an approximate date in mind for the ``0.1`` release? I'll knock out a few of these easy issues this afternoon.

Calama-Consulting:
Let's aim for the end of this week, to give me time to make sure the tutorial are solid. We will also need to resolve #9, @wholmgren can you handle this as well? I beleive we had decided on the option with a modified `setup.py` which allows a user to manually install spa if they download the package from github and add their spa C code.

wholmgren:
Sounds good!

wholmgren:
Add #24 to 0.1 milestones.

wholmgren:
Looks pretty easy to [make a conda package](http://conda.pydata.org/docs/build_tutorials/pkgs.html). Should we do that for 0.1 or wait for a future release? I don't use Anacondas but it seems to be all the rage these days, especially for people that are new to Python e.g. former PVLIB_MATLAB users.

bmu:
+1 for a conda package. I use anaconda on Windows and linux. However I think its fine to wait for the next release. Maybe I could create the package than.

Calama-Consulting:
Sounds good, it doesn't look too involved,  I will add a conda wrapper and upload to binstar in addition to pypi for 0.1 . I think it will increase the ability to uptake, especially for new windows users.

dacoex:
Please ping me when you release the package to pypi.

I will adapt the Ubuntu package accordingly and propose a change in the docs for pointing at this repository:
https://code.launchpad.net/~pythonxy/+archive/ubuntu/pythonxy-devel/+packages?field.name_filter=pvlib

bmu:
[...]

PR #7, opened 2015-02-28
TMY loading improvements

wholmgren:
Adds ability to open TMY3 files via a url.

Closes #5 

Cleans up some other aspects of the module including unneeded imports, slightly improved documentation.

~~**Not yet ready for merging** as it still needs new tests and ``coerce_year`` is not yet implemented.~~

PR #23, opened 2015-03-10
Issue17

jforbess:
Only cosmetic changes. Note that Location refers to tz as a string, while the TMY meta creates TZ as a float. To be resolved in 0.2.

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -17,7 +17,7 @@ from pvlib import tools
 
 
 
-def systemdef(tmy_meta, surftilt, surfaz, albedo, series_modules, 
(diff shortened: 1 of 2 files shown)


wholmgren:
Thanks. The only thing is that we can easily merge the ``pvsystem.py`` commit 995f6a9 but I think that the ``tmy.py`` commit b31e6bf is going to conflict with the many ``tmy`` changes in #7. So, can you update your PR to just have 995f6a9?

Addresses #17 and #15.

jforbess:
Ok, I removed the last commit, so no changes to tmy.py. I think it's good to go.

wholmgren:
Great, thank you!

PR #31, opened 2015-03-13
Move scipy import in clearsky.py

wholmgren:
As discussed in #13.

I briefly looked into adding a real test for this, but it wasn't immediately obvious how to do it. Instead, I ran the test suite on a virtualenv without scipy. Here's the output, as expected:

```
$ nosetests pvlib/test/test_clearsky.py -v
pvlib.test.test_clearsky.test_ineichen_required ... ERROR
pvlib.test.test_clearsky.test_ineichen_supply_linke ... ok
pvlib.test.test_clearsky.test_ineichen_solpos ... ok
pvlib.test.test_clearsky.test_ineichen_airmass ... ok
pvlib.test.test_clearsky.test_ineichen_keys ... ok
pvlib.test.test_clearsky.test_haurwitz ... ok
pvlib.test.test_clearsky.test_haurwitz_keys ... ok
pvlib.test.test_clearsky.test_disc_keys ... ok
pvlib.test.test_clearsky.test_disc_value ... ok

======================================================================
ERROR: pvlib.test.test_clearsky.test_ineichen_required
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/holmgren/.pyenv/versions/pvlib-3.4.3/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/holmgren/git_repos/pvlib-python/pvlib/test/test_clearsky.py", line 34, in test_ineichen_required
    clearsky.ineichen(times, tus)
  File "/Users/holmgren/git_repos/pvlib-python/pvlib/clearsky.py", line 134, in ineichen
    'supply your own turbidities.')
nose.proxy.ImportError: The Linke turbidity lookup table requires scipy. You can still use clearsky.ineichen if you supply your own turbidities.
-------------------- >> begin captured logging << --------------------
pvlib: DEBUG: creating Location object
pvlib: DEBUG: using PyEphem to calculate solar position
pvlib: DEBUG: tz_localize to US/Arizona and then tz_convert to UTC
pvlib: DEBUG: irradiance.extraradiation()
pvlib: DEBUG: Calculating ET rad using Spencer method
pvlib: DEBUG: using PyEphem to calculate solar position
pvlib: DEBUG: tz_localize to US/Arizona and then tz_convert to UTC
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
Ran 9 tests in 1.986s

FAILED (errors=1)
```

PR #34, opened 2015-03-19
make linke turbidity fixed in tests

wholmgren:
See discussion in #31.


bmu:
Ok, we have three tests with errors now, but the error message states that the error is raised, because scipy is missing. So it seems ok for now.

I opended a new issue on this topic: #36

PR #35, opened 2015-03-19
add single axis tracking

wholmgren:
This PR contains a Python implementation of PVLIB MATLAB's [``pvl_singleaxis.m``](https://github.com/Sandia-Labs/PVLIB_Matlab/blob/master/PVLIB/pvl_singleaxis.m). This isn't yet ready to merge (needs tests and PEP8), but I wanted to share progress and provide a place for feedback.

The code and comments are very similar to ``pvl_singleaxis.m``. I think that some things can be simplified, but I wanted to get something working first. Sandia put a lot of comments in the code and I'm inclined to keep most of them.

I believe that the AOI calculation is correct, but there's something wrong with the surface azimuth and tilt calculation. 

Here's an [IPython notebook](http://nbviewer.ipython.org/github/wholmgren/pvlib-python/blob/tracking/docs/tutorials/tracking.ipynb) that explores most of the parameters.


wholmgren:
A lot of changes since the initial PR.

* Fixed the problem with ``surface_tilt`` and ``surface_azimuth``. I'm not sure whether the MATLAB authors intended for ``surface_tilt`` to be with respect to surface horizontal plane or surface normal.
* Parameters and most of the inline code has been PEP8-ified. 
* The MATLAB authors required a latitude input and threw an error if the latitude was less than 0, but I think the new code works fine independent of the latitude.
* The sample notebook walks through each step of the algorithm and then tests the whole thing in many regimes.
* I still need to make some automated tests.

It would be helpful for somebody to make an initial review at this point.

alorenzo175:
What if the given elevation and azimuth don't have the same index? Could run into problems when doing things like finding wid.

wholmgren:
@alorenzo175 pointed out that the ``surface_tilt`` calculation was done wrong, and that ``surface_azimuth`` did not work for non-integer multiples of 90 degrees. I think I've fixed both problems in the latest commit, and added some explanatory comments in the code and notebook. The second half of the ``surface_azimuth`` calculation is now significantly different than that of the MATLAB code. 

I expect that the tests will fail since I didn't update the expected values. I'll make the other changes too.

wholmgren:
Updated the tests and added a few lines for @alorenzo175. Added introduction material to notebook.

wholmgren:
Added the same examples as the PVLIB MATLAB 1.2 help file to the tutorial. Updated internal variable names to be consistent with PEP8 and, I think, make the algorithm more understandable. Also replaced lots of ``arctan`` mapping code with ``arctan2``.

I think this is close to being ready to merge.

wholmgren:
Rebased so that the test run against all of the pandas versions. 

Removed the latitude kwarg since I don't think it's necessary.

I am happy to make changes, but if there are no objections to merging then let's go ahead and do it.

PR #41, opened 2015-03-25
Fixed inconsistent azimuth angle in the ephemeris function

### LINKED ISSUE #40: solarposition.ephemeris
returns solar azimuth angle in wrong definition with zero = south. But irradiance.aoi needs as input panel azimuth angles from north.

This function is discouraged anyway, so may it be better to remove it from the library?

uvchik:
Fixed #40.

### DIFF
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -35,9 +35,9 @@ def get_solarposition(time, location, method='pyephem', pressure=101325,
         'spa' uses the spa code: :func:`spa`
 
         'ephemeris' uses the pvlib ephemeris code: :func:`ephemeris`
-    pressure : float [default: 101325]
+    pressure : float
         Pascals.
-    temperature : float [default: 12]
+    temperature : float
         Degrees C.
     """
 
@@ -223,10 +223,10 @@ def ephemeris(time, location, pressure=101325, temperature=12):
     Other Parameters
     ----------------
 
-    pressure : float or DataFrame [default: 101325]
+    pressure : float or DataFrame
           Ambient pressure (Pascals)
 
-    temperature : float or DataFrame [default: 12]
+    temperature : float or DataFrame
           Ambient temperature (C)
 
     Returns
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -3,8 +3,8 @@ Calculate the solar position using a variety of methods/packages.
 """
 
 # Contributors:
-# Rob Andrews (@Calama-Consulting), Calama Consulting, 2014 
-# Will Holmgren (@wholmgren), University of Arizona, 2014 
+# Rob Andrews (@Calama-Consulting), Calama Consulting, 2014
+# Will Holmgren (@wholmgren), University of Arizona, 2014
 
 from __future__ import division
 import logging
@@ -15,12 +15,12 @@ import datetime as dt
 import numpy as np
 import pandas as pd
 import pytz
-    
+
 
 from pvlib.tools import localize_to_utc, datetime_to_djd, djd_to_datetime
 
 
-def get_solarposition(time, location, method='pyephem', pressure=101325, 
+def get_solarposition(time, location, method='pyephem', pressure=101325,
                       temperature=12):
     """
     A convenience wrapper for the solar position calculators.
@@ -40,11 +40,11 @@ def get_solarposition(time, location, method='pyephem', pressure=101325,
     temperature : float [default: 12]
         Degrees C.
     """
-    
+
     method = method.lower()
     if isinstance(time, dt.datetime):
-        time = pd.DatetimeIndex([time,])
-    
+        time = pd.DatetimeIndex([time, ])
+
     if method == 'spa':
         ephem_df = spa(time, location)
     elif method == 'pyephem':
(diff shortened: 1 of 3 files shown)


uvchik:
The results of ephemeris are now similar to the results of pyephem. Both results are feasible for the location "Berlin/Europe" but I didn't checked the details.

wholmgren:
Thanks. Did you change anything about the algorithm itself (other than south=180) or is everything just PEP8 changes? It's hard to see on the diff.

Are you interested in getting into the details, improving the function, and thoroughly testing it, or would you prefer to just go ahead and call it good? While I'm all for the careful approach on the principle of making a good and dependable library, I'm not sure if it's worth the time. Also, @alorenzo175 was working on implementing NREL's SPA algorithm with numba wrappers.

pvlib/solarposition.py
```diff
@@ -15,50 +15,55 @@
...
-    pressure : float
+        'pyephem' uses the PyEphem package (default): :func:`pyephem`
+
+        'spa' uses the spa code: :func:`spa`
+
+        'ephemeris' uses the pvlib ephemeris code: :func:`ephemeris`
+    pressure : float [default: 101325]
         Pascals.
-    temperature : float 
+    temperature : float [default: 12]
```
wholmgren: Most of the pvlib docstrings do not have defaults in the parameter descriptions. This is mostly because I personally don't like to put them there, and I removed them when cleaning up the docstrings. I think I might be in the minority, though, so maybe consider starting an issue for this. I'm ok with keeping this here, as long as we know about the inconsistencies.

pvlib/solarposition.py
```diff
@@ -292,86 +294,98 @@ def ephemeris(time, location, pressure=101325, temperature=12):
...
+                                    np.cos(EcLonR)))
+
+    HrAngle = LocAST - RtAscen
+    HrAngleR = np.radians(HrAngle)
+
+    HrAngle = HrAngle - (360 * ((abs(HrAngle) > 180)))
+    SunAz = np.degrees(np.arctan2(- 1 * np.sin(HrAngleR), np.cos(LatR) *
+                       (np.tan(DecR)) - np.sin(LatR)*(np.cos(HrAngleR))))
+    SunAz = SunAz + (SunAz < 0) * 360
+    # potential error in the following line
```
wholmgren: What's the error?

wholmgren:
Can you go ahead and run the modified code on the standard NREL test? See ``test_spa_physical`` and ``test_pyephem_physical``. Then we'll at least have something concrete to say about the accuracy.

uvchik:
I made different commits to make it easier to understand.

1. uvchik/pvlib-python@9b95720: This is the bug fix. I just reverted the changes from RA.

2. uvchik/pvlib-python@bbb8b4e: This is the test function. That means it works now.

3. I removed the default values from the docstring because I like consistence more than I like default values.

4. uvchik/pvlib-python@e991712: This commit contains all the pep8 changes, except the ones that are right next to one of the other changes (e.g. within the docstring). @wholmgren The changes above are just pep8. Trailing spaces were remove by my IDE automatically.

uvchik:
Physical feasibility test.

```python
DaFrOut_a = pvlib.solarposition.get_solarposition(time=data.index,
            location=location, method='pyephem')
            
DaFrOut_b = pvlib.solarposition.get_solarposition(time=data.index,
            location=location, method='ephemeris')
        
# zenith over azimuth
plt.plot(DaFrOut_a['azimuth'], 90 - DaFrOut_a['zenith'], '.')
plt.plot(DaFrOut_b['azimuth'], 90 - DaFrOut_b['zenith'], '.')   
plt.show()

# zenith over index(time)
plt.plot(90 - DaFrOut_a['zenith'], '-')
plt.plot(90 - DaFrOut_b['zenith'], '-')   
plt.show()
```

In the first plot both models look pretty much the same but in the second plot you can see an offset of one hour between the methods (two hours during the summer period).

Half an hour of this is corrected in line 312 (there is already o comment of @wholmgren )
```python
# ToDo: Will H: surprised to see the 0.5 here, but moving on...
UnivHr = DecHours + utc_offset  #- .5
```

If there would be a minus 1 than everything would be the same between the two models during the winter time. But I cannot explain it so I didn't changed it.

1. We have an offset of one hour.
2. We have different behaviour considering the summer/winter time.

Any ideas?

I'm absent until Monday.

uvchik:
Changed the offset from line 312 to "-1":

The plots from the code above:

Plot of  zenith over azimuth (blue = 'pyephem', green='ephemeris')
solarposition
Plot of  zenith over index(time) (blue = 'pyephem', green='ephemeris') [winter-summer-time]
winter-summer-time

wholmgren:
Ok, I just wanted to make sure that there were no other changes hidden in the PEP8 commit.

I'm not sure where exactly the problem is. My memory is that I never got the results that I expected from this function, didn't trust it, and after making minimal effort at compatibility with the Location class, I gave up on it in favor of making a pyephem wrapper. I thought that the utc_offset code handled daylight savings time, but apparently not (no DST is one of the few things that Arizona does right!). After that, @Calama-Consulting added the NREL spa and nobody seriously looked at ephemeris again until now.

I'm ok with merging this now and opening a new PR for addressing the remaining problems in 0.2. Maybe keep #40 open instead of making a new issue. Sound good @Calama-Consulting or @bmu?

uvchik:
The difference between the two sun rises of the green line is 23 hours. Summer time means that the time is set one hour ahead. At two o'clock the time is set to three o'clock. So the ephemeris keeps the real local time and the pyephem keeps the local standard time (winter time). I would prefer the second variant but I have to look how our weather data set is like.

Still don't understand the -1 in line 312.

@wholmgren I agree with merging it and moving the time shift problem to a new issue. I can do that next week.

jforbess:
I strongly agree that processing modules should keep to standard (winter)
time, unless we have thought through a lot of hard problems that apparently
python and pandas are still working through themselves regarding the
ambiguity of daylight savings.

wholmgren:
I'm going to go ahead and merge and we'll deal with the tz problem separately.

PR #43, opened 2015-03-27
timedelta and pandas compatibility

wholmgren:
@alorenzo175 complained that pvlib required ``pandas >= 0.15``, so I removed a ``pd.Timedelta`` dependency and was able to get the tests to pass on pandas 13.1. Conda wouldn't install 13.0 on my Python 3.4 environment, so I stopped there. 

I also added a pandas environment variable to the travis config to run the test suite against multiple pandas versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ python:
     - 2.7
     - 3.3
     - 3.4
+env:
+    - PD_VERSION=0.13.1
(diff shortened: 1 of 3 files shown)


bmu:
Really useful to test different pandas versions, as sometimes there are subtle errors. Maybe we will see more errors when we have pandas unit tests, e.g. for the index.

PR #45, opened 2015-04-01
fix built in ephemeris

### LINKED ISSUE #40: solarposition.ephemeris

wholmgren:
I fixed the ephemeris function (I think). The key changes are

* replace the local and utc hour calculation code as described in #40.
* multiply longitude by -1 as described in a comment but not actually done in the code.

Here's some sample output.

Tucson summer elevation
tucson summer

Berlin no DST (March 28, 2015)
berlin no dst

Berlin DST (March 30, 2015)
berlin dst

Berlin summer elevation and azimuth.
berlin summer az
berlin summer elev

Closes #40. cc @uvchik

### DIFF
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -219,10 +219,6 @@ def ephemeris(time, location, pressure=101325, temperature=12):
     ----------
     time : pandas.DatetimeIndex
     location : pvlib.Location
-
-    Other Parameters
-    ----------------
-
     pressure : float or DataFrame
           Ambient pressure (Pascals)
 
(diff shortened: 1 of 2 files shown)


wholmgren:
Here's a closer look at the refraction correction. The difference between pyephem and ephemeris is negligible above the horizon.

apparent diff

wholmgren:
Rebased to retest on the many pandas versions. I think this can be merged now or with minor review.

uvchik:
Good work. Thank you.

mariusko:
We are currently working on a problem related to the ephemeris calculations of the sun elevation. The problem is that the time at which ephemeris calculates the sunrise ( above 0°) is sometimes later than the first radiation data derived from our climate model.

We think the problem is related to different methods of calculating the exact time of the sunrise. 
So I would like to ask you if you happen to know about further documentation/explanation of how ephemeris calculates its outputs.

Thanks in advance

wholmgren:
@mariusko I suggest comparing both ephemeris and your climate model to actuals using e.g. google for a couple of locations on a couple of days. My guess is that your climate model does not use the correct equation of time. It was only in the last year or so that the WRF weather model started using the correct equation of time, but this is even less important for climate modeling than it is for weather modeling.

bmu:
@mariusko : The fact that there is irradiance before the sun is above the horizon, may be a hint, that you have a good climate model ;-) You often measure positive irradiance before the sun is above the horizon, due to scattering in the atmosphere and atmospheric refraction.  
Maybe this is one reason.

Issue #47, opened 2015-04-04
Time zones / Time references


bmu:
I think we should make sure, that time zones, other time references are used correctly. 
We can have time stamps given as UTC, really related to a time zone, local time (without DST), mean solar time, true solar time, ...? This is an everyday problem from my experience, so we should be able to handle all of these formats at least in the future. 
For now it could be ok to define, that all times should be given as UTC and leave all time conversion issue to the user.

As pandas has [support for Julian dates since 0.14](http://pandas-docs.github.io/pandas-docs-travis/whatsnew.html#whatsnew-0140-enhancements) ([this](https://github.com/pydata/pandas/pull/4041) is the pull request). I was thinking about creating a pull request to implement `to_solar_time` in pandas. However this is not easy to implement and I am not sure, if they accept it. But this way all time conversion would be in pandas an we could implement thinks based on pandas.

wholmgren:
I'm all for being helpful with timezones. This was my motivation for adding a ``tz`` attribute to the ``Location`` class. The ``solarposition`` module (with some functions in ``tools``) has an attempt at using tz knowledge from either the ``DatetimeIndex`` or the ``Location`` object. Do these fit into your scope or do you want to do something different? I did think that it would be nice if the ``Location`` constructor could use a web service to determine a IANA timezone from a lat/lon. (Or maybe even a lat/lon/tz from a city/state/country search, but that's another topic.)

Guaranteeing that time zones are handled correctly is hard unless you're very restrictive about it. My inclination is to leave the vast majority of the work to pandas. All localized times in pandas are stored as utc internally, so we can and should take advantage of that (as the description above does). 

We could require that all times be localized, or we could throw warnings whenever we detect that people are using non-localized times.

Concerning solar time... I just rely on the zenith and azimuth calculations for a given lat/lon. Is there an application for solar time other than for the actual calculation of those quantities of interest? I don't know how pandas would reasonably incorporate something that requires a specific lat/lon. Maybe we're talking about two different quantities? I think of solar time as the local time shifted to make the minimum zenith occur at 12:00 noon.

bmu:
I mean true or [apparent solar time](http://en.wikipedia.org/wiki/Solar_time), so I think we talk about the same quantities. It is often useful to plot some quantities against this time e.g. to detect the azimuth angle of a system or to detect misalignments between irradiance sensor and modules, ...
So I think it would be useful.  

[...]

PR #48, opened 2015-04-04
Python translation of NREL SPA code

alorenzo175:
Since the NREL SPA C code isn't included in the repo anymore, I thought I would translate the algorithm to python. The translation produces the same results as the C code. When using numpy arrays and operations, this code is just as fast as the cythonized C code. When using numba, the import can take a while because it has to compile, but it is much faster for subsequent calls. Numba can also release the GIL and can use multiple threads.

I also made some changes to the spa C cython files and spa function to take pressure and temperature inputs (instead of hard coded values) and also output apparent zenith and elevation.

This isn't quite ready for a merge yet; I still want to get delta_t automatically from the USNO instead of having a set value or making the user set it.

```python
In [21]:

print len(times)
%time npy = spa_python(times, loc, pressure=82000, temperature=11, delta_t=67.0, how='numpy')
DEBUG:pvlib:Calculating solar position with spa_python code
DEBUG:pvlib:tz_convert to UTC
DEBUG:pvlib:Reloading spa module without compiling
1441
CPU times: user 51 ms, sys: 3 ms, total: 54 ms
Wall time: 52.3 ms
In [22]:

print len(times)
%time nba = spa_python(times, loc, pressure=82000, temperature=11, delta_t=67.0, how='numba', numthreads=1)
DEBUG:pvlib:Calculating solar position with spa_python code
DEBUG:pvlib:tz_convert to UTC
DEBUG:pvlib:Reloading spa module, compiling with numba
DEBUG:pvlib:Only using one thread for calculation
1441
CPU times: user 3.72 s, sys: 16 ms, total: 3.74 s
Wall time: 3.74 s
In [23]:

print len(times)
%time nba = spa_python(times, loc, pressure=82000, temperature=11, delta_t=67.0, how='numba', numthreads=1)
DEBUG:pvlib:Calculating solar position with spa_python code
DEBUG:pvlib:tz_convert to UTC
DEBUG:pvlib:Only using one thread for calculation
1441
CPU times: user 34 ms, sys: 2 ms, total: 36 ms
Wall time: 35 ms
In [24]:

print len(times)
%time nba = spa_python(times, loc, pressure=82000, temperature=11, delta_t=67.0, how='numba', numthreads=4)
DEBUG:pvlib:Calculating solar position with spa_python code
DEBUG:pvlib:tz_convert to UTC
1441
CPU times: user 39 ms, sys: 1 ms, total: 40 ms
Wall time: 12.4 ms
In [25]:

print len(times)
%time eph = get_solarposition(times, loc, 'pyephem', pressure=82000, temperature=11)
DEBUG:pvlib:using PyEphem to calculate solar position
DEBUG:pvlib:tz_convert to UTC
1441
CPU times: user 155 ms, sys: 1 ms, total: 156 ms
Wall time: 154 ms
In [27]:

%time spac = get_solarposition(times, loc, 'spa_c', pressure=82000, temperature=11)
DEBUG:pvlib:using built-in spa code to calculate solar position
DEBUG:pvlib:tz_convert to UTC
CPU times: user 49 ms, sys: 3 ms, total: 52 ms
Wall time: 49.8 ms
```


wholmgren:
This is very cool and very well done. To be clear, @alorenzo175 used NREL's paper, not their code, to make this.

bmu:
This is great!! (I tried to do this myself a while ago, but I gave up after a while!)

One question:
A lot of the spa code is on time calculations. Do you think it is possible to separate the time calculations and the calculation of the sun position? I think (but I'm not sure), that true solar time is called `apparent sideral time` in the code (the code uses naming conventions of astronomers, and I am not an expert in this field ;-). see #47

alorenzo175:
> 'spa' is removed, or it stays but as the default and performs a try: spa_numba; except: spa_python.
> If 'spa' is removed, 'spa_python' is the default.

I would recommend that spa_numpy is the default. Numba can take a few seconds to compile and the numpy code is pretty fast. If someone is doing a lot of calculations and needs the speed, they can specify the use of numba and many threads.

> 'spa_c' becomes 'spa_nrel_c'.

I thought changing 'spa' to something else, but I didn't want to break the API. I think we should all agree on if/how to change it.

> maybe add **kwargs to this and pass to worker functions?

Good idea.

alorenzo175:
> One question:
> A lot of the spa code is on time calculations. Do you think it is possible to separate the time calculations and the calculation of the sun position? I think (but I'm not sure), that true solar time is called apparent sideral time in the code (the code uses naming conventions of astronomers, and I am not an expert in this field ;-). see #47

It is definitely possible to separate out the time calculations. I'll play around with it a little to see the best way to do it.

alorenzo175:
I added a `equation_of_time` output from the solar position algorithm. I think any conversion to true solar time should be another function in `solarposition`.

I also renamed `spa` to `spa_c`, made the numpy calculation the default method to remove the pyephem dependency, and added a function to get sunrise, sunset, and sun transit time.

I think it is ready to merge as soon as it is reviewed.

wholmgren:
This looks good to me. @bmu or @Calama-Consulting?

bmu:
I don't see, were `equation_of_time` is returned and the code is to complex at a first look, so I cannot say I have understood everything. However I would merge it in its present form.

wholmgren:
If anybody wants something changed further then they can make a new issue for it.

uvchik:
There are no errors but unused imports (import warnings, import pytz) and unused variables ```e``` using the following expression: ```except ImportError as e``` but the ```e``` is not used anymore.

Furthermore there some PEP8 issues but I don't know how strict we wanna be.

uvchik:
@alorenzo175 
Do you want to add this or shall I create a new pull request about these issues.

alorenzo175:
I can change it quickly.

Issue #37, opened 2015-03-20
consistent variable names and capitalization


wholmgren:
Let's try to agree on the variable names and capitalization rules for all terms in the library. PEP8 is a good place to start, but it may be reasonable to break those rules for some terms. 

I don't expect that we can come to agreement on everything in this issue, but let's at least write down the terms and rules for which we do agree and change the code to reflect whatever consensus emerges. 

A wiki would probably be a good place to put the results. 

In no particular order, here are some problems that I'm aware of:

* Acronyms such as DNI, GHI, DHI, AOI. 

* Beam irradiance and similar could be ``Eb``, ``beam``, ``beam_irrad``, ``irrad_beam``. I know that at least two those are currently in use.

* ``diffuse``, ``sky_diffuse``, ``diffuse_ground``, ``In_Plane_SkyDiffuse``, ``aoi_projection``, etc.

* ``airmass``, ``AM``, or ``am``? I think we may be using all three. How do we use a modifier like relative or absolute?

* ``lat``, ``latitude``

* ``tz``, ``TZ``

* ``surf_az``, ``surfazi``, ``surfaceazimuth``

* Variables that are closely tied to a particular equation (see many of the ``pvsystem`` functions).

* ``temp``, ``T``, ``Tcell``. 

* ``Vmp``, ``V_mp``, ``vmp``, ``v_mp``.

Fixing this without breaking things requires good tests.

uvchik:
I would use existing conventions like PEP8 because this is easier for new contributors. For variable names I like the convention table of pylint: http://pylint-messages.wikidot.com/messages:c0103

Pylint is a program to check your code. It marks all lines where you broke the rules. You can integrate pylint in many IDEs or editors (http://docs.pylint.org/ide-integration) or use it as a standalone tool to check your modules.

Breaking the rules makes it more difficult for contributors from other projects.

But it's your project, if the convention is clear I will keep to the code.

wholmgren:
I'll put a few of my preferences here, going from perhaps least to most controversial.

Everything that is not an acronym or a complicated variable in an equation should be lower case. This is just good PEP8, and most of the code actually follows this already. Of the list above, the only change would be that ``In_Plane_SkyDiffuse`` becomes ``in_plane_skydiffuse`` or similar.

Airmass should just be called ``airmass`` since it's one word and it's not that long.

Up to about 10 characters, I prefer longer names to shorter names. So, I prefer to spell out words like ``latitude`` and ``longitude``.

[...]

Issue #30, opened 2015-03-12
Port Tutorials 


Calama-Consulting:
Port of tutorials from sandia labs develop branch.

bmu:
I think we could shift this to 0.2 and release 0.1 of pvlib.

wholmgren:
Agreed.

bmu:
@Calama-Consulting ?

bmu:
I changed the milestone to 0.2

wholmgren:
I'm going to mark this 0.3 since I'm not sure what should be ported into new or existing notebooks and what should be put into the docstrings and sphinx documentation.

wholmgren:
I've had a lot of offline feedback that broken tutorials a big problem and I'd like to do a 0.2.1 release with updated tutorials. I don't plan to make them read any better, just get them to the point that they can run end-to-end without exceptions. #69 is also relevant to this.

NelisW:
Will try to help here. My take on the Sphinx vs notebook issue is that personally I prefer a worked example to get up to speed first and then dig into the API-level docs. Trying to get into a new product just by reading the API docs is just too hard. So, let's fix the notebooks first.

PR #51, opened 2015-04-18
pvsystem clean up

wholmgren:
The most important parts of this PR are

* Changed some parameter names in ``pvsystem.sapm`` to follow the rules discussed in #37 and the [wiki](https://github.com/pvlib/pvlib-python/wiki/Variables-and-style-rules). Would be good for people to look at this and make sure that they're happy with how it looks in practice.
* ``pvsystem.sapm`` documentation is more complete. Some module parameters were misnamed and we were missing some others.
* The [pvsystem tutorial](http://nbviewer.ipython.org/github/wholmgren/pvlib-python/blob/pvsystem-fixes/docs/tutorials/pvsystem.ipynb) goes from terrible to very bad, but might help with #30 and #15.

Less important changes:

* I added the "effective irradiance" ``Ee`` to the ``pvsystem.sapm`` output DataFrame. Not sure if this name follows the rules or not. It's a parameter that's very specific to the Sandia model, so that makes me think that it's better to leave as ``Ee``.
* ``pvsystem.sapm`` can now take a dict for a module specification.
* By line, the majority of this PR is PEP8 improvements (except the tutorial of course), especially focused on wrapping long lines. Many PEP8 issues still exist and are beyond the scope of this PR.
* Added a note about a difference with the Matlab function. The difference has existed since the original python port.


bmu:
I see one thing: According to the wiki `systemdef` should return `surface_azimuth` and `surface_tilt` instead of `surfaz` and `surftilt`.

wholmgren:
Thanks. Fixed that and made some other minor changes too.

Calama-Consulting:
This looks good! Thanks for the cleanup

Issue #53, opened 2015-04-20
git master/branch structure


wholmgren:
Now that we have our first release we need to make a decision regarding how to handle the git tree. This was discussed a bit in Sandia-Labs/PVLIB_Python/issues/33, and @uvchik brought it up in #15. 

I always look first to copy IPython and/or pandas. The [ipython system](https://github.com/ipython/ipython/wiki/Dev%3A-GitHub-workflow) is roughly:

* ``master`` should be the next major release. Now that ``0.1`` is on pypi, this would mean that the ``master`` version should be updated to something like ``0.2.0dev``. See [``release.py``](https://github.com/ipython/ipython/blob/master/IPython/core/release.py)
* Branches should be created for any maintenance releases. So, if we're going to bother with a ``0.1.1`` release then it should be developed in a ``0.1.x`` branch.

So far as I can tell, pandas just keeps bumping the version number by 0.0.1 or 0.1 according to the scope of whatever has been merged. Maybe they let some bigger PRs rot for awhile if they're trying to focus on a maintenance release.

In principle I like the IPython system, but in practice, I think that we should go with the pandas system at this early stage. The IPython system probably makes more sense if/when we get to 1.0 and 2.0 releases. However, I don't care too much about this (we just need a decision so that we can update the version file), and I am especially willing to go with the IPython system if somebody else volunteers to be the main administrator of the maintenance releases and development.

Whatever we decide on should be added to the [Development wiki](https://github.com/pvlib/pvlib-python/wiki/Development-information).

bmu:
I would prefer the pandas system (simply let `master` be the next release), because it is the most easiest. Pandas tags release candidates if they made bigger changes, before releasing the next major version. But I think up to now, this is not necessary for pvlib.

If we do so, the only thing we need to do after a release is to change `__version__` in `pvlib.version` to `<<next_release>>.dev` (`0.2.dev` in the actual case), as far as I can see.

uvchik:
For now, it might be enough to go on with the master branch and a few weeks before the release we branch a new 0.2-branch from the actual commit. Then we go on with developing the master. From this time bug fixes will be merged into the "master" and the "0.2" branch but new features will only be merged into the master branch. So we can make sure that the new version is stable. Then we release the 0.2 and go on with the master until we branch a 0.3 branch a few weeks before the 0.3 release, and so on....

bmu:
[...]

PR #57, opened 2015-04-24
Prepare 0.2

bmu:
updated version and prepared whatsnew.

We should add a whatsnew now, to document all changes in 0.2.
Should be independent of the outcome of the development workflow discussion in #53

### DIFF
--- a/docs/sphinx/source/whatsnew.rst
+++ b/docs/sphinx/source/whatsnew.rst
@@ -6,4 +6,5 @@ What's New
 
 These are new features and improvements of note in each release.
 
+.. include:: whatsnew/v0.2.0.txt
 .. include:: whatsnew/v0.1.0.txt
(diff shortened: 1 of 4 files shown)


wholmgren:
Oh good, I forgot about needing to change this.

I noticed that the new what's new file lists all contributors up to present instead of just the 0.2 contributors. I have a slight preference for the shorter list in the what's new file, and perhaps a longer "all contributors" list as a separate part of the documentation.

While you're working on the what's new files can you fix the release date for 0.1 (and make a note of the change in the 0.2 what's new file)?

bmu:
done. not sure if `Contributors`should be something like `Contributors for this release` (or "to this"?). Than a link to the full list of contributors could be given.
But we can add this later, not necessarily within this pull request.

PR #58, opened 2015-04-24
update haydavies kwargs

wholmgren:
Here's a possible improvement to ``haydavies``.

I'm using the ``haydavies`` function inside a loop that basically iterates over many different irradiance forecasts for the same time period. Since the times are always the same, the solar position and zenith angle projections are always the same. However, the ``haydavies`` function recalculated to zenith angle projections every time, and this is a relatively time consuming process since it involves calculating many sines and cosines. So, I added a new keyword argument, ``projection_ratio``, that can bypass those time consuming calculations.

A downside to this is that the ``haydavies`` function signature becomes slightly more complicated. Another downside is that ``haydavies`` gets a feature that functions with the same purpose do not have. There's already a lot of feature mismatch in ``irradiance.py``, so maybe this makes a problem worse or it conforms to an existing style and it's not a big deal.

So, I'm not sure if this should be merged or not. If not, I'll just integrate my version of ``haydavies`` into my other project.

Thoughts?

### DIFF
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -675,7 +675,8 @@ def klucher(surf_tilt, surf_az, DHI, GHI, sun_zen, sun_az):
     return sky_diffuse
 
 
-def haydavies(surf_tilt, surf_az, DHI, DNI, DNI_ET, sun_zen, sun_az):
+def haydavies(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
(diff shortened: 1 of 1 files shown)


bmu:
I think this is useful. However it would be nice, if all transposition models would have the same signature (so you can simply loop over the models for comparison), which differs now even more (you already mentioned that).

But it maybe better to include this and look for simplifications/possible changes in the other models later. So I would merge it the next days, if no other users/contributors comment against this merge.

PR #65, opened 2015-05-08
Doctest fixes

bmu:
Fixed existing doctests so that

`nosetests -v -w pvlib --with-coverage --cover-package=pvlib --with-doctest`

runs without failure.

Not sure if this work for all versions, as the string representation of pandas objects changed in recent versions (doctest compares the output, so older versions may fail).


wholmgren:
Great, looking forward to more of this.

PR #68, opened 2015-06-17
move disc and dirint from clearsky to irradiance

### LINKED ISSUE #42: disc algorithm location
The disc algorithm is currently in ``clearsky``, but I think ``irradiance`` is a better place for it (and the rest of the similar models that still need to be implemented #2 #4).

wholmgren:
Closes #42.

To do before merge:
- [x] Check/update tutorials


wholmgren:
Doesn't look like these functions are currently used in any of the tutorials. I also added this and other issues to the whatsnew doc. I think this is good to go @bmu.

bmu:
Ok, I'll merge this. The tests are passing and this is mostly copying of functions. Didn't checked the html documentation, but restructured text seems to be ok.

PR #74, opened 2015-07-10
Tutorials

wholmgren:
~~Not yet ready to merge~~ 

Closes #30. This PR is just about making the notebooks work from beginning to end. I'm not going to spend much time making them read better.

I wrote a new [``tmy_to_power``](http://nbviewer.ipython.org/github/wholmgren/pvlib-python/blob/tutorials/docs/tutorials/tmy_to_power.ipynb) notebook that replaces ``Test_Scripts``. 

Here's a nbviewer [link to my fork](http://nbviewer.ipython.org/github/wholmgren/pvlib-python/tree/tutorials/docs/tutorials/).

To do:
- [x] ET_Irradiance (move into irradiance)
- [x] TS_2_Irradiance_functions reworked and renamed tmy_and_diffuse_irrad_models
- [x] atmosphere
- [x] irradiance
- [x] pvsystem
- [x] solarposition
- [x] tmy
- [x] tracking
- [x] add 0.2.1 whatsnew

Note that I've already rebased this on #77.

~~@NelisW sorry if this steps on your toes. Would be great to divide and conquer if you have the time to work on one or more of these before ~Monday.~~

PR #78, opened 2015-07-12
notebooks updated - please check zenith angle calcs, strange values

NelisW:
All notebooks run to completion without breaking, but please note:

1.  Zenith angles appear suspect, airmass returnd from atmosphere.airmass_absolute() are mostly NaN. Can someone best acquanted with zenith angle algorithms please investigate.  I added some print code marked with #cjwcjw (my initials) to help with finding.

2.  Notebooks updated to provide correct named function parameters when calling functions, but the local names & column headers in the notebooks are still as original.  Ideally we should update local names to be similar to function parameter names.

3. No attempt was made to check the correctness of the results, I have to little experience in this field to recognise right from wrong.  Can someone just please run the notebooks (they should work now) and check the results?

4.  No attempts at additional documentation & text.  It hurts me to commit this :-( but I would rather commit this now than to wait another week.

5. Minor adjustments were made to the code. The most important is in pvsystem.retrieve_sam() to download and save model files locally. You can now download once and the use the local file afterward.

go well!


wholmgren:
Oh crap I think there was a misunderstanding. I tried to say in #74 that all of the tutorials in the pull request were now working. That's a lot of duplicated tedious work. I'm really sorry about not making that more clear. I will compare your new notebooks to my new notebooks and we can figure out how to move forward.

Most of the changes that are not notebook related should go into a separate PR. Adding the ability to save SAM files via the retrieve_sam function is interesting but will need to be discussed. 

I briefly looked at your notebooks that show zenith and airmass calculations. The calculations seemed ok to me: zenith angles are between 0 and 180 degrees, and airmass is only defined for zenith angles less than 90 degrees. What's not ok is that I accidentally made a plot (atmosphere cell 6) that labels airmass as zenith.

NelisW:
Will & Co

no problem, I should have read more carefully.  I think it is easiest if
you just ignore my changes, yours are probably more comprehensive.  I have
to apologise for not bringing my contribution in time.

Not related to this matter, I must say goodbye.  It became clear to me over
the weekend that personal obligations do not allow me to make a significant
contribution to this very exciting project.  I just cannot make enough time
available without sacrificing in other areas.   Perhaps later....

thanks for your warm and friendly attitude, this is a really nice team and
I really appreciate your ettiquete and kindness!

Go well and good luck.

wholmgren:
@NelisW I'm sorry to see you go and I hope we can work together again in the future!

PR #80, opened 2015-07-12
Updated pointers to current SAM files, updated column names, ... 

jforbess:
and added more replace chars for equipment names [Issue 75]. 

The calcparams_desoto and sapm functions won't work with the previous csv files, because Sandia changed some column names between CSVs, and the change is hardcoded. It seems worth forcing the upgrade, though.

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -537,7 +537,7 @@ def retrieve_sam(name=None, samfile=None):
             url = 'https://sam.nrel.gov/sites/sam.nrel.gov/files/sam-library-cec-modules-2015-6-30.csv'
         elif name == 'sandiamod':
             url = 'https://sam.nrel.gov/sites/sam.nrel.gov/files/sam-library-sandia-modules-2015-6-30.csv'
-        elif name == 'sandiainverter': # Still only one current inverter set, so left it as sandia, to avoid breaking old code
+        elif name in ['cecinverter', 'sandiainverter']: # Allowing either, to provide for old code, while aligning with current expectations
             url = 'https://sam.nrel.gov/sites/sam.nrel.gov/files/sam-library-cec-inverters-2015-6-30.csv'
         elif samfile is None:
             raise ValueError('invalid name {}'.format(name))
@@ -562,6 +562,13 @@ def retrieve_sam(name=None, samfile=None):
 
 def _parse_raw_sam_df(csvdata):
     df = pd.read_csv(csvdata, index_col=0, skiprows=[1,2])
+    colnames = df.columns.values.tolist()
+    parsedcolnames = []
+    for cn in colnames:
+        parsedcolnames.append(cn.replace(' ', '_'))
+
+    df.columns = parsedcolnames        
+
     parsedindex = []
     for index in df.index:
         parsedindex.append(index.replace(' ', '_').replace('-', '_')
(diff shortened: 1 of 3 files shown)


wholmgren:
I'm not too worried about the name changes since I doubt that anybody is accessing those columns directly.

Is there any advantage to adding the ability to choose the data set?

pvlib/pvsystem.py
```diff
@@ -534,11 +534,11 @@ def retrieve_sam(name=None, samfile=None):
...
 
         if name == 'cecmod':
-            url = 'https://sam.nrel.gov/sites/sam.nrel.gov/files/sam-library-cec-modules-2014-1-14.csv'
+            url = 'https://sam.nrel.gov/sites/sam.nrel.gov/files/sam-library-cec-modules-2015-6-30.csv'
         elif name == 'sandiamod':
-            url = 'https://sam.nrel.gov/sites/sam.nrel.gov/files/sam-library-sandia-modules-2014-1-14.csv'
-        elif name == 'sandiainverter':
-            url = 'https://sam.nrel.gov/sites/sam.nrel.gov/files/sam-library-sandia-inverters-2014-1-14.csv'
+            url = 'https://sam.nrel.gov/sites/sam.nrel.gov/files/sam-library-sandia-modules-2015-6-30.csv'
+        elif name == 'sandiainverter': # Still only one current inverter set, so left it as sandia, to avoid breaking old code
```
wholmgren: What if we do something like ``elif 'inverter' in name`` or ``elif name in ['sandiainverter', 'cecinverter']`` to cover both the old code and the expectations of new users?

pvlib/pvsystem.py
```diff
@@ -693,12 +695,12 @@ def sapm(module, poa_direct, poa_diffuse, temp_cell, airmass_absolute, aoi):
         (1 + module['Aimp']*(temp_cell - T0)) )
 
     dfout['v_oc'] = (( module['Voco'] +
-        module['#Series']*delta*np.log(Ee) + Bvoco*(temp_cell - T0) )
+        module['Cells in Series']*delta*np.log(Ee) + Bvoco*(temp_cell - T0) )
```
wholmgren: Surprised to see the spaces in the column name. I thought that ``_parse_raw_sam_df`` was supposed to remove them. Spaces are a problem for people that like to use ``DataFrame.column_name`` (not me).

wholmgren:
Oh, and add your change and name to the 0.2.1 whatsnew file.

pvlib/pvsystem.py
jforbess: The spaces are only removed from the index field, not the column names. I was actually tempted to change the column names to make them lowercase, so that they would be slightly more robust to Sandia's tweaks. I could make them lower and remove spaces, seems reasonable.

pvlib/pvsystem.py
jforbess: I'll add the latter.

pvlib/pvsystem.py
jforbess: I take it back. I don't think it's worth all of the changes to make them lowercase, which is why I didn't do it in the first place. I'll replace the spaces, though, since it's only a few variables.

pvlib/pvsystem.py
jforbess: Personally I don't see much value in old versions, because if you are paying that close attention to the different module or inverter parameters, you would probably do better to have a local copy that you version yourself, and can specify directly as a file (and would have to update the column names to match, but that's not hard at this point).

wholmgren: Ok, good points.

The doc string needs to be updated with something about how you can use either name and there is currently no difference between the two. I will merge after that.

wholmgren:
Thanks @jforbess!

PR #81, opened 2015-07-27
added link to paper.

dacoex:
added link to paper.

Maybe there can eb new page with publications

### DIFF
--- a/docs/sphinx/source/index.rst
+++ b/docs/sphinx/source/index.rst
@@ -38,7 +38,7 @@ Please see our
 `PVSC 2014 paper <http://energy.sandia.gov/wp/wp-content/gallery/uploads/PV_LIB_Python_final_SAND2014-18444C.pdf>`_
 and
 `PVSC 2015 abstract <http://energy.sandia.gov/about/publications/download-info/pvlib-python-2015/>`_ 
-(and the `notebook to reproduce the figures <http://nbviewer.ipython.org/github/pvlib/pvsc2015/blob/master/paper.ipynb>`_) therein for more information.
+(and the `notebook to reproduce the figures <http://nbviewer.ipython.org/github/pvlib/pvsc2015/blob/master/paper.ipynb>`_ therein) for more information.
 
 Installation
 ============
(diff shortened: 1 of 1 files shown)


docs/sphinx/source/index.rst
```diff
@@ -38,7 +38,7 @@ Please see our
 `PVSC 2014 paper <http://energy.sandia.gov/wp/wp-content/gallery/uploads/PV_LIB_Python_final_SAND2014-18444C.pdf>`_
 and
 `PVSC 2015 abstract <https://github.com/UARENForecasting/pvlib-pvsc2015/blob/master/pvlib_pvsc_42.pdf?raw=true>`_ 
-for more information.
+(and the `final paper <https://github.com/pvlib/pvsc2015>`_) for more information.
```
wholmgren: "final paper" doesn't seem like the right way to describe this. A better way might be to change "PVSC 2015 abstract" to "PVSC 2015 paper" with the correct pdf link, then replace "and the final paper" with something like  "GitHub repository". I don't see any benefit to keeping the link to the abstract.

wholmgren:
@dacoex do you want to make the suggested changes or something similar? If so, please also update the change and version files.

dacoex:
OK, I made the changes.

I guess in future there may be dedicated page for the references & project links...

wholmgren:
Thanks.

dacoex:
Easy, One cannot only talk, right?

docs/sphinx/source/index.rst
```diff
@@ -37,8 +37,10 @@ help updating them!
 Please see our 
 `PVSC 2014 paper <http://energy.sandia.gov/wp/wp-content/gallery/uploads/PV_LIB_Python_final_SAND2014-18444C.pdf>`_
 and
-`PVSC 2015 abstract <https://github.com/UARENForecasting/pvlib-pvsc2015/blob/master/pvlib_pvsc_42.pdf?raw=true>`_ 
-for more information.
+`PVSC 2015 abstract <http://energy.sandia.gov/about/publications/download-info/pvlib-python-2015/>`_
```
wholmgren: This is not the correct link and I should have looked at this more closely before merging. I fixed this with commit 263c596

PR #90, opened 2015-10-26
Python 3.5, reconfigure Travis

### LINKED ISSUE #87: test against python 3.5, pandas 0.17
Python 3.5 was released yesterday (Sep 14, 2015) and Pandas 0.17 is scheduled to be released sometime this week. I'll add them to the travis build matrix later this week. I don't expect any problems.

wholmgren:
This PR primarily does two things:
* Allows pvlib to be installed on Python 3.5
* Reconfigures the TravisCI set up to work around Continuum's crappy scipy/linux/python3 package.

I originally thought that there was an issue with scipy/linux/python3, but it turns out that it's actually a problem with the version of gcc that Continuum uses to build scipy. For more, see ContinuumIO/anaconda-issues#425. The Readme text has been updated accordingly. 

The Travis build matrix was getting to be too big and take too long, so this PR only tests the minimum pandas/numpy version on Python 2.7, and the latest versions for Python 2.7, 3.4, and 3.5. I think it is reasonable to say that pvlib is supported on the intermediate versions (Python 3.3, pandas 0.14-0.16) even if we're not explicitly testing for it.

One thing that I'm no longer testing is the numba accelerated SPA code on Python 3. It is still tested on Python 2.7. I'll change it once Continuum gets their SciPy packages fixed. I've already spent way too much time on this PR to hack this together too.

@bmu @Calama-Consulting I will merge this in ~24 hours if nobody objects. I think it's also worth the effort to make a minor release later this week.

Closes #87.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,40 +1,71 @@
+# Based on
+# http://conda.pydata.org/docs/travis.html
+# https://github.com/xray/xray/blob/master/.travis.yml
+# https://github.com/scipy/scipy/blob/master/.travis.yml
+
 language: python
-python:
-    - 2.7
-    - 3.3
-    - 3.4
-    - 3.5
-env:
-    - PD_VERSION=0.13.1
-    - PD_VERSION="0.14.*"
-    - PD_VERSION="0.15.*"
-    - PD_VERSION="0.16.*"
-    - PD_VERSION="0.17.*"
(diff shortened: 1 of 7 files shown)


wholmgren:
I going ahead with this but will revert or fix if there are any future objections.

PR #92, opened 2015-10-27
replace format with %s in logging calls

### LINKED ISSUE #89: make logging calls more efficient
I recently learned that I've been using the ``logging`` module incorrectly. The problem is that I've been using the pattern 

```python
logger.debug('array of numbers that you rarely want to see: {}'.format(numbers))
```

Regardless of the logging level, ``numbers`` will be formatted into a string. This is because the string formatting is part of the input to the logging function, so it happens before the logging module has a chance to decide if it actually needs to do the conversion and print the output. 

The correct way to do it is

```python
logger.debug('array of numbers that you rarely want to see: %s', numbers)
```

Now ``numbers`` is only converted into a string when the logging level is set to ``debug``.

This only makes a measurable impact in function time in a few places, but I want to change it throughout the library so that people don't learn bad habits from my examples. 

I had some utility functions in some other code that spent 75% of their time doing string formatting for rare logging calls instead of doing real work. Embarrassing.

wholmgren:
closes #89 

I'll rebase this on #90 once I work out yet another CI issue in that.

### DIFF
--- a/docs/sphinx/source/whatsnew.rst
+++ b/docs/sphinx/source/whatsnew.rst
@@ -6,6 +6,7 @@ What's New
 
 These are new features and improvements of note in each release.
 
+.. include:: whatsnew/v0.2.2.txt
 .. include:: whatsnew/v0.2.1.txt
 .. include:: whatsnew/v0.2.0.txt
 .. include:: whatsnew/v0.1.0.txt
--- a/docs/sphinx/source/whatsnew/v0.2.2.txt
+++ b/docs/sphinx/source/whatsnew/v0.2.2.txt
@@ -1,0 +1,27 @@
+.. _whatsnew_0220:
+
+v0.2.2 (November, 2015)
+-----------------------
+
+This is a minor release from 0.2.1.
+We recommend that all users upgrade to this version.
+
+Enhancements
+~~~~~~~~~~~~
+
+* Adds Python 3.5 compatibility (:issue:`87`)
+
+
+Bug fixes
+~~~~~~~~~
+
+* Uses the ``logging`` module properly by replacing ``format``
+  calls with ``args``. This results in a 5x speed increase for
+  ``tracking.singleaxis`` (:issue:`89`).
+* Adds a link to the 2015 PVSC paper (:issue:`81`)
+
+Contributors
+~~~~~~~~~~~~
+
+* Will Holmgren
+* dacoex
--- a/pvlib/atmosphere.py
+++ b/pvlib/atmosphere.py
@@ -237,8 +237,8 @@ def relativeairmass(zenith, model='kastenyoung1989'):
         AM = ( 1.0 / (np.cos(zenith_rad) +
             0.00176759*(z)*((94.37515 - z) ** - 1.21563)) )
     else:
-        pvl_logger.warning("{} is not a valid model type for relative airmass. The 'kastenyoung1989' model was used."
-                           .format(model))
+        pvl_logger.warning("%s is not a valid model type for relative airmass. The 'kastenyoung1989' model was used.",
+                           model)
         AM = ( 1.0 / (np.cos(zenith_rad) +
             0.50572*(((6.07995 + (90 - z)) ** - 1.6364))) )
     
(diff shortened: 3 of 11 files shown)


wholmgren:
Rebased. I'll wait a day before merging.

jforbess:
Did you get an estimate on how much of an improvement this is in
performance?

wholmgren:
Good question. 

I only tested ``tracking.singleaxis`` since it has by far the most debug calls with large arrays or dataframes. Calling format on a single number is pretty fast (more on this at the bottom). I ran the first 4 cells of [``tracking.ipynb``](http://nbviewer.ipython.org/github/pvlib/pvlib-python/blob/master/docs/tutorials/tracking.ipynb), then skipped down to [tracking.singleaxis examples](http://nbviewer.ipython.org/github/pvlib/pvlib-python/blob/master/docs/tutorials/tracking.ipynb#tracking.singleaxis-examples). I profiled the first cell in that section with ``%%timeit`` and ``%lprun``...

With the current ``pvlib:master (9b2b368)``:

```python
%%timeit
tracker_data = pvlib.tracking.singleaxis(pyephemout['apparent_zenith'], pyephemout['apparent_azimuth'],
                                         axis_tilt=0, axis_azimuth=180, max_angle=90,
                                         backtrack=True, gcr=2.0/7.0)

10 loops, best of 3: 138 ms per loop
```

And if I run ``line_profiler`` with

```python
%lprun -f pvlib.tracking.singleaxis pvlib.tracking.singleaxis(pyephemout['apparent_zenith'], pyephemout['apparent_azimuth'], axis_tilt=0, axis_azimuth=180, max_angle=90, backtrack=True, gcr=2.0/7.0)
```

I see that ~75% of the time is spent on the ``pvl_logger.debug`` statements from lines 263-284.

After a ``git checkout logging`` and...

```python
from importlib import reload
pvlib = reload(pvlib)
pvlib.tracking = reload(pvlib.tracking)
```

...with the current PR ``wholmgren:logging (3acc134)``:

```python
%%timeit
tracker_data = pvlib.tracking.singleaxis(pyephemout['apparent_zenith'], pyephemout['apparent_azimuth'],
                                         axis_tilt=0, axis_azimuth=180, max_angle=90,
                                         backtrack=True, gcr=2.0/7.0)

10 loops, best of 3: 26 ms per loop
```

More than 5x faster! Face, meet palm.

And if I run ``line_profiler`` with

```python
%lprun -f pvlib.tracking.singleaxis pvlib.tracking.singleaxis(pyephemout['apparent_zenith'], pyephemout['apparent_azimuth'], axis_tilt=0, axis_azimuth=180, max_angle=90, backtrack=True, gcr=2.0/7.0)
```

those logging calls take 0.0% of the time. Note that the logging call on line 100 still uses 1% of the time due to the fact that Python must still call ``apparent_zenith.head(), apparent_azimuth.head()`` before the ``logging`` module decides if string formatting is necessary. I suppose it could be removed and people could use a debugger if they're really that interested. Anyways, the longest call is now line 214: ``widc = pd.Series(index=times)`` with 25% of the time, and the next two ``widc`` assignment lines combining for another 25%. 

Finally some numbers to back up the claim that single value formatting is already reasonably fast and to justify the fact that I don't want to bother to benchmark the rest of the library:

```python
# numpy float
In [21]:
a = np.float64(1)
b = np.float64(2)

In [22]:
%%timeit
'{} {}'.format(a, b)
The slowest run took 15.96 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 1.91 µs per loop

# python int
In [23]:
a = 1
b = 2

In [24]:
%%timeit
'{} {}'.format(a, b)
The slowest run took 8.39 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 535 ns per loop

# python float
In [27]:
a = 1.
b = 2.

In [29]:
%%timeit
'{} {}'.format(a, b)
The slowest run took 10.21 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 916 ns per loop
```

So it's faster to format native python ints than native python floats than numpy floats. All of these times are much faster than any of the functions that they'd be found in.

jforbess:
Great, thanks.

wholmgren:
I added a 0.2.2 whatsnew doc with a note about this change. It can be renamed in the future if we go straight to 0.3.

Issue #84, opened 2015-08-20
Tutorilal: system modeling


dacoex:
NOTE: no bug but clarification question.

Reading the tutorials is is unclear how inverter connections are defined and the power of a whole PV system of more than 1 string is calculated.

The notebook shows that 1 module is picked and then DC power calculated.
http://nbviewer.ipython.org/github/pvlib/pvlib-python/blob/master/docs/tutorials/tmy_to_power.ipynb#DC-power-using-SAPM

But the next step
http://nbviewer.ipython.org/github/pvlib/pvlib-python/blob/master/docs/tutorials/tmy_to_power.ipynb#AC-power-using-SAPM

uses a micro-inverter on module level.

How would the whole process be done with a different inverter, say a 30kW multi-string inverter?

And how would these then be added into a system of 500kW?

If it's roof-top, then multiplying by number of inverters would be possible but on a flat roof or ground mounted case, row-to-row spacing would need to be accounted for.

wholmgren:
Feel free to make PRs with improved tutorials.

dacoex:
Well, that was the purpose of my comment: I am not really sure how it is done with pvlib.

Although I could see the potential applications already.

wholmgren:
There is currently no standardized way to do something like that in pvlib. Would be nice to see development in that direction.

dacoex:
Ok, so it means that your application of pvlib is rather for forecasting. There, single system is not that important but rather an upscale get of the typical behaviour, right?

My focus would be more system analysis.

Back to the topic: so the 30kW inverter would be simulated with a, say a number of 138 modules of 260Wp totalling to about 36kW.

So in section 
http://nbviewer.ipython.org/github/pvlib/pvlib-python/blob/master/docs/tutorials/tmy_to_power.ipynb#AC-power-using-SAPM
One would have to manually add the strings to series and parallels, correct?

jforbess:
Yes, I have put together a dict that I use that expands the systemdef, and
perhaps it is time to wrap it into the systemdef dict, because it is pretty
unhelpful to define how many modules you have if you don't define which
kind of modules they are, or what kind of inverter you are using.

One more item on the todo list.

jforbess:
My dict isn't anywhere to be found in the pvlib code, in case it wasn't
clear.

wholmgren:
Realistic system level modeling in pvlib needs expert help. I don't have the time or the expertise to do it right. There is discussion in #17 and #39 that is relevant.

[...]

PR #93, opened 2015-10-30
Location demotion and PVSystem/LocalizedPVSystem/ModelChain addition

wholmgren:
This code does not work. 

~~This PR is only to make it easier to share and discuss the ideas originally proposed in #17.~~ **Edit: this will eventually be merged.**

I'm going to flesh it out a bit more and provide some examples in the next week or so.

There are two main ideas in this PR:

1. Make the procedural code self contained. If you don't like objects, don't use them.  ``get_solarposition(times, location)`` becomes ``get_solarposition(times, latitude, longitude)`` and similar changes for a few other functions.
2. Create a class structure. Want the solar position for a ``PVSystem``? Call ``PVSystem.get_solarposition(times)``.

Why add classes at all? A ``PVSystem`` class can potentially help to separate the intrinsic and extrinsic parts of a system model. When you make a ``PVSystem`` object, you'd give it intrinsic attributes such as lat, lon, and module type. When you want to model the power of your system, you'd need to supply extrinsic data such as irradiance and temperature to the object instance methods. This logical separation seems useful to me, but your mileage may vary.

The proposed class hierarchy is:

``Location``
``PVSystem``
``SingleAxisTracker``

I think there is value in having a ``Location`` class because it does not require any system specs to do things like calculate solar position or model clearsky profiles. Maybe it's not worth the extra effort to make this distinction.

Any of these objects could be created by the metadata from a TMY file via the ``from_tmy`` classmethod (again, not tested).


bt-:
@wholmgren I recently discovered pvlib-python and I have been reading through the current conversations in #17 and #84.  I am fairly new to the scientific python stack, but I wanted to contribute a few thoughts.

I like the idea of creating a `PVSystem` class and your logic of extrinsic data and intrinsic attributes makes sense to me.

I think it would be useful to have the ability to run calculations for a location without needing to define a complete PV system.  Could this be achieved by creating `PVSystem` as an Abstract Base Class and having `Location` as a child class?  Or, maybe the `PVSystem` class is written in a way that it is simple to just create a `PVSystem` class object with only location attributes?

What type of attributes and methods do you envision for the `SingleAxisTracker` class?

wholmgren:
@bt- thanks for the feedback. 

We could choose to make the ``PVSystem`` constructor accept keyword arguments (probably with ``None`` defaults) for most/all attributes. Then it could be as well or as poorly specified as you'd like. We'd need to do something to make it clear that not everything is going to work if you don't fully specify your system. Maybe you just specify lat/lon so that you can use the solar position and clear sky methods. Maybe you just specify a module and inverter so that you can use the power functions. If a ``PVSystem`` can be as poorly specified as you'd like, then the usefulness of a separate ``Location`` class seems to be pretty low.

What would be the advantage of designating one of these classes as an abstract base class? On first consideration, it to me like an unnecessary complication.

Concerning ``SingleAxisTracker``, I hadn't thought of anything other than what's already in this PR's ``tracking.py``. It has its tracking function and then all of the inherited methods from ``PVSystem`` and ``Location``.

bt-:
@wholmgren I just read the PVLIB Python 2015 paper and I've looked at the code more since my last comment.  I think it probably does make sense to keep a `Location` class and let `PVSystem` inherit from it.

I'm interested to know what others think about the repository that @dacoex referenced and how much interest there is in general in developing more system modeling capabilities.

wholmgren:
@bt- I wrote a short book over at that issue in feedinlib, so maybe the conversation will shift there for awhile.

wholmgren:
Rebased on #98 and fixed/added enough stuff to make ``get_solarposition`` and ``get_clearsky`` work.

dacoex:
>I think that it's starting to feel more concrete, but there's still a ton of work to do if we want to make this >viable. 

What exactly?

I would remove the ```get_``` prefix.

wholmgren:
>What exactly?

If I make a to do list are you going to help do the work?

>I would remove the get_ prefix.

Usually I would agree, however, the ``get_`` prefix allows you do to things like:

```python
tucson_system.clearsky = tucson_system.get_clearsky(times)
```

Do you have an alternative suggestion?

dacoex:
>If I make a to do list are you going to help do the work?

Yes, but very step by step because my main work is not about modelling and code development.

I am currently interest in getting pvlib toward the status to have something like what was in the tutorial operational:
https://github.com/pvlib/pvlib-python/pull/88

> Do you have an alternative suggestion?

Not now. One would just have to check that the ```get_``` functions get too long.

wholmgren:
>I am currently interest in getting pvlib toward the status to have something like what was in the tutorial operational: #88

Go for it. I am looking forward to the pull request.

dacoex:
>Not now. One would just have to check that the get_ functions get too long.

Not this one. The only one with this issue would potentially be ```get_poa_horizontal_ratio```. But it's I minor issue. I think we should implement things. Most discussion has come to a good conclusion. The rest coudl be sorted on the way.

dacoex:
@wholmgren as follow-up to https://github.com/pvlib/pvlib-python/pull/93#issuecomment-155571354 I created a page for [Roadmap](https://github.com/pvlib/pvlib-python/wiki/Roadmap).

Hope this helps and does not add too much overhead.

wholmgren:
I added skeletons for ``ModelChain`` and ``LocalizedPVSystem``, wrote some [documentation](http://wholmgren-pvlib-python-new.readthedocs.org/en/pvsystem/index.html), and rebased on master. The [Modeling paradigms](http://wholmgren-pvlib-python-new.readthedocs.org/en/pvsystem/package_overview.html#modeling-paradigms) documentation is the most important update since it tries to summarize the long and hard to follow conversation in #17. The example code is closer to working, but it's not there yet.

The ``PVSystem`` implementation is a little bit better than it was. Any thoughts on how to handle the intrinsic parameters for functions such as ``ashraeiam`` and ``physicaliam``? These functions require arguments named b, K, L, and n. I don't really want to support attaching these kinds of vague attributes to ``PVSystem`` objects. I'm considering eliminating functions like this from the class specification and just leaving them in the procedural code. Maybe we don't need to support every function, especially if a custom ``ModelChain`` can be written to use it.

I don't really know what I'm doing with ``ModelChain``. I would be happy if someone else would take over the implementation of this class (*cough* @bmu @alorenzo175 *cough*). I added an ``orientation_strategy`` kwarg that drives ``self.surface_azimuth`` and ``self.surface_tilt``. Might be a bad idea, not sure at this point.

bmu:
I won't be able to do this in let's say the next months, sorry!

In my mind, `PVSystem` is a class the holds all the details on the system, so something like a system description. `ModelChain` combines all necessary steps to actually **model** the system given the information in `Location`. As far as I can see from your `PVSystem` implementation it holds a lot of steps, that I would see in `ModelChain`.

wholmgren:
@bmu thanks, your feedback is helpful. 

Some questions for ``ModelChain`` proponents @bmu @alorenzo175 or anybody else:
1. Would you put any methods at all on a ``PVSystem`` or would the class just be a formal collection of attributes?
2. Should ``ModelChain`` implement wrappers of procedural functions like ``sapm`` or should it just call the procedural functions as part of a ``run_model`` step?

@alorenzo175's example code this [comment](https://github.com/pvlib/pvlib-python/issues/17#issuecomment-156237181) relies on the ``location`` and ``system`` instances to do the work, while the instance of ``ModelChain`` ties together the intermediate results in a prescribed way. So I think that @bmu and @alorenzo175 are in disagreement about the scope of ``ModelChain``.

cwhanse:
> Any thoughts on how to handle the intrinsic parameters for functions such as  ashraeiam  and  physicaliam ? These functions require arguments named b, K, L, and n. 

I'd put them in module_parameters

pvlib/irradiance.py
```diff
@@ -376,7 +376,7 @@ def total_irrad(surface_tilt, surface_azimuth,
     model = model.lower()
     if model == 'isotropic':
         sky = isotropic(surface_tilt, dhi)
-    elif model == 'klutcher':
+    elif model in ['klucher', 'klutcher']:
```
alorenzo175: Should probably put a deprecation warning here if the old spelling is used.

wholmgren: This was already merged in #100 and then I rebased after the 0.2.2 release, so I didn't expect this to show up on the diff. Maybe I should have merged 0.2.2 into my branch instead of rebased.

The deprecation warning probably would have been nicer.

alorenzo175:
> @alorenzo175's example code this comment relies on the location and system instances to do the work, while the instance of ModelChain ties together the intermediate results in a prescribed way. So I think that @bmu and @alorenzo175 are in disagreement about the scope of ModelChain.

I think we might be in disagreement. I like ``PVSystem`` having the system specific methods so that one could play around with different ways of calculating one step without having to initialize a ``ModelChain``. ``ModelChain`` would do all the modeling by calling ``PVSystem`` methods.


> I don't really know what I'm doing with ModelChain. I would be happy if someone else would take over the implementation of this class (cough @bmu @alorenzo175 cough). I added an orientation_strategy kwarg that drives self.surface_azimuth and self.surface_tilt. Might be a bad idea, not sure at this point.

I'm not sure I'll have time to implement the class, but I'll try by the end of next week.

wholmgren:
I removed the ``location`` arg from the ``solarposition`` methods and replaced it with ``latitude``, ``longitude``, and ``altitude`` args. So, now you can use the solar position calculators without needing to construct the annoying location objects. As discussed elsewhere, the ``Location`` and ``LocalizedPVSystem`` classes will wrap ``get_solarposition``. 

A side effect (benefit?) is that the solar position calculators will now assume UTC if the input time is not localized to a timezone. No more ``location.tz`` magic.

Tests are now working for the modified clear sky and solar position signatures. I changed the clear sky testing solar position method to ``'nrel_numpy'`` and moved ``altitude`` from an arg to a kwarg. So, the tests are failing because the output has changed by a few W/m^2 and I haven't yet updated the expected output.

dacoex:
@wholmgren I lost the track on the refractoring.
Could you add some tasks to wiki (https://github.com/pvlib/pvlib-python/wiki/Roadmap#022)?
If there are small tasks that could be worked on I may be able to contribute. thx and rgds.

wholmgren:
@dacoex you could put together your idea of what the ``ModelChain`` class should look like. You can also make more comments on the code in this PR.

dacoex:
@wholmgren thanks for the reply.

What would you see as missing to make https://github.com/pvlib/pvlib-python/pull/88 work?

This would be my focus / interest.

wholmgren:
#88 already "works": you can checkout the code, run the notebook, and modify it to your liking. It's not merged due to a number of style issues that I commented on in that issue.

I think that what you've been requesting for quite a while is a standard, recommended approach to modeling in pvlib-python. First, this PR includes some incomplete documentation on that topic -- perhaps you want to help improve it. Next, you seemed to be a big proponent of the ``ModelChain`` class, so I thought that's what you'd want to work on to realize that goal. Finally, you could implement some kind of functional version of the proposed ``ModelChain`` class.

dacoex:
>#88 already "works": you can checkout the code, run the notebook, and modify it to your liking. It's not >merged due to a number of style issues that I commented on in that issue.

OK, so I will focus on this. Just be patient with me because I am more an occsional coder. Working more on porject development and other stuff as such.
Shall I pull @jforbess code and try to implement the comments?

>I think that what you've been requesting for quite a while is a standard, recommended approach to >modeling in pvlib-python. 

Not sure what you refer to specifically.

>Next, you seemed to be a big proponent of the ModelChain class, so I thought that's what you'd >want to work on to realize that goal. Finally, you could implement some kind of functional version of >the proposed ModelChain class.

Fine. When above part is done.

wholmgren:
From my [comment](https://github.com/pvlib/pvlib-python/pull/93#issuecomment-166201953) above...
>Tests are now working for the modified clear sky and solar position signatures. I changed the clear sky testing solar position method to 'nrel_numpy' and moved altitude from an arg to a kwarg. So, the tests are failing because the output has changed by a few W/m^2 and I haven't yet updated the expected output.

This is fixed now and all of the old tests are passing. I also removed pyephem and scipy from the minimum test configuration.

dacoex:
So are you planning to merge:
https://github.com/wholmgren/pvlib-python/blob/37a36da4de34bc03b3cdc9fd3962c6bf1acbceb5/pvlib/modelchain.py

wholmgren:
The version of ``modelchain.py`` that is initially merged will be very experimental at best and not functional at worse. However, I think that other people will have an easier time contributing to it if we can get the rest of this into the main repo. I am open to other ideas, though.

dacoex:
Sorry, misunderstanding. What is blocking us currently from merging it?

If you feel this is too experimental, I would propose to have a new branch *modelchain* in `pvlib/pvlib-python/`. This o would make contribution easier.

The same would be great for the system modelling though. I plan to fix the #84 but currently, I have to clone from @jforbess and then push eleswhere. If there was a branch `systemmodelling` under `pvlib/pvlib-python/` it would be easier.

Like feature-branches for both items.

Sorry, there may be better models for all this git branching, but really this is not my daily turf.

What do you think?

wholmgren:
> What is blocking us currently from merging it?

I'm working on the new [modeling paradigms](http://wholmgren-pvlib-python-new.readthedocs.org/en/pvsystem/package_overview.html#modeling-paradigms) section of the docs. The first of 3 sections is working. I might be ok with merging once this is done. Writing the first part of the docs exposed a few issues that I had not considered.

I'm generally not a fan of feature branches on the main repo. This is the sort of PR that might warrant one, but I'd probably still rather merge it into master once we know it does what it's supposed to do. Also, it's easy (once you know how or google it, at least) to clone from someone else's fork and then push a new branch to your own, or to make PRs across forks. 

Note that recent commits also close #105.

dacoex:
>>What is blocking us currently from merging it?

>I'm working on the new modeling paradigms section of the docs. The first of 3 sections is working. I >might be ok with merging once this is done. Writing the first part of the docs exposed a few issues that >I had not considered.

OK, find. I will be patient. Enough other things to do with/without pvlib ;-)

dacoex:
>>Also, it's easy (once you know how or google it, at least) to clone from someone else's fork and then push a new branch to your own, or to make PRs across forks. 

So like this?
https://github.com/dacoex/pvlib-python/tree/issue84

wholmgren:
> So like this?

That's the result of @jforbess's branch merged with pvlib/master. In that case, that's probably what you want, but you can just get the original branch also.

>OK, find. I will be patient. Enough other things to do with/without pvlib ;-)

If you checkout wholmgren/pvlib/pvsystem then you can use it right now and help me find problems.

wholmgren:
I think that this PR is nearly ready to merge, assuming that:

1. No one has strong objections to the direction that we’re moving in.
2. No one objects to merging these major api changes into the master branch.

See the [what’s new docs](http://wholmgren-pvlib-python-new.readthedocs.org/en/pvsystem/whatsnew.html#v0-3-0-2016) for a list of changes. I might have missed a few.

See the [modeling paradigms docs](http://wholmgren-pvlib-python-new.readthedocs.org/en/pvsystem/package_overview.html#modeling-paradigms) for examples. In the future, I'd like to greatly expand the examples that are rendered on rtd.

Most of the more complicated new code is tested. Some of the PVSystem functions are missing tests. Some of the optional arguments are not tested.

I doubt that I implemented things, especially ``ModelChain``, the way that people were thinking about them. I am definitely open to changes. I could make some changes as part of this PR and we could make more changes in future PRs.

I know that the diff is a bit long, but it is worth your time to read it and comment sooner than later. I'm hoping to merge this within a week.

dacoex:
>I think that this PR is nearly ready to merge, assuming that:

Wow, you are fast. Winter weather?

> Some of the PVSystem functions are missing tests. Some of the optional arguments are not tested.

How can I find the non-tested code?

>I doubt that I implemented things, especially ModelChain, the way that people were thinking about them. 

Why?

I thinnk this is just fine:
https://github.com/wholmgren/pvlib-python/blob/pvsystem/pvlib/modelchain.py#L5
http://wholmgren-pvlib-python-new.readthedocs.org/en/pvsystem/package_overview.html#modeling-paradigms

I would just make the difference between the two object oriented paradigm more clear. 
As I understand, is LocalizedPVSystem a convenience class, or very special modelchain.

this docstring describes it well:
```
See the :py:class:`PVSystem` class for an object model that
    describes an unlocalized PV system.
```
i would emphasize it more.

>See the modeling paradigms docs for examples. In the future, I'd like to greatly expand the examples that >are rendered on rtd.

I think this is the best way I could contribute and also learn the inner workings of the code.
I would suggest that the tutorials are converted into real rst docs. By this manner, they could be included in the testing framework.

The [Ipython Directive](http://matplotlib.org/sampledoc/ipython_directive.html#ipython-directive) could be used.
Ideally, II would prefer to deliver the code like here:
http://matplotlib.org/examples/lines_bars_and_markers/barh_demo.html

So finally, the best combination between both worlds is here:
http://yt-project.org/docs/dev/quickstart/introduction.html

A nice page with the notebooks at the bottom.

It come from this repo:
https://bitbucket.org/yt_analysis/yt/src/34ed15d98fe48c75d824fbf6246d8ab89c4e2398/doc/?at=yt
The only question I have is how these notebooks are tested.

wholmgren:
>How can I find the non-tested code?

Look at our [coveralls page](https://coveralls.io/github/pvlib/pvlib-python). Writing tests is also a good way to learn the inner workings of the code.

>Why?

There was some disagreement and it was hard for me to follow some of the discussion when rereading it.

>As I understand, is LocalizedPVSystem a convenience class, or very special modelchain.

``LocalizedPVSystem`` is totally unrelated to ``ModelChain``, but this goes to your point of the docs needing to be more clear. So, I'll work on it a little more.

>The Ipython Directive could be used.

I'm using it in the new docs, so I don't understand why you're saying this.

dacoex:
>> Why?
>There was some disagreement and it was hard for me to follow some of the discussion when rereading it.

Let's have a few model chains and then see if it works. I like your implementation of the basic version.

>I'm using it in the new docs, so I don't understand why you're saying this.

added #120 to move the doc related discussion there.

wholmgren:
A couple of dumb problems have sprung up...

1. The figures in the modeling paradigms section are no longer being rendered properly. I've spent more hours on this than I care to admit to and I'm still not sure why it stopped working. It could be a rtd/ipython problem.
2. One of the windows test suites now fails to run. I think this is because Continuum recently started bundling the MKL optimized libraries with their numpy/scipy packages.

dacoex:
>rtd/ipython problem.

And on your local machine?

wholmgren:
>And on your local machine?

Using IPython 4.0.1 it makes the images just fine, but has a sqlite error.

Using IPython 4.1.1, it works locally only if the build directory already exists. So, it works if you run ``make clean; make html; make html``.

dacoex:
What if we forget about the IPython dependancy?

If we include code like matplotlib did:
http://matplotlib.org/users/pyplot_tutorial.html

This is actually very convenient and reduces overhead.

wholmgren:
You'd be adding the overhead of matplotlib's plot directive, which looks a lot worse to me.

uvchik:
According to @wholmgren's [post at feedinlib](../../../oemof/feedinlib/issues/9#issuecomment-184275610) I will give some feedback.

I really like the idea of the ModelChain class. It lowers the entry barrier for new users and maybe we can shorten our feedinlib-pv-code by using this class.

So the `modelchain` is great but what is the idea to use oop instead of procedural programming? What is the real effort from `In [16]` to `In[33]` at http://wholmgren-pvlib-python-new.readthedocs.org/en/pvsystem/package_overview.html#modeling-paradigms

As I understand it one advantage of oop is that you can inherit. So we could create a new class `PVSite` (name is not important) which inherits the `PVSystem` class but has some more attributes such as `location`, `weather(optional)`,.... This class could have a method `run_model` and an attribute `ac`. The code would look as follows:

```python
my_pv = PVSite(module_parameters=my_mod, inverter_parameters=my_inv, location=my_loc)
my_pv.run_model(times)
annual_ac_energy = my_pv.ac.sum()
annual_dc_energy = my_pv.dc.sum()
```
I know that this is a matter of taste but I prefer attributes instead of using the methods like functions. If one method calculates `aoi` than the other method should use `self.aoi`. I thought the new thing of oop is that the object contains inputs and outputs. If there are too many attributes you can still inherit or split classes.

 These are just ideas. Thank you for your approach.

wholmgren:
Thanks @uvchik for your helpful comments.

>So the modelchain is great but what is the idea to use oop instead of procedural programming? 

I'm not sure if you only meant this as a lead-in to your next paragraph or if you wanted an answer, but... I was thinking about this too and I don't have a good answer. I mostly implemented it this way because that's what other people wanted to see. I do think that a basic model chain function should be added to this PR.

>I know that this is a matter of taste but I prefer attributes instead of using the methods like functions. If one method calculates aoi than the other method should use self.aoi. I thought the new thing of oop is that the object contains inputs and outputs.

There have been a lot of angry words spewed on this topic on the internet. I used to use a lot of side effects and no return values. My code inevitably turned into a nasty mess. Of course, it's possible to write bad code using any design pattern. 

I suppose that there's something to be said for the idea that if we're going to make both frameworks then the OO framework should be as high-level as possible so that it's most clearly separated from the functional framework.

As a compromise of sorts, the [``SingleAxisTracker.get_irradiance``](https://github.com/wholmgren/pvlib-python/blob/pvsystem/pvlib/tracking.py#L106) method currently looks for a couple attributes if they're not provided as kwargs. The idea here was that the system's attributes (surface_tilt and surface_azimuth) are probably linked to the solar position, so it might make sense to assign all of these attributes to the system during your modeling. I don't know if this is a good idea or not.

alorenzo175:
I've only briefly looked over the changes, but I'm happy with what I've seen. I think ``ModelChain`` probably still needs some work that can be deferred to the future like tracking the exact modeling steps and exporting to a file. 

I mostly agree with @uvchik  that ``ModelChain`` should set attributes in addition to returning the output. This reduces the variables one has to keep track of and it makes sense for how it will likely be used.

I also like the idea of a tracker base class.

uvchik:
I'm far away from being an expert in OOP. As I understand it one should divide the full System into its components. For a table you would have a DeskTop class, a TableLeg class and a Table class. Than you would create four TableLeg instances and a DeskTop instance and use them to create a Table instance. For the PV-System this is more abstract because you have weather conditions and locations which are also parts of your PVSimulation object. Furthermore some python "rules" say that a class should not have more than seven attributes and x methods. Otherwise you have to define smaller classes. That's how I understand it. The question is how we want to use it and what are the requirement of the users.

I would say that the most important thing is that we have a nice high level class that works and I see that `ModelChain` is already a useful class and for version 0.3 I would only do little changes.

@wholmgren, just decide what suggestions you want to implement.

* A ModelChain object for me is less speaking than a PVSimulation object or something similar. I would rename it.

* I still would inherit from the PV-System class (or from a Tracker class) as I explained in my last comment instead of adding the PVsystem as an attribute.

* Having al least `ac` and `dc` as attributes.

If we have classes with some attributes I could add a dump and a restore method to save and load the attributes of an object.

wholmgren:
Thanks for the comments and sorry for the delay in getting back to this. 

@uvchik, regarding your name and inheritance preferences... I'm going to go with what we have now but I am open to changing all of this in the future. After this is merged, please make a new issue(s) for these ideas if you'd like to continue to discuss them. I think some of this will change a lot in the future.

Some changes:

* Added a ``basic_chain`` function to the modelchain module.
* Added a ``basic_chain`` example to the Package Overview. Unfortunately, RTD is currently refusing to build documentation due to an error on their end.
* The ``ModelChain.run_model`` method now assigns all intermediate calculations as attributes of the mc object, while returning the tuple (dc, ac). I would be open to returning ``None``, but let's leave that for future discussions.
* Removed the unimplemented model chain code from early examples.
* Cleaned up modelchain documentation.
* Added a few tests (100% modelchain [coverage](https://coveralls.io/github/pvlib/pvlib-python), 100% location coverage, 93% pvsystem coverage (the omissions are not testable and/or very minor).
* Pinned the scipy version in the Appveyor builds to avoid a problem with Continuum's scipy build for Windows.

I'm going to merge this on Tuesday unless I hear objections (this time I really mean it).

uvchik:
Good work. Thank you. Yes I really think you should merge it, because it is a very helpful feature. 

By the way. I have been told to never return an attribute. An alternative way is to return the object itself to allow chaining.
```python
    self.dc = ....
    self.ac = ...
    return self

# Use the function:
mc = ModelChain(system, location, ...)
annual_energy = mc.run_model(times).ac.sum()
```

Just an idea for future development. No need to open an issue for me. I think we better wait a while and let people use and test this feature. Based on this experiences we can open new issues.

wholmgren:
Returning self is an excellent idea. I updated the code accordingly.

wholmgren:
Thanks to all for your ideas and help in developing this massive PR. I'm sure that it will take more work to get it right, so please install the new master and try it out. See you on the issue tracker...

dacoex:
Thanks to @wholmgren for being the driver and actaully implement it.
I think it offers huge potential in terms of workflow and simplification. 
Thanks again for the great work here.

uvchik:
:+1: Great feature, thank you @wholmgren .

PR #94, opened 2015-11-02
fixed imports and keys

jetheurer:
added `import datetime`; changed key from `Alpha_sc` to `alpha_sc`


wholmgren:
@jetheurer I took a very brief look at your notebooks and it's not clear to me what's different. Can you provide a description of what you've changed? git diff does not work well with IPython notebooks, unfortunately.

jetheurer:
They were very simple changes.  In one notebook I added `import datetime` and in another i changed the dictionary key from `Alpha_sc` to `alpha_sc`

But I ran all the notebooks again to be sure they worked -- so I think that's why the diff is so confusing.

wholmgren:
Thanks for clarification. That ``Alpha_sc`` issue is due to the new sam files in #80.

PR #98, opened 2015-11-05
Refactor clearsky.py

### LINKED ISSUE #95: move linke turbidity lookup out of clearsky.ineichen
I'd like to move the Linke turbidity look up code out of ``clearsky.ineichen`` and into its own function. The ``clearsky.ineichen`` API can remain the same, and it can call the new function to automatically pull in the turbidity data. Two advantages: you can easily access the turbidity data that is driving your pv models, you can cache the result so that you don't have to read and process the file over and over.

wholmgren:
This PR:
* Moves linke turbidity lookup out of ``clearsky.ineichen`` and thus closes #95.
* Improves the test suite.
* Improves the PEP8 coding style compliance.

I ran the new  ``assert_frame_equals`` tests against both the old and the new versions of ``clearsky.py``, so I'm confident that the refactoring has not changed the output of the ``ineichen`` function.

I still need to:
- [x] add some test functions for the new ``lookup_linke_turbidity`` function
- [x] add a numeric test for the ``haurwitz`` function.
- [x]  update what's new

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.2.2.txt
+++ b/docs/sphinx/source/whatsnew/v0.2.2.txt
@@ -10,11 +10,15 @@ Enhancements
 ~~~~~~~~~~~~
 
 * Adds Python 3.5 compatibility (:issue:`87`)
+* Moves the Linke turbidity lookup into ``clearsky.lookup_linke_turbidity``.
+  The API for ``clearsky.ineichen`` remains the same. (:issue:`95`)
 
 
 Bug fixes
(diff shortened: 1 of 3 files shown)


wholmgren:
Ok, I think this is now good to go, but I'll welcome any comments. I'll merge or Monday or Tuesday unless I hear objections.

Note that I also added a line to ``whatsnew`` for @jetheurer's contribution in #94. I should have remembered that needed to be done before merging #94.

cc @cwhanse

PR #112, opened 2016-01-20
Update pvsystem.py

pyElena21:
I think you need this change in tkinter module import in order to comply with Python 3. 

Cheers,
Elena


wholmgren:
Thanks for catching this and making the PR. We're going to need a try/except block to make it work on both Python 2 and 3. I think that the following should work. Can you update your PR to something like this?

```python
try:
    # python 2
    import Tkinter as tkinter
    from tkFileDialog import askopenfilename
except ImportError:
    # python 3
    import tkinter
    from tkinter.filedialog import askopenfilename

tkinter.Tk().withdraw()
csvdata = askopenfilename() 
```

wholmgren:
@pyElena21 Thanks for updating the code.

pyElena21:
np!

Issue #120, opened 2016-02-01
doc: update to version 3


dacoex:
In conjunction with [Location demotion and PVSystem/LocalizedPVSystem/ModelChain addition](https://github.com/pvlib/pvlib-python/pull/93#issuecomment-178206899) but now as separate issue:

>     The Ipython Directive could be used.
> I'm using it in the new docs, so I don't understand why you're saying this.

I think the original intention with notebooks was to offer quick interactive exploration.
But now we find it hard to keep them updated and tested.
Below a way which might help 

- [ ]  integrate the notebooks into testing, tools: [RunNotebook](https://github.com/ngoldbaum/RunNotebook), see also [[yt-dev] Documentation and Testing](http://lists.spacepope.org/pipermail/yt-dev-spacepope.org/2016-February/019902.html)

- [ ]  update and enhance the documentation to include the features of #93 

- [ ]  add system modelling tutorial, as in #84 but adapted to the changes in #93

wholmgren:
First, we need a new name for this issue. It sounds like what you really want is something along the lines of "Port notebook tutorials to rst with doctests". Whether or not it makes it into version 0.3 (not 3) is another matter.

Regarding the notebooks themselves, I am much more inclined to move the notebooks to their own repo with different administrators than to add CI testing for them here. Also, I'm am very, very, very reluctant to add dependencies on small projects such as RunNotebook. Have you used nbconverts --to-rst flag? I have not, but that seems like the place to start.

dacoex:
>It sounds like what you really want is something along the lines

Yes.
Do you agree with the idea of porting to rst & doctest?

I would then go ahead with this by the following activities:

* propose a structure on how existing tutorials will be included in the ToC
* porting

>Whether or not it makes it into version 0.3 (not 3) is another matter.

OK.

>Regarding the notebooks themselves, I am much more inclined to move the notebooks to their >own repo with different administrators than to add CI testing for them here. 

What a bout the following structure:

1. Essential tuturials that demonstrate the functionality of the package will be ported to rst
2. More application specific tutorials will be moved to the separate repo

I would include all existing notebooks and #84 in item 1 above.

Number 2 could be notebooks that :

* show the use of a specific data set
* demonstrate applications for a specific location or technology
* relate to a specific article / publication
* explain a complete application of a modelchain

Another question about documentation:

* To which extend shall modelchains be documented?
* To which exentd do we want to include cross-comparison with other software?

Looking forward on your feedback.

[...]

PR #121, opened 2016-02-10
Restructure documentation, port tutorials

dacoex:
This is the code for #120. Based on the pvsystem branch by @wholmgren 

The last commit now has a potential ToC as a list.

I think discussing using the list is easier and less effort.
Let's agree ToC level by ToC level.

To revewi, please also look at the RTD version:
[Contents (potential new ToC)](http://pvlib-python-dacoex.readthedocs.org/en/pvsystem_tutport/index.html#contents-potential-new-toc)


docs/sphinx/source/index.rst
```diff
@@ -60,11 +60,54 @@ Contents
...
+
+* What’s New
+* Overview: Why pvlib?
+    * Introduction
+    * Modeling paradigms
+    * User extensions
+    * What it is not
+    * Comparison with PVLIB_MATLAB
+* Installation
+* Examples
```
wholmgren: I'd remove the top level "Examples".

docs/sphinx/source/index.rst
```diff
@@ -60,11 +60,54 @@ Contents
...
+        * irradiance.py tutorial 
+    * Specific datasets
+        * TMY tutorial 
+        * TMY data and diffuse irradiance model 
+        * TMY to Power Tutorial 
+    * System Modelling
+        * pvsystem tutorial
+        * Tracking
+        * Modeling multiple systems with system losses
+    * Validations
```
wholmgren: I don't understand the "Validations" title. Are you planning to compare to real data? That sounds like a good idea, but the subheadings are not specific to data validation.

docs/sphinx/source/index.rst
dacoex: Was mimicked after
http://xarray.pydata.org/en/stable/index.html
http://xarray.pydata.org/en/stable/examples.html

docs/sphinx/source/index.rst
dacoex: Could be "Validations & Comparisons"

In the "SAPM" notebook, the different models are compared (= validated against each other).
This a strength of pvlib as compared to other tools albeit SAM & PVSyst can also do this...

docs/sphinx/source/index.rst
wholmgren: I see. I don't have a strong preference here.

docs/sphinx/source/index.rst
wholmgren: Ah, no, I would not use "validation" unless there was data.

dacoex: So it will be "Validations & Comparisons"?

docs/sphinx/source/index.rst
dacoex: I suggest to keep the top level.
This helps to reduce `maxdepth` to 1 once there are many sub-pages.

docs/sphinx/source/index.rst
wholmgren: No, I suggest removing the section entirely and putting the subitems under system modeling. If you really want the section then call it something like "SAPM and Single Diode Comparison"

dacoex: OK.

dacoex:
So it seems that we are set with the ToC for now?

wholmgren:
Yes, with the understanding that we'll probably want to change some of it once we start to see the content.

dacoex:
Sure. Seems stepwise is a better approach....

dacoex:
@wholmgren likly to pick this one up step by step.
Would you still agree to port the essential notebook based tutorials to python code / doctest so that they can be tested through an api restructuring as it happened with [v3.0](http://pvlib-python.readthedocs.io/en/latest/whatsnew.html#v0-3-0-march-21-2016)?

wholmgren:
The documentation has grown since this conversation started. I would like to see more documentation pages similar to our Time and time zones, Clearsky, and Forecasting pages. Transposition and solar position pages would be relatively straightforward to write. I think a good ModelChain page is very important, but more work. I recommend closing this PR and doing these (or something like them that you care more about) one PR at a time.

dacoex:
Thanks for responding.

Currently, my preference is with the pvsystem tutorial initiated by @jforbess

>I would like to see more documentation pages similar to our Time and time zones, Clearsky, and >Forecasting pages. 

This is what I was planning to convert the existing notebook into.

May question now:
**How do we ensure that the code in the documentation always up-to date?
Is it included in the tests?**

This shall be there to ensure that we do not damage the docs when refractoring the api of the core lib.

e.g. there are soem error in http://pvlib-python.readthedocs.io/en/latest/forecasts.html

_Some guidance would be appreciated before writing new pages._

Regarding ToC & structure, I see that in future there may be the followings technical pages:

```
- Fundamentals & Basics
   |---- Time and time zones
   |---- Clearsky
   |---- solar position
   |---- Transposition 
   |---- Input data
           |---- US TMY
           |---- Forecast & weather model (the whole section of Accessing Forecast Data to Weather Models deal with getting data)
           |---- other providers
           |---- local measurement & SCADA data (link to https://github.com/sandialabs/pecos )
- Modelling & Simulations
   |---- Trackers
   |---- Complete PV system
   |---- Sensitivity studies
   |---- Forecasting 
   |---- Cross-comparison & validation with other tools
   |---- etc.
- Analysis & Presentation
   |---- Statistisc ( ... should create a feature request issue because nothing there yet)
   |---- Plotting 
   |---- Reporting ( ... pvlib could be used to create a report like: 
                                http://solargis.com/assets/sample/SolarGIS-pvPlanner-sampleReport.xls
                                http://solargis.com/assets/sample/SolarGIS-pvPlanner-sampleReport.pdf) 
```

wholmgren:
I've never written any doctests but I think you can do them in ipython directive. It will be a lot of work, so good luck. As I said before, go ahead and write up one new page in a new pr. 

I've explained the error on the forecasting page multiple times.

dacoex:
>I've never written any doctests but I think you can do them in ipython directive. It will be a lot of >work, >so good luck. 

As alternative, we could use a separate code file being test & then use it as listing in the Spinx docs.

wholmgren:
http://matplotlib.org/sampledoc/ipython_directive.html

```
.. ipython::

   In [1]: x = 'hello world'

   # this will raise an error if the ipython output is different
   @doctest
   In [2]: x.upper()
   Out[2]: 'HELLO WORLD'
```

wholmgren:
There are some useful ideas in the discussion above, but the way to make progress at this point is starting and finishing one small PR before moving on to the next.

PR #141, opened 2016-03-23
Resolve #140 snlinverter night tare power issue

### LINKED ISSUE #140: Night time power issue for some microinverters
I am trying to use pvlib to compare some systems with microinverters versus string inverters. I have noticed that for many microinverters I am getting a positive AC power output when the DC power is 0. For example, using the inverter:
    `Enphase_Energy__M250_60_2LL_S2x___ZC____NA__208V_208V__CEC_2013_`

With 0V and 0W:
    `pvsystem.snlinverter(inverters.Enphase_Energy__M250_60_2LL_S2x___ZC____NA__208V_208V__CEC_2013_, 0, 0)`

I get:
    `1.7860471814922816`

I would expect to see a negative result with a magnitude of Pnt, which is what I see for most of the other string/central inverters I have tried. This is causing issues for my comparisons since my night time values for the microinverter are adding to the total power production, while the string and central inverters are being reduced. This can add up to quite a difference when accumulating over a year.

If I try the same module and same inverter in SAM, the "Power generated by system" is showing the negative Pnt value I expect for the inverter when the DC values are 0.

The code in pvsystem.py for this situation is
    `ac_power[ac_power < Pso] = - 1.0 * abs(Pnt)`

Shouldn't the comparison be between the DC power and Pso (p_dc < Pso)?

ejmiller2:
Fixing #140 by changing the start power (Pso) comparison to use the DC input power (p_dc) instead of the calculated AC power (ac_power).

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.3.1.txt
+++ b/docs/sphinx/source/whatsnew/v0.3.1.txt
@@ -15,8 +15,10 @@ Enhancements
 Bug fixes
 ~~~~~~~~~
 
+* Fixes snlinverter Pnt problem (:issue:`140`)
 
 
 Contributors
 ~~~~~~~~~~~~
 
+* ejmiller2
--- a/docs/sphinx/source/whatsnew/v0.3.1.txt
+++ b/docs/sphinx/source/whatsnew/v0.3.1.txt
@@ -15,7 +15,12 @@ Enhancements
 Bug fixes
 ~~~~~~~~~
 
-* Fixes night tare issue in snlinverter. When the DC input power (p_dc) to an inverter is below the inversion startup power (Ps0), the model should set the AC output (ac_power) to the night tare value (Pnt). The night tare power indicates the power consumed by the inverter to sense PV array voltage. The model was erroneously comparing Ps0 with the AC output power (ac_power), rather than the DC input power (p_dc).
+* Fixes night tare issue in snlinverter. When the DC input power (p_dc)
+  to an inverter is below the inversion startup power (Ps0), the model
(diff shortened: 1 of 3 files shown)


wholmgren:
Great. One easy request and one slightly harder request:

Easy: Add a brief note to the "Bug fixes" section of the ``pvlib/docs/sphinx/source/whatsnew/v0.3.1.txt`` file and add your name or username to the "Contributors" list. 

Slightly harder: Add a new test to ``pvlib/test/test_pvsystem.py`` that requires this new behavior in order for it to pass. You basically want a copy of [``test_snlinverter``](https://github.com/pvlib/pvlib-python/blob/master/pvlib/test/test_pvsystem.py#L352) but with one of the inverters and the input parameters that were giving you the wrong results. 

Thanks!

ejmiller2:
I'll give it a shot a bit later today.

ejmiller2:
All done (I think).

docs/sphinx/source/whatsnew/v0.3.1.txt
```diff
@@ -15,8 +15,10 @@ Enhancements
 Bug fixes
 ~~~~~~~~~
 
+* Fixes snlinverter Pnt problem (:issue:`140`)
```
wholmgren: I know I asked for brief, but maybe something more like "Fixes night tare issue in snlinverter. The night tare was erroneously activated when the AC output power, rather than the DC input power, was less than the inverter's self-consumption" or similar. I had to look up Pnt, so I'm sure others would need to as well.

wholmgren:
@ejmiller2 thank you for fixing this.

Issue #83, opened 2015-07-27
pvsystem.py - no continuous IV curve data


cbirkj:
The singlediode function, located in pvsystem.py, does not have the capability to return continuous current and voltage values.  I have modified the code to include a loop:

si = np.ones((NumPoints,1))
sv = np.linspace(-1,1,NumPoints)
if NumPoints >= 2:
    	for i in range(NumPoints):
    		V[i,:] = Voc.values * sv[i]
    		I[i,:] = I_from_V(Rsh=Rsh, Rs=Rs, nNsVth=nNsVth, V=V[i,:], I0=I0, IL=IL)

This may not be the most efficient way but it worked.

- Birk

wholmgren:
Thanks, Birk. 

I'm not sure how to best construct the API for this. I removed the ``kwargs``, and thus ``NumPoints``, from the from the ``singlediode`` function in the 0.2 API refactor. One approach is to use two different functions, one for labelled points and one for arbitrary points. What are you returning from the function? A dataframe and, optionally, a series? You could just embed the whole function in a comment using [triple ticks and the python declaration](https://help.github.com/articles/github-flavored-markdown/#syntax-highlighting).

Also, I think that ``i_from_v`` will work with vector inputs, and this will be a lot faster than iterating over each point.

What do others think?

wholmgren:
Here's my code to make IV curves. I'm still not sure if any of this should be included in pvlib or not. I think we should at least add something similar to this to the documentation. Maybe someone else has a better way to do this.

```python

def fivepoints_to_frame(pnt):
    """
    Converts a 1 dimensional, dict-like singlediode or sapm result
    to a 5 row DataFrame with columns current and voltage.
    Users can iterate over the rows of a multidimensional
    singlediode or sapm result, if necessary.
    """
    ivframe = {'i_sc': (pnt['i_sc'], 0),
               'p_mp': (pnt['i_mp'], pnt['v_mp']),
               'i_x': (pnt['i_x'], 0.5*pnt['v_oc']),
               'i_xx': (pnt['i_xx'], 0.5*(pnt['v_oc']+pnt['v_mp'])),
               'v_oc': (0, pnt['v_oc'])}
    ivframe = pd.DataFrame(ivframe, index=['current', 'voltage']).T
    ivframe = ivframe.sort_values(by='voltage')
    
    return ivframe

resistance_shunt = 16
resistance_series = 0.094
nNsVth = 0.473
saturation_current = 1.943e-09
[...]

PR #148, opened 2016-04-14
implement simplified solis clear sky model

wholmgren:
Here's an implementation of Ineichen's "simplified Solis" clear sky model (see the references below). Though I've only just implemented it and have not tested it against any data or the existing ``ineichen`` algorithm, it seems like it could be a useful addition to pvlib. I like the fact that it is simple to program, continuous, and has good limiting behavior. I could easily be missing something, though, since my experience with it is currently limited to the plots below. Does anyone else have any experience with this model? Maybe someone at Sandia, @cwhanse?

For this to be merged we would need to:

- [x] agree that it is a useful addition
- [x] think about the function's API (what is required, what is optional? I think that the existing ``ineichen`` function does too much.).
- [x] consider adding some helping functions such as aod converters. These can be added separately.
- [x] hook it up the ``Location.get_clearsky`` method
- [x] fill in the empty tests

[1] P. Ineichen, "A broadband simplified version of the Solis clear sky model," Solar Energy, 82, 758-762 (2008).

[2] P. Ineichen, "Validation of models that estimate the clear sky global and beam solar irradiance," Solar Energy, 132, 332-344 (2016).

aod0 1_pw0 5
aod0 01_pw0 5
aod0 1_pw5
aod0 01_pw5
dhi
dni
ghi

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.3.3.txt
+++ b/docs/sphinx/source/whatsnew/v0.3.3.txt
@@ -1,6 +1,6 @@
 .. _whatsnew_0330:
 
-v0.3.3 (June xx, 2016)
+v0.3.3 (June 15, 2016)
 -----------------------
 
 This is a minor release from 0.3.2.
@@ -32,6 +32,7 @@ Enhancements
   (:issue:`190`)
 * Improve speed of ``singlediode`` function by using ``v_from_i`` to
   determine ``v_oc``. Speed is ~2x faster. (:issue:`190`)
+* Adds the Simplified Solis clear sky model. (:issue:`148`)
 
 
 Bug fixes
--- a/pvlib/clearsky.py
+++ b/pvlib/clearsky.py
@@ -9,6 +9,7 @@ import logging
 logger = logging.getLogger('pvlib')
 
 import os
+from collections import OrderedDict
 
 import numpy as np
 import pandas as pd
@@ -330,7 +331,7 @@ def _linearly_scale(inputmatrix, inputmin, inputmax, outputmin, outputmax):
 
 
 def simplified_solis(apparent_elevation, aod700=0.1, precipitable_water=1.,
-                     pressure=101325., dni_extra=1364., return_raw=False):
+                     pressure=101325., dni_extra=1364.):
     """
     Calculate the clear sky GHI, DNI, and DHI according to the
     simplified Solis model [1]_.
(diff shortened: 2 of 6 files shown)


cwhanse:
We haven’t implemented either the Solis or Pierre’s simplied Solis models.  I agree that it is a useful addition, and would like to port it to Matlab when I can get to it.

Cliff

wholmgren:
I think this is nearly ready to merge. 

@alorenzo175 this may be useful for your work, so you might want to review the code before it goes in.

alorenzo175:
Looks like all the equations follow the paper. Does it really fail if pw < 0.2? Why put that restriction but leave off the AOD and pressure restrictions?

wholmgren:
Thanks for double checking. 

You can see the algorithm fail for very small PW in the lowest sections of the contour plots in my initial comment. This is because PW is parameterized in powers of ``ln(w)``, and ``ln(w)`` goes to ``-inf`` as ``w`` goes to 0. The AOD parameterization only contains powers of AOD, so that has good lower limit behavior. The ``ln(p/p0)`` pressure parameterization should work for any reasonable user input. I think that it's reasonable for users to expect the algorithm to work for ``AOD --> 0`` and ``PW --> 0`` (clean, dry atmosphere), but not ``p --> 0`` (high flying solar drones?).

I'm less worried about the upper limits since I don't think that they will be reached using real data and because users should naturally be skeptical of the upper limits. 

I just noticed that Fig 3 of Ineichen 2008 shows PW from 0.1 to 10 cm, so maybe 0.1 is a better choice for the limit. I'll try to reproduce that figure and then decide.

wholmgren:
I might as well share my attempt at reproducing the figure. Unfortunately, the paper does not give enough information to reproduce it with much accuracy, but the differences are large enough that I want to take more time to double check the work against Ineichen's excel tool:

http://www.unige.ch/energie/fr/equipe/ineichen/solis-tool/

The excel tool is "an improved version" so that comparison may or may not turn out right either. 

tau_3d

wholmgren:
I'm going to merge this once the tests pass. Here's a gist with a notebook that demonstrates the usage: https://gist.github.com/wholmgren/fb5043a7d3cd111cf38ce7a7b7779550

I'm fairly confident that the algorithm is implemented as written in the papers above, and I've decided that makes it good enough to merge. However, be aware that there are some discrepancies with the results in the Excel spreadsheet. I'd like to revisit this later, and I hope that others might be interested enough to do their own validations of the implementation once it's merged and released.

The discussion in #83 led me to change the API. Here's how it works now: If you pass a Series into the function, you'll get back a DataFrame, thus making it easy to handle time series data with pandas. If you pass scalars or arrays, you'll get back an OrderedDict of scalars or arrays, thus making it easy to use the function in more complex code. In either case you can use e.g. result['ghi'] to get the GHI, but further indexing/slicing is up to the user. 

@mikofski and @MLEEFS might be interested in trying this out when I release 0.3.3 later today.

Issue #62, opened 2015-04-30
Add examples to dostrings


bmu:
We should decide whether we want to add examples to the docstrings of the functions or not.
While (I think) examples are not included in every pandas object, they are in numpy. [numpydoc](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard) has an *optional* section for examples.

I would like to include examples and also I would like to use doctests.
For the moment we have some examples, but some doctests fail 
(see output of `nosetests -v -w pvlib --with-coverage --cover-package=pvlib --with-doctest`).
I once had a discussion with pandas developers (see pydata/pandas#3439) and they didn't see their examples as tests. 
However for pvlib, `doctest` may have an additional benefit for testing purposes and also it tests if our documentation is up to date with the state of the objects we document.

wholmgren:
We definitely need more examples in the docs. I don't know if we need an example for absolutely everything, though. In any case, we probably will want to change the documentation formatting to 1 page per function.

I don't have any experience using ``doctest``, but I'm ok with trying it out. I would say that we should not rely on ``doctest`` for anything other than ensuring that the docs are up to date with the library, and that anything that is in a doctest should be in the main tests.

mikofski:
Any decision on this?

wholmgren:
@mikofski I support adding brief examples to the function docstrings where they would be useful. Personally, I find documentation pages structured like [ModelChain](https://pvlib-python.readthedocs.io/en/latest/modelchain.html) and [Clearsky](https://pvlib-python.readthedocs.io/en/latest/clearsky.html) to be more useful, so that's where I've focused my documentation efforts. But I encourage people that want small examples for each function to take on the effort. So I lean toward leaving this generic issue open. What do you think?

abhinav9414:
can I try to do this sir.

cwhanse:
> can I try to do this sir.

Certainly. Are you somewhat familiar with PV system modeling? That would help with forming examples.

It would be best to take this up in small pieces. Maybe the file `iam.py` is a place to begin.

abhinav9414:
ok ill start with iam.py file thanks for the boast.

dgowdaan-cmyk:
[...]

PR #149, opened 2016-04-15
use sphinx autosummary to make one doc page per function

wholmgren:
I'd like to change to API documentation format to have:
1. a separate summary page for each module
2. a separate detailed page for each function

see the new documentation here:

http://wholmgren-pvlib-python-new.readthedocs.org/en/doc-reorg2/modules.html

I think it's a lot easier to navigate this. Separate pages for each function also makes it reasonable to add example documentation for each function (like the matlab library already has) and was discussed in #62.


alorenzo175:
Looks good except for the scrollbars on the tables. You might try some custom CSS so that the descriptions wrap like in https://github.com/qtile/qtile/commit/3eadbbf20adf975ed2ac502f59a6a7d854b7de91.

alorenzo175:
Also, is there a way to have the +- boxes in the menu on the left for the autosummary items (like in Time and time zones)?

wholmgren:
I applied that scroll bar patch. Nice find. The some of the tables look a little messy, but I prefer this to scrolling. 

I noticed earlier that the +/- symbols were not being created, but I couldn't figure out how to fix it.

autosummary is creating new pages for imported functions and classes. Not sure if this due to a sphinx bug or if I'm doing something wrong.

wholmgren:
I forgot to add that I am fairly sure that the duplicated pages are due to a sphinx bug, and I'm waiting to merge this until they fix it.

wholmgren:
closed in favor of #258.

PR #150, opened 2016-04-16
added versioneer for versioning

alorenzo175:
[Versioneer](https://github.com/warner/python-versioneer) is nice because you never have to (or forget to) update the ``__version__`` string. To make a new release version, you just add a git tag. When changes are made after a tag, the git commit is indicated in the version, so it might be something like ``0.3.0+17.g39a420e`` indicating the code is based on git commit ``39a420e`` which is 17 commits after the ``0.3.0`` tag.

The versioneer code is kept in _version.py so that version.py can be left alone for backwards compatibility. This seems kind of redundant, but might help a few people.


wholmgren:
Thanks, I've wanted this a few times recently. Add a note and your name to the 0.3.1 whatsnew.

wholmgren:
For future reference, also see also commit 73e1df1bc

PR #151, opened 2016-04-18
refactor ModelChain. add SingleDiode modelchain

wholmgren:
I refactored ``ModelChain`` into a parent ``ModelChain`` class with subclasses ``SAPM`` and ``SingleDiode``. 

In doing so, I moved some solar position, irradiance, and weather prep work into ``ModelChain.prepare_inputs`` and this method is called during ``run_model``.


wholmgren:
closed in favor of #179.

PR #164, opened 2016-05-10
convert indexing arrays to int to avoid future deprecation

alorenzo175:
This PR just gets rid of some deprecation warnings using numpy 1.11 from non-integer indices. The default for the bins in irradiance.py is now 0 (versus the old NaN), which may or may not be a problem. Maybe the default should be something else?

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.3.3.txt
+++ b/docs/sphinx/source/whatsnew/v0.3.3.txt
@@ -35,7 +35,8 @@ Bug fixes
 * Fix snlinverter and singlediode documentation. They incorrectly said that
   inverter/module must be a DataFrame, when in reality they can be any
   dict-like object. (:issue:`157`)
-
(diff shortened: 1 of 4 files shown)


wholmgren:
The 0 will turn into a -1, which will be a valid indexer and as you said may be a problem. Alternatively, we could use an int that is two larger than the number of elements in each array, so that invalid values ultimately result in an IndexError. The problem here is that I think this means that nans in the input data will result in IndexErrors, which is also unfortunate. I suppose we can add nan filters, but that's one more complication to test.

I think that we can use defaults of ``pd.Series(np.nan, index=times, dtype=np.object)``, but I'm not sure what the performance hit would be. It's probably minor for most applications.

Annoying that numpy doesn't support integer nans.

wholmgren:
I'd like to resolve this one way or another, but I don't care what we do. Thoughts?

PR #166, opened 2016-05-12
Implement the Erbs model

wholmgren:
Here's an implementation of the Erbs model based on ``pvl_erbs.m``. I was going to demonstrate this during the Santa Clara workshop, but didn't plan the time very well. Sorry.

It's nearly a 1:1 port of the MATLAB function.

As currently written, the user must supply the day of the year, but we should consider that it instead be inferred from the index of the GHI Series. A potential downside of this is that it would require that ghi is a Series, instead of a numpy array. @mikofski might have something to say about that.

I'm open to making this change now, or merging as-is.

The PR also starts a 0.3.3 whatsnew file.

A gist with a simple example... 

https://gist.github.com/wholmgren/21c9a60547a3f240aeb1a1919f8829b8

Here's the diff minus the unimportant whitespace changes...

https://github.com/pvlib/pvlib-python/pull/166/files?w=0

### DIFF
--- a/docs/sphinx/source/whatsnew.rst
+++ b/docs/sphinx/source/whatsnew.rst
@@ -6,6 +6,7 @@ What's New
 
 These are new features and improvements of note in each release.
 
+.. include:: whatsnew/v0.3.3.txt
 .. include:: whatsnew/v0.3.2.txt
 .. include:: whatsnew/v0.3.1.txt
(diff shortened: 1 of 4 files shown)


wholmgren:
I'm going to merge as-is but, as always, make an issue/PR if you want to see something different.

Issue #159, opened 2016-05-06
add a function that multiplies voltages and currents for arrays


wholmgren:
I'd like to consider adding a function to ``pvsystem.py`` that would multiply the voltages and currents returned by ``pvsystem.sapm`` and ``pvsystem.singlediode``. Of course, the motivation for this is to make it easier to calculate array parameters.

```python
def multiply_voltage_current(data, voltage_multiplier=1, current_multiplier=1): 
    # as written, only works with a DataFrame
    # could make it work with a dict, but it would be more verbose
    data = data.copy()
    voltages = ['v_mp', 'v_oc']
    currents = ['i_mp' ,'i_x', 'i_xx', 'i_sc']
    data[voltages] *= voltage_multiplier
    data[currents] *= current_multiplier
    data['p_mp'] *= voltage_multiplier * current_multiplier
    
    return data
    
    
# within the PVSystem class...

    def multiply_voltage_current(self, data):
     
        return multiply_voltage_current(
            data, voltage_multiplier=self.series_modules,
            current_multiplier=self.parallel_modules)
```

Maybe the function can also be reused for some system losses.

wholmgren:
Closed by #171.

PR #171, opened 2016-05-13
implement voltage, current, power scaling function and method

wholmgren:
This PR implements a simple scaling function and hooks it up to PVSystem and ModelChain. I'm open to changing things.

I'd like to merge early next week.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.3.3.txt
+++ b/docs/sphinx/source/whatsnew/v0.3.3.txt
@@ -10,6 +10,8 @@ Enhancements
 ~~~~~~~~~~~~
 
 * Adds the Erbs model. (:issue:`2`)
+* Adds the ``scale_voltage_current_power`` function and ``PVSystem`` method
+  to support simple array modeling. (:issue:`159`)
 
 
 Bug fixes
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -342,6 +342,8 @@ class ModelChain(object):
                                    self.airmass['airmass_absolute'],
                                    self.aoi)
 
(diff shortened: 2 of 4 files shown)


wholmgren:
I'll merge tomorrow unless someone comments, which you should, because I'm confident that there's a non-0 number of users that will find this useful if done right and annoying if done wrong. 

original issue: #159.

jforbess:
It looks good to me. I was trying to decide if there was any value in calculating voltage and current multipliers separately, but I think there is not, because any adjustments on either that are needed (electrical shading) can happen before the dc parameters are fed to the inverter. 

I appreciate the clarification you provided over the definiton of series_modules and parallel_modules, but one recommended change is to say that parallel_modules are the number of strings in parallel, not the number of *modules or* strings in parallel. I think that wording just confuses. The parameter name is confusing in general, but I've made my peace with it. :)

wholmgren:
Thanks for the feedback. One quick proposal and then a longer discussion...

Quick: What if I change the individual parameter descriptions to something like "see system topology discussion above"?

Long: I agree that the parameter names and descriptions are confusing. Now is a good time to change them since they didn't actually do anything before. Well, they didn't do anything useful in pvlib code, but maybe users did something with them in their own code. In any case, they're not my creation and I have no attachment to them. I would support something like ``series_modules --> modules_per_string`` and ``parallel_modules --> strings``. We can also merge this and then address the name issue separately.

jforbess:
Quick: Sure, that's fine.

Long: If no one else complains, I would definitely prefer
modules_per_string and strings, for clarity. I don't mind series_modules,
but parallel_modules is so confusing and annoyingly wrong in concept.

Issue #142, opened 2016-04-07
add __repr__ methods to classes


wholmgren:
We should add [``__repr__``](https://docs.python.org/3/reference/datamodel.html#object.__repr__) methods to all of pvlib's classes.  ``Location`` implements ``__str__``, but this should probably be changed to ``__repr__``.

wholmgren:
closed by #174.

PR #145, opened 2016-04-12
Issue#144-fixed the azimuth calculation of rotated PV panel

mayudong:
The results are consistent with PVsyst.

Also the tests are updated to cover this issue

PR #174, opened 2016-05-16
"add __repr__() function to class PVSystem, LocalizedPVSystem, ModelChain, SingleAxisTracker, Location and add test for same"

JohannesOos:
According to issue #142

### DIFF
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -265,6 +265,14 @@ class ModelChain(object):
 
         # calls setter
         self.orientation_strategy = orientation_strategy
+        
+    def __repr__(self):
+        return ('ModelChain for: '+ str(self.system) + 
+                ' orientation_startegy: ' + str(self.orientation_strategy) +
+                ' clearsky_model: ' + str(self.clearsky_model) +
+                'transposition_model: ' + str(self.transposition_model) + 
+                ' solar_position_method: ' + str(self.solar_position_method) +
+                'airmass_model: ' + str(self.airmass_model))
 
     @property
     def orientation_strategy(self):
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -141,7 +141,10 @@ class PVSystem(object):
         super(PVSystem, self).__init__(**kwargs)
         
     def __repr__(self):
-        return 'PVSystem with tilt:' + str(self.surface_tilt) + ' and azimuth: ' + str(self.surface_azimuth)
+        return ('PVSystem with tilt:' + str(self.surface_tilt) +
+                ' and azimuth: ' + str(self.surface_azimuth) +
+                ' with Module: ' + str(self.module) +
+                ' and Inverter: ' + str(self.inverter))
 
     def get_aoi(self, solar_zenith, solar_azimuth):
         """Get the angle of incidence on the system.
@@ -439,7 +442,9 @@ class LocalizedPVSystem(PVSystem, Location):
 
         # get and combine attributes from the pvsystem and/or location
         # with the rest of the kwargs
-
+        
+        self.location = location
+        
         if pvsystem is not None:
             pv_dict = pvsystem.__dict__
         else:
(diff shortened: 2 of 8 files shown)


wholmgren:
Thanks for submitting this. I'd lean towards including more attributes for PVSystem and add some for LocalizedPVSystem, but I don't have a specific recommendation for which to include and exclude. 

Also, please keep the line length to 79 or fewer characters.

JohannesOos:
Thanks for advise.
Updated length of lines.
Updated more output and added self.location for LocalizedPVSystem

wholmgren:
Please consolidate this PR and #175. The PR will automatically be updated when you push new commits to GitHub, so there's no need to create a second one for the same topic.

We'll need tests similar to this one:

https://github.com/pvlib/pvlib-python/blob/master/pvlib/test/test_location.py#L37

JohannesOos:
Thanks very much for the advise. Am new to the github workflow, so thanks for the patience.
Closed #175 and put changed on this thread.

JohannesOos:
Regarding testing:
No experience with it, so will leave it open.

pvlib/modelchain.py
```diff
@@ -8,7 +8,7 @@
 
 import pandas as pd
 
-from pvlib import solarposition, pvsystem, clearsky, atmosphere
+from pvlib import solarposition, pvsystem, clearsky, atmosphere, location
```
wholmgren: Is this new import needed?

pvlib/pvsystem.py
```diff
@@ -400,7 +406,9 @@ def __init__(self, pvsystem=None, location=None, **kwargs):
 
         # get and combine attributes from the pvsystem and/or location
         # with the rest of the kwargs
-
+        
+        self.location = location
```
wholmgren: I think I know why you're adding this ``location`` attribute (to avoid copy/paste location representation code), but I don't think we should do it. The input ``location``'s attributes are reassigned as attributes of the new object. Adding a separate ``location`` attribute duplicates the information and potentially causes confusion. Also, changes to the localized object's location-like attributes would not be reflected in the ``__repr__`` return.

pvlib/tracking.py
```diff
@@ -143,6 +148,8 @@ def __init__(self, pvsystem=None, location=None, **kwargs):
         # get and combine attributes from the pvsystem and/or location
         # with the rest of the kwargs
 
+        self.location = location
```
wholmgren: same as above

pvlib/modelchain.py
```diff
@@ -262,6 +262,14 @@ def __init__(self, system, location,
 
         # calls setter
         self.orientation_strategy = orientation_strategy
+        
+    def __repr__(self):
```
wholmgren: nice. edit the PR name to reflect this addition.

wholmgren:
Running the tests should be as easy as: conda or pip install ``nose``, then run ``nosetests pvlib -v`` from your pvlib-python directory. I think that you can create tests for your new methods fairly easily by following the simple test example linked to above. I encourage you to try it.

JohannesOos:
Thanks for the guidance on the testing - I did pattern matching from the example. The test worked on my machine, not sure though.
added the tests for ModelChain.__repr__(), Location.__repr__(), PVSystem.__repr__() and LocalizedPVSystem.__repr__().

pvlib/modelchain.py
JohannesOos: Done. Thanks.

wholmgren:
The tests look good, though we still need a couple for the trackers. 

Did you see the comments about the unnecessary location import and undesirable location attribute? They should show up in-line here, otherwise you can look at the [files changed tab](https://github.com/pvlib/pvlib-python/pull/174/files).

JohannesOos:
Added the tests of __repr__() for the trackers

JohannesOos:
Location in trackers adjusted - did not realize the override before. Thanks!

wholmgren:
You'll need to make the same changes to your ``LocalizedPVSystem``: remove ``self.location = location`` and add the individual attributes to the ``repr``. You also added a ``location`` import to ``modelchain`` that isn't used and should be removed.

Is there any reason to keep the ``Location.__str__`` method?

Finally, you're going to need to rebase your commits on the latest version pvlib and resolve a merge conflict along the way. See this [comment](https://github.com/pvlib/pvlib-python/pull/145#issuecomment-209056356) and google for more.

pvlib/modelchain.py
JohannesOos: You are right, there is no need for that import

JohannesOos:
Did the rebase, but think I killed some of my changed on the way. So will re-commit the modelchain and location file

wholmgren:
@JohannesOos we're close... I'm still hoping to get rid of the ``self.location`` assignment in ``LocalizedPVSystem``. Also, did you see that one of your tests is failing?

JohannesOos:
Adjusted the tests and replaced self.location with the lat and lon attributes inherited from location. 
Thanks for the continuous advise.

wholmgren:
Looks good to me. I'll merge tomorrow unless anyone objects.

JohannesOos:
Cool, thanks for the help on getting me up to speed with github

wholmgren:
Thanks @JohannesOos!

PR #175, opened 2016-05-17
#142 __repr__ for tracking classes

JohannesOos:


JohannesOos:
consolidated with #174 according to advise

PR #190, opened 2016-06-13
add v_from_i, improve singlediode speed by 2x

wholmgren:
I implemented a ``v_from_i`` function, similar to ``i_from_v``, by following Jain and Kapoor (2004) and PVLIB MATLAB pvl_singlediode.m. This lets us replace the ``singlediode`` function's numerical solver for v_oc with an analytic solution. Thus, the ``singlediode`` function is about 2x faster.

The analytic solution differs from the numerical solution by a little more than 1 part in 1000, which is similar to the tolerance of the ``_golden_sect_DataFrame`` numerical solver.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.3.3.txt
+++ b/docs/sphinx/source/whatsnew/v0.3.3.txt
@@ -24,6 +24,10 @@ Enhancements
   (:issue:`169`)
 * Add ``__repr__`` method to PVSystem, LocalizedPVSystem, ModelChain,
   SingleAxisTracker, Location. (:issue:`142`)
+* Add ``v_from_i`` function for solving the single diode model.
+  (:issue:`190`)
+* Improve speed of ``singlediode`` function by using ``v_from_i`` to
+  determine ``v_oc``. Speed is ~2x faster. (:issue:`190`)
 
 
 Bug fixes
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1405,6 +1405,10 @@ def singlediode(module, photocurrent, saturation_current,
               'i_0': saturation_current,
               'i_l': photocurrent}
 
+    # to do:
+    # replace with a call to scipy.optimize or minimize(_dpower_dcurrent)
+    # and then calculate v_mp, and then calculate p_mp = i_mp*v_mp
+    # might need a partial(_dpower_dcurrent) for all params but I
     p_mp, v_mp = _golden_sect_DataFrame(params, 0, module['V_oc_ref']*1.14,
                                         _pwr_optfcn)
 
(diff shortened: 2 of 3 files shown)


cwhanse:
Will,
 For v_from_i be sure to test for a module with a high value of Voc, e.g., Rsh = 500, Rs = 10, nNsVth = 4.06 (n = 1.4, Ns = 116, Vth = 0.025), Io = 6e-10, Iph = 1.2.  These values should give Voc around 90 and are representative of a CdTe module.

The numerical issue is to handle overflow in the lambertw function.  

Cliff

cwhanse:
What do you think about replacing the golden section determination of Pmp by a lambertw approach?  Is the Pmp determination the limiting performance step in this subroutine?  We used a Pmp from lambertw approach in PVLib matlab to get past a piece of custom optimization for which we couldn't release source code.  It still uses an iteration so maybe we can reuse the golden section search code.

wholmgren:
The Pmp determination is the limiting step and it would be great to speed it up in any way.

Over the weekend I benchmarked some of [Scipy's optimization routines](http://docs.scipy.org/doc/scipy/reference/optimize.html) for Pmp but could not find anything as fast as what we already have. The fastest was [scipy.optimize.fmin_tnc](http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin_tnc.html#scipy.optimize.fmin_tnc) and it was 50% slower than our method. I also implemented a dpower/dcurrent function based on Jain and Kapoor and supplied it to the optimization routines. If I remember correctly, the optimizer is about twice as slow if it has to compute the gradients numerically, though it varies a bit depending on which optimizer you choose. 

In any case, I pushed what I did to my scipyoptimize branch. Here's the crux of it: https://github.com/wholmgren/pvlib-python/blob/scipyoptimize/pvlib/pvsystem.py#L1422 

It might be possible to speed up that loop with multithreading or numba JIT. That scipy function should release the GIL and allow for efficient multithreading, but it won't make any difference if the functions that it's calling are Python functions. It might not be too hard to write the these simple functions in cython, though. The first thing I would do is decorate the function with numba.jit and see if it helps (see spa.py for examples, though it does a little bit of work first to avoid making numba a pvlib requirement). If that's all new to you, then welcome to Python's multithreading hell.

wholmgren:
I added ``test_v_from_i_big`` with your parameters and it passed on my 64 bit mac and the 32 bit appveyor windows build. Would that have failed on matlab?

cwhanse:
Matlab PVLib currently uses a bisection to find Vmp, then gets Imp from Vmp.  Its faster than the Matlab root finder, and guarantees convergence.  The function operated on by the bisection (to find a zero) involves Lamberts W.

I can implement in python next week (I hope) when I’m back from catching fish.

Cliff

wholmgren:
Great, thanks. I'm going to merge this in the meantime.

Issue #191, opened 2016-06-13
perez transposition function drops nighttime data


wholmgren:
The irradiance.perez function drops nighttime data instead of returning 0s. You can see this in cell 37 of 

https://github.com/pvlib/pvlib-python/blob/master/docs/tutorials/irradiance.ipynb

Note that the plot only goes from sunrise to sunset, instead of spanning the full day. 

I have a fix for this in the works.

wholmgren:
closed by #192

PR #192, opened 2016-06-13
fix dropped nighttime values in perez, propagate airmass to perez in ModelChain

### LINKED ISSUE #172: ModelChain does not propagate airmass to perez transposition function
I think it can be fixed by adding ``airmass=self.airmass['airmass_relative']`` to the ``self.system.get_irradiance`` call. Probably need to add a test for this, too.

wholmgren:
fixes #172 and #191.


pvlib/irradiance.py
```diff
@@ -1100,31 +1094,29 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    # down, and there is still var.DHI, we set the airmass to the
+    # airmass value on the horizon (approximately 37-38).
     # var.AM(var.solar_zenith >=90 & var.DHI >0) = 37;
 
     # var.DNI_ET[var.DNI_ET==0] = .00000001 #very hacky, fix this
 
     # delta is the sky's "brightness"
     delta = dhi * airmass / dni_extra
 
-    # keep only valid times
```
wholmgren: the ebin.index changes from here to the end of the function are what really matters in this diff.

Issue #143, opened 2016-04-07
create ModelChain(s) that implement very simple models


wholmgren:
It could be useful to create a ModelChain or two that implements some kind of extremely basic system modeling. Something like ``power = user_def_const * poa_global``. The PVUSA model could be useful too. You all probably have more experience with sort of thing than I do.

A couple of motivations: 
* The simple models would make it easier to quantify just how much value the more sophisticated models add in different situations.
* The simple models might be a more gentle introduction to the library.
* It's easier to fit simple models to data, or at least to get started fitting.

I'm not yet sure if/how PVSystem would need to be modified to support this kind of thing. Maybe some additional keyword arguments in its constructor. Maybe you can reuse the module and inverter parameters arguments.

wholmgren:
To be a little more concrete, the ``ModelChain`` refactor might look something like:

```python
class ModelChain():

    def prepare_inputs():
        self.solar_position = self.location.get_solarposition(self.times)

        self.airmass = self.location.get_airmass(
            solar_position=self.solar_position, model=self.airmass_model)

        if irradiance is None:
            irradiance = self.location.get_clearsky(
                self.solar_position.index, self.clearsky_model,
                zenith_data=self.solar_position['apparent_zenith'],
                airmass_data=self.airmass['airmass_absolute'])
        self.irradiance = irradiance

        self.total_irrad = self.system.get_irradiance(
            self.solar_position['apparent_zenith'],
            self.solar_position['azimuth'],
            self.irradiance['dni'],
            self.irradiance['ghi'],
            self.irradiance['dhi'],
            model=self.transposition_model)

        if weather is None:
            weather = {'wind_speed': 0, 'temp_air': 20}
        self.weather = weather

    def run_model():
        raise NotImplementedError('you must subclass MC and implement this method')


class SAPMChain(ModelChain):

    def run_model():
        self.prepare_inputs()
        # followed by all of the SAPM calls


class BasicChain(ModelChain):

    def run_model():
        self.prepare_inputs()
        self.dc = self.total_irrad['global'] * self.poa_global_to_dc_power_const
        self.ac = self.dc * self.dc_to_ac_const


class PVUSAChain(ModelChain):

    def run_model():
        self.prepare_inputs()
        # the pvusa eqn

```

dacoex:
hello again ;.)

I'd done this with the 2.x code but still quite experimental.

I would recommend to use:
[...]

PR #179, opened 2016-05-22
refactor ModelChain. add SingleDiode modelchain

wholmgren:
Continuing from #151...

I refactored ``ModelChain`` into a parent ``ModelChain`` class with subclasses ``SAPM`` and ``SingleDiode``. In doing so, I moved some solar position, irradiance, and weather prep work into ``ModelChain.prepare_inputs``. ``prepare_inputs`` is called during ``run_model``. 

I also rebased this code on the changes that added support for ``SingleAxisTracker`` and voltage/current/power scaling.

I think that this is close to ready, but would appreciate testing and comments.

PR #194, opened 2016-06-19
refactor ModelChain with methods instead of classes

wholmgren:
I proposed refactoring ModelChain into a base class, a SAPM class, and a SingleDiode class in #179. This likely is a bad idea. It severely limits the flexibility of future ModelChains by promoting a single model choice, the DC power model, to the level of a class choice. It is also likely to make user code unnecessarily complicated.

Here, I propose a new way to extend ModelChain that may be more flexible, more powerful, and easier to use. The strategy is to assign predefined or user-defined functions to standard method names. For example, ModelChain.dc_model will be set to ModelChain.sapm or ModelChain.singlediode. This is basically a controlled version of "monkey patching", which is generally regarded as a bad thing in Python. I'm not sure what the problems would be, though.

Here's a gist with a notebook that demonstrates the usage:

https://gist.github.com/wholmgren/3fabacd3003c2a549b420b5e79f55e95

Some highlights:

```python
times = pd.date_range(start='20160101 0000-0700', end='20160104 0000-0700', freq='5min')
location = Location(32.2, -111, altitude=700)

# sapm model, just like in 0.3
# uses ModelChain.infer_dc_parameters to automatically determine which model to use
system = PVSystem(module_parameters=get_sapm_module_parameters(),
                  inverter_parameters=get_cec_inverter_parameters())
mc = ModelChain(system, location)
mc.run_model(times)

# singlediode model with cec module parameters
system = PVSystem(module_parameters=get_cec_module_parameters(),
                  inverter_parameters=get_cec_inverter_parameters())
mc = ModelChain(system, location)
mc.run_model(times)

# user defined parameters and functions
module_parameters = {'a': 3, 'b': 0.001, 'c': 1, 'd': -0.05}
system = PVSystem(module_parameters=module_parameters)
# pvusa_mc_wrapper is a user-defined function. see gist for details
mc = ModelChain(system, location, dc_model=pvusa_mc_wrapper, ac_model=pvusa_ac_mc_wrapper)
mc.run_model(times)
```


wholmgren:
I added modelchain options for pvwatts dc/ac/losses models, aoi models, and spectral models.

I've done a lot of small tests of the code and I think it's fairly robust, but I've not extensively tested it under real world conditions. It will be at least 4 weeks before I can push the code to its breaking point. Would you rather have only modestly tested modelchain code merged and released this week or better tested code in 4 weeks? I prefer to merge and release now and let you do the real world testing and file bug reports.

jforbess:
I suspect you'll get the best feedback if you release now, and I think we
are still at the stage where it is reasonable to expect possible bugs in
new features, so I vote release early and often!

wholmgren:
Thanks for your input, Jessica. 

Closes #143.

PR #208, opened 2016-07-07
updated first_solar_spectral_correction within atmosphere.py

MLEEFS:
I updated first_solar_spectral_correction within atmosphere.py to account for changes in model that are reflected in the IEEE PVSC 43 publication. I also added data flags/filters in order to make sure that the function doesn't "blow up" when given extraordinary inputs.

Minimum precipitable water floor is set to 0.1cm. This presents the function from going toward -Inf as precipitable water goes to toward zero.

An absolute airmass ceiling is placed at AM = 10. This prevents function from destabilizing at very high AM. AM > 10 occur very close to sunset/sunrise when irradiance levels are very low.

Warning flags are now printed if Pwat is really high (above 8 cm), or if absolute air mass is very low (below .58)

### DIFF
--- a/pvlib/atmosphere.py
+++ b/pvlib/atmosphere.py
@@ -339,7 +339,7 @@ def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
     .. math::
 
         M = c_1 + c_2*AMa  + c_3*Pwat  + c_4*AMa^.5
-            + c_5*Pwat^.5 + c_6*AMa/Pwat
+            + c_5*Pwat^.5 + c_6*AMa/Pwat^.5
 
     Default coefficients are determined for several cell types with
     known quantum efficiency curves, by using the Simple Model of the
@@ -348,7 +348,7 @@ def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
     Pwat where:
 
        * 0.5 cm <= Pwat <= 5 cm
-       * 0.8 <= AMa <= 4.75 (Pressure of 800 mbar and 1.01 <= AM <= 6)
+       * 1.0 <= AMa <= 5.0 
        * Spectral range is limited to that of CMP11 (280 nm to 2800 nm)
        * spectrum simulated on a plane normal to the sun
        * All other parameters fixed at G173 standard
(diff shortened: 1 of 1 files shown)


wholmgren:
Thanks, @MLEEFS. Looks good, just a few simple things...

It's best to replace the print statements with [warnings.warn](https://docs.python.org/3.5/library/warnings.html#warnings.warn). This way, users can silence the warnings if they want to. I am ambivalent about input warnings for pvlib python, but if you want to use them, that's the way to do it.

You should also replace the limit setting with the [np.maximum](http://docs.scipy.org/doc/numpy/reference/generated/numpy.maximum.html) and np.minimum functions. For example:

```
pw = np.maximum(pw, 0.1)
```

These numpy functions should be much faster than the list comprehensions and I think they're more readable.

I'm a little surprised to see that the tests still passed without any modification, so your changes to the coefficients must amount to pretty small changes in output over the tested range (1-5 cm and 1-5 airmasses).

MLEEFS:
Hi @wholmgren ,

I made the changes you requested. Can you confirm that they have been made? I am still getting used to the mechanics of GIT.

pvlib/atmosphere.py
```diff
@@ -406,20 +407,58 @@ def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
...
-    .. [2] Marion, William F., et al. User's Manual for Data for Validating
+    .. [2] Lee, Mitchell, and Panchula, Alex. "Spectral Correction for
+       Photovoltaic Module Performance Based on Air Mass and Precipitable 
+       Water." IEEE Photovoltaic Specialists Conference, Portland, 2016 
+    .. [3] Marion, William F., et al. User's Manual for Data for Validating
        Models for PV Module Performance. National Renewable Energy
        Laboratory, 2014. http://www.nrel.gov/docs/fy14osti/61610.pdf
     """
 
+    import pdb
```
wholmgren: You'll need to delete these pdb lines. They prevent the test suite from completing.

pvlib/atmosphere.py
```diff
@@ -435,6 +474,6 @@ def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
     AMa = airmass_absolute
     modifier = (
         coeff[0] + coeff[1]*AMa  + coeff[2]*pw  + coeff[3]*np.sqrt(AMa) +
-        + coeff[4]*np.sqrt(pw) + coeff[5]*AMa/pw)
+        + coeff[4]*np.sqrt(pw) + coeff[5]*AMa/np.sqrt(pw))
```
wholmgren: looks like I made a mistake when I first ported this and put two + signs here (one on each line). Let's keep only the first one.

MLEEFS:
the "bonus" lines of code should be removed

wholmgren:
Thanks @MLEEFS. merging...

Issue #201, opened 2016-06-24
replace limits set with numpy/pandas slicing with np.maximum and np.minimum 


wholmgren:
A number of functions use numpy/pandas slicing or clip to set their limits. This leads to unnecessary requirements on the input data, unnecessary type promotion of the output data, and/or unnecessary code in the functions. Many of the limits can instead be set with [np.maximum](http://docs.scipy.org/doc/numpy/reference/generated/numpy.maximum.html) and [np.minimum](http://docs.scipy.org/doc/numpy/reference/generated/numpy.minimum.html). 

The easy functions to fix include:
- [x] haurwitz
- [x] beam_component
- [x] haydavies
- [x] reindl
- [x] king

Other functions that rely on bins or masks need array-like inputs or coercion and require a little more effort. These include:
- [x] perez
- [x] disc
- [ ] dirint
- [x] erbs
- [ ] singleaxis
- [x] relativeairmass
- [x] ashraeiam
- [x] physicaliam
- [x] snlinverter
- [ ] pyephem_earthsun_distance

adriesse:
I'd like to understand this.  Could you give one or two problematic 
examples?

On 2016-06-24 21:22, Will Holmgren wrote:
>
> A number of functions use numpy/pandas slicing or clip to set their 
> limits. This leads to unnecessary requirements on the input data, 
> unnecessary type promotion of the output data, and/or unnecessary code 
> in the functions. Many of the limits can instead be set with 
> np.maximum 
> <http://docs.scipy.org/doc/numpy/reference/generated/numpy.maximum.html> 
> and np.minimum 
> <http://docs.scipy.org/doc/numpy/reference/generated/numpy.minimum.html>.
>
> The easy functions to fix include:
>
> relativeairmass
> ashraeiam
> physicaliam
> snlinverter
> haurwitz
> beam_component
> haydavies
> reindl
> king
>
> Other functions that rely on bins or masks need array-like inputs and 
> require a little more effort. These include:
>
> perez
> disc
> dirint
> erbs
> singleaxis
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub 
> <https://github.com/pvlib/pvlib-python/issues/201>, or mute the thread 
> <https://github.com/notifications/unsubscribe/AIlYQ1nxcHZn0kjoykd4Gjlwg9sjLut_ks5qPC6IgaJpZM4I-DWp>.
>

-- 
[...]

PR #220, opened 2016-07-20
improve function argument ordering in pvsystem

### LINKED ISSUE #197: reorder function arguments to put variables first and parameters second
I think we should reorder a handful of function arguments to put the most variable-like arguments first and most parameter-like arguments last. I'm guessing that a lot people will think that I'm just screwing with the API for no reason, but I do have a few reasons!

1. consistency is important (and underrated)
2. we can make better use of default arguments. 

Here are the functions that I know about so far:

```python
ashraeiam(b, aoi) ---> ashraeiam(aoi, b=0.05)

# updated to reflect Cliff's comment below
physicaliam(K, L, n, aoi) ---> physicaliam(aoi, n=1.526, K=4., L=0.002)

# see #200
singlediode(module, photocurrent, saturation_current, resistance_series, resistance_shunt, nNsVth) --->
singlediode(photocurrent, saturation_current, resistance_series, resistance_shunt, nNsVth)

# I just noticed that the matlab sapm api is very different. # see #198
sapm(module, poa_direct, poa_diffuse, temp_cell, airmass_absolute, aoi) ---> 
sapm(poa_direct, poa_diffuse, temp_cell, airmass_absolute, aoi, module)

snlinverter(inverter, v_dc, p_dc) ---> snlinverter(v_dc, p_dc, inverter)
```

wholmgren:
This PR improves some function signatures as discussed in #197.

Also addresses the input/output type issues described #201 for ashraeiam, physicaliam, and snlinverter.

Adds test_tools.py. 

I'll merge tomorrow unless there are comments.

Closes #197.

### DIFF
--- a/docs/sphinx/source/package_overview.rst
+++ b/docs/sphinx/source/package_overview.rst
@@ -105,10 +105,10 @@ to accomplish our system modeling goal:
         temps = pvlib.pvsystem.sapm_celltemp(total_irrad['poa_global'],
                                              wind_speed, temp_air)
         effective_irradiance = pvlib.pvsystem.sapm_effective_irradiance(
-            module, total_irrad['poa_direct'], total_irrad['poa_diffuse'],
-            am_abs, aoi)
-        dc = pvlib.pvsystem.sapm(module, effective_irradiance, temps['temp_cell'])
-        ac = pvlib.pvsystem.snlinverter(inverter, dc['v_mp'], dc['p_mp'])
+            total_irrad['poa_direct'], total_irrad['poa_diffuse'],
+            am_abs, aoi, module)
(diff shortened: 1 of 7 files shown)


jforbess:
It looks good at a high level, but I noticed that you didn't do the same type of reordering for the sapm function. I see that you have just updated sapm to better align with Matlab, but the intention was only to align algorithmically, right, so we'd still want the parameters ordered with effective_irradiance and temperature first?

wholmgren:
Good catch @jforbess. The functions introduced in #218 should be changed, too.

wholmgren:
@jforbess I updated the sapm functions. I'll let this sit for another day.

PR #226, opened 2016-07-24
Fix numerical overflow in v_from_i for some conditions

### LINKED ISSUE #225: overflow error in v_from_i for some conditions
PR #190 replaced a numerical solution for v_oc with an analytical solution. @cwhanse warned that the analytical solution is prone to numerical overflow for modules with large Voc. Cliff's suggested test passed, though, so I figured that we were in good shape and the difference could be attributed to something about Python vs. Matlab. Well, I found a situation where the analytical solution does fail...

```python
In [76]:     module_parameters = modules['Canadian_Solar_CS5P_220M'].copy()
    ...:     module_parameters['EgRef'] = 1.121
    ...:     module_parameters['dEgdT'] = -0.0002677

In [77]: IL, I0, Rs, Rsh, nNsVth = pvlib.pvsystem.calcparams_desoto(1000, 30, module_parameters['alpha_sc'], module_parameters, module_parameters['EgRef'], module_parameters['dEgdT'])

In [78]: IL, I0, Rs, Rsh, nNsVth
Out[78]: (5.1366949999999996, 1.8739027472625636e-09, 1.065, 381.68, 2.681527737715915)

In [79]: I = 0

# the line in v_from_i that fails
In [80]: I0 * Rsh / nNsVth * np.exp(Rsh * (-I + IL + I0) / nNsVth)
Out[80]: inf
```

Following pvlib matlab's v_from_i, we can calculate the lambertw term in log space...

```python
In [81]:     logargW = (np.log(I0) + np.log(Rsh) - np.log(nNsVth) +
    ...:                Rsh * (-I + IL + I0) / nNsVth)

In [82]:     w = logargW
    ...:     for i in range(0, 3):
    ...:         w = w * (1 - np.log(w) + logargW) / (1 + w)
    ...:     lambertwterm_log = w
    ...:

In [83]: lambertwterm_log
Out[83]: 709.43905979862768
```

I'll put this in a PR soon.

wholmgren:
Closes #225.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,7 +27,7 @@ install:
     - cmd: conda info -a
 
     # install depenencies
-    - cmd: conda create -n test_env --yes --quiet python=%PYTHON_VERSION% pip numpy scipy=0.16.0 pandas nose pytest pytz ephem numba
+    - cmd: conda create -n test_env --yes --quiet python=%PYTHON_VERSION% pip numpy scipy pandas nose pytest pytz ephem numba
     - cmd: activate test_env
     - cmd: python --version
     - cmd: conda list
--- a/docs/sphinx/source/whatsnew/v0.4.0.txt
+++ b/docs/sphinx/source/whatsnew/v0.4.0.txt
@@ -65,6 +65,8 @@ Bug fixes
   rather than requiring that the solution for all points be within the
   error tolerance. Results in test scenarios changed by 1-10%.
   (:issue:`221`)
+* Fixed a numerical overflow error in pvsystem.singlediode's v_oc
+  determination for some combinations of parameters. (:issue:`225`)
 * dirint function yielded the wrong results for non-sea-level pressures.
   Fixed. (:issue:`212`)
 * Fixed a bug in the day angle calculation used by the 'spencer' and 'asce'
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1843,8 +1843,8 @@ def v_from_i(resistance_shunt, resistance_series, nNsVth, current,
         w = w * (1 - np.log(w) + logargW) / (1 + w)
     lambertwterm_log = w
 
-    lambertwterm = np.where(np.isinf(lambertwterm), lambertwterm_log,
-                            lambertwterm)
+    lambertwterm = np.where(np.isfinite(lambertwterm), lambertwterm,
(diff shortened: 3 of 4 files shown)


wholmgren:
The new v_from_i code yields different numbers on windows than on mac/linux, or at least it does on the appveyor testing service. It would be great if someone running windows can try to debug this. I think you can just run the code pasted in #225 using the parameters printed there as well as the parameters in the test_v_from_i_bigger function (see files changed). No need to git checkout and install my branch.

cwhanse:
I'll do it but I'll need some instructions, I clicked around unsuccesfully for awhile to see the error or test that failed.

wholmgren:
Here's a link to the error at the bottom of one of the appveyor builds: 

https://ci.appveyor.com/project/wholmgren/pvlib-python-fv2to/build/1.0.428/job/tx70e02w2w8vdrr0#L552

and a link to the test in the git diff:

https://github.com/pvlib/pvlib-python/pull/226/files#diff-134bccec9f71c38ba6da34b38c35ac27R368

cwhanse:
On Win64 the line 

produces

54.304092877335222

which is different from the asserted test value by

0.00013404354376689298

At the intermediate steps:
logargW is 677.4553451926995

lambertwterm is 670.9466091769514+0j   (type issue???)

Can you post the values from the code that finds the asserted value?

wholmgren:
Interesting that your system returns a much closer but still different result. 

I reran this on my mac...

pvsystem.v_from_i(190, 1.065, 2.89, 0, 7.05196029e-08, 10.491262)

produces

54.303958833791455

and logargW is 677.4553451926995

lambertwterm is 670.9466555588004+0j 

Scipy's lambertw always returns a complex number type, even if the complex part is 0. It probably would be better to drop the complex part at the ``lambertwterm = lambertw(argW)`` line than at the end.

cwhanse:
After moving the .real I get the same values you posted.

wholmgren:
Hm. I am guessing that the difference is due to something about the precision of real numbers and complex numbers, and that is going to depend a user's platform and compiler. Or, for many of us, the version of Continuum's numpy/scipy packages. 

I removed the scipy=0.16.0 version restriction on our appveyor test suite, and the test now passes. The version restriction was only present due to a problem with the matlab file reader in Continuum's scipy 0.16.1-0.17. 

Thanks for your help, Cliff.

PR #230, opened 2016-08-04
Fix #228: dhi/ghi typo

### LINKED ISSUE #228: Typo in klucher irradiance model
This change looks like a typo:
https://github.com/pvlib/pvlib-python/commit/1f968a034f74e46f021f3a6b85ea86c5e1ad6cf8#diff-9e2b3ebf5d45c478fee41a88b0e5f44eR671

emulbreh:


wholmgren:
Thanks @emulbreh. As @adriesse pointed out, we should have a better test. I forgot that the irradiance module's tests do not currently assert that the test output is equal to an expected number, and I should not have changed this code without such a test in place. Sorry. Can you add a line to pvlib/test/test_irradiance.py's test_klucher_series_float that asserts that the output is equal to whatever your fixed code produces? You'll probably want to use the assert_allclose function since you're comparing floats.

wholmgren:
Thanks for the fix and adding the extra test code! Closes #228.

PR #232, opened 2016-08-04
Replace nosetests references with py.test

emulbreh:


docs/sphinx/source/installation.rst
```diff
@@ -154,7 +154,7 @@ referred to as *conda environments*, but they're the same for our purposes.
    ``conda create --name pvlibdev python pandas scipy``
 #. **Activate** the new conda environment: ``source activate pvlibdev``
 #. **Install** additional packages into your development environment:
-   ``conda install jupyter ipython matplotlib seaborn nose flake8``
+   ``conda install jupyter ipython matplotlib seaborn pytest flake8``
```
wholmgren: We actually currently need both nose and pytest due to our use of a pandas test function that requires nose.

setup.py
```diff
@@ -46,7 +46,7 @@
                     'six',
                     ]
 SETUP_REQUIRES = ['pytest-runner']
-TESTS_REQUIRE = ['pytest', 'nose']
+TESTS_REQUIRE = ['pytest']
```
wholmgren: Same as above.

wholmgren:
Thanks. I missed the notification that this was updated.

Issue #233, opened 2016-08-11
ModelChain with weather data


uvchik:
Some time ago we started the [feedinlib](https://github.com/oemof/feedinlib) project to calculated the output of wind and pv power plants with a given weather data set. The pv part of the feedinlib is an application of the pvlib similar to the ModelChain, which did not exist when we started.

In my opinion a ModelChain with weather data is better located within the pvlib, so I would like to move this functionality from the feedinlib to the pvlib. The structure should be similar to the existing ModelChain while the content could be base on the existing [feedinlib model](https://github.com/oemof/feedinlib/blob/dev/feedinlib/models.py#L56).

My idea is to define a weather class containing the weather data and the meta data of the weather data set similar to the [weather class of the feedinlib](https://github.com/oemof/feedinlib/blob/dev/feedinlib/weather.py#L11). A weather object could be passed to a new ModelChain or to an expanded ModelChain.

I would like to discuss the idea before creating a PR to avoid obsolete work.

What do you think of this idea?

wholmgren:
If I'm understanding correctly, feedinlib's FeedinWeather is similar to pvlib's Location but with a data attribute that contains weather data, and then the rest of feedinlib works on that object. In contrast, ModelChain takes a Location and a PVSystem object on initialization, and then the run_model method takes separately defined weather and irradiance data. Of course, one can assign a data attribute to a Location object and then pass the attribute to run_model (I've used this pattern in the past). It seems to me that this kind of data organization is best left to the user, but I could be wrong! Can you say more about why you don't want to simply adopt the pvlib approach here? 

On a somewhat related note, I don't particularly care for the fact that run_model has separate irradiance and weather inputs. It may well be better to have a single dict-like weather input that includes irradiance.

uvchik:
First I have to say that I have overlooked the possibility to pass a weather object to the model chain, so some of the things I wrote above are obsolete now. It definitely makes sense for me to enhance the ModelChain instead of writing something new.

I am pretty sure that we will find a way to keep the API as it is and still make it possible to pass a weather DataFrame containing wind_speed, temperature and(!) irradiation.

One problem is the calculation of the irradiation in plane if you use hourly weather data or measurement data. Sometimes we had a morning and evening peak due to timing differences between the direct irradiation and the height of the sun or due to measurement deviations. In the feedinlib we calculated something like an average angle ors even cut the direct irradiation for very small angles.

But I think these are details and we could easily discuss this in a PR if you do not have any general objections.

[...]

PR #239, opened 2016-09-01
Making the irradiation and weather DataFrame of the ModelChain more flexible

uvchik:
This is the PR for #233. It is just a proposal. I am still not sure if this approach makes sense.

There are two changes:
 * The weather DataFrame can contain the irradiation columns and will be used instead of the irradiance DataFrame if no irradiance DataFrame is passed to the method.

* The irradiance (or weather) DataFrame does not have to contain all types of irradiance (ghi, dni, dhi) but just two of them or only `ghi`.

The new  `tools.check_df` function (does not exist so far) should check if some columns in the DataFrame exist and some do not and returns True or False. If this function doesn't make sense we could change the if condition to:
```python
# Using the function
if tools.check_df(self.irradiance, ['col1', 'col2'], not_exist=['col3'])

# Using the normal if condition
if (self.irradiance['col1'] is None and 
    self.irradiance['col2'] is None and
    self.irradiance['col3'] is not None):
```

The prepare_irradiance method makes it more convenient to pass weather data to the ModelChain but it blows up the code. So what do you think?

This is just a proposal to get some feedback some work is still missing:
 * adapt documentation

 * add new docstring / adept docstring of existing methods

 * write / adapt tests (my problem is that I have already 111 errors with the actual master branch)

 * write `tools.check_df` function (@gnn Could you write an efficient way to do it?) or adept if clause

 * adapt "whatsnew"-file

@gnn could you have a look from the programming point of view?

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.4.2.txt
+++ b/docs/sphinx/source/whatsnew/v0.4.2.txt
@@ -6,6 +6,12 @@ v0.4.2 ()
 This is a minor release from 0.4.0. ....
 
 
+Bug fixes
+~~~~~~~~~
+
+* Fixed typo in __repr__ method of ModelChain and in its regarding test. (commit: b691358b)
+
+
 API Changes
 ~~~~~~~~~~~
 
--- a/docs/sphinx/source/whatsnew/v0.4.2.txt
+++ b/docs/sphinx/source/whatsnew/v0.4.2.txt
@@ -1,0 +1,24 @@
+.. _whatsnew_0420:
+
+v0.4.2 ()
+------------------------
+
+This is a minor release from 0.4.0. ....
+
+
+API Changes
+~~~~~~~~~~~
+
+* The run_model method of the ModelChain will use the weather parameter of all weather data instead of splitting it to irradiation and weather. The irradiation parameter still works but will be removed soon.(:issue:`239`)
+
+
+Enhancements
+~~~~~~~~~~~~
+
+* Adding a complete_irradiance method to the ModelChain to make it possible to calculate missing irradiation data from the existing columns [beta] (:issue:`239`)
+
+
+Code Contributors
+~~~~~~~~~~~~~~~~~
+
+* Uwe Krien
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -673,6 +673,12 @@ class ModelChain(object):
 
         Assigns attributes: times, solar_position, airmass, total_irrad, aoi
         """
+        # Add columns that does not exist and overwrite existing columns
+        # Maybe there is a more elegant way to do this. Any ideas?
+        if weather is not None:
+            self.weather = self.weather.combine_first(weather)
+            self.weather.update(weather)
+
         # The following part could be removed together with the irradiance
         # parameter at version v0.5 or v0.6.
         # **** Begin ****
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -652,7 +652,7 @@ class ModelChain(object):
                                    self.weather.dni *
                                    tools.cosd(self.solar_position.zenith))
 
-    def prepare_inputs(self, times, irradiance=None, weather=None):
+    def prepare_inputs(self, times=None, irradiance=None, weather=None):
         """
         Prepare the solar position, irradiance, and weather inputs to
         the model.
@@ -695,13 +695,8 @@ class ModelChain(object):
                 self.weather[column] = irradiance.pop(column)
         # **** End ****
 
-        # Add columns that does not exist and overwrite existing columns
-        # Maybe there is a more elegant way to do this. Any ideas?
-        if weather is not None:
-            self.weather = self.weather.combine_first(weather)
-            self.weather.update(weather)
-
-        self.times = times
+        if times is not None:
+            self.times = times
 
         self.solar_position = self.location.get_solarposition(self.times)
 
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -323,7 +323,7 @@ class ModelChain(object):
 
     def __repr__(self):
         return ('ModelChain for: ' + str(self.system) +
-                ' orientation_startegy: ' + str(self.orientation_strategy) +
+                ' orientation_strategy: ' + str(self.orientation_strategy) +
                 ' clearsky_model: ' + str(self.clearsky_model) +
                 ' transposition_model: ' + str(self.transposition_model) +
                 ' solar_position_method: ' + str(self.solar_position_method) +
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -605,7 +605,7 @@ class ModelChain(object):
             fd*self.total_irrad['poa_diffuse'])
         return self
 
-    def prepare_irradiance(self, dni_determination_method='dirint'):
+    def complete_irradiance(self, dni_determination_method='dirint'):
         """
 
         Parameters
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -686,11 +686,7 @@ class ModelChain(object):
         # Add columns that does not exist and overwrite existing columns
         # Maybe there is a more elegant way to do this. Any ideas?
         if weather is not None:
-            if self.weather is None:
-                self.weather = weather
-            else:
-                self.weather = self.weather.combine_first(weather)
-                self.weather.update(weather)
+            self.weather = weather
 
         # The following part could be removed together with the irradiance
         # parameter at version v0.5 or v0.6.
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -7,6 +7,7 @@ the time to read the source code for the module.
 """
 
 from functools import partial
+import logging
 import warnings
 import pandas as pd
 
@@ -607,50 +608,54 @@ class ModelChain(object):
             fd*self.total_irrad['poa_diffuse'])
         return self
 
-    def complete_irradiance(self, dni_determination_method='dirint'):
+    def complete_irradiance(self, times, weather):
         """
+        Determine the missing irradiation columns. Only two of the following
+        data columns (dni, ghi, dhi) are needed to calculate the missing data.
+
+        This function is not save at the moment. Results can be too high or
+        negative. Please contribute and help to improve this function on
+        https://github.com/pvlib/pvlib-python
 
         Parameters
         ----------
-        dni_determination_method : string
-            something....
+        times : datetime index
+            Date and time index of the irradiation data
+        weather : pandas.DataFrame
+            Table with at least two columns containing one of the following data
+            sets: dni, dhi, ghi
 
         Returns
         -------
+        pandas.DataFrame
+            Containing the missing column of the data sets passed with the
+            weather DataFrame.
 
         """
-        icolumns = self.weather.columns
-
-        if {'ghi'} <= icolumns and not {'dhi', 'dni'} <= icolumns:
-            if dni_determination_method == 'dirint':
-                return_value = irradiance.dirint(self.weather.ghi,
-                                                 self.solar_position.zenith,
-                                                 self.times)
-                self.weather['dhi'] = return_value
-            elif dni_determination_method == 'disc':
-                return_value = irradiance.disc(self.weather.ghi,
-                                               self.solar_position.zenith,
-                                               self.times)
-                self.weather['dni'] = return_value.dni
-            elif dni_determination_method == 'erbs':
-                return_value = irradiance.erbs(self.weather.ghi,
-                                               self.solar_position.zenith,
-                                               self.times)
-                self.weather['dni'] = return_value.dni
-                self.weather['dhi'] = return_value.dhi
-
-        if {'ghi', 'dhi'} <= icolumns and not {'dni'} <= icolumns:
-            self.weather['dni'] = ((self.weather.ghi -
-                                    self.weather.dhi) /
-                                   tools.cosd(self.solar_position.zenith))
-        elif {'dni', 'dhi'} <= icolumns and not {'ghi'} <= icolumns:
-            self.weather['ghi'] = (self.weather.dni *
-                                   tools.cosd(self.solar_position.zenith) +
-                                   self.weather.dhi)
-        elif {'dni', 'ghi'} <= icolumns and not {'dhi'} <= icolumns:
-            self.weather['dhi'] = (self.weather.ghi -
-                                   self.weather.dni *
-                                   tools.cosd(self.solar_position.zenith))
+        self.weather = weather
+        self.times = times
+        self.solar_position = self.location.get_solarposition(self.times)
+        icolumns = set(self.weather.columns)
+        wrn_txt = "This function is not save at the moment.\n"
+        wrn_txt += "Results can be too high or negative.\n"
+        wrn_txt += "Help to improve this function on github."
+        wrn_txt += "https://github.com/pvlib/pvlib-python"
+        warnings.warn(wrn_txt, UserWarning)
+        if {'ghi', 'dhi'} <= icolumns and 'dni' not in icolumns:
+            logging.debug('Estimate dni from ghi and dhi')
+            self.weather.loc[:, 'dni'] = (
+                (self.weather.loc[:, 'ghi'] - self.weather.loc[:, 'dhi']) /
+                tools.cosd(self.solar_position.loc[:, 'zenith']))
+        elif {'dni', 'dhi'} <= icolumns and 'ghi' not in icolumns:
+            logging.debug('Estimate ghi from dni and dhi')
+            self.weather.loc[:, 'ghi'] = (
+                self.weather.dni * tools.cosd(self.solar_position.zenith) +
+                self.weather.dhi)
+        elif {'dni', 'ghi'} <= icolumns and 'dhi' not in icolumns:
+            logging.debug('Estimate dhi from dni and ghi')
+            self.weather.loc[:, 'dhi'] = (
+                self.weather.ghi - self.weather.dni *
+                tools.cosd(self.solar_position.zenith))
 
     def prepare_inputs(self, times=None, irradiance=None, weather=None):
         """
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -679,7 +679,11 @@ class ModelChain(object):
         # The following part could be removed together with the irradiance
         # parameter at version v0.5 or v0.6.
         # **** Begin ****
+        wrn_txt = "The irradiance parameter will be removed soon."
+        wrn_txt += "Please use the weather parameter to pass a DataFrame with "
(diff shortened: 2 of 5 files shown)


wholmgren:
Thanks for the nice start, @uvchik. I like the ideas. A few things to think about...

Should dni_determination_method be a ModelChain argument rather than a prepare_inputs (or similar) argument? I think this change would make it more consistent with the existing API. The ModelChain __init__ does have a transposition_model argument, though it's not actually used at this time. 

Should we get rid of the irradiance kwarg all together and insist that all of the data be put into weather? Or perhaps we should allow either, but set self.irradiance = self.weather[['dni', 'ghi', 'dhi']] if only weather is supplied.

I think you'll need to check for ``is None`` rather than relying on the truthiness of your array-like DataFrames.

Is your check_df function only going to check for keys/columns or is it going to consider nans as well? If it's only going to check for keys/columns then I don't know if separate function adds much over a inline try/except or set logic. 

```python
In [3]: df = pd.DataFrame(columns=['ghi', 'dni'])

In [4]: columns = set(df.columns)

In [5]: {'ghi', 'dni'} <= columns and not {'dhi'} <= columns
Out[5]: True
```

I'll try to get a fix for those forecast module errors up in the next few days so that they don't clutter your test reports.

uvchik:
> Should dni_determination_method be a ModelChain argument rather than a prepare_inputs (or similar) argument?

I just forgot the `**kwargs` parameter in the run_model method. I think this could be a good compromise. We can define our default stuff for normal users but allow advanced users to pass additional arguments directly with the run_model call.

```python
def run_model(self, times, irradiance=None, weather=None, **kwargs):
    self.prepare_inputs(times, irradiance, weather, **kwargs)

def prepare_inputs(self, times, irradiance=None, weather=None, **kwargs):
    self.prepare_irradiance(**kwargs)

def prepare_irradiance(self, dni_determination_method='dirint', **kwargs):
    if  dni_determination_method is something do something...

# The call:
my_chain.run_model(times, irradiance_df, dni_determination_method='erbs') 
```
We could use this structure for default values in other methods/functions, too.

> Should we get rid of the irradiance kwarg all together and insist that all of the data be put into weather? 

For me that would be okay but I think we are not allowed to change the API in minor releases so we have to wait until `0.5`. Defining `self.irradiance = self.weather` is no problem because it just sets a pointer so we will have one DataFrame with two names. In the long run one DataFrame would be enough for me.

```python
>>> import pandas as pd
>>> a = pd.DataFrame({'A':[1,2,3],'B':[4,5,6]})
>>> a
   A  B
0  1  4
1  2  5
2  3  6
>>> b = a
>>> b['C']=[7,8,9]
>>> a
   A  B  C
0  1  4  7
1  2  5  8
2  3  6  9
```

You are right, I will use an inline method to check the columns. I like your example with the `set`.

wholmgren:
I'm hesitant to hide this model choice in **kwargs. To keep ModelChain as explicit as possible, I prefer something more like these two options:

1. Add a dni_determination_method keyword argument to the ModelChain constructor. The default value could be 'required' or 'supplied' and other values could be as you've described above. If dni is required and not supplied then an exception would be raised somewhere. If dni is supplied but dni_determination_method='erbs' or similar then the dni should be recalculated (possibly with a warning) or an exception should be raised.

2. Create prepare_irradiance as you've proposed, but do not add it to the run_model or the misnamed prepare_inputs method (more on this later). Users that want code that is robust to varying irradiance inputs can add a ``irradiance = mc.prepare_irradiance(irradiance)`` line, and users that want code that breaks with unexpected input can skip it. I like code that breaks with unexpected input because it usually indicates that I've done something wrong somewhere else or that I don't understand all of my input data.

Regarding the misnamed prepare_inputs, I've previously considered renaming prepare_inputs to prepare_intermediates. I've also considered removing the assumption of clear sky if irradiance=None. I didn't follow through only for lack of time, but I'd like to make these changes in 0.5. This may or may not change your opinion on how the proposed functionality should be implemented.

> For me that would be okay but I think we are not allowed to change the API in minor releases so we have to wait until 0.5.

True. Whatever we decide to do, let's make it non-breaking for now and reconsider the breaking change later on. 

> Defining self.irradiance = self.weather is no problem because it just sets a pointer so we will have one DataFrame with two names. 

My only concern was that people might be confused when they see non-irradiance data in the irradiance attribute. It's probably not a big deal.

uvchik:
> I've also considered removing the assumption of clear sky if irradiance=None. I didn't follow through only for lack of time, but I'd like to make these changes in 0.5. This may or may not change your opinion on how the proposed functionality should be implemented.

That changes a lot. What about defining an irradiation (or weather) class. Then we can deal with all that attributes defining an irradiation (or weather) object and the ModelChain will get its dni, dhi and ghi regardless how they have been created.

The best thing is, that it is still possible to pass a weather DataFrame to the run_model because `irradiation.dni` (or `weather.dni`) works for DataFrames and will work for the object.

This might be a way to not overload the ModelChain but keep all functionalities like clearsky and my prepare_irradiation stuff.

wholmgren:
A weather class might be useful and I agree that it should be optional. Keep in mind that you'll need to support item lookup in addition to attribute lookup. The vast majority of pvlib uses item lookup, including ModelChain. I think it should be doable if you define one or more of the __get*__ methods.

I'm not sure how long it would take to develop this class, so I'd also be ok with following one of the approaches outlined above for now and then changing it later. Documentation and rigorous testing always takes a lot longer than I expect. My main concern is that the default behavior remains that an error is raised if a required irradiance value does not exist.

uvchik:
Okay, so let us keep it simple by now and let us conclude.

1. In the long run the irradiance parameter will be removed in favour of the weather parameter
2. I will add a FutureWarning make users aware of this change
3. The prepare_irradiance method will be optional `mc.prepare_irradiance(weather)`
4. Do you like the name `prepare_irradiance`? It's more like filling or completing the weather data but you are the native speaker :smirk:
5. Something we did not discussed by now: If `weather` is passed to the `run_model` it assigns the `self.weather` attribute. Afterwards we can pass a new pv-system and run the model again. _In my opinion the `self.weather` attribute should be used if no new weather parameter is passed._ Next time we keep the pv-system and just run the `run_model` again with a new weather DF. _In my opinion it should overwrite the self.weather attribute._

Please point out with which statement you agree/disagree.

wholmgren:
1-3: I agree.

4: I don't particularly like the name, but I'm not opposed to it. ``check_and_fill_irradiance``? I don't love that either. In any case, I would recommend keeping irradiance in the name and keeping this particular method focused on irradiance. An e.g. ``complete_weather_inputs``  method could be created in the future that calls the irradiance method.

5: I agree. We should make this clear in the documentation, though.

dacoex:
>4: I don't particularly like the name, but I'm not opposed to it. check_and_fill_irradiance? I don't love that >either. 
Alternative `setup_irradiance` or `assign_irradiance_values`

adriesse:
I would raise a word of caution about calculating the missing component from the two that are given among GHI, DNI and DHI.  In real-world data they rarely add up perfectly.  If you have only one of the three, ignorance is bliss and you can model the other two so they do add up perfectly.  But if you have two, then the third one may end up being negative or some other unreasonable value.  Also, dividing by cos(zenith) where zenith -> 90 is a sensitive operation.

uvchik:
@wholmgren Thanks for you feedback. I soon will provide my second approach. Could take some weeks because I have some deadlines within the next three weeks.

@adriesse Thank you for bringing up this point. I almost forgot about this point but it was my concern when I started [issue 233](https://github.com/pvlib/pvlib-python/issues/233#issuecomment-240698516).

I often got global and diffuse irradiation measured on a horizontal surface and had to calculate DNI using them. I got negative values for `ghi-dhi > 0` if `zenith > 90°` and very(!) high values if `89.99° > zenith < 90°`. I even got these problems using reanalysis data from a weather data set.

I think the pvlib would be a could place to talk about these problems and offer tools to deal with it.
One reason for this behaviour is the timestamp of the weather data set. This timestamp can stand for the average value of the previous hour or for the average value of the next hour. The pvlib uses this timestamp to calculate the position of the sun. In the evening of a day this might result into `zenith > 90` (or slightly below 90°) while in the previous hour was some direct irradiation so that the average value of `ghi-dhi` is above zero.

What can we do to avoid this? Providing an average zenith? Providing a "middle" zenith for timestamp+/-30 minutes? What if it is caused by a measurement error, shall we cut small angles above the horizon?

Of course we could leave it to the users but in my opinion the pvlib is a good place to discuss these problems and offer solutions.

adriesse:
There are many different difficult aspects to this, so indeed I think the users of these tools need to be knowledgable.  I think it's ok to discuss here, but probably you want to have a larger group.

For your specific example, you can keep DNI values somewhat sane by not allowing them to be greater than a value obtained from a clear-sky model.  Calculating the angles for the midpoint of the averaging interval is common practice (I hope).  For long time steps like one hour, you can split the intervals that includes sunrise/set into two parts (one before and one after the event).

cwhanse:
Ø  Calculating the angles for the midpoint of the averaging interval is common practice (I hope).

It ought to be common practice.  SAM does it, as does PVsyst I think, but other applications?  We ought to build this capability into pvlib I think, as part of model chain – a kwarg switch perhaps?

wholmgren:
I agree with all of the above, though it's up to @uvchik to decide the scope of this particular addition. I have slight preference for finishing this small, non-breaking, non-default-changing pull request, and then moving on to address the time resolutions in a separate issue and pull request. I am ok with addressing it all at once, too. I will request that we add detailed documentation with examples when we address the time resolution issue, though.

uvchik:
I would suggest the following steps.

1. I will warning if `DNI` is calculated from `GHI` and `DHI`, that this operation is not safe by now.
2. Finishing this PR with tests, documentation...
3. Let it merge by @wholmgren if everything works fine
4. Open a new PR to create a new function "safly_calculate_dni_from_ghi_and_dhi" (this might not be the final name though :smirk:). Implementing all ideas to avoid strange values and control them by parameters.
5. If the new function works we can discuss what would be a good default way to implement this in the ModelChain. Then we can remove the warning from step 1.

wholmgren:
I like the plan. The code mostly looks good, but let me know when you're ready for a few line by line comments.

uvchik:
You can pass a datetime index and an incomplete weather data set to the `complete_irradiance` method. Times and weather will be set as attribute so that you do not have to set them again. In that case you will be able to call the `run_model` method without any parameters.

```python
mc = ModelChain(system, location)
mc.complete_irradiance(times, weather=my_weather_data)
mc.run_model()
```

I soon will open a new PR to write a function that safely calculates `dni` from `ghi` and `dhi` (and so on) and integrate that function in the `complete_irradiance` method.

@wholmgren What do you think by now. Any corrections or improvements?

uvchik:
AppVeyor fails, but I think this is a general problem. I found the following issue.

https://github.com/conda-forge/hdf5-feedstock/issues/41

pvlib/modelchain.py
```diff
@@ -316,6 +317,10 @@ def __init__(self, system, location,
         self.losses_model = losses_model
         self.orientation_strategy = orientation_strategy
 
+        self.weather = pd.DataFrame()
```
wholmgren: Why make this a DataFrame instead of None (like the values below)? I'm not sure if one way is better than the other, but consistency might be more important.

pvlib/modelchain.py
```diff
@@ -602,32 +607,101 @@ def effective_irradiance_model(self):
             fd*self.total_irrad['poa_diffuse'])
         return self
 
-    def prepare_inputs(self, times, irradiance=None, weather=None):
+    def complete_irradiance(self, times, weather):
+        """
+        Determine the missing irradiation columns. Only two of the following
+        data columns (dni, ghi, dhi) are needed to calculate the missing data.
+
+        This function is not save at the moment. Results can be too high or
```
wholmgren: save --> safe

pvlib/modelchain.py
```diff
@@ -602,32 +607,101 @@ def effective_irradiance_model(self):
...
+        negative. Please contribute and help to improve this function on
+        https://github.com/pvlib/pvlib-python
+
+        Parameters
+        ----------
+        times : datetime index
```
wholmgren: DatetimeIndex

pvlib/modelchain.py
```diff
@@ -602,32 +607,101 @@ def effective_irradiance_model(self):
...
+        weather : pandas.DataFrame
+            Table with at least two columns containing one of the following data
+            sets: dni, dhi, ghi
+
+        Returns
+        -------
+        pandas.DataFrame
+            Containing the missing column of the data sets passed with the
+            weather DataFrame.
+
```
wholmgren: unneeded extra line here

pvlib/modelchain.py
```diff
@@ -602,32 +607,101 @@ def effective_irradiance_model(self):
...
+        """
+        self.weather = weather
+        self.times = times
+        self.solar_position = self.location.get_solarposition(self.times)
+        icolumns = set(self.weather.columns)
+        wrn_txt = "This function is not save at the moment.\n"
```
wholmgren: save --> safe

pvlib/modelchain.py
```diff
@@ -602,32 +607,101 @@ def effective_irradiance_model(self):
...
+        wrn_txt += "Results can be too high or negative.\n"
+        wrn_txt += "Help to improve this function on github."
```
wholmgren: Need a space or \n after the .

pvlib/modelchain.py
wholmgren: maybe better to write as 

```
wrn_text = ("text" +
            "text" +
             ...)
```

instead of using +=

pvlib/modelchain.py
```diff
@@ -602,32 +607,101 @@ def effective_irradiance_model(self):
...
+        # Add columns that does not exist and overwrite existing columns
+        # Maybe there is a more elegant way to do this. Any ideas?
+        if weather is not None:
+            self.weather = self.weather.combine_first(weather)
+            self.weather.update(weather)
+
+        # The following part could be removed together with the irradiance
+        # parameter at version v0.5 or v0.6.
+        # **** Begin ****
+        wrn_txt = "The irradiance parameter will be removed soon.\n"
```
wholmgren: same += comment as above.

pvlib/modelchain.py
```diff
@@ -602,32 +607,101 @@ def effective_irradiance_model(self):
...
 
-        Assigns attributes: times, solar_position, airmass, irradiance,
-        total_irrad, weather, aoi
+        Assigns attributes: times, solar_position, airmass, total_irrad, aoi
         """
-
-        self.times = times
+        # Add columns that does not exist and overwrite existing columns
+        # Maybe there is a more elegant way to do this. Any ideas?
+        if weather is not None:
```
wholmgren: Might be best for pvlib to only use a simple assignment. Users can do this in their own code before passing the parameter or making the assignment.

pvlib/modelchain.py
```diff
@@ -602,32 +607,101 @@ def effective_irradiance_model(self):
...
+        # The following part could be removed together with the irradiance
+        # parameter at version v0.5 or v0.6.
+        # **** Begin ****
+        wrn_txt = "The irradiance parameter will be removed soon.\n"
+        wrn_txt += "Please use the weather parameter to pass a DataFrame with "
+        wrn_txt += "irradiance (ghi, dni, dhi), wind speed and temp_air"
+        if irradiance is not None:
+            warnings.warn(wrn_txt, FutureWarning)
+            for column in irradiance.columns:
+                self.weather[column] = irradiance.pop(column)
```
wholmgren: why do you pop the column? I think this would mutate the input data, which is usually not a nice thing to do.

uvchik: 6a1a61dfd2

pvlib/modelchain.py
uvchik: 721743d972f4

pvlib/modelchain.py
uvchik: f702c73ef4dd

pvlib/modelchain.py
uvchik: 721743d97

pvlib/modelchain.py
uvchik: 721743d97

pvlib/modelchain.py
uvchik: 721743d97

uvchik:
This is how it works now. This could be an example, but it is not working due to undefined input data. A working example would be too overloaded in my opinion. I could add the following code to the docstring using skip_doctest or just leave it as an information here. It just shows that you have to pass the weather data and the index only once.

```python       
Examples
--------

>>> from pvlib.modelchain import ModelChain

>>> # my_weather containing 'dhi' and 'ghi'. 
>>> mc = ModelChain(my_system, my_location)
>>> mc.complete_irradiance(my_datetime_index, my_weather)
>>> mc.run_model()

>>> # my_weather containing 'dhi', 'ghi' and 'dni'. 
>>> mc = ModelChain(my_system, my_location)
>>> mc.run_model(my_datetime_index, my_weather)
```

uvchik:
Last comment (hopefully).

Using pandas.DataFrame, I do not understand why we pass the DatetimeIndex and the columns separately. In my applications it looks like this:

```python       
mc = ModelChain(my_system, my_location)
mc.run_model(my_weather.index, my_weather)
```

I think it is good to force the people to pass weather data with a working DatetimeIndex. If we do so we do not have to pass them separately.

But this might be a new discussion.

wholmgren:
I think adding your simple example to the docstring would be useful. Just let me know if I should merge it now or wait for the change.

Passing the times separately was needed to support the default clear sky and weather data. I am open to changing these behaviors in a separate issue/PR.

docs/sphinx/source/whatsnew/v0.4.2.txt
```diff
@@ -0,0 +1,24 @@
...
+
+* The run_model method of the ModelChain will use the weather parameter of all weather data instead of splitting it to irradiation and weather. The irradiation parameter still works but will be removed soon.(:issue:`239`)
+
+
+Enhancements
+~~~~~~~~~~~~
+
+* Adding a complete_irradiance method to the ModelChain to make it possible to calculate missing irradiation data from the existing columns [beta] (:issue:`239`)
+
+
```
wholmgren: Thanks for finding and fixing the typo in the ModelChain repr. You could add a note for that here, too.

uvchik: It was not clean to fix it within this PR but I was too lazy to do it separately. So I added this fix linking it to its commit. I guess there is no commit-role in your sphinx documentation similar to your issue-role, so it will not be linked.

uvchik:
I hope everything is fine now :smile: 

If you find typos etc. you can fix them you directly or let me know. Otherwise feel free to merge.

wholmgren:
Great, thanks so much! I'll check out the next PR soon.

uvchik:
The merge confused me but I think now everything is in order again.

Issue #242, opened 2016-09-12
lines in snlinverter table in docs doesn't wrap


mikofski:
the table in the docs for [`pvlib.pvsystem.snlinverter`](http://pvlib-python.readthedocs.io/en/latest/modules.html#pvlib.pvsystem.snlinverter) doesn't wrap, causing the text before and after the table and in the table itself to extend off the page.

see [tables in sphinx/docutils docs](http://www.sphinx-doc.org/en/stable/rest.html#tables) and [this SO Q/A](http://stackoverflow.com/questions/16570087/multiple-lines-of-text-in-single-cell-of-simple-table),

>simple table style does not support wrapping blocks. Use the grid style instead

wholmgren:
Thanks for pointing that out. You're already working on the fix, right???

dacoex:
I recommend using a CSV-table.
I found it easier to handle & to maintain.

Here, it does not wrap either but use a scrollbar:
http://pvlib-python.readthedocs.io/en/latest/variables_style_rules.html

Back when I wrote the page, I faced a similar issue.
It was due to the CSS of RTD.
See also:
http://stackoverflow.com/questions/34194728/csv-table-formatting-in-python-docstrings-sphinx-multiple-lines-in-one-cell

--> So check also against a "blank" sphinx project.

wholmgren:
I think we need to make the table inline if we want the function documentation to be properly rendered using help, ?, or ??

wholmgren:
@mikofski I'd prefer that we close this issue, but am ok with leaving it open if you think you'll return to it at some point.

mikofski:
@wholmgren PR #243 fixes this issue enough for me, so I also prefer to close this issue. Thanks!

PR #243, opened 2016-09-13
fix snlinverter and extraradiation doc string problems

wholmgren:
@mikofski here's a partial fix for #242. I tried using a fancy table but it didn't make a difference. So, I moved the simple table out of the parameter description and into the notes section, which does improve the rendering:

http://wholmgren-pvlib-python-new.readthedocs.io/en/fixsnldoc/modules.html#pvlib.pvsystem.snlinverter

Maybe something about the RTD theme? In any case, I think this is an improvement and will merge soon unless there are objections or better ideas.

Here's the fancy table format that I tried.

```
    +--------+------------------------------------------------------------+
    | Column |  Description                                               |
    +========+============================================================+
    | Pac0   | AC-power output from inverter based on input power         |
    |        | and voltage (W)                                            |
    +--------+------------------------------------------------------------+
    | Pdc0   | DC-power input to inverter, typically assumed to be equal  |
    |        | to the PV array maximum power (W)                          |
    +--------+------------------------------------------------------------+
    | Vdc0   | DC-voltage level at which the AC-power rating is achieved  |
    |        | at the reference operating condition (V)                   |
    +--------+------------------------------------------------------------+
    | Ps0    | DC-power required to start the inversion process, or       |
    |        | self-consumption by inverter, strongly influences inverter |
    |        | efficiency at low power levels (W)                         |
    +--------+------------------------------------------------------------+
    | C0     | Parameter defining the curvature (parabolic) of the        |
    |        | relationship between ac-power and dc-power at the          |
    |        | reference operating condition, default value of zero gives |
    |        | a linear relationship (1/W)                                |
    +--------+------------------------------------------------------------+
    | C1     | Empirical coefficient allowing Pdco to vary linearly       |
    |        | with dc-voltage input, default value is zero (1/V)         |
    +--------+------------------------------------------------------------+
    | C2     | Empirical coefficient allowing Pso to vary linearly with   |
    |        | dc-voltage input, default value is zero (1/V)              |
    +--------+------------------------------------------------------------+
    | C3     | Empirical coefficient allowing Co to vary linearly with    |
    |        | dc-voltage input, default value is zero (1/V)              |
    +--------+------------------------------------------------------------+
    | Pnt    | AC-power consumed by inverter at night (night tare) to     |
    |        | maintain circuitry required to sense PV array voltage (W)  |
    +--------+------------------------------------------------------------+
```

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.4.1.txt
+++ b/docs/sphinx/source/whatsnew/v0.4.1.txt
@@ -20,6 +20,7 @@ Documentation
 
 * Fixed a typo in the pvsystem.sapm returns description. (:issue:`234`)
 * Replaced nosetests references with py.test. (:issue:`232`)
+* Improve the rendering of the snlinverter doc string. (:issue:`242`)
 
 
 Code Contributors
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -43,8 +43,7 @@ def extraradiation(datetime_or_doy, solar_constant=1366.1, method='spencer',
 
     Parameters
     ----------
-    datetime_or_doy : int, float, array, date, datetime, datetime64,
-                      Timestamp, DatetimeIndex
+    datetime_or_doy : numeric, array, date, datetime, Timestamp, DatetimeIndex
         Day of year, array of days of year, or datetime-like object
 
     solar_constant : float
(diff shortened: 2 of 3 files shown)


mikofski:
This is great! Sorry I haven't been contributing.

dacoex:
@wholmgren seems like a RTD CSS issue. But haven checked with the native Sphinx theme.

wholmgren:
Tests are failing because of unrelated forecast module problems, so going ahead with the merge.

Issue #244, opened 2016-10-04
Location Timezone support for python3 newstr


lukewitmer:
Things have been quiet for a while. Hope the fall is going well for all of you so far! 

The following code 

```
from __future__ import (absolute_import, division,
                        print_function, unicode_literals)
from builtins import *
from pvlib.location import Location
tz = -10
str_tz = 'Etc/GMT+' + str(-1*int(tz))
my_loc = Location(20.0, -155.9, str_tz, 105.0, "my_place")
```

gives the following exception:

```
Traceback (most recent call last):
  File "/Users/lucas/Documents/Code/sim/pv.py", line 52, in <module>
    my_loc = Location(20.0, -155.9, str_tz, 105.0, "my_place")
  File "/usr/local/lib/python2.7/site-packages/pvlib/location.py", line 76, in __init__
    raise TypeError('Invalid tz specification')
TypeError: Invalid tz specification
```

The most straightforward fix is, in location.py, to check isinstance(tz, basestring) instead of str on line 66. This makes it compatible across python 2 and 3.

lukewitmer:
The fix is just one word in location.py, but I can't make the change as I seemingly don't have permission to push my branch for a PR. I'm fine if someone else simply makes the fix, but if I can have permission to contribute to branches for PRs, that would be nice :) thanks!

```
git -c diff.mnemonicprefix=false -c core.quotepath=false -c credential.helper=sourcetree push -v --tags --set-upstream origin refs/heads/Issue244-location-compatibility:refs/heads/Issue244-location-compatibility 
Pushing to https://github.com/pvlib/pvlib-python.git
remote: Permission to pvlib/pvlib-python.git denied to lukewitmer.
fatal: unable to access 'https://github.com/pvlib/pvlib-python.git/': The requested URL returned error: 403
Completed with errors, see above
```

wholmgren:
To maybe make things a little more clear for future readers, the problem is that in the example above, ``str_tz = 'Etc/GMT+' + str(-1*int(tz))`` is not actually a pure ``str``. Rather, the ``from __future__ import unicode_literals`` statement means that Python 2 will interpret that "string" as unicode (same behavior as the default in Python 3). In Python 2, this presents a problem for the Location class constructor since it's looking for ``str`` but getting ``unicode`` instead. 

[...]

PR #245, opened 2016-10-04
Issue244: Fix for isinstance str in python 3.

lukewitmer:
location timezone support for python3 newstr and unicode args

### DIFF
--- a/pvlib/location.py
+++ b/pvlib/location.py
@@ -63,7 +63,7 @@ class Location(object):
         self.latitude = latitude
         self.longitude = longitude
 
-        if isinstance(tz, str):
+        if isinstance(tz, basestring):
             self.tz = tz
             self.pytz = pytz.timezone(tz)
         elif isinstance(tz, datetime.tzinfo):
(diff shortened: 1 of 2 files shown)


wholmgren:
I think one problem is that basestring doesn't exist in Python 3. It looks like you can import six and use string_types as described here:

http://stackoverflow.com/questions/11301138/how-to-check-if-variable-is-string-with-python-2-and-3-compatibility

Might be some unrelated pandas 0.19 as well as forecast module problems, too.

lukewitmer:
Interesting. Good point on basestring not being in python 3. I'll take a more detailed look at it tonight or tomorrow, as well as actually testing against python3 (instead of in python2 with future imports which is what I tested before submitting the PR).

wholmgren:
I reran the travis tests for this PR now that #247 is merged. It looks like the only issue here is the basestring issue on Python 3 and hopefully the fix is as simple as advertised. I'll wait to release 0.4.1 if you can try that in the next 24 hours. Please also add a note and your name or username to the whatsnew file.

wholmgren:
@lukewitmer are you still interested in pursuing this fix or can we close it and #244? I'm trying to clean up some issues and PRs. I can also take a quick attempt at it if you check the "allow edits from maintainers" box at the bottom of the right most column. Thanks!

wholmgren:
@lukewitmer feel free to reopen if you want to finish this.

PR #249, opened 2016-10-17
Safely calculate dni from dhi and ghi.

uvchik:
This PR is a little bit confusing, because it works on top of PR #239. The changes of this PR start with commit 6907b6c. It will be easier to read once PR #239 is merged.

The aim of this PR is to safely calculate `DNI` from `GHI` and `DHI`. There are different approaches and we may collect them within this function. A `method` parameter (or something similar) can be used to switch between the different approaches.

As this PR is a spin-off from PR #239 there are already two comments regarding this PR.

@adriesse wrote:
> For your specific example, you can keep DNI values somewhat sane by not allowing them to be greater than a value obtained from a clear-sky model. Calculating the sun angles for the midpoint of the averaging interval is common practice (I hope). For long time steps like one hour, you can split the intervals that includes sunrise/set into two parts (one before and one after the event).

@cwhanse wrote:
> Ø Calculating the angles for the midpoint of the averaging interval is common practice (I hope). It ought to be common practice. SAM does it, as does PVsyst I think, but other applications? We ought to build this capability into pvlib I think, as part of model chain – a kwarg switch perhaps?

A rough approach is to use the horizontal value for low angles, as there might not be relevant irradiation at small angles (high zenith angles) or set it to zero.

```python
dni = (ghi - dhi) / tools.cosd(zenith)
dni[zenith > 88] = ghi[zenith > 88] - dhi[zenith > 88]
# or
dni[zenith > 88] = 0
```

1. What methods do we want to implement?
2. How do we want to switch between methods?

@birgits will help to implement this feature.


wholmgren:
Can you rebase this on the squashed commit created by merging #239? That will make it easier to review.

uvchik:
Replaced by #250.

PR #250, opened 2016-10-18
Safely calculate dni from dhi and ghi

uvchik:
This PR is a little bit confusing, because it works on top of PR #239. The changes of this PR start with commit 6907b6c. It will be easier to read once PR #239 is merged.

The aim of this PR is to safely calculate `DNI` from `GHI` and `DHI`. There are different approaches and we may collect them within this function. A `method` parameter (or something similar) can be used to switch between the different approaches.

As this PR is a spin-off from PR #239 there are already two comments regarding this PR.

@adriesse wrote:

> For your specific example, you can keep DNI values somewhat sane by not allowing them to be greater than a value obtained from a clear-sky model. Calculating the sun angles for the midpoint of the averaging interval is common practice (I hope). For long time steps like one hour, you can split the intervals that includes sunrise/set into two parts (one before and one after the event).

@cwhanse wrote:

> Ø Calculating the angles for the midpoint of the averaging interval is common practice (I hope). It ought to be common practice. SAM does it, as does PVsyst I think, but other applications? We ought to build this capability into pvlib I think, as part of model chain – a kwarg switch perhaps?

A rough approach is to use the horizontal value for low angles, as there might not be relevant irradiation at small angles (high zenith angles) or set it to zero.

``` python
dni = (ghi - dhi) / tools.cosd(zenith)
dni[zenith > 88] = ghi[zenith > 88] - dhi[zenith > 88]
# or
dni[zenith > 88] = 0
```
1. What methods do we want to implement?
2. How do we want to switch between methods?

@birgits will help to implement this feature.


wholmgren:
sorry for overlooking this @uvchik. I'd suggest some kind of filtering for the simple calculation but you could convince me that it's the user's responsibility. The comments above suggested to me that you were thinking of doing something more for time averages and sunrise/sunset transitions. Is that the case?

birgits:
Sorry for having neglected this PR for so long. I have been working on it and am still having some problems, but I will post what I've got so far at the beginning of next week.

> The comments above suggested to me that you were thinking of doing something more for time averages and sunrise/sunset transitions.

This is right. When calculating the DNI from GHI and DHI it may happen that for zenith angles close to 90°, so sunrise/sunset, the calculated DNI is unreasonable high or negative. The method we want to implement will "correct" those wrong values using different approaches.

birgits:
This is a first suggestion for how to correct calculated DNI values that are unreasonably high or even negative in sunrise/sunset transitions. For the dataset I used the two methods implemented right now give nearly the same results.
@uvchik and I decided that it would be good to let the user know for how many timesteps the DNI had to be corrected, maybe serving as an indication on how good the quality of the given weather dataset is. We also found out that if corrections were only conducted for sunrise or sunset hours, a likely reason is that the timezone is not correct, and decided to also give out a warning in that case.

adriesse:
I clicked on a bunch of the links but couldn't easily see what the main method was.

adriesse:
I am afraid that you could see problematic DNI values well before 88 degrees with real measurements due to cosine and alignment errors.  Maybe you  could make a gradual constraint that goes from the clear sky value at 80 degrees and zero at 90 degrees?  In real data you may also need to constrain the DNI to non-negative values.

birgits:
Thank you @adriesse for your comments and suggestions, I will try to implement them! 
I was also wondering, since you suggested to take the clearsky DNI as an upper bound to get "sane" values during sunrise/sunset transitions, if the clearsky DNI should always be higher than the calculated or measured DNI, as is the case for GHI? I couldn't really find anything on that.

adriesse:
Since the clear sky DNI you use here is from a model, it is going to represent some kind of average conditions, so certainly the true DNI could be a bit higher than the model values.  I suppose you could even add 10% to the model value and still achieve your objective of clipping unrealistic values.  

A separate comment: in some cases it might be preferable to replace suspicious values with nan's.

birgits:
> A separate comment: in some cases it might be preferable to replace suspicious values with nan's.

I could add a parameter that can be used to specify if values should be corrected or set to nan. What do you think? 

About setting the clearsky DNI as upper bound - to me it seems a bit random to add 10% to the modelled clearsky value, but maybe this could also be an optional parameter?

adriesse:
I think an option for nan's would be nice.  I have no opinion about the 10%.

wholmgren:
Thanks @birgits. I agree with @adriesse's comments. A couple of implementation things:

The function signature should use basic inputs such as clearsky_dni and zenith rather than a Location. This gives users more more flexibility, keeps the function as simple as possible, and makes it faster. I'm guessing that most users will already have that data available in their script or interpreter session, too.

I'd also prefer that we not add the timezone speculation code. Generally speaking, users have a responsibility to get their timezones right. However, we could consider a more robust, dedicated function for checking timezones in a separate issue/pr (there was also some talk of this in one of the IO package/module threads). 

It might help if you (and others) provide some examples of some real data that you're trying to clean up.

birgits:
> The function signature should use basic inputs such as clearsky_dni and zenith rather than a Location. 

I agree with you. I just thought it would be easier for the user to only have to create a location object and not have to worry about how to get the zenith and which clearsky model to use, but have the definition take care of that. It took me quite a while anyway to go through all the clearsky models and try to understand how I can use them and what inputs I needed and I was very happy when I found that the location object does all that for me already. But you are right, it will be more straight forward the other way.

birgits:
I tried to incorporate everything we previously discussed and am happy about any further suggestions and comments. I will start working on the tests now.

pvlib/irradiance.py
```diff
@@ -1945,3 +1945,106 @@ def _get_dirint_coeffs():
...
+    dni = dni_tmp.copy()
+    
+    # cutoff negative values
+    if set_to_nan:
+        dni[dni < 0] = float('nan')
+    else:
+        dni[dni < 0] = 0
+    
+    # correct unreasonable values
+    if method == 'clearsky':
```
wholmgren: Could we get rid of ``method`` and instead always apply a zenith cutoff and optionally apply a clear sky tolerance if clearsky_dni is supplied?

pvlib/irradiance.py
```diff
@@ -1945,3 +1945,106 @@ def _get_dirint_coeffs():
...
+        if set_to_nan:
```
wholmgren: I don't understand why there is different logic for the ``set_to_nan = True`` and ``set_to_nan = False`` cases. I expected a single if block that set a e.g. ``fill_value`` and then that value was applied to the same criteria. Am I missing something?

birgits:
Thank @wholmgren for your input! 

I uploaded a few days from the [reanalysis data](https://github.com/pvlib/pvlib-python/files/779528/irr_data.zip) that we use where the problem of negative or very high DNI values occurs. I also included the clearsky DNI, zenith and calculated DNI (without correction), all calculated using the pvlib, as well as a plot. I hope that this is what you were looking for @wholmgren?

pvlib/irradiance.py
birgits: Sure, this seems like a good option!

pvlib/irradiance.py
birgits: They should both be using the same logic, you are right. I only forgot to include the tolerance value in the `set_to_nan = True`, otherwise they are the same. But the `fill_value` way is more elegant so I'll change it.

pvlib/irradiance.py
```diff
@@ -1945,3 +1945,106 @@ def _get_dirint_coeffs():
...
+                    (clearsky_dni * clearsky_tolerance)
+    elif method == 'cutoff':
+        if set_to_nan:
+            dni[(zenith > 88) & (dni != 0)] = float('nan')
+        else:
+            dni[zenith > 88] = 0
+    else:
+        raise ValueError('Invalid method: %s', method)
+
+    # debugging info how many values were corrected
```
birgits: Sure, I can also take it out if you like. I think it is a nice debugging feature, though. Maybe I could reduce it down to 2-3 lines?

pvlib/irradiance.py
uvchik: What about replacing the `method` parameter by a numeric parameter.

```diff
- dni[(zenith >= 80.0) & (zenith <= 90.0) & (dni > clearsky_dni)] = float('nan')
+ new_parameter = 80.0
+ dni[(zenith >= new_parameter) & (zenith <= 90.0) & (dni > clearsky_dni)] = float('nan')

- dni[(zenith > 88) & (dni != 0)] = float('nan')
+ another_new_parameter = 88
+ dni[(zenith > another_new_parameter) & (dni != 0)] = float('nan')
```

I do not have a name for the new parameters though :flushed: but you can set the value to `90` and nothing will happen. If it saves time we could check whether the value is `90` and skip the lines.

birgits:
I took out the different methods and instead added two new parameters for the zenith angles between which the unreasonable values are set to NaN or corrected, as suggested by @wholmgren and @uvchik. What do you think?

A few things I want to point out:
* I chose the default values for lower_cutoff_zenith and upper_cutoff_zenith from an early comment by @adriesse. They are open for discussion.
* Negative DNI values as well as nonzero values above the upper_cutoff_zenith (above 88 for example) are set to NaN. Before there was also the option to set them to zero.
* In the modelchain the function irradiance.dni is now by default called with the calculated clearsky_dni.

uvchik:
1. I like the dni-function.

2. It might be controversial but I would remove the `**kwargs` attribute from both functions (`dni` and  `complete_irradiance`) and change the following lines in `complete_irradiance`:

```diff
-                clearsky_dni=kwargs.get('clearsky_dni',
-                                        self.location.get_clearsky(times).dni),
-                clearsky_tolerance=kwargs.get('clearsky_tolerance', 1))
+                clearsky_dni=self.location.get_clearsky(times).dni),
+                clearsky_tolerance=1))
```
People should use this function or do their own stuff.

3. A NaN to Zero is not important for me. Anyway we might want to add it to the `ModelChain` but not to the `dni` function.

4. You have to rebase the actual pvlib-dev branch on top of yours to solve the conflict.

birgits:
Thank you @uvchik for your comments. @wholmgren and @adriesse do you have any more comments or suggestions?

wholmgren:
I am generally ok with these additions. I'll make minor comments later this week.

pvlib/irradiance.py
```diff
@@ -2048,3 +2048,70 @@ def _get_dirint_coeffs():
...
+        tolerance by how much the calculated DNI value can be greater than
+        the clearsky value before it is identified as an unreasonable value.
+        Default: 1.
+
+    upper_cutoff_zenith : float
+        Zenith angle above which nonzero DNI values are set to NaN.
+        Default: 88.
+
+    lower_cutoff_zenith : float
+        Zenith angle above which DNI values greater the clearsky DNI (plus
```
wholmgren: "plus" should be "times"

pvlib/irradiance.py
```diff
@@ -2048,3 +2048,70 @@ def _get_dirint_coeffs():
...
+
+    Returns
+    -------
+    dni : Series
+        The modeled direct normal irradiance.
+
+    """
+
+    # calculate DNI
+    dni_tmp = (ghi - dhi) / tools.cosd(zenith)
```
wholmgren: why do you need dni_tmp and dni? maybe a holdover from a previous implementation?

pvlib/irradiance.py
```diff
@@ -2048,3 +2048,70 @@ def _get_dirint_coeffs():
...
+    dni[dni < 0] = float('nan')
+
+    # set non-zero DNI values for zenith angles >= upper_cutoff_zenith to NaN
+    dni[(zenith >= upper_cutoff_zenith) & (dni != 0)] = float('nan')
+
+    # correct DNI values for zenith angles between lower_cutoff_zenith and
+    # upper_cutoff_zenith that are greater than the clearsky tolerance
+    if clearsky_dni is not None:
+        dni[(zenith >= lower_cutoff_zenith) & (zenith <= upper_cutoff_zenith) &
+            (dni > (clearsky_dni * clearsky_tolerance))] = (clearsky_dni *
```
wholmgren: You're doing the same calculation twice, so it may be slightly faster and slightly more readable if you define and use ``max_dni = clearsky_dni * clearsky_tolerance``.

pvlib/irradiance.py
```diff
@@ -2048,3 +2048,70 @@ def _get_dirint_coeffs():
...
+
+    # calculate DNI
+    dni_tmp = (ghi - dhi) / tools.cosd(zenith)
+    dni = dni_tmp.copy()
+
+    # cutoff negative values
+    dni[dni < 0] = float('nan')
+
+    # set non-zero DNI values for zenith angles >= upper_cutoff_zenith to NaN
+    dni[(zenith >= upper_cutoff_zenith) & (dni != 0)] = float('nan')
```
wholmgren: The doc string suggested to me that the comparison should be strictly greater than, rather than greater than or equal to.

pvlib/irradiance.py
```diff
@@ -2048,3 +2048,70 @@ def _get_dirint_coeffs():
...
+
+    # correct DNI values for zenith angles between lower_cutoff_zenith and
+    # upper_cutoff_zenith that are greater than the clearsky tolerance
+    if clearsky_dni is not None:
+        dni[(zenith >= lower_cutoff_zenith) & (zenith <= upper_cutoff_zenith) &
```
wholmgren: same as above regarding comparison type

pvlib/irradiance.py
```diff
@@ -2048,3 +2048,70 @@ def _get_dirint_coeffs():
         [0.743440, 0.592190, 0.603060, 0.316930, 0.794390]]
 
     return coeffs[1:, 1:, :, :]
+
+
+def dni(ghi, dhi, zenith, clearsky_dni=None, clearsky_tolerance=1,
```
wholmgren: Would we be better off with a more descriptive name? ``dni_from_components``? I'm not sure.

pvlib/irradiance.py
```diff
@@ -2048,3 +2048,70 @@ def _get_dirint_coeffs():
...
+        upper_cutoff_zenith=88.0, lower_cutoff_zenith=80.0, **kwargs):
```
wholmgren: no kwargs here

pvlib/irradiance.py
```diff
@@ -2048,3 +2048,70 @@ def _get_dirint_coeffs():
...
+    lower_cutoff_zenith : float
+        Zenith angle above which DNI values greater the clearsky DNI (plus
+        allowed tolerance) are corrected. Only applies if 'clearsky_dni' is not
+        None. Default: 80.
+
+    Returns
+    -------
+    dni : Series
+        The modeled direct normal irradiance.
+
```
wholmgren: no extra space needed

pvlib/modelchain.py
```diff
@@ -623,7 +623,7 @@ def effective_irradiance_model(self):
             fd*self.total_irrad['poa_diffuse'])
         return self
 
-    def complete_irradiance(self, times=None, weather=None):
+    def complete_irradiance(self, times=None, weather=None, **kwargs):
```
wholmgren: as discussed elsewhere, probably no kwargs here, too.

pvlib/irradiance.py
birgits: I find the name dni_from_components a bit confusing since components could mean different things. Maybe we just name it after what it does `dni_from_ghi_and_dhi`?

pvlib/irradiance.py
birgits: Would that work and be correct English? 

> Zenith angle from which on nonzero DNI values are set to NaN.

pvlib/irradiance.py
birgits: Done

pvlib/irradiance.py
birgits: Done

pvlib/irradiance.py
birgits: Done

pvlib/irradiance.py
birgits: Done

pvlib/modelchain.py
birgits: Done. 
I had to set a default value for the clearsky_tolerance and chose 1.1, again according to an early comment from @adriesse. Would you suggest to have the clearsky_tolerance as an input into the complete_irradiance function, so it can be changed by the user or should the user write his/her own function in case he/she wants to change those parameters?

adriesse:
Good work.  It occurs to me that the upper and lower cutoff names might not be entirely intuitive. How about: zenith_threshold_for_clearsky_limit, zenith_threshold_for_nan or something like that?

pvlib/irradiance.py
uvchik: According to `aoi` we could keep `dni`, but I like `get_dni`or `calculate_dni` definitely better than `dni_from_something`. We haven't used it so far and in my opinion the parameters should be part of the API documentation and not of the name.

birgits:
Sounds good @adriesse! How about `zenith_threshold_for_zero_dni` for the `upper_cutoff_zenith,` since the DNI for zenith angels above that value should be zero?

birgits:
And I don't know what to do about the failing tests...  All tests pass on my computer. Does anyone know what's going wrong?

wholmgren:
see #326 regarding the failing tests. I will merge the fix and retrigger the build.

pvlib/irradiance.py
wholmgren: Seems that ``dni`` is moderately acceptable to all.

pvlib/irradiance.py
wholmgren: Zenith angles greater than or equal to upper_cutoff_zenith will be set to NaN.

birgits:
Thank you fixing the failing tests @wholmgren! Do you think the PR is ready to be merged?

wholmgren:
Looks good to me. Can you add a note and your name to the 0.4.5 whats new file?

wholmgren:
thanks @birgits and @uvchik!

PR #260, opened 2016-11-12
Improve precision of v_from_i for large arguments

cwhanse:
Number of iterations of Newton method should increase with the order of
the argument, to maintain precision.  Not expected to have a significant
effect on most calculated IV curves.


wholmgren:
Looks like the tests are failing for two reasons: 1. ``order`` needs to be a scalar, and 2. ``order`` needs to be an integer. Something like this might work

```
order = np.floor(np.max(np.log10(w))).astype('int')
```

Finally, please add a note and your name to the docs/sphinx/source/whatsnew/v0.4.2.txt file too.

cwhanse:
Closing this to submit a clean set of files

PR #266, opened 2016-11-16
260

cwhanse:
Fix typing, add whatsnew note

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.4.2.txt
+++ b/docs/sphinx/source/whatsnew/v0.4.2.txt
@@ -41,6 +41,7 @@ Enhancements
 * Added name attribute to ModelChain and PVSystem. (:issue:`254`)
 * Restructured API section of the documentation so that there are
   separate pages for each function, class, or method. (:issue:`258`)
+* Improved convergence of lambertw function (:issue: '260')
 
 
 Code Contributors
@@ -49,3 +50,4 @@ Code Contributors
 * Uwe Krien
 * Will Holmgren
 * Volker Beutner
+* Cliff Hansen
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1843,7 +1843,7 @@ def v_from_i(resistance_shunt, resistance_series, nNsVth, current,
     # evaluation (above) results in NaN from overflow, 3 iterations
     # of Newton's method gives approximately 8 digits of precision.
     w = logargW
-    order = np.log10(w)
+    order = np.ceil(np.log10(w)).astype(int)
     for i in range(0, 3*order):
         w = w * (1 - np.log(w) + logargW) / (1 + w)
     lambertwterm_log = w
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1843,7 +1843,8 @@ def v_from_i(resistance_shunt, resistance_series, nNsVth, current,
     # evaluation (above) results in NaN from overflow, 3 iterations
     # of Newton's method gives approximately 8 digits of precision.
     w = logargW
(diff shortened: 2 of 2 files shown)


mikofski:
@cwhanse why not use [`scipy.optimize.newton`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.newton.html#scipy.optimize.newton) here in [`pvsystem.v_from_i` on line 1846](../../pvlib/pvlib-python/blob/master/pvlib/pvsystem.py#L1841-L1848)?

```python
from scipy.optimize import newton
# w+log(w)=logargW. The initial guess is w=logargW. Where direct
# evaluation (above) results in NaN from overflow, 3 iterations
# of Newton's method gives approximately 8 digits of precision.
lambertwterm_log = newton(lambda w: w + np.log(w) - logargW, logargW)
```

cwhanse:
This was transcribed from Matlab.  No reason not to use the scipy function, but I doubt there's any performance advantage - after I fix the failed checks, of course.

mikofski:
True, I checked, performance at least for the test case is exactly the same. But IMO `scipy.optimize.newton` will be easier to maintain. For example, you wouldn't have to update the number of iterations to depend on order as you did in #260

wholmgren:
I think the newton function will only work with scalar inputs. I'm guessing that it's more efficient to do it Cliff's original way for array input of non-trivial length. 

I did not know you could highlight a block of code like that in a github link!

mikofski:
I chose `newton` specifically because this example uses scalar inputs. The equivalent multivariate solver is [`scipy.optimize.fsolve`](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.optimize.fsolve.html#scipy.optimize.fsolve) based on [MINPACK](https://en.wikipedia.org/wiki/MINPACK)’s `hybrd` and `hybrj` which is a modified version of [Powell's method](https://en.wikipedia.org/wiki/Powell's_method) and the Newton-Raphson methods.

From the GNU Scientific Library:
>This is a modified version of Powell’s Hybrid method as implemented in the HYBRJ algorithm in MINPACK. Minpack was written by Jorge J. Moré, Burton S. Garbow and Kenneth E. Hillstrom. The Hybrid algorithm retains the fast convergence of Newton’s method but will also reduce the residual when Newton’s method is unreliable. The algorithm uses a generalized trust region to keep each step under control. 

```python
from scipy import optimize
# w+log(w)=logargW. The initial guess is w=logargW. Where direct
# evaluation (above) results in NaN from overflow, 3 iterations
# of Newton's method gives approximately 8 digits of precision.
lambertwterm_log = optimize.fsolve(
    func=lambda w: w + np.log(w) - logargW,
    x0=logargW,
    fprime=lambda w: 1 + 1/w)
```

If even more flexibility is needed, then [`scipy.optimize.root`](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.optimize.root.html#scipy.optimize.root) is the general interface to multivariate equation solvers. You can specify the `method` argument, EG: `lm` is Levenberg-Marquardt for least square fitting of non-square systems of equations.

@cwhanse the equivalent of this in MATLAB would be the Optimization Toolbox's `fsolve` which I re-implemented in [the MATLAB FEX Newton-Raphson solver](https://www.mathworks.com/matlabcentral/fileexchange/43097-newton-raphson-solver)

mikofski:
I'm okay with it as it is:

1. I don't want to block merging it,
2. it is working well and is identical to the SciPy methods,
3. and it's already quite compact and short,

but IMO I think that mature, well established, easily available methods are usually more transparent to users and easier to maintain than custom re-implementations.

cwhanse:
Closing this to submit a clean pull request

PR #270, opened 2016-11-30
initial io module prototype

dacoex:
prototype for #261 & #29 (can these issues be merged?)

The data is here:
cd MY-GITHUB-CLONES
git clone https://github.com/dacoex/pvlib_data


adriesse:
Thanks for submitting this.  The file maccrad format doesn't look like it was designed for ease of parsing.

I can offer a couple of comments from looking at the file and code:

The header section of the maccrad file looks like it could be of variable length, so you probably don't want to hardcode the 40.  The leading # can be used to let read_csv skip the variable header section, but unfortunately the column names would be lost that way too.  Is the number of columns always the same?  In that case I would suggest using header=None and providing a standard name for each column.

Your dateparser looks like it returns text.  Does this produce a datetime index for the dataframe?

I the time zone for the data in the file?  "# Time reference: Universal time (UT)"  Maybe the dataframe could be localized with this?

dacoex:
Let's first start with a general feedback regarding design of io module.
I moved all comments related to the actual macc-rad format to #271
(Soda-is changed their website again after these example files were downloaded.)

- Does the sub-module structure  correspond to the intentions discussed in 261 & #29?
of the if the file structures correspond 
- Is there anything that shall be changed to facilitate the addition of new file formats?

I will use #271 to finalise the first input format and as we go improve design of the io module

wholmgren:
Thanks @dacoex. Please remake your macc-rad PR against the new io branch discussed in #272.

dacoex:
closed in favour of #272 

maccrad no in #271

PR #272, opened 2016-12-01
make iotools package

wholmgren:
I made a new ``io`` branch to try to streamline the process of multiple users contributing their new io tools and to make it easier for all of us to test it them without hurting the master branch. I also went ahead and moved the tmy readers into the package and updated the tests accordingly. I named the package ``iotools`` since Cliff expressed a preference for that name, but it's not yet set in stone. I think this is mostly consistent with what we talked about in #261 and #29. Let me know if I missed something concerning the structure.

Please make your new IO tools pull requests against this branch rather than master.

Finally, I'm not going to merge into master until we're ready to make a series of changes to the API. I want to do at least 0.4 release with non-API breaking changes before then.

Closes #29. Closes #261.

Issue #271, opened 2016-12-01
io module: add macc-rad reader


dacoex:
**Upstream data source**
http://www.soda-pro.com/web-services/radiation/cams-radiation-service

**Git repository with examples**
https://github.com/dacoex/pvlib_data/tree/master/MACC-RAD
(Soda-pro website was changed after the example data was downloaded.)

**Related**
Issues: #261 & #29
PR: #270 

**Previous comments**
taken from:
https://github.com/pvlib/pvlib-python/pull/270#issuecomment-264102908

> The header section of the maccrad file looks like it could be of variable length, so you probably don't want to hardcode the 40. The leading # can be used to let read_csv skip the variable header section, but unfortunately the column names would be lost that way too. Is the number of columns always the same? In that case I would suggest using header=None and providing a standard name for each column.
> 
> Your dateparser looks like it returns text. Does this produce a datetime index for the dataframe?
> 
> I the time zone for the data in the file? "# Time reference: Universal time (UT)" Maybe the dataframe could be localized with this?

dacoex:
>The file maccrad format doesn't look like it was designed for ease of parsing.

It says that there is some standard befind:

_ Encoding partly from D2.8.III.13-14 INSPIRE Data Specification on Atmospheric Conditions and Meteorological Geographical Features – Technical Guidelines, 2013-12-10; and CF (Climate and Forecast) metadata (2013-11-11) CF Standard Names registry of ObservablePropertyValue_

>Is the number of columns always the same?

Yes.

>Your dateparser looks like it returns text. Does this produce a datetime index for the dataframe?

Yes.

>I the time zone for the data in the file? "# Time reference: Universal time (UT)" Maybe the dataframe >could be localized with this?

Thanks, I'd forgot but now implemented.
The functions also add metadata to the dataframe (df.df_name = "MACC-RAD v2.7 Solar Irradiation") and set a default location name (loc.name = "MACC-RAD v2.7 Solar Irradiation @ lat (deg. N), lon (deg. E): 44.083, 5.059")

If the file structure & functions settle down I will add 
- io module documentation
- whatsnew

wholmgren:
Please add the discussion above to the new PR that I requested in #271, and then close the issue.

dacoex:
closed in favour of #274

PR #274, opened 2016-12-02
iotools component for maccrad

dacoex:
See: #272 for reason.

Previous discussion in #271 

@wholmgren I hope I got this git thing right... But anyway, we are in a separate branch than master.


wholmgren:
Thanks. Some general comments:

I'd prefer that we tackle one format per pull request. This will make each one go faster and be less likely to become stale and neglected. 

It looks like you're duplicating the new code in the pvlib-python/io branch instead of adding to it and that's what led to the merge conflict. To build on what's already in the pvlib-python/io branch, you'll need to fetch pvlib-python's io branch and then rebase your changes on it (or merge it into your branch). Depending on your git experience, it might be easier to make a fresh clone, checkout the io branch, and copy over the files related to macc or pvsyst. 

There are probably too many data files and they're probably too long. We only want to include the bare minimum of data required to test the code and to let users run an example or two. 

The ``iotools/iotools.py`` module should be renamed ``iotools/util.py`` to avoid confusion.

I don't think we should create a dependency on the timezonefinder package. That's probably best left to users to implement themselves, if desired.

PR #280, opened 2016-12-05
iotools: reader for pvsyst (hourly)

dacoex:
a separate PR for hourly pvsyst output as requested in https://github.com/pvlib/pvlib-python/pull/274#issuecomment-264486119

- only 1 example file
- renamed to util


wholmgren:
Also interested in the feature, but this PR would need to be reopened against the master branch and might need some coordination with other pvsyst-related features that have been implemented/discussed.

Issue #282, opened 2016-12-08
Implement dirindex decomposition model


mikofski:
The DIRINDEX model is a variation of the DIRINT model for decomposition that considers the clear sky index. Reference Daryl Myers Solar Radiation CRC press, chapter 5.

wholmgren:
go for it.

wholmgren:
closed by #283.

PR #283, opened 2016-12-12
Implement dirindex decomposition model

anomam:
* should address #282 posted by @mikofski

### DIFF
--- a/pvlib/test/test_irradiance.py
+++ b/pvlib/test/test_irradiance.py
@@ -371,11 +371,17 @@ def test_erbs_all_scalar():
 
 
 def test_dirindex():
-    times = pd.DatetimeIndex(['2014-06-24T12-0700', '2014-06-24T18-0700'])
-    ghi = pd.Series([1038.62, 254.53], index=times)
-    zenith = pd.Series([10.567, 72.469], index=times)
+    clearsky_data = tus.get_clearsky(times, model='ineichen',
+                                     linke_turbidity=3)
+    ghi = pd.Series([0, 0, 1038.62, 254.53], index=times)
+    ghi_clearsky = clearsky_data['ghi']
+    dni_clearsky = clearsky_data['dni']
+    zenith = ephem_data['apparent_zenith']
     pressure = 93193.
-    dirindex_data = irradiance.dirindex(ghi, zenith, times, pressure=pressure,
-                                        temp_dew=10)
+    tdew = 10.
+    dirindex_data = irradiance.dirindex(ghi, ghi_clearsky, dni_clearsky,
+                                        zenith, times, pressure=pressure,
+                                        temp_dew=tdew)
 
-    print 'DIRINDEX data: ', dirindex_data
+    print 'DIRINDEX data: '
+    print dirindex_data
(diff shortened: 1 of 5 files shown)


wholmgren:
looks great @anomam. ping me when your test is working. 

also, please go ahead and set up a new v0.4.3.txt whatsnew file and entry in the [whatsnew.rst](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/whatsnew.rst) file.

anomam:
Thanks @wholmgren . Will do!
By the way, I was wondering if you would know where I could find some (validation) data about the DIRINDEX model that could help me make sure that  my implementation is correct?

wholmgren:
I do not have any data or a reference implementation for DIRINDEX, but maybe someone else can comment. I think it's a straightforward extension of DIRINT, though, so I'm not sure if it's really necessary.

wholmgren:
oh, and also add your new function to the api.rst file

pvlib/test/test_irradiance.py
```diff
@@ -365,3 +368,29 @@ def test_erbs_all_scalar():
...
+    pressure = 93193.
+    tdew = 10.
+    out = irradiance.dirindex(ghi, ghi_clearsky, dni_clearsky,
+                              zenith, times, pressure=pressure,
+                              temp_dew=tdew)
+    dirint_close_values = irradiance.dirint(ghi, zenith, times,
+                                            pressure=pressure,
+                                            use_delta_kt_prime=True,
+                                            temp_dew=tdew).values
+    expected_out = np.array([np.nan, 0., 748.31562753, 630.72592644])
```
wholmgren: Does this nan arise from 0/0? If so, I think most users would probably prefer to get a 0 here instead. Something like ``dni_dirindex[(dni_dirindex < 0) | dni_dirindex.isnan()] = 0.`` should work.

pvlib/test/test_irradiance.py
```diff
@@ -365,3 +368,29 @@ def test_erbs_all_scalar():
...
         assert_allclose(v, expected[k], 5)
+
+
+def test_dirindex():
+    clearsky_data = tus.get_clearsky(times, model='ineichen',
+                                     linke_turbidity=3)
+    ghi = pd.Series([0, 0, 1038.62, 254.53], index=times)
+    ghi_clearsky = clearsky_data['ghi']
+    dni_clearsky = clearsky_data['dni']
+    zenith = ephem_data['apparent_zenith']
```
wholmgren: Your test is going to be more robust you hard code these input values. The values themselves are fine, it's only that they're derived at runtime from functions that could produce slightly different results in the future and thus might make your new test fail for no good reason. You'll can see that the inputs are precisely defined in the dirint tests that actually check values.

anomam: Got it!

pvlib/test/test_irradiance.py
anomam: It comes from the relative airmass function and its use in the dirint model from what I saw. I think that ``airmass`` returns ``NaN`` for night times.

wholmgren: Oh right, thanks for reminding me. My suggestion was bad anyways since it would have yielded 0s for any nan inputs.

PR #284, opened 2016-12-13
implement detect_clear function

wholmgren:
This PR adds a ``clearsky.detect_clearsky`` function. The function is a port of the matlab library's ``pvl_detect_clear_times``. The main differences are: 

* no support for unequal times -- you'll need to resample your irregular data before passing it to the function, or you can implement the feature yourself!
* automatically determines sample_interval
* requires a reference clear sky series instead calculating one from a user supplied location and UTCoffset. I suggest that users that want functionality closer to the matlab function should implement a ``Location.detect_clearsky`` method.
* parameters are controllable via keyword arguments instead of hardcoded
* option to return individual test components

@cwhanse provided me with a short test file and expected output from the matlab function. We do not think that it covers all cases, but it's the best we have for now. Cliff has also been a big help in tracking down some issues with my initial implementation and helping me understand how the function works. Thanks!

I hope to merge this by the end of the week, so please let me know sooner than later if you have comments or concerns. Maybe an 0.4.3 release with just this and #283?

to do:
- [x] test for NotImplementedError for irregular data
- [x] probably need tests for more of the arguments
- [x] test that a RuntimeWarning is generated if the clear sky scaling for loop fails to break
- [x] ensure that documentation example works
- [x] update documentation to make input names more general. State that it was designed for GHI but can work with other inputs if the filter settings are adjusted.
- [x] update whatsnew (waiting on #283 for git simplicity)


wholmgren:
Some examples from the NREL MIDC OASIS irradiance sensor here at the U. Arizona campus. 

```python
location = pvlib.location.Location(32.229961, -110.95388, altitude=741, name='OASIS GHI', tz='US/Arizona')

# data available at http://www.nrel.gov/midc/ua%5Foasis/
# import not shown here
ghi = data.resample('1min', how='first')
cs = location.get_clearsky(ghi.index)

clear_samples = pvlib.clearsky.detect_clearsky(ghi, cs['ghi'], cs.index, 10)

cs['ghi'].plot(label='clear sky profile')
ghi.plot(label='measured ghi')
ghi[clear_samples].reindex(ghi.index).plot(label='detected clear')
plt.legend(ncol=3)
plt.ylabel('irradiance W/m**2')
plt.ylim(0, None)
plt.title(location.name)
```

wx2o3h1aheq7aaaaabjru5erkjggg
wfnr0dfych1jaaaaabjru5erkjggg
ffp1xiuwcrnkmtbblgktzjkwqzyje2szmn0sd5tvi8snoj4c5ihxaf gqufbd yrzzjhgiwtjsgyyyiseybimmgclbomm vhk214kgbarc4dblfvwnglgz9zjkyqzciewszmkgxcbfkmtjblwnszpgyunpoexhhjjdhgggomiy9elhsyy4wxxhhjjdemrqxjyiwxxhhjjdhggmcabmyyy4wxxhhjjgmzjoexxhhjjdhgggmaaxiyy4wxxh

erased99:
Hi, I'm trying to replicate the code for a college assignment, but I can't get a similar result. I'm actually just getting errors. I would like to know if it is possible to help me? (And yes, I'm new to programming).

kandersolar:
Hi @erased99, this github repository is for the development of pvlib-python.  Questions about PV modeling or programming should go to a forum like StackOverflow, for example using the pvlib tag: https://stackoverflow.com/questions/tagged/pvlib

Issue #160, opened 2016-05-06
port pvl_adrinverter


wholmgren:
This is a relatively simple function, so it should be easier to port.

Including the database can be as simple as adding it to `pvlib/data` and adding it to the `manifest`. Ideally we'd convert the database to something other than .mat (.csv or .h5), but that can also be done later.

This was previously mentioned in #2, but I want to highlight it here. 

cc @adriesse

adriesse:
A bit busy with other things at the
      moment, but feel free to ask questions about this.  
      
      Have a great workshop everyone!
      
      Anton
      
      
      On 2016-05-07 0:31, Will Holmgren wrote:
    
    
      This is a relatively simple function, so it should be easier to
        port.
      Including the database can be as simple as adding it to pvlib/data
        and adding it to the manifest. Ideally we'd
        convert the database to something other than .mat (.csv or .h5),
        but that can also be done later.
      This was previously mentioned in #2,
        but I want to highlight it here. 
      cc @adriesse
      —
        You are receiving this because you were mentioned.
        Reply to this email directly or view
          it on GitHub

VolkrB:
Hi all, 
I took the liberty of taking the Driesse inverter database .mat file and converting it to a pd.df and then to .csv. 

Question: I noticed that some of the inverters dont have model names, but do have the manufacturer and I can't directly check the .mat file because I don't have matlab installed here. (I converted the .mat with scipy)

Example: I found that there is some data where the manufacturer, Dow Chemical, doesn't have a model name. Could someone confirm this.

I want to be sure that my conversion went smoothly before I pull request

Thanks!

adriesse:
Thanks for the initiative!  I see four inverters from Dow Chemical without model descriptions in my source files.

VolkrB:
Great! That matches up with what I have. Going to see if I can get this csv to be compatible with pvsystem.retrieve_sam() and then port the pvl_adrinverter function. Let me know if you guys have any thoughts or concerns about this

adriesse:
I'll support your efforts.

wholmgren:
Awesome. In the short term, I agree that it makes sense to add a csv file to the ``data`` folder and ``name == 'adrinverter'`` option to the ``retrieve_sam`` function. I suppose someone might object to putting this in a SAM reader, but I don't think it's a big deal. In the long term, we should consider moving this kind of functionality to the new iotools subpackage (#261 and links therein).

[...]

PR #288, opened 2017-01-04
port pvl_adrinverter

VolkrB:
#160  In short:

- Ported mat -> csv adrinverter data file

pvsystem.py
- included import of csv in pvsystem.py retrieve_sam function
- added adrinverter function (two lines are a bit ugly, because i am storing coefficients in a string instead of making a column for each)

modelchain.py
- added adrinverter function to wrapper

test_modelchain.py
- removed xfail for adrinverter
- added tests using csv data (Zigor inverter)
(had to also change the single diode test, since i replaced cecinverter with adrinverter)

test_pvsystem.py
- added test for adrinverter function

#### Need to document after everything looks good

I've compared inverter data from the adrinverter function to the sandiainverter function to see if, at least, the results are of the same magnitude and follow a similar trend. They do, however would be nice to see results from the matlab version to test against.

let me know what i can change/improve etc


pvlib/pvsystem.py
```diff
@@ -1122,6 +1125,9 @@ def retrieve_sam(name=None, path=None):
         elif name == 'sandiamod':
             csvdata = os.path.join(
                 data_path, 'sam-library-sandia-modules-2015-6-30.csv')
+        elif name == 'adrinverter':
+            csvdata = os.path.join(
```
wholmgren: looks like this might fit on a single line that's less than 80 characters.

pvlib/pvsystem.py
```diff
@@ -2037,6 +2043,37 @@ def snlinverter(v_dc, p_dc, inverter):
     return ac_power
 
 
+def adrinverter(v_dc, p_dc, inverter):
+
+    ce_clean = [','.join(inverter['ADRCoefficients'][1:-1].split())]
+    ce_list = np.array([float(s) for s in ce_clean[0].split(',')])
```
wholmgren: ``inverter['ADRCoefficients']`` should be parsed at import time

pvlib/test/test_modelchain.py
```diff
@@ -33,8 +33,8 @@ def cec_dc_snl_ac_system(sam_data):
     module_parameters['b'] = 0.05
     module_parameters['EgRef'] = 1.121
     module_parameters['dEgdT'] = -0.0002677
-    inverters = sam_data['cecinverter']
-    inverter = inverters['ABB__MICRO_0_25_I_OUTD_US_208_208V__CEC_2014_'].copy()
+    inverters = sam_data['adrinverter']
```
wholmgren: maybe I'm forgetting something, but don't we want a new test here instead of changing the existing test?

pvlib/data/adr-library.csv
```diff
@@ -0,0 +1,1762 @@
+Name,Manufacturer,Model,Source,Vac,Vintage,Pacmax,Pnom,Vnom,Vmin,Vmax,ADRCoefficients,Pnt,Vdcmax,Idcmax,MPPTLow,MPPTHi,TambLow,TambHi,Weight
+0,0,0,0,0.0,0.0,0.0,0,0,0,0,0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
```
wholmgren: are these 2 lines of 0s here because of the ``_parse_raw_sam_df`` function's ``skip_rows`` argument? I think we should keep this file clean and make the pvlib parser smarter. I don't remember the logic for that parser (I don't even remember if I wrote it).

pvlib/test/test_modelchain.py
wholmgren: on a second read, maybe this is ok because all of the methods are still being tested in some way. Is that right?

pvlib/pvsystem.py
VolkrB: sure

pvlib/pvsystem.py
VolkrB: Yea, seems like that's where it should be

pvlib/test/test_modelchain.py
VolkrB: Yea they are, I think the coverage is the same- this was something I was debating as well.

pvlib/data/adr-library.csv
VolkrB: Yea exactly. I could write a conditional statement in ```_parse_raw_sam_df``` to only skip rows for certain csvs, but maybe you can think of something better than that?

wholmgren: I'm struggling to think of a good solution for this and the array parsing issue. One relatively simple option is to replace the 0 lines with metadata lines like in the sam files. So, we'd want a units line and then I guess a copy of the column names. Or maybe just empty cells instead of 0s would be better? The end of the parser function could have a conditional on the presence of 'ADRCoefficients' for the array parsing issue discussed above.

pvlib/test/test_modelchain.py
wholmgren: ok with me to leave it alone then

VolkrB:
Updated the csv and moved ```inverter['ADRCoefficients']``` parsing to ```pvsystem._parse_raw_sam_df```

pvlib/pvsystem.py
```diff
@@ -1177,6 +1176,15 @@ def _parse_raw_sam_df(csvdata):
 
     df.index = parsedindex
     df = df.transpose()
+    if 'ADRCoefficients' in df.index:
+        ad_ce = 'ADRCoefficients'
+        df.loc[ad_ce] = df.loc[ad_ce].map(lambda x:
+                                          np.array([float(s) for s in
+                                                    [','.join(x[1:-1]
+                                                              .split())]
+                                                    [0].split(',')]))
```
VolkrB: I realize this is a pretty hackie, ugly, line of code. Trying to see if I can get this to be more elegant.

wholmgren: I think this works: lambda x: list(map(float, x.strip(' []').split(' ')))

not sure if there are pros/cons to adding np.array around the list. 

also, you don't have to stay justified with the opening parenthesis. you can make only a 4 space indent on the line below the opening parenthesis. that alone would help readability!

I'd get rid of the empty else block, too.

other than that, I think you only need to add documentation and then we'll be set.

pvlib/pvsystem.py
```diff
@@ -2037,6 +2043,37 @@ def snlinverter(v_dc, p_dc, inverter):
...
+    p_nt = inverter['Pnt']
+
+    pdc = p_dc/p_nom
+    vdc = v_dc/v_nom
+    poly = np.array([pdc**0, pdc, pdc**2, vdc-1, pdc*(vdc-1),
+                     pdc**2*(vdc-1), 1/vdc-1, pdc*(1./vdc-1),
+                     pdc**2*(1./vdc-1)])
+
+    p_loss = np.dot(ce_list, poly)
+    ac_power = p_nom * (pdc-p_loss)
```
adriesse: the name "ac_power" seems to be breaking the naming pattern...

pvlib/pvsystem.py
adriesse: Maybe a code comment should go here to explain what this line is supposed to do.  Does is really all have to be done in one line?

adriesse:
I reviewed the code and didn't see any mistakes.  Since all the coefficients are normalized, it would pretty easy to create some synthetic test cases.  For example, each coefficient with a value of 4% produces a loss of Pnom * 4% at Pdc==Pnom and Vdc==Vnom.  Depending on which particular coefficient it is, this loss will vary with Pdc, Vdc, both, or neither.

VolkrB:
I cleaned up that big ugly parsing line based on @wholmgren suggestions. 

@adriesse: I always struggle with variable naming, so perhaps you, @wholmgren  and I can all agree on something. Also, could you review my pushed documentation and suggest proper definition, references, etc? (i guess this would also be a good place to discuss the naming of vars)

adriesse:
The shorter parsing line is definitely easier to parse, but it still wouldn't hurt to throw a comment in there which would be even easier to parse for readers.

I see that the ac_power name came from the other inverter function and is used in matlab.  I was anticipating something like pac or p_ac, but there doesn't seem to be a risk of misunderstanding so I don't mind if you leave it.

Can you just copy the documentation from the matlab version?

VolkrB:
Updated docs/whatsnew. Let me know if I need to add/change etc.

pvlib/pvsystem.py
```diff
@@ -1059,6 +1059,7 @@ def retrieve_sam(name=None, path=None):
         * CEC module database
         * Sandia Module database
         * CEC Inverter database
+        * Antone Driesse Inverter database
```
wholmgren: Anton

pvlib/pvsystem.py
```diff
@@ -2048,30 +2051,40 @@ def snlinverter(v_dc, p_dc, inverter):
 
 def adrinverter(v_dc, p_dc, inverter):
     r'''
-    Converts DC power and voltage to AC power using Sandia's
-    Grid-Connected PV Inverter model.
+    PVL_ADRINVERTER Converts DC power and voltage to AC power
```
wholmgren: no PVL_ADRINVERTER

pvlib/pvsystem.py
```diff
@@ -2048,30 +2051,40 @@ def snlinverter(v_dc, p_dc, inverter):
...
     p_dc : numeric
-        A scalar or DataFrame of DC powers, in watts, which are provided
-        as input to the inverter. Pdc must be >= 0.
+         A scalar or pandas series of DC powers, in watts, which are provided
+         as input to the inverter. If Vdc and Pdc are vectors, they must be
+         of the same size. p_dc must be >= 0. (W)
 
     inverter : dict-like
-        A dict-like object defining the inverter to be used.
+        A struct defining the inverter to be used, giving the
```
wholmgren: dict-like

pvlib/pvsystem.py
```diff
@@ -2048,30 +2051,40 @@ def snlinverter(v_dc, p_dc, inverter):
...
+        inverter performance parameters according to the model
+        developed by Anton Driesse[1].
```
wholmgren: space before [1]

pvlib/pvsystem.py
```diff
@@ -2048,30 +2051,40 @@ def snlinverter(v_dc, p_dc, inverter):
...
         A set of inverter performance parameters are provided with
-        pvlib, or may be generated from the library using retrievesam. 
+        pvlib, or may be generated from the library using retrievesam.
```
wholmgren: maybe something like "loaded from the supplied data table using retrieve_sam" would be more accurate.

pvlib/pvsystem.py
```diff
@@ -2048,30 +2051,40 @@ def snlinverter(v_dc, p_dc, inverter):
...
         See Notes for required keys.
 
     Returns
     -------
     ac_power : numeric
-        Modeled AC power output given the input DC voltage, v_dc, and
-        input DC power, p_dc.
+        A numpy array of modeled AC power output given the input
```
wholmgren: "A numpy array" is not quite right. It returns a Series if the input data is also a Series.

pvlib/pvsystem.py
```diff
@@ -1178,7 +1180,8 @@ def _parse_raw_sam_df(csvdata):
     df = df.transpose()
     if 'ADRCoefficients' in df.index:
         ad_ce = 'ADRCoefficients'
-        df.loc[ad_ce] = df.loc[ad_ce].map(lambda x:list(
+        # parses string into list containing floats:
```
```
# for each inverter, parses a string of coefficients like
# ***insert example***
# into a list containing floats
```

pvlib/pvsystem.py
VolkrB: (Sorry Anton)

pvlib/pvsystem.py
```diff
@@ -1122,6 +1127,8 @@ def retrieve_sam(name=None, path=None):
         elif name == 'sandiamod':
             csvdata = os.path.join(
                 data_path, 'sam-library-sandia-modules-2015-6-30.csv')
+        elif name == 'adrinverter':
+            csvdata = os.path.join(data_path, 'adr-library.csv')
```
adriesse: The parameters in this particular file were generated using the CEC efficiency measurements available to me as of 2013-10.  It might be good to add the date suffix.  I'm looking into doing an update.

pvlib/pvsystem.py
```diff
@@ -2037,6 +2051,108 @@ def snlinverter(v_dc, p_dc, inverter):
...
+    pdc = p_dc/p_nom
+    vdc = v_dc/v_nom
+    poly = np.array([pdc**0, pdc, pdc**2, vdc-1, pdc*(vdc-1),
+                     pdc**2*(vdc-1), 1/vdc-1, pdc*(1./vdc-1),
+                     pdc**2*(1./vdc-1)])
+
+    p_loss = np.dot(np.array(ce_list), poly)
+    ac_power = p_nom * (pdc-p_loss)
+
+    p_nt = -1*np.absolute(p_nt)
```
adriesse: Earlier versions of the cec/sam/sandia database were not consistent: most inverters had positive p_nt but many were negative.  The latest sam database I looked at had only positve values.

pvlib/pvsystem.py
VolkrB: ok I will add a date suffix

pvlib/pvsystem.py
VolkrB: so ```p_nt = -1*np.absolute(p_nt)``` should just be ```-1*p_nt``` ?

adriesse: That was just background info.  The absolute() shouldn't be needed, but makes it more robust.

VolkrB:
I ran flake8 and added a date suffix

wholmgren:
Looks good to me. We can ignore the failing travis builds. 

@adriesse -- you get to make the final stamp of approval since your name is on it!

pvlib/test/test_pvsystem.py
```diff
@@ -577,6 +578,16 @@ def test_snlinverter(sam_data):
...
+def test_adrinverter(sam_data):
+    inverters = sam_data['adrinverter']
+    testinv = 'Zigor__Sunzet_3_TL_US_240V__CEC_2011_'
+    vdcs = pd.Series(np.linspace(0,50,3))
+    idcs = pd.Series(np.linspace(0,11,3))
+    pdcs = idcs * vdcs
+
+    pacs = pvsystem.adrinverter(vdcs, pdcs, inverters[testinv])
+    assert_series_equal(pacs, pd.Series([-0.25000, 105.635043, 503.769061]))
+
```
adriesse: This test may catch some types of future bugs, but it's not a good PV example.  The current, voltage and power ranges are way too small for this Zigor model inverter.  

Another idea might be to use some voltage and current combinations found in the pdf file on the CEC/gosolar website and look for the results of the model to be approxiamtely equal.

adriesse:
Other than the test everything looks good to me.

VolkrB:
I don't mind updating the test. I've found a nice reference on gosolar and working on it now

VolkrB:
@adriesse could you comment on the changes to the test when you get a chance?

adriesse:
Sure.  So this is indeed a more realistic set of test conditions.  Did you calculate the expected outputs using an independent method, e.g. Excel?

Unfortunately you dropped the case that tests the night-time tare functionality and this got me thinking that the clipping condition isn't tested either.  I guess this wouldn't show up in the code coverage metrics since it is done with a "where" rather than "if", but it still seems like a good idea to include them.

A test case with scalar inputs, like for pvl_snlinverter, would be good as well.

We could stop here, or we could go further--but that might lead to changes to the inverter code.  What should happen when Pdc or Vdc is less than zero?  There are in fact many realistic and non-realistic combinations of Pdc and Vdc where the model would not produce a realistic output, i.e. it should not be used to extrapolate beyond the normal operating range of the inverter. 

What do you think about producing nan's for out-of-range conditions?  This behavior could be triggered by an optional parameter "range_check", or even multiple parameters for different checks.  Another option would be to return one of more columns of flags.

VolkrB:
"Sure. So this is indeed a more realistic set of test conditions. Did you calculate the expected outputs using an independent method, e.g. Excel?"
- I used/compared values from gosolarcalifornia.

"Unfortunately you dropped the case that tests the night-time tare functionality and this got me thinking that the clipping condition isn't tested either. I guess this wouldn't show up in the code coverage metrics since it is done with a "where" rather than "if", but it still seems like a good idea to include them."
- Seems like I would just have to add an upper and lower limit to the pd.Series.

"What do you think about producing nan's for out-of-range conditions? This behavior could be triggered by an optional parameter "range_check", or even multiple parameters for different checks. Another option would be to return one of more columns of flags."
- I'm not opposed to that. Should the snlinverter have something like this as well?

cwhanse:
> What do you think about producing nan's for out-of-range conditions? This behavior could be triggered by an optional parameter "range_check", or even multiple parameters for different checks. Another option would be to return one of more columns of flags.

I think nan should be reserved for combinations of inputs where the model equation fails to produce a real value, or produces a non-physical result, such as negative power below the night tare.  If the model equation operates and produces meaningless results, I'd rather let the results themselves communicate that failing.

We could consider building an array check function, like is implemented in SAM, that compares the string voltage to the inverter input range, etc., and gives a warning when the array specification appears wrong.

wholmgren:
Definitely return nan if the output is non-physical. Otherwise it's up to model to define good limiting behavior and it's up to the user to not do something dumb. 

Part of the problem with warnings is that we need to be consistent throughout the library. Users see a warning about something in one function and then they expect that they'll be warned about bad behavior everywhere. It's so much easier to be consistently silent.

adriesse:
Ok, then let's get physical.  Inverters simply cannot operate beyond a certain input voltage range nor inject power into a PV array.  The reasons vary with topology and components used.  I would propose (untested):
```
def adrinverter(v_dc, p_dc, inverter, vtol=0.10):
...
v_limit_upper = np.nanmax([Vmax, Vdcmax, MPPTHi]) * (1 + vtol)
v_limit_lower = np.nanmin([Vmin, MPPTLow]) * (1 - vtol)

ac_power = np.where(v_dc > v_limit_upper, np.nan, ac_power)
ac_power = np.where(v_dc < v_limit_lower, np.nan, ac_power)

ac_power = np.where(p_dc < 0.0, np.nan, ac_power)
```

Regarding the tests:

> "Sure. So this is indeed a more realistic set of test conditions. Did you calculate the expected outputs using an independent method, e.g. Excel?"
> 
>     I used/compared values from gosolarcalifornia.
> 
I'm not sure you we understood each other here.  Where did the values [1161.5745, 1116.4459, 382.6679] come from?

> "Unfortunately you dropped the case that tests the night-time tare functionality and this got me thinking that the clipping condition isn't tested either. I guess this wouldn't show up in the code coverage metrics since it is done with a "where" rather than "if", but it still seems like a good idea to include them."
> 
>     Seems like I would just have to add an upper and lower limit to the pd.Series.
> 
I'm not sure exactly what you mean by that, but it could be right.

I also meant to say earlier that I don't see why the tests are defined with voltage and _current_, which are then multiplied to get power.  Current isn't seen explicitly in the original data or in the model code, so I suggest specifying the power levels directly.

VolkrB:
> I'm not sure you we understood each other here. Where did the values [1161.5745, 1116.4459, 382.6679] come from?

http://www.gosolarcalifornia.ca.gov/equipment/inverter_tests/summaries/Ablerex%20ES%202200-US-240.pdf I picked several input voltages on this doc and then compared the output efficiency to the code's output i.e p_dc/p_ac. They agreed within 1%. I guess I could be more precise in my testing, perhaps someone could check if the numbers agree with the matlab version? 

> I'm not sure exactly what you mean by that, but it could be right.

I mean 1A, 0W as the lower bound and something like 1A, 2500W as the upper. That should take care of tare and clipping conditions?

VolkrB:
Made changes based on input from @adriesse

pvlib/pvsystem.py
```diff
@@ -2051,7 +2051,7 @@ def snlinverter(v_dc, p_dc, inverter):
     return ac_power
 
 
-def adrinverter(v_dc, p_dc, inverter):
+def adrinverter(v_dc, p_dc, inverter, vtol=0.10):
```
adriesse: vtol : numeric
A unit-less fraction that determines how far the efficiency model is allowed to extrapolate beyond the inverter's normal input voltage operating range.

VolkrB: updated

adriesse:
I think this is good, Volker.  With the added comment for the optional parameter I think we could stop there.

adriesse:
I wonder how many people are actually watching this thread...

jforbess:
I'm reading it. I get all the threads in my email, and read them at some
regular frequency.

wholmgren:
It's not immediately clear to me if/how the vtol parameter is being tested. Maybe a comment in the test function would be enough. We do need at least one more test that sets the parameter, though.

adriesse:
There is one case with v_dc below the threshold determined by vtol, and another with v_dc above.  These both produce nan outputs.

I can image tons of additional tests, and garbage in/out scenarios but don't have a good feel for what's appropriate in this context.

wholmgren:
Thanks for clarifying, Anton. In that case I recommend one additional test that uses the same input but with vtol set sufficiently large so that the returned values are no longer nan. This way we have a test that actually sets the value of the parameter instead of relying on the default value, thus testing both the presence of the keyword and that its value is respected.

adriesse:
That seems a bit like testing the compiler's handling of parameter defaults, doesn't it?!

I have another suggestion. How about putting the inverter parameters in the code rather than reading them from the file?  This removes a dependency and also makes clear to the reader what the power and voltage ranges are for the test.

wholmgren:
My primary concern is testing that the function can be called when the keyword is used. This ensures that a test will fail if anyone ever tries to remove ``vtol`` or change ``vtol`` to something else. 

Regarding testing the value of the keyword, I am not at all concerned about the interpreter, but rather I am concerned that the value is actually used later in the code. For example, in prototyping the function there could have been a hardcoded ``vtol = 0.1`` line that was not removed. You would be right to say it should be obvious when reviewing the code, but I make these kinds of mistakes all the time. I am pretty sure that this has happened in pvlib python before. I think it was in singlediode, desoto, or maybe the earlier version of the sapm function.

I like the idea of hardcoding the test parameters for the ``adrinverter`` function. That makes me realize that we're not explicitly testing the ``retrieve_sam`` function for any of the data types. For this PR, we'd only need to worry about testing it with the 'adrinverter' option.

wholmgren:
Hi @VolkrB, can you merge/rebase your PR and add one more simple test using ``vtol`` as described above? Any test that puts ``vtol`` in the function call would be fine with me. Testing using hardcoded values and testing ``retrieve_sam`` can be implemented later. Thanks!

VolkrB:
Sorry I haven't looked at this in a while. I hope the merge/push looks ok

wholmgren:
There are some errors in the test suite that are related to the combination of changing the default ModelChain inverter and the introduction of the vtol/nans.

VolkrB:
oops. Seems like that test has non-physical results. I put in the appropriate nans in the params

pvlib/test/test_modelchain.py
wholmgren: I'm only now realizing that this change to the ``cec_dc_snl_ac_system`` fixture makes the name no longer correct. I do think it would be worthwhile to make a second test called ``cec_dc_adr_ac_system`` or something like it. Then you can revert the change to the ``test_dc_models`` fixture, while also updating the ``test_ac_models`` ``ac_systems`` dictionary.

pvlib/pvsystem.py
```diff
@@ -2037,6 +2051,120 @@ def snlinverter(v_dc, p_dc, inverter):
...
+    ======   ============================================================
+    p_nom    The nominal power value used to normalize all power values,
+             typically the DC power needed to produce maximum AC power output,
+             (W).
+
+    v_nom    The nominal DC voltage value used to normalize DC voltage values,
+             typically the level at which the highest efficiency is achieved,
+             (V).
+
+    pac_max  The maximum AC output power value, used to clip the output if
```
VolkrB: should the column header be longer or the variables shorter?

pvlib/test/test_modelchain.py
VolkrB: good catch

pvlib/pvsystem.py
wholmgren: A longer Column header by one = sign. You'll also need to add one space between all of the Column and the Description sections so that the table stays properly formatted. I guess you should probably also make the Description header longer and/or make the text fit within the existing Description header. 

I'm pretty sure that Sphinx would build the table in its present form and spit out a warning. It's not a big deal if it's just one table, but every silly warning makes it slightly harder to identify other issues with the documentation. If you actually build the docs and don't see a warning then we can ignore all this.

VolkrB:
adrinverter is not being tested in the singlediode dc modelchain, but it's being tested in the ac_model chain. Should I write a new test function for including it in the dc modelchain or maybe there's a better solution?

wholmgren:
No need to write a new test function for the dc modelchain. I'm confused about why the ``'singlediode'`` ``test_dc_models`` test passes, though. It's supposed to check to the second decimal, but you've (I think inadvertently) changed the value in the first decimal.

VolkrB:
failed for check_less_precise = 3, I updated with the correct result

wholmgren:
Thanks. The test errors are unrelated to this PR, so I think we're ready to merge. Agreed?

VolkrB:
Looks good on my end

adriesse:
Agreed.

wholmgren:
thanks @VolkrB and @adriesse!

PR #298, opened 2017-01-17
Improve convergence of lambertw for very large arguments

cwhanse:


cwhanse:
The test failures don't look to be caused by my one edit which changes range(0,3) to range(9) in pvsystem.

wholmgren:
Is it easy to add a test that requires this change in order to pass? test_pvsystem.py has the following function that you could copy/paste, rename, and then update with your parameters:

```python
@requires_scipy
def test_v_from_i_bigger():
    # 1000 W/m^2 on a Canadian Solar 220M with 20 C ambient temp
    # github issue 225
    assert_allclose(54.303958833791455, output, atol=1e-5)
```

For reference, this is a continuation of #268, #266, #260.

cwhanse:
I haven't been able to reproduce the failure, so let's close this pull request.  We can always re-open if we find the failure conditions again.

Issue #237, opened 2016-08-12
Localized time index not working when using 'nrel_c' method


anomam:
Hi,

I'm not sure if that was already covered in an other issue (couldn't find it) or intentional in the code, but when using the function `get_solarposition()` with `method='nrel_c'`, it doesn't seem to matter whether the time index is localized or not.
The function will assume that the time index is at 'UTC' no matter what. Which goes against what was described in http://pvlib-python.readthedocs.io/en/latest/timetimezones.html.

So I ended up converting my timezone to 'UTC' before calculation, and then back to local after the calculation

Example:
`pvlib_loc = location.Location(latitude, longitude, tz='Etc/GMT+8')`
`localized_index = index.tz_localize('Etc/GMT+8')`
`utc_index = localized_index.tz_convert('UTC')`
`solarAngles = pvlib_loc.get_solarposition(utc_index, method='nrel_c')`
`solarAngles.index = localized_index`

wholmgren:
I'd welcome a pull request if you want to fix it or at least document the bug. Personally, I'd vote for removing the nrel_c option all together, but that's probably a separate issue. Is there a reason you're using nrel_c instead of the nrel_numpy or nrel_numba options?

mikofski:
At the very least this should be changed in the docstring I think.

Issue #168, opened 2016-05-13
spa_c_files don't build with VC14 (Python 3.5)


mikofski:
windows 7 amd64
Cython (0.24)
Python 3.5.1 (v3.5.1:37a07cee5969, Dec  6 2015, 01:54:25) [MSC v.1900 64 bit (AMD64)] on win32

if I try to build spa_c for Python-3.5 I get the following:

```
$ py -3.5 setup.py build_ext --inplace
running build_ext
building 'spa_py' extension
creating build
creating build\temp.win-amd64-3.5
creating build\temp.win-amd64-3.5\Release
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -IC:\Python35\include -IC:\Python35\include "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.10240.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" "-IC:\Program Files (x86)\Windows Kits\8.1\include\\shared" "-IC:\Program Files (x86)\Windows Kits\8.1\include\\um" "-IC:\Program Files (x86)\Windows Kits\8.1\include\\winrt" /Tcspa_py.c /Fobuild\temp.win-amd64-3.5\Release\spa_py.obj
spa_py.c
c:\users\mmikofski\documents\projects\pvlib-python\pvlib\spa_c_files\spa.h(88): error C2032: '__timezone': function cannot be member of struct '<unnamed-tag>'
spa_py.c(854): error C2059: syntax error: '('
spa_py.c(1700): error C2059: syntax error: '('
spa_py.c(1702): error C2275: 'PyObject': illegal use of this type as an expression
c:\python35\include\object.h(110): note: see declaration of 'PyObject'
spa_py.c(1702): error C2059: syntax error: ')'
spa_py.c(1702): error C2223: left of '->ob_refcnt' must point to struct/union
spa_py.c(1702): error C2065: '_py_decref_tmp': undeclared identifier
spa_py.c(1702): warning C4312: 'type cast': conversion from 'int' to 'PyObject *' of greater size
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\BIN\\amd64\\cl.exe' failed with exit status 2
```

[...]

PR #301, opened 2017-02-02
Allowing spa_c() to use localized time index

anomam:
Addresses #237 :
``spa_c()``'s docstring specifies that ``time`` can be localized, but the code only assumes 'UTC' time.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.4.4.txt
+++ b/docs/sphinx/source/whatsnew/v0.4.4.txt
@@ -13,9 +13,11 @@ Documentation
 * Fixes the Forecasting page's broken links to the tutorials.
 * Fixes the Forecasting page's broken examples. (:issue:`299`)
 * Fixes broken Classes link in the v0.3.0 documentation.
+* Fixes timezone issue in solarposition spa_c function (:issue:`237`)
 
 
 Contributors
 ~~~~~~~~~~~~
 
 * Will Holmgren
+* Marc Anoma
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -173,7 +173,11 @@ def spa_c(time, latitude, longitude, pressure=101325, altitude=0,
 
     pvl_logger.debug('using built-in spa code to calculate solar position')
 
-    time_utc = time
+    # if localized, convert to UTC. otherwise, assume UTC.
+    try:
+        time_utc = time.tz_convert('UTC')
+    except TypeError:
+        time_utc = time
 
     spa_out = []
 
@@ -184,7 +188,7 @@ def spa_c(time, latitude, longitude, pressure=101325, altitude=0,
                                 hour=date.hour,
                                 minute=date.minute,
                                 second=date.second,
-                                timezone=0,  # must input localized or utc time
+                                timezone=0,  # date uses utc time
                                 latitude=latitude,
                                 longitude=longitude,
                                 elevation=altitude,
@@ -193,7 +197,7 @@ def spa_c(time, latitude, longitude, pressure=101325, altitude=0,
                                 delta_t=delta_t
                                 ))
 
-    spa_df = pd.DataFrame(spa_out, index=time_utc)
+    spa_df = pd.DataFrame(spa_out, index=time)
 
     if raw_spa_output:
         return spa_df


anomam:
Hi @wholmgren , I just implemented a quick fix for an issue I raised a while back. Let me know what you think!
Also, I just noticed that continuous integration is not triggered in this PR when I push a commit, did I mess up something?

Thanks!

mikofski:
I don't think the `nrel_c` option is tested in the build server though, or if it is, only python27, it only builds with msvc90 (vs2008), I wasn't able to build it for Python-3.5 using msvc140 (vs2015). I didn't try Python-3.3 or 3.4. see #168

wholmgren:
@mikofski is correct -- the spa_c algorithm is not tested on travis or appveyor. You can run the tests on your own machine if the spa files are present when you build pvlib python. I don't want to support spa_c at all, but so long as we do, you should probably add a new test. Still, I am ok merging this in any state. Let me know when you want me to go ahead and click the button.

I'd still like to know why you Sunpower folks are using nrel_c instead of nrel_numpy or nrel_numba. nrel_c is slower than both on every machine I've tested them on.

anomam:
@wholmgren thanks for your reply.
I guess I personally started using nrel_c because I was using the SPA website originally, and maybe I also had the assumption that the C code would be faster.
But it sounds like there's no reason to use it if nrel_numpy provides the exact same results even faster. Why do you guys keep it in the code considering all that?

wholmgren:
Good question! I remember discussing this when we added the nrel_num* methods though I can't seem to find it right now. I think the consensus was that we should keep nrel_c since it was already working and some people might prefer nrel_c because the c code was written by the renewable energy gods on the hill in Golden. Maybe we should add a note to the function documentation with the relative speeds. I guess the benchmarks are somewhat buried at the bottom of the solar position jupyter notebook. I'd suggest that you rerun the benchmarks on your own platform, though.

In any case, it don't think it hurts to go ahead with this PR since you've already made the change. Up to you.

adriesse:
I think it's great to have and keep multiple implementations that should and do give the same results.  Great confidence booster!  You could have a test that runs both and compares.

wholmgren:
Good point on the confidence booster. The test suite does this in a if A=B and A=C then B=C kind of way. The problem is that it's difficult (impossible?) to run those tests via travis/appveyor since we can't redistribute NREL's spa files. I haven't manually run those tests in a long time.

mikofski:
You can add `nrel_c` to `travis.yml` like this:

1. Add a script called `get_spa.py` to pvlib's top level. Here's an example, you can change the "name" and "company" fields to whatever you think is appropriate:

```python
#! /usr/env/python

"""
get spa c source from NREL for travis
"""

import requests
import os
import logging

# logger
logging.basicConfig(level=logging.DEBUG)
LOGGER = logging.getLogger(__file__)
# constants
SPAC_URL = r'https://midcdmz.nrel.gov/apps/download.pl'  # spa.c source url
# register with NREL to download spa.c source
PAYLOAD = {
    'name': 'pvlib-python',  # <-- okay to set this
    'country': 'US',
    'company': 'pvlib,  # <-- okay to set this
    'software': 'SPA'
}
SPAH_URL = r'http://midcdmz.nrel.gov/spa/spa.h'  # spa.h source url
PVLIB_PATH = os.environ['PVLIB_PATH']  # path to PVLIB on travis
SPA_C_FILES = os.path.join(PVLIB_PATH, 'pvlib', 'spa_c_files')

if __name__ == "__main__":
    # get spa.c source
    LOGGER.debug('post payload to url: %s\n\tpayload:\n%s', SPAC_URL, PAYLOAD)
    r = requests.post(SPAC_URL, data=PAYLOAD)
    LOGGER.debug('post response: %r', r)
    # save spa.c source to PVLIB/SPA_C_FILES folder
    with open(os.path.join(SPA_C_FILES, 'spa.c'), 'wb') as f:
        f.write(r.content)
    LOGGER.debug('saved file: %r', f.name)
    # get spa.c source
    LOGGER.debug('get url: %s', SPAH_URL)
    r = requests.get(SPAH_URL)
    LOGGER.debug('get response: %r', r)
    # save spa.c source to PVLIB/SPA_C_FILES folder
    with open(os.path.join(SPA_C_FILES, 'spa.h'), 'wb') as f:
        f.write(r.content)
    LOGGER.debug('saved file: %r', f.name)
    LOGGER.debug('exiting %s', __file__)  # exit
```

2. Add a `before_script` section to your `travis.yml` to build the `nrel_c` binaries.

```yaml
before_script:
  - pip install requests cython
  - export PVLIB_PATH=/home/travis/build/pvlib/pvlib-python
  - echo $PVLIB_PATH
  - python /home/travis/build/pvlib/pvlib-python/get_spa.py
  - BUILD_DIR=$PWD
  - echo leaving $BUILD_DIR
  - cd $PVLIB_PATH/pvlib/spa_c_files/
  - echo entered $PWD
  - python setup.py build_ext --inplace
  - cd $BUILD_DIR
  - echo entered $PWD
```

These scripts are adapted from [UncertaintyWrapper](../../../SunPower/UncertaintyWrapper), so they may be some redundancy or require some minor edits.

wholmgren:
Interesting. Up to you all to make it happen if you want it. Not sure if we'd want a separate Travis build for this or not.

mikofski:
>the renewable energy gods on the hill 

:laughing: :rofl:

anomam:
Thanks for all the inputs everybody.
Since it seems like the 'nrel_num*' methods are already doing the job I've moved on to other urgent things but I'll try to come back to it when I have more time to work on the continuous integration update and the new tests to write (not trivial to me). Or feel free to take over this PR if you want!

Thanks,
- Marc

wholmgren:
I can merge this if you remove the update to the whatsnew doc. I can update the whatsnew for you when the next file gets into the repository.

I don't want to add the travis script at this time. Tracking down errors on the build matrix is already too time consuming for our current level of support.

mikofski:
wondering if this is closed now, since I saw you branched it and made another commit (c881445)? Did you already merge it with #326 (2e2f7a8)?

wholmgren:
No, I was going to make a PR to @anomam's branch, but got distracted. It can go into 0.4.5 if you resolve the conflict ASAP.

mikofski:
I can't edit @anomam branch, but I think you can edit a pull request, it's like a GitHub option somewhere - https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/ - ah looks like Marc would have to allow it. Sorry :frowning:

wholmgren:
closing in favor of #333.

anomam:
Really sorry about this guys, I hope I didn't block anything and that you were able to do what you needed to do! @wholmgren

wholmgren:
No worries. I should have merged it back when it was easy to do so. The github "squash-and-merge" process dropped your name from the commit for some dumb reason. Sorry.

Issue #302, opened 2017-02-04
add discussion and examples of Angstrom and Kasten formulas to docs


mikofski:
Related to issue #269 and a follow on to PR #278 which adds the Kasten pyrheliometric formula to calculate Linke turbidity factor given broadband AOD and precipitable water, the Bird-Hulstrom broadband AOD correlation and Angstrom formulas to calculate alpha exponent from AOD at two wavelengths and to calculate AOD given one measurement and alpha.

This issue proposes to add a brief discussion and some examples to the clear sky documentation. The discussion can also mention other methods not currently included in pvlib-python.

mikofski:
Any comments on how to approach this? How long of a discussion? What's the main goal? I'd like to close this out, so any pointers would help a lot. thx

Here are some references with links to more:
* [Use of Measured Aerosol Optical Depth and Precipitable Water to Model Clear Sky Irradiance](https://www.researchgate.net/publication/318470239_Use_of_Measured_Aerosol_Optical_Depth_and_Precipitable_Water_to_Model_Clear_Sky_Irradiance)
* [Evaluation and Correction of the Impact of Spectral Variation of Irradiance on PV Performance](https://www.researchgate.net/publication/309648716_Evaluation_and_Correction_of_the_Impact_of_Spectral_Variation_of_Irradiance_on_PV_Performance)

wholmgren:
A paragraph or two in clearsky.rst. Only enough to give people a sense of what’s possible.

PR #312, opened 2017-02-15
minor documentation updates

### LINKED ISSUE #297: pvlib python communication forums
What do people think about these new communication forums?

- pvlib tag on stackoverflow: http://stackoverflow.com/questions/tagged/pvlib
- pvlib python google group: https://groups.google.com/forum/#!forum/pvlib-python
- pvlib python slack organization: https://pvlib-python.slack.com/

Here are my thoughts:

First, all credit goes to @mikofski for pushing this forward and setting up some things for us to try! Thanks!

I am concerned that too many forums could be counter productive, but I think we're big enough now that we could benefit from more than GitHub issues. Watching the project on GitHub does allow people to stay up to date on every little thing in the pvlib python world, but that is way too much information for many people. Spreading the communication over multiple platforms lets people focus on what they really care about at the cost of making things more complicated and potentially diluting the viewership of any given item.

I think it would be nice if we would use GitHub for bug tracking and development discussions, people would ask usage questions on Stack Overflow, and we’d use a mailing list for new version announcements. 

I know that a few people have tried to ask pvlib questions on stack overflow in the past with varying degrees of success. I am going to start watching Mark's new pvlib tag. I don't have much experience with google groups, but at this point I prefer stackoverflow for usage questions. In any case, while we're getting started with stack overflow and/or google groups, I think that people should feel free to open and then close a GitHub issue that links to the new post.

On a related note, [Sibbell](https://sibbell.com/about/) is a nice tool that will send you an email when there are new releases of any of your starred repositories. 

I like Slack for quick, informal conversations in which I don’t have to stress about if I’m going to come across the wrong way in a public forum. I don't like the constant din of interruptions on Slack, though.

I also get requests for help via email from time to time. I don't like answering questions via email because only a handful of people have an opportunity to learn from the exchange.

I don’t know what the ideal approach is, but I think we should try something new and see if it catches on!

Finally, we should add some notes to the documentation once we come to a consensus.

wholmgren:
A handful of minor updates to the documentation. Closes #297 (communication forums).
Partially addresses #302 (LT/AOD examples). Adds some notes about recommended solar position methods (relevant to #301). Updates the doc environment to use newer packages.

It's rendered here:

http://wholmgren-pvlib-python-new.readthedocs.io/en/docupdates/index.html

Let know if you have any suggestions.

### DIFF
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -38,26 +38,33 @@ def get_solarposition(time, latitude, longitude,
     Parameters
     ----------
     time : pandas.DatetimeIndex
+
     latitude : float
+
     longitude : float
+
     altitude : None or float
         If None, computed from pressure. Assumed to be 0 m
         if pressure is also None.
+
     pressure : None or float
         If None, computed from altitude. Assumed to be 101325 Pa
         if altitude is also None.
+
     method : string
-        'pyephem' uses the PyEphem package: :func:`pyephem`
-
-        'nrel_c' uses the NREL SPA C code [3]: :func:`spa_c`
-
         'nrel_numpy' uses an implementation of the NREL SPA algorithm
-        described in [1] (default): :func:`spa_python`
+        described in [1] (default, recommended): :py:func:`spa_python`
 
         'nrel_numba' uses an implementation of the NREL SPA algorithm
-        described in [1], but also compiles the code first: :func:`spa_python`
+        described in [1], but also compiles the code first:
+        :py:func:`spa_python`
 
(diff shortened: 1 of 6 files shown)


pvlib/solarposition.py
```diff
@@ -38,26 +38,33 @@ def get_solarposition(time, latitude, longitude,
     Parameters
     ----------
     time : pandas.DatetimeIndex
+
```
mikofski: @wholmgren did you really mean to add all of these extra newlines? Should I merge them, or was it a typo? I can't find the rationale in your comments, and it's inconsistent with docstrings everywhere else. Please let me know. Thanks!

PR #327, opened 2017-05-25
update solar_position docstring

mikofski:
* change units for pressure to mbars


adriesse:
Does this have any repercussions for the solar position calculations?  It seems that Pa are used everywhere else in PVLIB.  Is this converted to mbar where necessary?   `pvlib.solarposition.get_solarposition` appears to use Pa.

mikofski:
No, the conversion is already handled in `pvlib.solarposition.spa_python` by dividing pressure by 100(mbar/PA) https://github.com/pvlib/pvlib-python/blob/master/pvlib/solarposition.py

wholmgren:
Good catch, Mark. I suppose we could also change the API to make it more consistent, but I prefer the quick doc fix. Can you add a note to the whatsnew?

mikofski:
Hi @wholmgren , I added a somewhat verbose note in "whats new for v0.4.5" b/c I wanted it to be absolutely clear that this docstring change re: units of pressure in the low-level `spa.py` module should have **no effect on almost all users**.

>Update docstring in `pvlib.spa.solar_position` - change units of pressure to millibars. *NOTE*: units of pressure in `pvlib.solar_position.spa_python` and `pvlib.solar_position.spa_c` are still Pascals. This update should have no effect on most users, since it only applies to the low-level `spa.py` module.

wholmgren:
thanks mark

PR #330, opened 2017-06-02
Change method of multi-inheritance to allow kwargs in parent class of LocalizedPVSystem and LocalizedSingleAxisTracker

uvchik:
If keyword arguments are present a `TypeError` will be raised, so it is better to remove it from the `super` call.

As I normally initialise classes using a parameter dictionary, so it often happens that I will have additional keys. Therefore it is convenient if the PVSystem class takes `**kwargs` but they should not be passed on.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.4.6.rst
+++ b/docs/sphinx/source/whatsnew/v0.4.6.rst
@@ -8,7 +8,8 @@ Bug fixes
 ~~~~~~~~~
 
 * Method of multi-inheritance has changed to make it possible to use kwargs in
-  the parent class (:issue:`330`)
+  the parent classes of LocalizedPVSystem and LocalizedSingleAxisTracker
+  (:issue:`330`)
 
 
 Enhancements
--- a/docs/sphinx/source/whatsnew/v0.4.6.rst
+++ b/docs/sphinx/source/whatsnew/v0.4.6.rst
@@ -7,6 +7,9 @@ v0.4.6 ()
 Bug fixes
 ~~~~~~~~~
 
+* Method of multi-inheritance has changed to make it possible to use kwargs in
+  the parent class (:issue:`330`)
+
 
 Enhancements
 ~~~~~~~~~~~~
@@ -24,3 +27,4 @@ Contributors
 ~~~~~~~~~~~~
 
 * Will Holmgren
+* Uwe Krien
--- a/docs/sphinx/source/whatsnew/v0.4.6.rst
+++ b/docs/sphinx/source/whatsnew/v0.4.6.rst
@@ -1,0 +1,26 @@
+.. _whatsnew_0460:
+
+v0.4.6 ()
+---------
+
+
+Bug fixes
+~~~~~~~~~
+
+
+Enhancements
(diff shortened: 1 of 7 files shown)


uvchik:
My fault, so i reverted the commit.

For me it would be nice to make it possible to add additional parameters to the dictionary and pass it using the double star operator (see below). But may be it was a deliberate decision to forbid superfluous parameter.

```python
location = {
        'altitude': 34,
        'latitude': latlon.loc[int(key[2:]), 'st_y'],
        'longitude': latlon.loc[int(key[2:]), 'st_x']}

system = {
    'module_parameters': sandia_modules[module_name],
    'inverter_parameters': sapm_inverters[inverter_name],
    'surface_azimuth': 180,
    'surface_tilt': 30,
    'albedo': 0.2,
    'additional_information': 'something' 
    }
mc = ModelChain(PVSystem(**system), Location(**location))
```

If you want to allow additional parameters in this class I could think about, how to did, if not just close this PR. I do not use it all the time, but I like it especially for system parameters.

wholmgren:
I've run into this issue, as well, but haven't given much thought to the best approach. I am open to changing things, though.

uvchik:
What I read about multiple inheritance, one can call the init-methods of the parent classes explicitly to avoid such behaviour. After I have changed everything, all tests work fine, except of one. But I think the test is wrong and not my changes, because `my name` is passed to the method and so the expected name attribute should be `my name` and not `None`.

@wholmgren May you check, if I changed the test correctly (commit: 37d84ec).

wholmgren:
Thanks. After (re)reading a bunch of stuff on multiple inheritance in python, I think your proposal is a better way to handle these classes.

I agree that the test asserts the wrong result. It looks like the ``PVSystem.__init__`` consumes the ``name`` kwarg before it gets to ``Location``, then the ``Location__init__`` sets the ``name`` to ``None``. Your code passes all kwargs to both ``__init__`` methods, so the composite object ends up with the right attribute value regardless of the initialization order.

Can you start a 0.4.6 whats new file and describe this fix?

uvchik:
Thank you for your feedback.

I added a whatsnew file as `.rst` as it is a reStructuredText file. This makes file handling a little easier.

I hope I described my changes comprehensible.

docs/sphinx/source/whatsnew/v0.4.6.rst
```diff
@@ -0,0 +1,30 @@
...
+
+v0.4.6 ()
+---------
+
+
+Bug fixes
+~~~~~~~~~
+
+* Method of multi-inheritance has changed to make it possible to use kwargs in
+  the parent class (:issue:`330`)
```
wholmgren: We should probably specify that this is for ``LocalizedPVSystem`` and ``LocalizedSingleAxisTracker``.

wholmgren: Would also be good to update the PR name to something like "allow kwargs in parent class of LocalizedPVSystem and LocalizedSingleAxisTracker"

uvchik:
Is it okay like this?

wholmgren:
Yes, thanks!

Issue #338, opened 2017-06-13
Set negative values of AOI loss to 0 rather than np.nan


cwhanse:
The current AOI loss models set negative values to np.nan, rather than zero.  When np.nans are propagated to the effective irradiance model, the beam irradiance (and the effective irradiance) becomes np.nan rather than 0.  As a consequence, pvlib cannot compute effective irradiance when AOI>90, i.e., when the sun is behind the array plane.  In this circumstance there's still diffuse irradiance on the array plane.

I think it would be better to implement the AOI loss models as

f(AOI) = model value if AOI<90
f(AOI) = 0 otherwise.

jkfm:
Thanks Cliff for opening an issue! It refers to my question in the [google group](https://groups.google.com/forum/#!topic/pvlib-python/AiWA6v4x3GM).

If I find time in the next couple of days, and if there is consent on the issue, I will start working on a pull request.

cwhanse:
(waves at Johannes)  you’re welcome ☺

wholmgren:
I can see arguments for nans or 0s. I think nan could be the right result because maybe it doesn't make sense to talk about an AOI modifier when the sun is behind the array plane. In a similar fashion, pvlib's relative airmass function returns nan when the zenith angle is greater than 90. 

On the other hand, I can think of at least a few of my results where I've failed to account for these nans, so that's a good indicator that something should change.

One thing is for certain: ModelChain fails to account for the nans in its effective_irradiance calculation and that should be fixed as soon as possible.

wholmgren:
The code below should allow you to monkey patch your modelchain objects until we get this fixed:

```python
import types
def effective_irradiance_model_corrected(self):
    fd = self.system.module_parameters.get('FD', 1.)
    self.effective_irradiance = self.spectral_modifier * (
        self.total_irrad['poa_direct']*self.aoi_modifier.fillna(0) +
        fd*self.total_irrad['poa_diffuse'])
    return self

mc = ModelChain(system, location, **your_normal_kwargs)
mc.effective_irradiance_model = types.MethodType(effective_irradiance_model_corrected, mc)
mc.run_model(times, weather=weather)  # as if nothing has changed
```

AC power without the patch:

![download-8](https://user-images.githubusercontent.com/4383303/27109764-ad3911c0-5059-11e7-8d10-8708f30a6308.png)

and with the patch:

[...]

PR #339, opened 2017-06-14
Incidence Angle Modifier zero instead of np.nan (#338)

jkfm:
This PR changes the three IAM-models to return 0 instead of np.nan for aoi >= 90 (ashraeiam, physicaliam) and <0 (sapm_aoi_loss). 

I also updated the tests to account for the changes. The test for physicaliam now shows, that angles of -90 and 90 return 0, whereas an angle of 0 returns 1, in accordance to the model description in De Soto et al..

Comments are welcome!

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -697,7 +697,7 @@ def ashraeiam(aoi, b=0.05):
     ----------
     aoi : numeric
         The angle of incidence between the module normal vector and the
-        sun-beam vector in degrees.
+        sun-beam vector in degrees. Angles of nan will result in nan.
 
     b : float
         A parameter to adjust the modifier as a function of angle of
@@ -762,7 +762,7 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
     aoi : numeric
         The angle of incidence between the module normal vector and the
         sun-beam vector in degrees. Angles of 0 are replaced with 1e-06
-        to ensure non-nan results.
+        to ensure non-nan results. Angles of nan will result in nan.
 
     n : numeric
         The effective index of refraction (unitless). Reference [1]
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -149,9 +149,6 @@ class PVSystem(object):
 
         self.racking_model = racking_model
 
-        # needed for tying together Location and PVSystem in LocalizedPVSystem
-        super(PVSystem, self).__init__(**kwargs)
-
     def __repr__(self):
         attrs = ['name', 'surface_tilt', 'surface_azimuth', 'module',
                  'inverter', 'albedo', 'racking_model']
@@ -584,7 +581,8 @@ class LocalizedPVSystem(PVSystem, Location):
                           list(loc_dict.items()) +
                           list(kwargs.items()))
 
-        super(LocalizedPVSystem, self).__init__(**new_kwargs)
+        PVSystem.__init__(self, **new_kwargs)
+        Location.__init__(self, **new_kwargs)
 
     def __repr__(self):
         attrs = [
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -761,7 +761,8 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
     ----------
     aoi : numeric
         The angle of incidence between the module normal vector and the
-        sun-beam vector in degrees.
+        sun-beam vector in degrees. Angles of 0 are replaced with 1e-06
+        to ensure non-nan results.
 
     n : numeric
         The effective index of refraction (unitless). Reference [1]
@@ -811,6 +812,10 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
     spa
     ashraeiam
     '''
+    zeroang = 1e-06
+
+    aoi = np.where(aoi == 0, zeroang, aoi)
+
     thetar_deg = tools.asind(1.0 / n*(tools.sind(aoi)))
 
     tau = (np.exp(- 1.0 * (K*L / tools.cosd(thetar_deg))) *
@@ -818,8 +823,6 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
             ((tools.sind(thetar_deg + aoi)) ** 2) +
             ((tools.tand(thetar_deg - aoi)) ** 2) /
             ((tools.tand(thetar_deg + aoi)) ** 2))))))
-
-    zeroang = 1e-06
 
     thetar_deg0 = tools.asind(1.0 / n*(tools.sind(zeroang)))
 
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -819,19 +819,36 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
 
     aoi = np.where(aoi == 0, zeroang, aoi)
 
+    # angle of reflection
     thetar_deg = tools.asind(1.0 / n*(tools.sind(aoi)))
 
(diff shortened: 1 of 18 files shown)


adriesse:
Could we have the identical code in each of the three functions to do the same thing?  That is, to deal with the negative values and the larger than 90 angles?  I don't know whether np.maximum, np.where or np.clip is better, but I would lean toward clip.

The SAPM model doesn't like negative angles, the others are fine because of the cosine.  That should be handled consistently as well.

pvlib/pvsystem.py
```diff
@@ -811,6 +812,10 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
     spa
     ashraeiam
     '''
+    zeroang = 1e-06
+
+    aoi = np.where(aoi == 0, zeroang, aoi)
```
wholmgren: Why did you add this line? I understand what it does, but not why we need it now when we didn't previously.

It's not clear to me that we actually need zeroang at all. I think its only purpose is to avoid infs, but maybe those be more cleanly handled with np.where or similar.

pvlib/pvsystem.py
```diff
@@ -732,7 +732,7 @@ def ashraeiam(aoi, b=0.05):
 
     iam = 1 - b*((1/np.cos(np.radians(aoi)) - 1))
 
-    iam = np.where(np.abs(aoi) >= 90, np.nan, iam)
+    iam = np.where(np.abs(aoi) >= 90, 0, iam)
     iam = np.maximum(0, iam)
```
wholmgren: is the np.maximum line still necessary?

wholmgren:
@adriesse brings up a separate but closely related point regarding the meaning of negative values of aoi in pvlib. I tend to agree that the aoi loss functions should all start with something like ``aoi = np.abs(aoi)``, but that's open for discussion. I thought that we had previously discussed this elsewhere, but I can't find the discussion. If we're going to change it, now is the time.

cwhanse:
If there are negative values for AOI, that indicates a problem elsewhere.  I’m not in favor of using abs(aoi) to address it.

pvlib/pvsystem.py
jkfm: I think it is, because for angles close to +/-90° the function would return a negative value for the iam (the intercept for b=0.05 is at 87.21°: https://www.desmos.com/calculator/y8wvzrx3b9).

pvlib/pvsystem.py
jkfm: As I understood it, if the input angle is 0, tau will be nan. 
zeroang = 1e-06 is defined for all cases, because iam is defined as tau(theta) / tau(0°), so that tau(0°) has to be calculated anyway. I didn't actually add the line, but moved it up from further down.

I guess we could define tau0 as a constant in the function and handling aoi == 0 with np.where after the calculations are done.

adriesse: I am surprised that the model would not work for an angle of exactly zero degrees.  Could this have something to do with the equation discrepancy mentioned in the comments further above?

cwhanse:
I vote for IAM=0 where AOI>=90, which is consistent with its definition and its use in the modeling process.

Where AOI=0 (or very close to 0) in the physical IAM model the formula needs to switch to the limit expression in my earlier comment. That switch should be the use of zeroang.

Sent from my iPhone

On Jun 30, 2017, at 3:20 PM, Anton Driesse <notifications@github.com<mailto:notifications@github.com>> wrote:


@adriesse commented on this pull request.

________________________________

In pvlib/pvsystem.py<https://github.com/pvlib/pvlib-python/pull/339#discussion_r125122764>:

> @@ -811,6 +812,10 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
     spa
     ashraeiam
     '''
+    zeroang = 1e-06
+
+    aoi = np.where(aoi == 0, zeroang, aoi)


I am surprised that the model would not work for an angle of exactly zero degrees. Could this have something to do with the equation discrepancy mentioned in the comments further above?

-

pvlib/pvsystem.py
jkfm: I don't think it has to do with the function discrepancy. The unmodified function for theta_r is not defined for angles > ~40°.

The tau(theta)-function in turn is just not defined for theta = 0. It seems this was ignored in the paper...

jkfm:
I would also prefer IAM to return zero for angles outside -90/90 degrees.

pvlib/pvsystem.py
```diff
@@ -831,7 +834,7 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
```
adriesse: I agree with your suggestions, especially after having consulted my D&B book.  I would suggest calculating reflectance and absorption separately before combining them in order to clarify the code.

jkfm: I think I applied both your suggestions correctly. Could you take a look?

adriesse: It wasn't quite correct, so I thought it might be easier to communicate it this way:
```
    # angle of refraction
    thetar_deg = tools.asind(tools.sind(aoi) / n)

    # reflectance and transmittance for normal incidence light
    rho_zero = ((1-n) / (1+n)) ** 2
    tau_zero = np.exp(-K*L)

    # reflectance for parallel and perpendicular polarized light
    rho_para = (tools.tand(thetar_deg - aoi) /
                tools.tand(thetar_deg + aoi)) ** 2
    rho_perp = (tools.sind(thetar_deg - aoi) /
                tools.sind(thetar_deg + aoi)) ** 2

    # transmittance for non-normal light
    tau = np.exp(-K*L / tools.cosd(thetar_deg))

    # iam is ratio of non-normal to normal incidence transmitted light
    # after deducting the reflected portion of each
    iam = ((1 - (rho_para + rho_perp) / 2) / (1 - rho_zero)
          * tau / tau_zero )


    # angles near zero produce nan, but iam is defined as one
    small_angle = 1e-06
    iam = np.where(np.abs(aoi) < small_angle, 1.0, iam)

    # angles at 90 degrees can produce tiny negative values, which should be zero
    # this is a result of calculation precision rather than the physical model
    iam = np.where(iam <  0, 0, iam)    

    # for light coming from behind the plane, none can enter the module
    iam = np.where(aoi > 90, 0, iam)

    # negative angles are not logical
    iam = np.where(aoi <  0, np.nan, iam)

```

jkfm: Thanks for the detailed explanation! So does this refer to reference 2 in the Docstring (I don't have access so cannot check)?

adriesse: Ref [1] simply describes what is in ref [2].

adriesse:
Another observation: contrary to the statement in ref [1] the default value of 2 mm for the glass thickness is definitely at the thin extreme.  The value I currently see most often is 3.2 mm. Previously I believe 4 mm was more common.

jkfm:
I agree with the default value, but I could not find any sources on it. Do you think, we should change the default nevertheless?

pvlib/pvsystem.py
jkfm: I implemented your suggestion except for setting negative angles to np.nan, to keep consistency with the other IAM functions for now...

adriesse: Consistency has some appeal...

adriesse:
The source for the 2mm default has no sources either...

cwhanse:
If it helps, the model output is essentially insensitive to the thickness parameter.  You can double from 0.002 to 0.004 and the IAM changes very slightly.

For myself, I'd prefer the default to remain the value in the reference paper, simply for traceability back to literature.  I do see merit in changing to 3.2mm, though.

wholmgren:
I suggest keeping the default set to 2 mm. I'll merge tomorrow unless there are objections.

PR #340, opened 2017-06-15
Support ideal PV devices (#324)

### LINKED ISSUE #324: Ideal PV devices are not supported
One cannot define `resistance_series = 0` or `resistance_shunt = numpy.inf` as "ideal" device parameters in `pvlib.pvsystem.singlediode()`. See the error below for example.

Is this feature valued enough for me to make a PR that adds this capability?

Input:
```python
import sys
print("Python version: %s.%s.%s" % sys.version_info[:3])

import numpy as np
print("Numpy version: " + np.version.version)

import scipy as sp
print("Scipy version: " + sp.version.version)

import pvlib

i_ph_A = 2.13643772
i_rs_A = 2.10905501e-05
n_mod_V = 9.91331584
r_s_Ohm = 3.63866316
r_s_Ohm = 0.
g_p_Mho = 2.45880685e-03

pvlib.pvsystem.singlediode(i_ph_A, i_rs_A, r_s_Ohm, 1./g_p_Mho, n_mod_V)
```
with output:
```
Python version: 3.6.1
Numpy version: 1.12.1
Scipy version: 0.19.0
/Users/markcampanelli/anaconda3/lib/python3.6/site-packages/pvlib/pvsystem.py:1943: RuntimeWarning: divide by zero encountered in true_divide
  I = -V/(Rs + Rsh) - (nNsVth/Rs)*lambertwterm + Rsh*(IL + I0)/(Rs + Rsh)
/Users/markcampanelli/anaconda3/lib/python3.6/site-packages/pvlib/pvsystem.py:1943: RuntimeWarning: invalid value encountered in double_scalars
  I = -V/(Rs + Rsh) - (nNsVth/Rs)*lambertwterm + Rsh*(IL + I0)/(Rs + Rsh)
Out[1]:
OrderedDict([('i_sc', nan),
             ('v_oc', 112.87971015555131),
             ('i_mp', nan),
             ('v_mp', 0.02227117920245893),
             ('p_mp', nan),
             ('i_x', nan),
             ('i_xx', nan)])
```

thunderfish24:
Based on feedback, I decided to wait on implementing my full algorithm that uses the absolute residual in the sum of currents at the diode node to determine when an explicit computation for an ideal device is a better approximation for a very near ideal device. This PR documents one really edgy case where this appears to have been a better numerical computation.

By reformulating some of the LambertW formulas, this PR does make many computations more stable, so it appears to be harder to get into a poorly behaved edge case.

In future work, I propose changing the shunt resistance parameter Rsh to a shunt conductance parameter Gsh=1./Rsh, which appears to be much better behaved numerically for ideal and near-ideal devices, i.e., Gsh is approaching/at zero. This should avoid a lot of logic bloat in considering yet another limiting case in pvsystem.i_from_v() where Rsh is so large that it is essentially infinite and the shunt term in the model must be dropped out. (Kudos to Paul Basore for recommending that I switch to Gsh in my early modeling days at NREL.)

 - [x] Closes #324 
 - [x] Tests added / passed
 - [x] Passes ``git diff upstream/master -- pvlib | flake8 --diff``
 - [x] Fully documented, including the appropriate `docs/sphinx/source/whatsnew` file for all changes and the `docs/sphinx/source/api.rst` for new API


thunderfish24:
I updated pvsystem.i_from_v() to use a shunt conductance derived from the shunt resistance input, without changing the function interface. Rsh=np.inf with Rs>0 test case should now be passing. Note that computing short-circuit current with IL=0 (from zero irradiance) also appears to be improved.

I'm going to start on the analogous changes to pvsystem.v_from_i(), and hopefully get some stability tests written for parameters approaching the ideal cases.

cwhanse:
Mark,

Looks good so far.  Thanks for taking this on.

wholmgren:
Thanks @thunderfish24. I don't have time to comment on the substance at this point, but here are some more style-oriented things that I noticed with a quick glance: 

- [ ] functions need descriptive documentation
- [ ] keep the line length to 79 characters or less
- [ ] set your editor to trim extra white space from blank lines and the end of lines
- [ ] use required arguments for required inputs (instead of keyword arguments)
- [ ] more short tests are better than fewer long tests. pytest.parametrize can be useful here
- [ ] it appears to me that the functions will work with any numeric input type (ndarrays, Series, DataFrames, scalars) so perhaps "numeric" is more appropriate. In any case, "numpy.float64 or python scalar" can probably be replaced with "scalar"

flake8 is a good tool for tracking down some style issues.

thunderfish24:
@wholmgren Thanks for the additional tips.

thunderfish24:
Here is an example of computational instability in the existing v_from_i() as Rsh->inf, and its correction in v_from_i_alt() . For this particular parameter set, the Rsh value has to be fairly large before the voltage computation goes off the rails.

```python
import pvlib
import numpy as np

Rsh = 190.
Rs = 1.065
nNsVth = 2.89
I0 = 7.05196029e-08
IL = 10.491262

I = 0.
Voc_ideal = pvlib.pvsystem.v_from_i_alt(Rsh=np.inf, Rs=Rs, nNsVth=nNsVth, I=I, I0=I0, IL=IL)
print('Voc_ideal_V = ' + str(Voc_ideal))
print()

Rsh_pvlib_limit_vec = Rsh*np.logspace(0, 15, num=15)
for Rsh_mod in Rsh_pvlib_limit_vec:
    V_pvlib = pvlib.pvsystem.v_from_i(Rsh_mod, Rs, nNsVth, 0., I0, IL)
    print('Rsh_mod_Ohm = ' + str(Rsh_mod) + ', V_pvlib_V = ' + str(V_pvlib) + ': current_sum_A = ' + str(pvlib.pvsystem.current_sum_at_diode_node(V_pvlib, I, IL, I0, nNsVth, Rs, Rsh_mod)))

print()

for Rsh_mod in Rsh_pvlib_limit_vec:
    V_pvlib_alt, meta_dict = pvlib.pvsystem.v_from_i_alt(Rsh=Rsh_mod, Rs=Rs, nNsVth=nNsVth, I=I, I0=I0, IL=IL, return_meta_dict=True)
    print('Rsh_mod_Ohm = ' + str(Rsh_mod) + ', V_pvlib_alt_V = ' + str(V_pvlib_alt) + ': current_sum_A = ' + str(meta_dict['current_sum_at_diode_node']) + ', inf_Rsh_idx = ' + str(meta_dict['inf_Rsh_idx']))
    print('inf_Rsh_idx = ' + str(meta_dict['inf_Rsh_idx']))

print()
print()

V = 0.
Isc_ideal = pvlib.pvsystem.i_from_v_alt(Rsh=np.inf, Rs=Rs, nNsVth=nNsVth, V=V, I0=I0, IL=IL)
print('Isc_ideal_A = ' + str(Isc_ideal))
print()

Rsh_pvlib_limit_vec = Rsh*np.logspace(0, 15, num=15)
for Rsh_mod in Rsh_pvlib_limit_vec:
    V_pvlib = pvlib.pvsystem.v_from_i(Rsh_mod, Rs, nNsVth, Isc_ideal, I0, IL)
    print('Rsh_mod_Ohm = ' + str(Rsh_mod) + ', V_pvlib_V = ' + str(V_pvlib) + ': current_sum_A = ' + str(pvlib.pvsystem.current_sum_at_diode_node(V_pvlib, Isc_ideal, IL, I0, nNsVth, Rs, Rsh_mod)))

print()

for Rsh_mod in Rsh_pvlib_limit_vec:
    V_pvlib_alt, meta_dict = pvlib.pvsystem.v_from_i_alt(Rsh=Rsh_mod, Rs=Rs, nNsVth=nNsVth, I=Isc_ideal, I0=I0, IL=IL, return_meta_dict=True)
    print('Rsh_mod_Ohm = ' + str(Rsh_mod) + ', V_pvlib_alt_V = ' + str(V_pvlib_alt) + ': current_sum_A = ' + str(meta_dict['current_sum_at_diode_node']) + ', inf_Rsh_idx = ' + str(meta_dict['inf_Rsh_idx']))
    print('inf_Rsh_idx = ' + str(meta_dict['inf_Rsh_idx']))
```

```
Voc_ideal_V = 54.38378255010209

Rsh_mod_Ohm = 190.0, V_pvlib_V = 54.3039588338: current_sum_A = -1.09656728142e-12
Rsh_mod_Ohm = 2239.66040611, V_pvlib_V = 54.3770866821: current_sum_A = -1.35305239235e-11
Rsh_mod_Ohm = 26400.4143931, V_pvlib_V = 54.3832150488: current_sum_A = 6.02720963416e-11
Rsh_mod_Ohm = 311199.804321, V_pvlib_V = 54.3837344102: current_sum_A = 8.45458869689e-10
Rsh_mod_Ohm = 3668325.68488, V_pvlib_V = 54.3837784678: current_sum_A = -5.59872070431e-09
Rsh_mod_Ohm = 43241072.5954, V_pvlib_V = 54.383782208: current_sum_A = -1.56670771722e-08
Rsh_mod_Ohm = 509712201.103, V_pvlib_V = 54.3837823868: current_sum_A = 4.86196644805e-07
Rsh_mod_Ohm = 6008327554.32, V_pvlib_V = 54.3837890625: current_sum_A = -2.36503490364e-05
Rsh_mod_Ohm = 70824280686.0, V_pvlib_V = 54.3837890625: current_sum_A = -2.36420655035e-05
Rsh_mod_Ohm = 834854406545.0, V_pvlib_V = 54.3837890625: current_sum_A = -2.36413627759e-05
Rsh_mod_Ohm = 9.84100189054e+12, V_pvlib_V = 54.359375: current_sum_A = 0.0882310508952
Rsh_mod_Ohm = 1.16002643635e+14, V_pvlib_V = 54.25: current_sum_A = 0.474587242214
Rsh_mod_Ohm = 1.3674027787e+15, V_pvlib_V = 54.0: current_sum_A = 1.30466118631
Rsh_mod_Ohm = 1.61185150666e+16, V_pvlib_V = 64.0: current_sum_A = -281.859291829
Rsh_mod_Ohm = 1.9e+17, V_pvlib_V = 256.0: current_sum_A = -2.08298484277e+31

Rsh_mod_Ohm = 190.0, V_pvlib_alt_V = 54.303958833791285: current_sum_A = -1.0965672814222671e-12, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 2239.66040611, V_pvlib_alt_V = 54.377086682099616: current_sum_A = -1.3530523923499516e-11, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 26400.4143931, V_pvlib_alt_V = 54.38321504875785: current_sum_A = 6.027209634162922e-11, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 311199.804321, V_pvlib_alt_V = 54.383734410163015: current_sum_A = 8.454588696889601e-10, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 3668325.68488, V_pvlib_alt_V = 54.38377846777439: current_sum_A = -5.598720704305367e-09, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 43241072.5954, V_pvlib_alt_V = 54.38378220796585: current_sum_A = -1.5667077172235486e-08, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 509712201.103, V_pvlib_alt_V = 54.38378255010209: current_sum_A = -1.066950876690535e-07, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 6008327554.32, V_pvlib_alt_V = 54.38378255010209: current_sum_A = -9.051411744122236e-09, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 70824280686.0, V_pvlib_alt_V = 54.38378255010209: current_sum_A = -7.6787984118612e-10, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 834854406545.0, V_pvlib_alt_V = 54.38378255010209: current_sum_A = -6.515229496271385e-11, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 9.84100189054e+12, V_pvlib_alt_V = 54.38378255010209: current_sum_A = -5.536902638802866e-12, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 1.16002643635e+14, V_pvlib_alt_V = 54.38378255010209: current_sum_A = -4.794731684004577e-13, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 1.3674027787e+15, V_pvlib_alt_V = 54.38378255010209: current_sum_A = -5.0429730941841095e-14, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 1.61185150666e+16, V_pvlib_alt_V = 54.38378255010209: current_sum_A = -1.4032135621217036e-14, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 1.9e+17, V_pvlib_alt_V = 54.38378255010209: current_sum_A = -1.0944371470875724e-14, inf_Rsh_idx = True
inf_Rsh_idx = True


Isc_ideal_A = 10.491258702613099

Rsh_mod_Ohm = 190.0, V_pvlib_V = -11.1725640177: current_sum_A = 0.0
Rsh_mod_Ohm = 2239.66040611, V_pvlib_V = -11.1658058955: current_sum_A = 0.0
Rsh_mod_Ohm = 26400.4143931, V_pvlib_V = -11.0861950324: current_sum_A = 0.0
Rsh_mod_Ohm = 311199.804321, V_pvlib_V = -10.156299235: current_sum_A = 1.7763568394e-15
Rsh_mod_Ohm = 3668325.68488, V_pvlib_V = -3.07774074453: current_sum_A = 1.7763568394e-15
Rsh_mod_Ohm = 43241072.5954, V_pvlib_V = -0.225844280917: current_sum_A = 1.42108547152e-14
Rsh_mod_Ohm = 509712201.103, V_pvlib_V = -0.0188397606685: current_sum_A = 1.75859327101e-13
Rsh_mod_Ohm = 6008327554.32, V_pvlib_V = -0.00159774131316: current_sum_A = 2.0872192863e-12
Rsh_mod_Ohm = 70824280686.0, V_pvlib_V = -0.000156491063535: current_sum_A = 2.46078712962e-11
Rsh_mod_Ohm = 834854406545.0, V_pvlib_V = -0.000260397326201: current_sum_A = 2.90061308306e-10
Rsh_mod_Ohm = 9.84100189054e+12, V_pvlib_V = -0.00293508917093: current_sum_A = 3.41758088496e-09
Rsh_mod_Ohm = 1.16002643635e+14, V_pvlib_V = -0.0345864892006: current_sum_A = 4.00655846278e-08
Rsh_mod_Ohm = 1.3674027787e+15, V_pvlib_V = -0.407692909241: current_sum_A = 4.43121054516e-07
Rsh_mod_Ohm = 1.61185150666e+16, V_pvlib_V = -4.80576324463: current_sum_A = 2.72938665802e-06
Rsh_mod_Ohm = 1.9e+17, V_pvlib_V = -56.6489257812: current_sum_A = 3.36790649413e-06

Rsh_mod_Ohm = 190.0, V_pvlib_alt_V = -11.17256401767651: current_sum_A = 0.0, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 2239.66040611, V_pvlib_alt_V = -11.165805895484587: current_sum_A = 0.0, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 26400.4143931, V_pvlib_alt_V = -11.08619503244795: current_sum_A = 0.0, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 311199.804321, V_pvlib_alt_V = -10.156299235004479: current_sum_A = 1.7763568394002505e-15, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 3668325.68488, V_pvlib_alt_V = -3.077740744526781: current_sum_A = 1.7763568394002505e-15, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 43241072.5954, V_pvlib_alt_V = -0.2258442809169594: current_sum_A = 1.4210854715202004e-14, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 509712201.103, V_pvlib_alt_V = -0.018839760668470262: current_sum_A = 1.758593271006248e-13, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 6008327554.32, V_pvlib_alt_V = -0.0015977413131622598: current_sum_A = 2.0872192862952943e-12, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 70824280686.0, V_pvlib_alt_V = -0.00015649106353521347: current_sum_A = 2.460787129621167e-11, inf_Rsh_idx = False
inf_Rsh_idx = False
Rsh_mod_Ohm = 834854406545.0, V_pvlib_alt_V = 1.1329781557378737e-10: current_sum_A = -1.3383072428041487e-11, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 9.84100189054e+12, V_pvlib_alt_V = 1.1329781557378737e-10: current_sum_A = -1.13509202037676e-12, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 1.16002643635e+14, V_pvlib_alt_V = 1.1329781557378737e-10: current_sum_A = -9.592326932761353e-14, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 1.3674027787e+15, V_pvlib_alt_V = 1.1329781557378737e-10: current_sum_A = -8.881784197001252e-15, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 1.61185150666e+16, V_pvlib_alt_V = 1.1329781557378737e-10: current_sum_A = 0.0, inf_Rsh_idx = True
inf_Rsh_idx = True
Rsh_mod_Ohm = 1.9e+17, V_pvlib_alt_V = 1.1329781557378737e-10: current_sum_A = 0.0, inf_Rsh_idx = True
inf_Rsh_idx = True
```

thunderfish24:
@wholmgren @cwhanse I think this is in a state where a preliminary review would be helpful now. Some notes--
- I used a sdm_ prefix (for single diode model) on my three new functions: sdm_current_sum(), sdm_v_from_i(), and sdm_i_from_v(). One migration strategy would be to introduce them as is, mark the old ones as deprecated, and then switch over after a grace period and after we get some mileage on the new functions. I do not really know the extent of the other code paths and user use cases (e.g., Monte Carlo simulation?) that I am touching here.
- My functions call into v_from_i() and i_from_v() at this point to keep the code DRY. Note: I have code that avoids having to do log transforms for large LambertW inputs that overflow, but I don't think it is ready for pvlib yet, esp. because it should be at least as good as scipy.labertw().
- Because of a peculiarity in numpy.where(), sdm_v_from_i() and sdm_i_from_v() always return an numpy.ndarray, which will be rank-0 for scalar-only input. I actually like this consistency of output.
- I did not have luck passing pandas.DataFrame to the sdm_current_sum() unit test. It seems to behave differently than numpy.ndarray and pandas.Series, which concerns me, but I am unfamiliar with pandas. I don't really see the typical user needing this lower-level function anyhow.
- I'm not sure how much pandas-input unit testing to do in sdm_v_from_i() and sdm_i_from_v(). Can I just assume/hope that the numpy.asarray calls will do their job? Perhaps I should not just in case someone tries to remove the casting.
- Remaining TODO's for this PR concern numerical stability testing and benchmarks. The new code is indeed perhaps ~10x slower with better accuracy in some really edgy cases, but I cannot say if/how this would affect current consumers.

wholmgren:
I don't think a grace period is needed if we bump from version 0.4.x to 0.5.0. We would want to make sure that the new functions pass all of the old tests, though.

Is the 10x performance degradation important? From only looking at the code I am guessing that it's still plenty fast. For example, 10x slower is probably acceptable if it doesn't significantly impact the runtime of ``singlediode``.

Is ``return_meta_dict`` helpful for anything besides debugging? It seems to me that the few cases where someone would want that data could be better handled by manually computing it or using the debugger. Or did I miss something?

I don't think you need to worry about the pandas testing for these functions.

It would help if you could handle most of the errors reported when you run
```
git diff upstream/master | flake8 --diff --ignore N806
```
It makes things easier in the long run. Some of your longer lines could be shortened by using positional arguments instead of keyword arguments.

thunderfish24:
@wholmgren Thanks for the feedback. I have had to put this on hold for a bit while I finish up a paper.

thunderfish24:
I have decided that I would rather forgo including benchmarks and stability tests into the code, and hopefully use this PR as documentation of sufficient testing. It looks like the new code is roughly 2x slower:

'''python
Rsh = 20.
Rs = 0.1
nNsVth = 0.5
I = 3.
I0 = 6.e-7
IL = 7.

N = 10000

t_start = time.time()
for k in range(N):
    pvlib.pvsystem.v_from_i(Rsh, Rs, nNsVth, I, I0, IL)
t_end = time.time()
reg_delta_t = t_end - t_start
print('pvlib.pvsystem.v_from_i: average seconds per computation ' + str(reg_delta_t/N))

t_start = time.time()
for k in range(N):
    pvlib.pvsystem.sdm_v_from_i(Rsh=Rsh, Rs=Rs, nNsVth=nNsVth, I=I, I0=I0, IL=IL)
t_end = time.time()
alt_delta_t = t_end - t_start
print('pvlib.pvsystem.sdm_v_from_i: average seconds per computation ' + str(alt_delta_t/N))

print('pvlib.pvsystem.sdm_v_from_i is {} times slower'.format(alt_delta_t/reg_delta_t))
'''
gives
'''python
pvlib.pvsystem.v_from_i: average seconds per computation 7.457981109619141e-05
pvlib.pvsystem.sdm_v_from_i: average seconds per computation 0.00014008252620697023
pvlib.pvsystem.sdm_v_from_i is 1.8782901719380174 times slower
'''
and
'''python
Rsh = 20.
Rs = 0.1
nNsVth = 0.5
V = 0.
I0 = 6.e-7
IL = 7.

N = 10000

t_start = time.time()
for k in range(N):
    pvlib.pvsystem.i_from_v(Rsh, Rs, nNsVth, V, I0, IL)
t_end = time.time()
reg_delta_t = t_end - t_start
print('pvlib.pvsystem.i_from_v: average seconds per computation ' + str(reg_delta_t/N))

t_start = time.time()
for k in range(N):
    pvlib.pvsystem.sdm_i_from_v(Rsh=Rsh, Rs=Rs, nNsVth=nNsVth, V=V, I0=I0, IL=IL)
t_end = time.time()
alt_delta_t = t_end - t_start
print('pvlib.pvsystem.sdm_i_from_v: average seconds per computation ' + str(alt_delta_t/N))

print('pvlib.pvsystem.sdm_i_from_v is {} times slower'.format(alt_delta_t/reg_delta_t))
'''
gives
'''python
pvlib.pvsystem.i_from_v: average seconds per computation 5.51548957824707e-05
pvlib.pvsystem.sdm_i_from_v: average seconds per computation 0.00012906100749969482
pvlib.pvsystem.sdm_i_from_v is 2.3399737352184955 times slower
'''

thunderfish24:
@wholmgren "Is return_meta_dict helpful for anything besides debugging? It seems to me that the few cases where someone would want that data could be better handled by manually computing it or using the debugger. Or did I miss something?"

The `meta_dict` would probably be used largely for debugging, but I claim that reconstructing this info is not so easily done outside the function call. The default behavior is also the behavior the vast majority of users would want/expect.

thunderfish24:
@wholmgren @cwhanse Assuming this passes CI, it is ready for review for merging into master. I am strongly in favor of having a deprecation period for the existing `pvsystem.i_from_v` and `pvsystem.v_from_i` functions, which will no longer be used internally after this PR (except where these functions' code needs to be ported to the new functions).

pvlib/test/test_pvsystem.py
```diff
@@ -465,30 +822,28 @@ def test_singlediode_series_ivcurve(cec_module_params):
...
+                            ('i_x', array([0., 2.88414114, 5.74622046])),
+                            ('i_xx', array([0., 2.04340914, 3.90007956])),
+                            ('v', array([[0., 0., 0.],
+                                         [0., 4.98443481, 9.96886962],
+                                         [0., 5.14765242, 10.29530483]])),
+                            ('i', array([[0., 0., 0.],
+                                         [3.01079860e+00, 2.88414114e+00,
+                                          3.10862447e-14],
+                                         [6.00726296e+00, 5.74622046e+00,
+                                          0.00000000e+00]]))])
```
thunderfish24: Be gone nan's from zero irradiance calculations!

thunderfish24:
Looking into these failures on only some python configurations.

image

thunderfish24:
@wholmgren Cliff had concerns about documenting my changes to the published algorithm. Any suggestions here? I'd like to get a thorough review after posting a description somewhere good.

wholmgren:
The test failures are unrelated... I need to make them more reliable, sorry. I like the badge.

>  I am strongly in favor of having a deprecation period for the existing pvsystem.i_from_v and pvsystem.v_from_i functions

I don't understand this. It seems to me that your functions are more robust and there are no API changes to worry about. The only downside is that they are ~2x slower, but that's probably not going to be noticed by most users. Are you not confident that all of the above is true? 

This is my understanding of what you'd like to see:
1. Mark the older ``i_from_v`` and ``v_from_i`` functions as deprecated in documentation
2. Release code with a minor version change.
2. Users update their code to use ``sdm_i_from_v`` and ``sdm_v_from_i``
2. Replace the code in the older functions with the code from the new ``sdm*`` functions. Keep ``sdm*`` functions for compatibility.
2. Release code with a bigger version change.
2. Users update their code to use ``i_from_v`` and ``v_from_i``
2. Remove ``sdm*`` functions
2. Release code with a bigger version change.

That's asking users to change their code a lot. It would also take a year for this to play out at our current pace. Maybe I missed something and you could correct me by copying/editing the list above. 

I could be convinced to add the new functions with a deprecation warning stating that they will replace the older functions in the next release. This way people would know that they should only use the new functions on a provisional basis. The warning could include a request for feedback.

> Cliff had concerns about documenting my changes to the published algorithm. Any suggestions here? I'd like to get a thorough review after posting a description somewhere good.

Were those concerns on Github? If so, I cannot find them. Maybe you could add some documentation to the [pvsystem notebook](https://github.com/pvlib/pvlib-python/blob/master/docs/tutorials/pvsystem.ipynb)? Or make a Github gist? You could post graphs here, too. 

The parameters in your new functions should use the same long names that the older functions use, to be consistent with the [variable and style rules](http://pvlib-python.readthedocs.io/en/latest/variables_style_rules.html). It's fine to redefine them within the functions if it improves readability.

cwhanse:
>> Cliff had concerns about documenting my changes to the published algorithm. Any suggestions here? I'd like to get a thorough review after posting a description somewhere good.

> Were those concerns on Github? If so, I cannot find them. Maybe you could add some documentation to the pvsystem notebook? Or make a Github gist? You could post graphs here, too.

Not on GitHub.  For algorithm and documentation concerns, Mark and I will work through them by email first.

thunderfish24:
@wholmgren Sorry I missed the parameter naming requirement. I will get that fixed.

In terms of the user expectations I have one correction: The expectation is that users altogether REMOVE usage of i_from_v and v_from_i and PERMANENTLY SWITCH to `sdm_i_from_v` and `sdm_v_from_i`. The major version bump is the step where the old functions are removed.

That said, you may have convinced me that deprecation of `i_from_v` and `v_from_i` is overkill in this particular situation. Depending on one's viewpoint, what we have essentially done is either fix the computation for certain inputs (patch) or extend the domain of the inputs (minor). It can be argued that neither of these breaks the input signature (major). The performance loss is more of a grey area, but 2x probably isn't horrible and would probably disrupt way fewer people than the deprecation.

That that said, I am now looking into some ramifications to `pvsystem.singlediode` that I think I missed when I updated the unit test for that function. I'll get back to you on that. I am also checking the soundness of my algorithms and formulas with Cliff's help.

cwhanse:
I am not in favor of changing the function name.  I thought Mark’s work would improve the computation but I didn’t, and don’t, see a compelling reason to jettison the existing function names.

There’s some merit in considering the prefix ‘sdm_’ etc. but only if we are also considering adding other diode-like models, e.g., two-diode ‘ddm_’ etc.  We’re not at that point so I think the prefix is unnecessary.

thunderfish24:
What I have done improves the computation in some edgy cases, but I claim these edgy cases could realistically be hit in parameter fitting algorithms that return values that are in the edge case scenario.

Also, my code aims to fix the current `pvsystem.singlediode` computing nan's when IL=0. (Again, is a device ever in complete darkness? Probably almost never, but users probably set IL=0 at nighttime.)

pvlib/test/test_pvsystem.py
cwhanse: I suspect the NaNs arise from this line in calcparams_desoto:

    Rsh = Rsh_ref * (irrad_ref / poa_global)

We should change to use np.divide here

pvlib/pvsystem.py
```diff
@@ -1929,23 +1938,253 @@ def i_from_v(resistance_shunt, resistance_series, nNsVth, voltage,
...
+    V = np.asarray(voltage, np.float64)
+    I0 = np.asarray(saturation_current, np.float64)
+    IL = np.asarray(photocurrent, np.float64)
+
+    # This transforms any ideal Rsh=np.inf into Gsh=0., which is generally more numerically stable
+    Gsh = 1./Rsh
+
+    # argW cannot be float128
+    argW = \
+Rs*I0/(nNsVth*(Rs*Gsh + 1.))*np.exp((Rs*(IL + I0) + V)/(nNsVth*(Rs*Gsh + 1.)))
```
thunderfish24: @cwhanse Notice how this formulation is numerically better behaved in limit of ideal case where Rs=0 and/or Gsh=0 (aka Rsh=inf).

pvlib/pvsystem.py
```diff
@@ -1929,23 +1938,253 @@ def i_from_v(resistance_shunt, resistance_series, nNsVth, voltage,
...
     lambertwterm = lambertw(argW).real
 
     # Eqn. 4 in Jain and Kapoor, 2004
-    I = -V/(Rs + Rsh) - (nNsVth/Rs)*lambertwterm + Rsh*(IL + I0)/(Rs + Rsh)
+    I = (IL + I0 - V*Gsh)/(Rs*Gsh + 1.) - (nNsVth/Rs)*lambertwterm
```
thunderfish24: @cwhanse Notice how this formulation is numerically better behaved in limit of ideal case where Rs=0 and/or Gsh=0 (aka Rsh=inf).

cwhanse: Let's leave the original in as a comment to accompany the reference, it helps if someone wants to understand the source of the equation.

pvlib/pvsystem.py
cwhanse: I agree, but let's leave the original formula in as a comment, help connect the dots.

thunderfish24:
@cwhanse @wholmgren So there is some question as to if the computation of model residuals that I use to gauge the quality of solutions is sufficiently well computed numerically (as opposed to being mathematically sound, which seems sufficiently obvious to me). Perhaps we should simplify the logic for this PR and only check if the series resistance is zero (respectively, parallel resistance is infinity) and compute using the explicit solution in only those cases. We could leave the very near ideal cases where the LambertW computation seems inaccurate to future work. This would mean that this PR should be guaranteed to not make any computations worse.

This said, I would claim that I have seen no better evidence that the LambertW solution is sufficiently well computed numerically (as opposed to being mathematically sound, which is well documented in the literature), and maybe we should write a paper that, for many cases, numerically checks the residuals of the implicit model equation for both solutions using "provable" interval analysis, and maybe even some numerical analysis proofs that could put all these computations on a firmer standing.

thunderfish24:
TODOs at this point:

  - [ ] Remove using residual checks for numerical edge cases with "near ideal" devices until there is more proof that it is indeed better.
  - [ ] Use standard variable name conventions and merge new functions into old function to maintain existing API.
  - [ ] Show connection to published LambertW-related equations in Jain and Kapoor 2004, even though numerically they are now re-formulated for better numerical stability.
  - [ ] Make dimension matching clearer in the code in ideal cases, and make sure unit tests cover these cases so that they break if dimensioning is removed.

pvlib/test/test_pvsystem.py
```diff
@@ -353,37 +353,224 @@ def test_PVSystem_calcparams_desoto(cec_module_params):
     assert_allclose(nNsVth, 0.473)
 
 
-@requires_scipy
-def test_v_from_i():
```
thunderfish24: Now in the test fixture.

pvlib/test/test_pvsystem.py
```diff
@@ -353,37 +353,224 @@ def test_PVSystem_calcparams_desoto(cec_module_params):
...
-    output = pvsystem.v_from_i(20, .1, .5, 3, 6e-7, 7)
-    assert_allclose(7.5049875193450521, output, atol=1e-5)
-
-
-@requires_scipy
-def test_v_from_i_big():
```
thunderfish24: Now in the test fixture.

pvlib/test/test_pvsystem.py
```diff
@@ -353,37 +353,224 @@ def test_PVSystem_calcparams_desoto(cec_module_params):
...
+      'I0': 7.05196029e-08,
+      'IL': 10.491262,
+      'V_expected': 54.303958833791455
+    }])
+def fixture_v_from_i(request):
+    return request.param
 
 
 @requires_scipy
-def test_v_from_i_bigger():
```
thunderfish24: Now in the test fixture.

pvlib/test/test_pvsystem.py
```diff
@@ -353,37 +353,224 @@ def test_PVSystem_calcparams_desoto(cec_module_params):
...
+      'I0': 6.e-7,
+      'IL': 7.,
+      'I_expected': -299.7383436645412
+    }])
+def fixture_i_from_v(request):
+    return request.param
 
 
 @requires_scipy
-def test_i_from_v():
```
thunderfish24: Now in the test fixture.

docs/sphinx/source/api.rst
```diff
@@ -201,10 +201,10 @@ Functions relevant for the single diode model.
 .. autosummary::
    :toctree: generated/
 
-   pvsystem.singlediode
    pvsystem.calcparams_desoto
-   pvsystem.v_from_i
    pvsystem.i_from_v
+   pvsystem.singlediode
+   pvsystem.v_from_i
```
thunderfish24: Alphabetized.

pvlib/test/test_pvsystem.py
thunderfish24: This is "fixed" in the sense that Rsh is computed as +inf when poa_global is zero, and now the i_from_v and v_from_i code can properly handle this infinity.

pvlib/pvsystem.py
```diff
@@ -1860,51 +1880,111 @@ def v_from_i(resistance_shunt, resistance_series, nNsVth, current,
...
+                       (Gsh[idx] * nNsVth[idx]))[idx_w]
+
+            # Three iterations of Newton-Raphson method to solve
+            #  w+log(w)=logargW. The initial guess is w=logargW. Where direct
+            #  evaluation (above) results in NaN from overflow, 3 iterations
+            #  of Newton's method gives approximately 8 digits of precision.
+            w = logargW
+            for _ in range(0, 3):
+                w = w * (1 - np.log(w) + logargW) / (1 + w)
+            lambertwterm[idx_w] = w
```
thunderfish24: Looking at the coverage report, we never hit lines 1937-1949, even though one of the pre-existing unit tests (copied to the fixture in this PR) is supposed to cover this branch of the code. It may be that recasting in terms of Gsh helps avoid the overflow. @cwhanse Any ideas here?

thunderfish24: The unit test was from #226, I think.

thunderfish24:
@cwhanse @wholmgren Ready again for review. Looking for some feedback about testing LambertW input overflow in `v_from_i()`. Does anyone have any other realistic parameter combinations for unit testing this?

pvlib/pvsystem.py
cwhanse: Not sure why the parameters from #225 don't cause entry into that code block.  They cause overflow on the np.exp() call.

```
def test_v_from_i_bigger():
    # 1000 W/m^2 on a Canadian Solar 220M with 20 C ambient temp
    # github issue 225
'''

Comment (thunderfish24) [reply pvlib/pvsystem.py]: So it appears that the Gsh recast helps the existing test's computation stay finite:

In[5]: pvlib.pvsystem.v_from_i(190, 1.065, 2.89, 0, 7.05196029e-08, 10.491262)
Using LambertW with argW=[  1.64103617e+294] and lambertwterm=[ 670.94665556]
Out[5]: 54.303958833791285

However, the original parameters from #225  serve as a good test, I think:

In[5]: pvlib.pvsystem.v_from_i(381.68, 1.065, 2.681527737715915, 0, 1.8739027472625636e-09, 5.1366949999999996)
Using LambertW with argW=[ inf] and lambertwterm=[ inf]
/Users/markcampanelli/Documents/Programming/pvlib-python/pvlib/pvsystem.py:1926: RuntimeWarning: overflow encountered in exp
  np.exp((-I[idx] + IL[idx] + I0[idx]) / (Gsh[idx]*nNsVth[idx]))
Out[4]: 58.19323124611128

Note that in followup work I might be interested in checking how accurate lambertw() is for such large inputs.

I will add this parameter set as a unit test and verify the coverage

File: pvlib/test/test_pvsystem.py [hunk 10]
```diff
@@ -353,37 +353,235 @@ def test_PVSystem_calcparams_desoto(cec_module_params):
...
+    {  # Can handle all python scalar inputs with bigger LambertW arg
+      #  1000 W/m^2 on a Canadian Solar 220M with 20 C ambient temp
+      #  github issue 225
+      'Rsh': 381.68,
+      'Rs': 1.065,
+      'nNsVth': 2.681527737715915,
+      'I': 0.,
+      'I0': 1.8739027472625636e-09,
+      'IL': 5.1366949999999996,
+      'V_expected': 58.19323124611128
```
thunderfish24: @cwhanse Would you mind running this parameter set though the Matlab version to "independently" verify my V_expected value?

cwhanse: 58.193231246111300

thunderfish24: Awesome! Thanks.

thunderfish24:
Note: Some informal followup performance testing suggests that the additional computation and indexing logic causes these new computations to take slightly more than 2x time.

thunderfish24:
@wholmgren Still waiting on the Landscape result to verify increased code coverage with the new test. Other than that, I think it good to go except for any further reviewers you might try to muster.

cwhanse:
Makes sense –roughly a fixed computation cost to solve the ideal (or non-ideal case) using numpy vectorized operations.

Maybe we should switch to have the non-ideal as the default branch, and the ideal as the exception?  Currently the code always computes the ideal branch then checks for the non-ideal.

My guess is that the ideal branch isn’t the typical use.

thunderfish24:
So I currently use the ideal computation first as an "easy" way to get the correct shape of the output. It's probably worth investigating if a bunch of "input shape logic" is ultimately faster here.

thunderfish24:
Coverage of my changes is good now.

thunderfish24:
@cwhanse It looks like I may be able to use numpy.broadcast to some advantage here.

thunderfish24:
@cwhanse I did not see the speedup that I hoped for (still about 2-2.5x slower), but I the code is cleaner now. I think the improved computational stability is worth it at this point.

thunderfish24:
@wholmgren Note that the test failure before the last build appears to be unrelated to my changes.

thunderfish24:
@mikofski Would you be able to do a quick review of this PR?

mikofski:
@thunderfish24 I'll try to take a quick look next week

thunderfish24:
@wholmgren Can you provide a second code review so that perhaps we can get this merged?

wholmgren:
Yes, sorry, I have a number of pvlib things I’m aiming to get to by the end
of the week.

wholmgren:
@thunderfish24 this looks great! Thank you for your patience and flexibility as we worked this out.

thunderfish24:
Thanks @wholmgren.

PR #345, opened 2017-06-30
add explicit tests for aoi and aoi projection functions

wholmgren:
PR #339 discusses changing the angle of incidence modifier for values greater than 90 or less than 0. Reviewing the code made me realize that we're not explicitly testing pvlib's aoi calculation functions. So, this PR adds some simple tests of ``irradiance.aoi`` and ``irradiance.aoi_projection``. I think it's a reasonable set of parameters, but let me know if you think I should add another test case.


pvlib/test/test_irradiance.py
```diff
@@ -422,3 +422,22 @@ def test_dni():
...
+    (30, 180, 30, 180, 0, 1),
+    (30, 180, 150, 0, 180, -1),
+    (90, 0, 30, 60, 75.52248781407006, 0.25),
+    (90, 0, 30, 170, 119.49870423110367, -0.49240387650610395)])
+def test_aoi_and_aoi_projection(surface_tilt, surface_azimuth, solar_zenith,
+                                solar_azimuth,  aoi_expected,
+                                aoi_proj_expected):
+    aoi = irradiance.aoi(surface_tilt, surface_azimuth, solar_zenith,
+                         solar_azimuth)
+    assert_allclose(aoi, aoi_expected, atol=1e-6)
```
adriesse: Might as well drop all those decimal digits above.

Issue #344, opened 2017-06-30
_calc_d function of clearsky.py doesn't use passed argument


KonstantinTr:
It seem's that **w** is not used within function...

![image](https://user-images.githubusercontent.com/16244350/27731439-93fc2352-5d95-11e7-8830-5f0498f811d7.png)

wholmgren:
True, it can be removed. Feel free to submit a pull request.

alainakafkes:
Hi! I'm doing OpenSourceFriday PRs today, so I'll tackle this. :)

PR #346, opened 2017-06-30
Removes var w from _calc_d (Issue #344)

alainakafkes:
First commit removes var w from declaration of _calc_d, second commit removes var w from invocation of _calc_d.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.4.6.rst
+++ b/docs/sphinx/source/whatsnew/v0.4.6.rst
@@ -18,7 +18,7 @@ Enhancements
 
 API Changes
 ~~~~~~~~~~~
-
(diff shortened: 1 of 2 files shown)


wholmgren:
thanks @alainakafkes. Can you add a short note to the [v0.4.6.rst](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/whatsnew/v0.4.6.rst) whatsnew file?

alainakafkes:
@wholmgren Certainly! Would what I did go under the API Changes header?

wholmgren:
yes, API changes is the right place to put it.

wholmgren:
Great, thanks @alainakafkes! closes #344

Issue #343, opened 2017-06-30
cloud_cover_to_transmittance_linear function of forecast.py doesn't use passed 'offset' parameter


KonstantinTr:
Should 'offset' be used instead of 0.75 in the expression?

![image](https://user-images.githubusercontent.com/16244350/27729252-d095e026-5d8d-11e7-8ba9-31452d05445f.png)

wholmgren:
Thanks, good catch. Can you submit a pull request with a new test for this behavior?

KonstantinTr:
Sure. I'm new to git so it'll take some time to get how to do it :)

wholmgren:
closed by #348.

PR #348, opened 2017-07-12
here are fix and test for issue #343

KonstantinTr:
Now passed 'offset' value is used instead of 0.75 default value.


wholmgren:
Thanks @KonstantinTr. Sorry for the delay in getting to this. Can you add a note and your name to the 0.4.6 whatsnew file?

wholmgren:
I'll go ahead and merge this and then add the note myself. closes #343

PR #364, opened 2017-08-12
Handle overflow condition in pvlib.clearsky.haurwitz

cwhanse:
Revised test to operate on zenith rather than time, to remove dependence on solarposition


cwhanse:
Ready for review

wholmgren:
Thanks, Cliff. Looks good, I only have a couple of small style nitpicks...

The line with the calculation is too long (> 79 characters).

No spaces around equal signs inside a function call, so change ``data = clearsky_ghi`` to ``data=clearsky_ghi``.

Please delete the old, commented test code from the test file.

Can you create a new ``docs/sphinx/source/whatsnew/v0.5.1.rst`` file with a note about the change?

I see that there is a preexisting mistake in the function's documentation. The documentation claims that the function will return a Series, however, it always returns a DataFrame (I think this was done to be consistent with other clear sky functions). Can you go ahead and fix this?

I see that you first tried to use ``np.where`` instead of array indexing. Why did you switch? I'm guessing that ``np.where`` has better behavior for scalar inputs and it should preserve nans. It might also be faster because it only calculates the mask once and it avoids indexing. I think it's also fine the way it is.

cwhanse:
Will do.

np.where() evaluates both outcomes for all arguments before it applies the condition, so the overflow warning would still appear.

wholmgren:
Thanks, Cliff!

PR #400, opened 2017-11-30
Add GTI DIRINT model

### LINKED ISSUE #396: Add GTI DIRINT model
We should consider adding an implementation of the GTI DIRINT model described in Marion 2015:

https://www.sciencedirect.com/science/article/pii/S0038092X15005757

In short, the GTI DIRINT model converts plane of array irradiance to DNI, DHI, and GHI.

I have a partial implementation that I've been sitting on for a few months. I'll try to get it into a PR soon.

### LINKED ISSUE #540: irradiance.disc has unbounded clearness index calculation
In `pvlib.irradiance.disc` the clearness index `kt` is calculated as `kt = ghi / I0h`. Negative values of `kt` are removed but large values are not.  In rare circumstances, `ghi` (e.g., from a clear sky model) greatly exceeds `I0h` yielding `kt` values upwards of 80.  These values lead to a numpy overflow warning at

`delta_kn` = a + b + np.exp(c*am)`

because `c` is very large.

**To demonstrate:**
`import pvlib`
`from pandas import Timestamp`
` ts = Timestamp('2016-07-19 06:11:00-0600', tz='US/Mountain', freq='T')`
`pvlib.irradiance.disc(ghi=1.0, zenith=89.99, datetime_or_doy=ts)`

**Expected behavior**
`irradiance.disc` should use the new `clearness_index` function which sets an upper bound on `kt`

**Versions:**
 - ``pvlib.__version__``: 0.5.1

wholmgren:
This PR adds an implementation of the GTI DIRINT model. 

I'm not confident about my API design/changes, but I'm making the PR now due to new interest in issue #396. See the whatsnew diff for a list of the changes to existing functions. I did this to avoid copy/pasting code, but maybe there's a better way. Let me know what you think. 

~The algorithm works well for angles less than 90 degrees. It struggles for angles greater than 90 degrees. I'm not sure if that's because I made a mistake. I've only used synthetic data to test it so far.~
Edit: it now seems to work ok for angles greater than 90 degrees and it struggles for angles within a few degrees of 90.

Here's a notebook with examples:

https://gist.github.com/wholmgren/fe4d72ebb2c0e6481680bee5d9041b06

 - [x] Closes #396. Closes #540 

other: 

- [x] Fix AOI < 90 case or add comment regarding performance
- [x] Remove debug kwarg
- [x] tests for new kwargs in ``disc`` and ``dirint``
- [x] tests for clearness index functions
- [x] fix python 2.7 min failure
- [x] decide zero or nan and remove ``zero_instead_of_nan`` block
- [x] remove tutorial


wholmgren:
And thanks to @cwhanse for reviewing a previous version of this code and helping me figure out some problems with the implementation.

jforbess:
For what it's worth, once I figured out what you did to add the params to
the disc and dirint functions, I thought it was a good idea.

I used it fairly successfully, but didn't test it strongly myself, because
I didn't have measured GHI (thus the exercise).

wholmgren:
Sure. I'd do it like this on the command line (adapted from #443):

1. add a new "remote" so you can get my versions of pvlib-python on github. This will associate the name "wholmgren" with the url, but does nothing more: ``git remote add wholmgren https://github.com/wholmgren/pvlib-python.git``
2. fetch the data from the ``wholmgren`` remote (this doesn't touch your data or any local changes you've made. it just stores it in a hidden directory): ``git fetch wholmgren``
3. make a new local branch that's based on ``wholmgren/gtidirint``: ``git checkout -b gtidirint wholmgren/gtidirint``

I have no idea if the GitHub desktop client or similar can do this kind of thing. Maybe others can comment if they have experience.

Alternatively, you can skip git entirely by navigating to my ``gtidirint`` branch on github and downloading a zip file of the code (green button at right): https://github.com/wholmgren/pvlib-python/tree/gtidirint

adriesse:
Thanks!  The command line version worked.
I got it to run also, however it appears the doc string needs updating.

pvlib/irradiance.py
```diff
@@ -1400,6 +1405,268 @@ def dirindex(ghi, ghi_clearsky, dni_clearsky, zenith, times, pressure=101325.,
...
+    # make coeffs that is at least 30 elements long
+    # for loop below will limit iterations if necessary
+    coeffs = np.empty(max(30, max_iterations))
+    coeffs[0:3] = 1
+    coeffs[3:10] = 0.5
+    coeffs[10:20] = 0.25
+    coeffs[20:] = 0.125
+
+    # initialize diff
+    diff = pd.Series(9999, index=times)
```
adriesse: Would it be a good idea to convert any remaining 999 to nan at the end?

pvlib/irradiance.py
```diff
@@ -1400,6 +1405,268 @@ def dirindex(ghi, ghi_clearsky, dni_clearsky, zenith, times, pressure=101325.,
...
+def gti_dirint(poa_global, aoi, solar_zenith, solar_azimuth,
+               surface_tilt, surface_azimuth,
+               times, pressure=101325.,
+               use_delta_kt_prime=True, temp_dew=None, albedo=.25,
+               model='perez', model_perez='allsitescomposite1990',
+               calculate_gt_90=True, max_iterations=30):
+    """
+    Determine GHI, DNI, DHI from POA global using the GTI DIRINT model.
+
+    Parameters
```
adriesse: Parameters need updating.

wholmgren: updated.

pvlib/irradiance.py
wholmgren: to convert the ghi, dni, dhi to nan where diff remains 9999? alternatively, we could do this for any point that does not converge. This seems to easily occur near aoi = 90.

cwhanse:
@wholmgren I agree with the refactoring described above.  Private functions, with `disc` or `dirint` in the name, strike me as an effective way to break up the calculation without obscuring the relationship between this code and the reference publications.

I think the `clearness_index` function has been needed for some time.

wholmgren:
I implemented the public functions ``clearness_index`` and ``clearness_index_zenith_independent``, plus a handful of private functions. I think this makes the ``gti_dirint`` code much more explicit and it does not hurt the ``dirint`` and ``disc`` code.

I still don't get good ``git_dirint`` results for AOI between 80 and 90 degrees (see linked tutorial above) and I'm not sure why.

I'd appreciate a review or two of the API before writing the tests.

pvlib/irradiance.py
```diff
@@ -1178,7 +1178,89 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    [1] Maxwell, E. L., "A Quasi-Physical Model for Converting Hourly
+    Global Horizontal to Direct Normal Insolation", Technical
+    Report No. SERI/TR-215-3087, Golden, CO: Solar Energy Research
+    Institute, 1987.
+    """
+    cos_zenith = tools.cosd(solar_zenith)
+    I0h = extra_radiation * np.maximum(cos_zenith, min_cos_zenith)
+    kt = ghi / I0h
+    kt = np.maximum(kt, 0)
+    kt = np.minimum(kt, 0.82)  # consider replacing with 1 or nan assignment
```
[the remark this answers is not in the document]
wholmgren: I can add that I copied this limit from the kt_prime calculation in ``dirint``, which is justified with a reference to SRRL code. I don't know if that's good enough or if we should take a different approach here.

pvlib/irradiance.py
```diff
@@ -1509,6 +1706,337 @@ def dirindex(ghi, ghi_clearsky, dni_clearsky, zenith, times, pressure=101325.,
...
+    surface_azimuth : numeric
+        Surface azimuth angles in decimal degrees. surface_azimuth must
+        be >=0 and <=360. The Azimuth convention is defined as degrees
+        east of north (e.g. North = 0, South=180 East = 90, West = 270).
+
+    pressure : numeric, default 101325.0
+        The site pressure in Pascal. Pressure may be measured or an
+        average pressure may be calculated from site altitude.
+
+    use_delta_kt_prime : bool, default True
```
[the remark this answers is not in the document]
wholmgren: > any other numeric value uses the index

I'm confused about the comment quoted above. Currently we only allow True/False as options. If True, the stability index is calculated and used. If False, it is not. Are you suggesting that we allow people to provide their own stability index as inputs in ``dirint`` and ``gti_dirint``?

pvlib/irradiance.py
```diff
@@ -1345,31 +1464,124 @@ def dirint(ghi, zenith, times, pressure=101325., use_delta_kt_prime=True,
...
+
+    # Perez eqn 5
+    dni = disc_out['dni'] * dirint_coeffs
+
+    return dni
+
 
-    # wholmgren:
-    # the use_delta_kt_prime statement is a port of the MATLAB code.
-    # I am confused by the abs() in the delta_kt_prime calculation.
```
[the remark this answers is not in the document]
wholmgren: agreed. old comment that is being removed.

pvlib/irradiance.py
```diff
@@ -1323,6 +1426,22 @@ def dirint(ghi, zenith, times, pressure=101325., use_delta_kt_prime=True,
         DewPtTemp is not provided, then dew point improvements are not
         applied.
 
+    kt : None or array-like, default None
```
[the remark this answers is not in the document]
wholmgren: thanks.

pvlib/irradiance.py
```diff
@@ -1178,7 +1178,89 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    solar_zenith : numeric
+        True (not refraction-corrected) solar zenith angle in decimal
+        degrees.
+
+    extra_radiation : numeric
+        Irradiance incident at the top of the atmosphere
+
+    min_cos_zenith : numeric, default 0.065
+        Minimum value of cos(zenith) to allow when calculating global
+        clearness index `kt`. Equivalent to 86.273 degrees.
```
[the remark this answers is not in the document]
wholmgren: good idea

pvlib/irradiance.py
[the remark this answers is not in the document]
cwhanse: Disregard "any other numeric value" please, sloppy wording on my part.

pvlib/irradiance.py
```diff
@@ -1182,7 +1182,98 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    I0h = extra_radiation * np.maximum(cos_zenith, min_cos_zenith)
+    # consider adding
+    # with np.errstate(invalid='ignore', divide='ignore'):
+    # to kt calculation, but perhaps it's good to allow these
+    # warnings to the users that override min_cos_zenith
+    kt = ghi / I0h
+    kt = np.maximum(kt, 0)
+    # Limit copied from the kt prime limit in dirint, which was justified
+    # with reference to SRRL code. consider replacing with 1 or nan
+    kt = np.minimum(kt, 0.82)
```
cwhanse: I've looked into the source of 0.82 as the upper bound.  It does not appear in the references, which don't specify an upper bound.  The value 0.82 comes from a Fortran implementation of the DIRINT model at NREL, and appears to have been added as a modification to the original algorithm.

I recommend we keep the upper bound, but make it a kwarg (like `min_cos_zenith`) with default of 0.82, and point at NREL SRRL's Fortran code for DIRINT as the source (although the file appears no longer available through their website.)

An upper bound makes sense, but with the caution that the value will depend on the time averaging.  0.82 is probably an hourly value.  For a minute, clearness index > 1 is certainly possible if the GHI is enhanced by cloud effects.

wholmgren: Thanks for the good comment. We should definitely make a kwarg. I'd recommend a maximum value of 2 for a few reasons. My general feeling is that the defaults should be as permissive as possible and the user has some responsibility to QA/QC their data before putting it into this function. Also, subhourly data seems to be common these days.

cwhanse: 2 is OK with me.  G. Yordanov has a few conference papers looking at the magnitude, duration and frequency of occurrence of over-irradiance, down to sub-second responses from a photodiode sensor. From memory, a maximum value of 2 would be a generous limit.  2 would likely also prevent the overflow warning reported in #540

wholmgren: I will add the kwarg to ``clearness_index`` and ``clearness_index_zenith_independent`` with default values of 2. Should it also be a kwarg in ``disc``, ``dirint``, ``dirindex`` (like ``min_cos_zenith`` and ``max_zenith``)?

adriesse:
Just  a general thought here triggered by the discussion around 0.82.  It would be nice in this case and others to have a quick way to set up a model to run the way the original author proposed it, although we may all agree it is not the best way for today's toolbox.  How about an optional kwarg `settings_from` which could have the name of a paper or institute or author and provide all the relevant settings?

cwhanse:
@adriesse that's an idea worth pursuing.  Maybe not in this particular pull request, but something to keep in mind if we have a model that has many 'settings' arising from ambiguous or conflicting information in the references.

I could get really excited about a `settings_from` method at the  `ModelChain` level. Suppose we could `init` a `ModelChain` object with `settings_from=<PAN file name>` or some similar concept?  The `settings_from` would import model selections, kwarg values, etc. that have been saved, and would go a long way towards promoting reproducibility.

pvlib/irradiance.py
wholmgren: They do call ``clearness_index`` but that doesn't mean that we have to expose all of its options. It's not clear to me if ``disc``, ``dirint``, ``dirindex`` should enforce a different limit.

cwhanse: Right, they do with this pull request.  I think that `disc` and `dirint` should enforce a limit on `kt`.  `dirindex` should The challenge here is that the limit is up to us.  The only precedent we have is old SRRL Fortran code for `dirint` which is where 0.82 comes from.  I'm fairly sure that value is appropriate for hourly data but not for shorter time steps.  I agree with you that kt<2 is generally applicable and is better than 0.82.

I'd say let's go ahead and expose the kwarg.

wholmgren:
@adriesse this sounds like #433. I'm not sure what it would mean in this specific case, though. Unbounded output?

adriesse:
I guess the ideas are related.  It would be nice to be able to reproduce original results without too much trouble as well as raise awareness when uncharted territory is entered.

pvlib/irradiance.py
wholmgren: Updating the unit tests has me thinking that we should enforce a limit of 0.82 or 1 or we should return NaN. 

For ``disc``, the input conditions that yield ``kt > 1.25`` will produce ``dni = 0``. This because ``Kn`` becomes negative and we set negative values to 0.

For ``dirint``, the input conditions that yield ``kt_prime > 1`` will produce ``dni = NaN``.

Here's the line in ``dirint`` that currently limits ``kt_prime`` to 0.82:

https://github.com/pvlib/pvlib-python/blob/master/pvlib/irradiance.py#L1358

These lines assign bins for ``kt_prime`` values [0, 1]:

https://github.com/pvlib/pvlib-python/blob/master/pvlib/irradiance.py#L1381-L1387

These lines will ensure that ``kt_prime`` values > 1 will become NaN:

https://github.com/pvlib/pvlib-python/blob/master/pvlib/irradiance.py#L1425-L1428

One could argue that returning NaN is the right thing to do because the authors did not anticipate the model being applied for clearness index > 1. Or one could argue that we should limit the clearness index to 0.82 or 1 and move on.

cwhanse: I think you are suggesting that the call to `clearness_index` in, e.g., `disc`, passes a value `max_kt`=1 (or otherwise) that is appropriate for `disc`. If this is correct, then I'm OK.  If you are suggesting that the default value for `max_kt` is set to prevent wonky output from `disc` then I'm not OK.

The appropriate value for each model is less clear.  The only source we have is the SRRL code for `dirint` which is an uncommented addition to the original program.  Since I am confident that none of these models meant to predict DNI from GHI during over-irradiance conditions, an upper bound of 0.82, or 1, seems appropriate to me.

On the other hand, for `clearness_index` itself, I like your suggestion of `max_kt`=2, for physical reasons.

I don't think `disc` etc. should return `NaN` unless its inputs are meaningless, or in the case of `dirint` are outside the model's domain of bins.

wholmgren: Sorry for being unclear. We agree (I think): within ``disc`` and ``dirint``, we should call ``clearness_index`` and ``clearness_index_zenith_independent`` with ``max_clearness_index=1``. The functions retain the default max value of 2. I'll update the PR this morning.

wholmgren:
I removed the ``gti_dirint.ipynb`` tutorial from this pull request and put it in this [gist](https://gist.github.com/wholmgren/fe4d72ebb2c0e6481680bee5d9041b06). The latest version explores why the ``gti_dirint`` function seems to yield bad results for AOI > 80 (roughly speaking) in some conditions. In short, the iterative algorithm is doing what it's supposed to do: minimize the difference between the modeled and measured POA global. However, this can lead to a less accurate modeled GHI/DNI/DHI at higher AOI. Refactoring the code has made me more confident in the implementation, but that doesn't guarantee it's correct.

wholmgren:
See link below for rendered docs for new functions:

https://wholmgren-pvlib-python-new.readthedocs.io/en/gtidirint/api.html#irradiance

(scroll to bottom of the Irradiance section).

My recommendation is to merge it, though I am happy to take further suggestions if there are any.

pvlib/irradiance.py
```diff
@@ -1513,6 +1737,332 @@ def dirindex(ghi, ghi_clearsky, dni_clearsky, zenith, times, pressure=101325.,
...
+            * ``dni``: the modeled direct normal irradiance in W/m^2.
+            * ``dhi``: the modeled diffuse horizontal irradiance in
+              W/m^2.
+
+    References
+    ----------
+    .. [1] B. Marion, A model for deriving the direct normal and
+           diffuse horizontal irradiance from the global tilted
+           irradiance, Solar Energy 122, 1037-1046.
+           http://dx.doi.org/10.1016/j.solener.2015.10.024
```
mikofski: please use ``:doi:`10.1016/j.solener.2015.10.024` ``

wholmgren: thanks. updated.

Issue #411, opened 2018-01-29
Test singlediode() implementations using a "gold" function and data set


markcampanelli:
#409 Suggests Bishop [1988] as a truly explicit method to calculate i-from-v. Bishop's method is not subject to Newton convergence failure or to LambertW argument "blow up", handles ideal device parameters without extra logic, has no dependency on scipy, and naturally vectorizes via standard numpy functions' vectoization. The trade-off is that the explicit device terminal current computation occurs at the diode junction voltage (V_d := V + I R_s), not the device terminal voltage. (The device terminal voltage is "back-calculated" as V = V_d - I Rs.) Thus, additional computation work must be done to determine currents at specified terminal voltages, inc. I_sc @ V_sc=0, I_mp @ V_mp, and I_oc=0 @ V_oc.

The Bishop method is a very good candidate for a "gold" function implementation for testing the accuracy and speed of [various implementations of `i_from_v()`](#410) that pvlib may wish to offer users or refactor to. By inverting the dataset, it could also be used to test `v_from_i()` implementations.
```python
singlediode(photocurrent, saturation_current, resistance_series, resistance_shunt, nNsVth, ivcurve_pnts=None)
```

I think that after this `bishop88` function is implemented (presumably by #409), we should create and commit a corresponding "gold" dataset that we can use for better testing, refactoring, and optimization of the `singlediode()` function. I think we should try to cover a large number (all?) of the devices in the CEC database over a combination of temperature and irradiance ranges (perhaps the IEC 61853-1 matrix?). This dataset should also be valuable to the greater PV modeling community who wish to test/contribute new computational algorithms.

Outstanding questions:
(1) What should the voltage range be? Specifically, the diode voltages will give some terminal voltages less than zero, and how much beyond V_oc should we try to extend?
(2) Can/should we also include I_sc, I_x, I_mp, I_xx, and V_oc points, and what about the slope at I_sc and slope at V_oc?
(3) Does python have an interval analysis package up to the task of verifying our computations, e.g., [intPy](https://pypi.python.org/pypi/IntPy)?
(4) Is it feasible to run such a comprehensive test suite on every PR?

cc @mikofski @adriesse

wholmgren:
I am a big +1 on testing pvlib against "gold datasets" with every PR. The implementation probably depends on the size and scope of a dataset. Adding numeric checks to the test suite to guard against unknown changes was a big effort, so I'd recommend starting small.

adriesse:
[...]

PR #412, opened 2018-01-30
[DEMO: DO NOT MERGE] Brentq and Halley's method algorithm illustrations

markcampanelli:
This is a PR to illustrate usage of brentq() in place of LambertW for i_from_v() function.

v_from_i() should be very similar and I hope to add it soon.

 - [ ] Closes issue #xxxx


thunderfish24:
Turns out I was using some wonky pytest version and I needed to conda install the latest in my virtual env to get my local tests to run properly.

pvlib/pvsystem.py
```diff
@@ -1952,11 +1952,144 @@ def v_from_i(resistance_shunt, resistance_series, nNsVth, current,
...
+        all_scalar_or_rank0_inputs = False
+        rank0_output = False
+        scalar_output = False
+        # Make sure that we can index numpy arrays
+        conductance_shunt, resistance_series, nNsVth, voltage, saturation_current, photocurrent = np.broadcast_arrays(conductance_shunt, resistance_series, nNsVth, voltage, saturation_current, photocurrent)
+
+    # Compute initially with zero r_s_Ohm
+    current_ic = photocurrent - saturation_current*np.expm1(voltage/nNsVth) - conductance_shunt*voltage
+    
+    f = lambda current, conductance_shunt, resistance_series, nNsVth, voltage, saturation_current, photocurrent: current_sum_diode_node(voltage, current, conductance_shunt, resistance_series, nNsVth, saturation_current, photocurrent)
```
mikofski: IMO if you should online use lambda for an anonymous inline function, if you are going to write it on it's own line, then you should write it as a function instead so either do this:
```
def f(lambda current, conductance_shunt, resistance_series, nNsVth, voltage, saturation_current, photocurrent): 
    return current_sum_diode_node(voltage, current, conductance_shunt, resistance_series, nNsVth, saturation_current, photocurrent)
```
or move this down to where you use f on line 2075.

see [pep 8](http://pep8.org/#programming-recommendations)

mikofski: Also would be lovely to see these lines wrapped down to < 80 characters so I can read them without scrolling. thanks!

thunderfish24: This code is for illustration only, so please go easy on the over-conciseness with my lambda functions. I did add many comments :). That said, I'm sorry the formatting isn't better: I haven't figured out yet how to set up VSCode to auto format and for personal projects I just let my editor wrap my code to the window on my 14" laptop.

thunderfish24:
Looks like it is easier to hit edge cases when picking brentq bracketing intervals for the v_from_i() case, so I decided to illustrate my original Halley's method (i.e., scipy.optimize.newton()) implementation instead.

To be clear: I would want to do a lot more testing to show that either of these methods *always* converge (cf. #411). However, IMHO, the implementations are so simple (but perhaps slower due to vectorization wrappers) that I remain hopeful.

thunderfish24:
So, now that I've added range checks for [0, Voc] and [0, Isc], I no longer get the failing unit test apparently due to bad bracketing on brentq(). That makes me feel more confident in this "fool proof" approach for positive power quadrant only. The downside is that very many of the `v\_from\_i()` unit tests (and some `i\_from\_v()`) don't use inputs in this quadrant, and the functions raise() in those cases.

@mikofski I have to admit that I'm not totally sure how the bishop-based solution offers much beyond this.

wholmgren:
Maybe best to first make and merge a new PR that puts the tests in the right quadrant. I made many of the numerical tests very quickly thinking that something was better than nothing to ensure stability.

pvlib/pvsystem.py
```diff
@@ -1814,11 +1815,206 @@ def _pwr_optfcn(df, loc):
...
+        # Make sure that we can index numpy arrays
+        conductance_shunt, resistance_series, nNsVth, current, saturation_current, photocurrent = np.broadcast_arrays(
+            conductance_shunt, resistance_series, nNsVth, current, saturation_current, photocurrent)
+
+    # Check current range to ensure good root bracketing
+    if np.any(current < 0.) or np.any(i_sc(conductance_shunt, resistance_series, nNsVth, saturation_current, photocurrent) < current):
+        raise ValueError("Current input out of range [0., Isc], current = {}, Isc = {}".format(current, i_sc(conductance_shunt, resistance_series, nNsVth, saturation_current, photocurrent)))
+
+    # Compute initially for ideal device with zero resistance_series and zero conductance_shunt
+    b = nNsVth*(np.log(photocurrent + saturation_current - current) - np.log(saturation_current))
```
thunderfish24: It's not clear how many iterations this might save, rather than just using Voc.

pvlib/pvsystem.py
```diff
@@ -1952,11 +2148,174 @@ def v_from_i(resistance_shunt, resistance_series, nNsVth, current,
...
+        # Make sure that we can index numpy arrays
+        conductance_shunt, resistance_series, nNsVth, voltage, saturation_current, photocurrent = np.broadcast_arrays(
+            conductance_shunt, resistance_series, nNsVth, voltage, saturation_current, photocurrent)
+
+    # Check voltage range to ensure good root bracketing
+    if np.any(voltage < 0.) or np.any(v_oc(conductance_shunt, resistance_series, nNsVth, saturation_current, photocurrent) < voltage):
+        raise ValueError("Voltage input out of range [0., Voc], voltage = {}, Voc = {}".format(voltage, v_oc(conductance_shunt, resistance_series, nNsVth, saturation_current, photocurrent)))
+
+    # Compute initially for ideal device with zero resistance_series and zero conductance_shunt
+    b = photocurrent - saturation_current*np.expm1(voltage/nNsVth)
```
thunderfish24: It's not clear how many iterations this might save, rather than just using Isc.

mikofski:
@thunderfish24 , I'd really appreciate your review of [scipy PR 8357](../../../scipy/scipy/pull/8357) I think it does a good job at addressing your desire to vectorize these (newton, halley, secant) methods. The difference in speed is tremendous, using numpy outperforms cython by 10x, and numpy is a lot easier to use. You can post comments on [scipy PR 8357](../../../scipy/scipy/pull/8357), [scipy issue 8354](../../../scipy/scipy/issues/8354) or on the [scipy-dev maillist](https://mail.python.org/mailman/listinfo/scipy-dev). I'm concerned that if we don't push our agenda, then we'll be forced to use cython and that will increase the burden on us to maintain esoteric code. ditto for you @wholmgren , please comment on scipy PR8357. thanks!

markcampanelli:
This was for demo purposes only. Closing.

PR #426, opened 2018-02-13
Gold I-V dataset with benchmarking

markcampanelli:

 - [x] Partly addresses issue #411


thunderfish24:
Here is some code for others to reproduce computations. To run this do the following:
```bash
$ pip install pyinterval # first time only
$ python gold_dataset.py # Will rewrite test.json file with gold dataset
```
`pvlib/data/gold/worst_pvlib_computaitons.json` contains the worst computations from my laptop on python 1.0.0 with numpy 1.14.0 and pvlib 0.5.1+34.g329450a.dirty

cwhanse:
For starters, alpha_sc is undefined for the Bishop cell but I think that's not material.  You've omitted the '-1' term from IL - Io*(exp((V + I Rs)/nNsVth) - 1) - (V + I Rs)/Rsh; that is likely causing the large error where the exponential is O(1).

thunderfish24:
@mikofski I *think* you could test out some of your new functions here using this function call:
```python
gauge_against_gold_dataset({"i_from_v": pvsystem.i_from_v, "v_from_i": pvsystem.v_from_i}, json_filepath="test.json"))
```
That will use the default `method="gold"` in your new `v_from_i` and `i_from_v` functions. You can try the others out by wrapping a lambda function around these functions choosing other values for the method argument. I can eventually incorporate a `method` argument, but this API is not merged into master yet.

I'm also not quite sure what will happen for the gold I-V points outside of the positive power quadrant for functions that don't support that. Seems like it would be better to add some sort of filter argument rather than ignore all NaN's in the comparisons, because in general usage the NaN's could be from a computational error.

thunderfish24:
I'm thinking about adding a "single large cell" device and a "multi-cell mini-module" device to the list used to generate the gold dataset in this PR. I'm open to suggestions, esp. because these types of devices don't seem to be in the CEC database provided by pvlib.

@adriesse I have to admit that I'm torn about adding a negative series resistance in the mix.

I would like to leave the Isc, R@Isc, Ix, Pmp, Ixx, R@Voc, and Voc gold value generation and checks to a followup PR. The data structures and gauge function herein are supposed to be forwards compatible with those additions.

I suppose a third PR would actually implement the "complete" gauge function in the unit tests, once we've had a chance to exercise it on various implementations and gain confidence.

thunderfish24:
Looking for yet more feedback:
- Should we put the variables' units in the gold dataset's dictionary for posterity? (I think yes.)
- Should we be using a relative or absolute tolerance for the residuals? (I think absolute, for simplicity, and then the relative can be derived by the consumer with flexible normalization.)

thunderfish24:
Note that I renamed `test.json` to `gold_dataset.json` and formatted it nicely for review. There are lots of code comments now, but I'm waiting on the final docstrings until more eyes have been on the code.

Diving into the anomalous computations now...

thunderfish24:
So, some investigation has indicated that the anomaly in the "worst" Bishop cell I-V curve is due to numerical instability in the `lambertw()` function for large input arguments that do not quite overflow the output. If I force the computation of that portion of the I-V curve into the "alternate" Newton-Raphson solver on the log-transformed equation, then the anomaly disappears! What's super interesting is how localized the problem is in the I-V curve. This suggests that maybe we should take more than 21 points per gold curve.

I did a bit of inquiry a while back with Ken Roberts on a more stable alternative to LambertW. He came up with [this solution](https://arxiv.org/pdf/1504.01964.pdf) based on the log of the [Wright Omega function](https://en.wikipedia.org/wiki/Wright_Omega_function). However, I hypothesize that a simple Newton-Raphson or Halley's method might be sufficiently robust.

BTW, my "naive" `brentq()` algorithm failed pretty miserably on the gold dataset due to lack of sign change in the function at the bracket endpoints, even after I tried to restrict the data to the positive quadrant. I will probably revisit my Newton/Halley method and see how it fares before trying to fix up my bracketing approach.

adriesse:
Interesting developments!  This reinforces my feeling that the gold implementation is needed more than the gold data set.  Any candidates for a better/sleeker/sexier algorithm can then run millions of tests.

cwhanse:
@thunderfish24 I'm not saying your wrong but I still can't duplicate v_test[8] = -0.0888234583382328 in your 'v_from_i' dataset.  I'm using scipy.special.lambertw (scipy 0.19.0), the parameters listed in the structure for 'v_from_i' and lines copied from pvsystem.v_from_i.  Maybe we have python configuration issues.  For the point in question, argw = 1.22045782051e+304 and lambertw overflows.  With the log transform I compute v_test[8] = 0.184436430257. What do you get?

wholmgren:
@thunderfish24 @cwhanse consider posting your environment's ``pd.show_versions()`` and ``pvlib.__version__``

thunderfish24:
Versions are now recorded in the gold_dataset.json header info :). Looks
like at least some of my versions are different than Cliff’s. I will try to
follow up further later today.

cwhanse:
`pvlib.__version__
'0.5.1'`
`pd.show_versions()

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.1.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 23 Stepping 10, GenuineIntel
byteorder: little
LC_ALL: None
LANG: en
LOCALE: None.None

pandas: 0.20.1
pytest: 3.0.7
pip: 9.0.1
setuptools: 27.2.0
Cython: 0.25.2
numpy: 1.12.1
scipy: 0.19.0
xarray: None
IPython: 5.3.0
sphinx: 1.5.6
patsy: 0.4.1
dateutil: 2.6.0
pytz: 2017.2
blosc: None
bottleneck: 1.2.1
tables: 3.2.2
numexpr: 2.6.2
feather: None
matplotlib: 2.0.2
openpyxl: 2.4.7
xlrd: 1.0.0
xlwt: 1.2.0
xlsxwriter: 0.9.6
lxml: 3.7.3
bs4: 4.6.0
html5lib: 0.999
sqlalchemy: 1.1.9
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
pandas_gbq: None
pandas_datareader: None`

wholmgren:
My bad. I didn't notice because the diff is so big that it's not rendered by default. I'd consider a csv format for your "iv_curves" attribute. 

Do we really need to use hex values? It seems to me that we could use floats and sacrifice a bit of precision without losing anything important. There's a big gain in readability for using floats if we're using a human-readable format like json. I'd prefer a binary format to json+hex.

thunderfish24:
I failed to say that I’m on Ubuntu 16.04.

thunderfish24:
I tried uninstalling pyinterval and boosting the tolerance in the lamberw to `tol=1e-15`. Nothing changes. The lambertw function file in scipy hasn't changed in 3-5 years, but I cannot say for sure that the compiled C code underneath it hasn't changed in a recent scipy release. I might put together a toy example where the two LambertW solution methods diverge for me and post an issue in scipy.

@wholmgren I'm happy to write a "translator" for the putting gold dataset output in a more human readable number format. However, I wouldn't make the claim "gold" for users with all sorts of different machine hardware if we didn't use hex for storage, and the thought of less structured csv files makes me shutter a bit.

wholmgren:
Re: csv format... there are pros and cons. "tidy" csv data is quite structured and easily compressed.

I guess my idea of a "gold" data set is different than yours. I just don't care about the precision of the data set beyond what is practical. Practical is way less than the precision of floats and a couple digits beyond what would ever be meaningful in any use case. Maybe I want a "silver" dataset, or maybe I want a "gold" dataset in the sense of "gold bars are made of 99.5% gold". 

Whatever you all decide is most useful is fine with me. Just my 2 zinc pennies.

We were tracking a subtle scipy lambertw bug for awhile a couple of years ago. I can't remember the details very well, but it depended on exactly how you compiled scipy. The anaconda distribution included the bug, but my local scipy build did not. Maybe related.

thunderfish24:
@wholmgren I will make a file output option that is a "human readable" and "spreadsheet compatible" csv format, with floats in decimal format and significant figures that reasonable for most PV modeling purposes. However, I probably won't write an importer for this.

When it comes to taking as much guesswork as possible out of computational discrepancies, I'd prefer to keep the current lossless, json format generated using 100% reliable intervals.

thunderfish24:
@cwhanse Sorry for the delay in responding. It looks like my system is indeed computing differently. `lambertw()` does not overflow on my machine for that value. I will submit an issue to scipy.

```python
import numpy
import scipy.special
import pandas

pandas.show_versions()
```

    
    INSTALLED VERSIONS
    ------------------
    commit: None
    python: 3.6.4.final.0
    python-bits: 64
    OS: Linux
    OS-release: 4.13.0-32-generic
    machine: x86_64
    processor: x86_64
    byteorder: little
    LC_ALL: None
    LANG: en_US.UTF-8
    LOCALE: en_US.UTF-8
    
    pandas: 0.22.0
    pytest: None
    pip: 9.0.1
    setuptools: 38.4.0
    Cython: None
    numpy: 1.14.0
    scipy: 1.0.0
    pyarrow: None
    xarray: None
    IPython: 6.2.1
    sphinx: None
    patsy: None
    dateutil: 2.6.1
    pytz: 2017.3
    blosc: None
    bottleneck: None
    tables: None
    numexpr: None
    feather: None
    matplotlib: 2.1.1
    openpyxl: None
    xlrd: None
    xlwt: None
    xlsxwriter: None
    lxml: None
    bs4: None
    html5lib: 1.0.1
    sqlalchemy: None
    pymysql: None
    psycopg2: None
    jinja2: 2.10
    s3fs: None
    fastparquet: None
    pandas_gbq: None
    pandas_datareader: None

```python
scipy.special.lambertw(1.22045782051e+304)
```

    (700.1850943212547+0j)

```python
logargW = numpy.log(1.22045782051e+304)

#  w+log(w)=logargW. The initial guess is w=logargW. Where direct
#  evaluation (above) results in NaN from overflow, 3 iterations
#  of Newton's method gives approximately 8 digits of precision.
w = logargW
for _ in range(0, 3):
    w = w * (1. - numpy.log(w) + logargW) / (1. + w)
w
```

    693.6431367052406

```python
# Check convergence by increasing iterations
w = logargW
for _ in range(0, 6):
    w = w * (1. - numpy.log(w) + logargW) / (1. + w)
w
```

    693.6431367052406

mikofski:
@thunderfish24 nice work hunting down that discrepancy. FYI: [machine epsilon](https://en.wikipedia.org/wiki/Machine_epsilon) is generally _not_ guaranteed to be the same for different CPUs.

cwhanse:
@thunderfish24 @mikofski FWIW WolframAlpha gives

lambertw(1.220457820510 × 10^304) = 693.6431367052407

and

693.6431367052406 * exp(693.6431367052406) = 1.220457820510 × 10^304

Looks like scipy.special.lambertw only tested small numbers.

In PVLib matlab our pvl_lambertw uses Halley's method and gives these exact results.  We implemented our own (adapted from someone else's code) because Matlab put lambertw in the Symbolic toolbox.  I suggest we do that in pvlib-python also.

wholmgren:
> I will make a file output option that is a "human readable" and "spreadsheet compatible" csv format, with floats in decimal format and significant figures that reasonable for most PV modeling purposes. However, I probably won't write an importer for this.

It's fine with me if you leave this for future work to be done by someone that is more interested in that format.

thunderfish24:
I started work on that this weekend. I should have something ready very soon. It’s a totally valid request to have something “reader friendly” for a first implementation.

thunderfish24:
This is a pretty good working version of my first PR for gold dataset for the single diode model. (Sorry docstrings still incomplete ATM, and looks like there might be Python 2 issues based on CI.)

I added some benchmark timing info and a csv output. Most of the code is now in pvlib-python/pvlib/data/gold/sdm.py, but I plan to move that code into pvsystem.py when I do the documentation and spit and polish.

Run `python sdm.py` to do all the things. 

@wholmgren Do you know the best git command for squashing my commits?

mikofski:
[`git rebase -i <commit-to-rebase-on>`](https://git-scm.com/docs/git-rebase) for interactive [rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) - choose `squash` for each commit

thunderfish24:
I hit a snag adding a single, large area cell to the set of devices for which I compute gold I-V data.

@mikofski Can you provide a ballpark for series resistance in Ohm/cm^2 for Sunpower cells that are ~153 cm^2 (like [this one](https://us.sunpower.com/sites/sunpower/files/uploads/resources/sp-gen2-solar-cell-ds-en-a4-160-506760e_1.pdf))? The Bishop 1988 value of  1.33 Ohm/cm^2 is perhaps unrealistic and the following short-circuit current computes to -inf (1.33 Ohm/cm^2 x 153 cm^2 = 203.49 Ohm):
```python
pvsystem.i_from_v(1147500.0, 203.49, 0.03103854670836424, 0.0, 1.4808680805002758e-07, 0.45447800000000005)
``

**Comment by mikofski** [comment]
For a single cell, try the values in [PVMismatch pvmismatch.pvcell](https://github.com/SunPower/PVMismatch/blob/master/pvmismatch/pvmismatch_lib/pvcell.py/#L16)

**Comment by thunderfish24** [comment]
@wholmgren So I am looking for a better place to put `sdm.py`, currently it is at `data/gold/sdm.py`, but I think python files should reside there. I could move all the code into pvsystem.py, but that file is already bloated. Other than this, I'm down to needing to write the docstrings and a README.md for the HOW TO of regenerating and bench-marking against the gold data sets. (Also, perhaps you can suggest how I broke the Python2 tests?)

In #436, @cwhanse mentioned breaking out all-the-things by modeling step, so I could rename the file and put it in a subdirectory path like `dc_module_iv/sdm/gold.py`, and also put the README.md in there. Likewise, the data file could be renamed and reside in `data/dc_module_iv/sdm/gold.json`. I'm not sure if this directory structure has good parity with Cliff's idea of wrappers that access the various implementations for a given modeling step. (Here, the step is [dc_moduel_iv](https://pvpmc.sandia.gov/modeling-steps/2-dc-module-iv/).) In the future, we could further refine the path (for example, `sdm/desoto` and `sdm/pvsyst`), because we should allow extension to different "flavors" of single diode model (SDM), which probably will only change the computation of the five parameters of the "fundamental" SDM as operating conditions change. Likewise, there could be a `dc_module_iv/ddm/...` for double diode models. (Nit: I think I would actually prefer the prefix `dc_device_iv` instead of `dc_module_iv`, because we sometimes model single cells at this step.)

cc @mikofski because you mentioned reorganizing the code in #235. Also, @adriesse.

File: pvlib/data/gold/sdm.py [hunk 1]
```diff
@@ -0,0 +1,598 @@
...
+
+    # When requested, write dataset in csv file, skips empty path strings too
+    if csv_filepath:
+        with open(csv_filepath, 'w') as fp:
+            fp.write(gold_dataset_csv)
+
+    return gold_dataset_csv
+
+
+# TODO This module will be pulled into pvsystem.py and this removed
```
wholmgren: I'm not sure I agree with this. I like the way you're doing it now, especially with a readme that lives in this directory.

pvlib/pvsystem.py
```diff
@@ -1835,6 +1836,76 @@ def _pwr_optfcn(df, loc):
     return I * df[loc]
 
 
+def sdm_sum_current(resistance_shunt, resistance_series, nNsVth, current,
```
wholmgren: In the spirit of functions that start with verbs and of avoiding most abbreviations, I think the name should be ``sum_singlediode_current``

wholmgren: This needs an api.rst entry.

pvlib/test/test_pvsystem.py
```diff
@@ -383,6 +384,27 @@ def test_PVSystem_calcparams_desoto(cec_module_params):
     assert_allclose(nNsVth, 0.473)
 
 
+def test_sdm_sum_current():
```
wholmgren: I vote for putting this in a new ``test_gold_sdm.py`` module. And the import above.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,598 @@
...
+        default_cell_area_cm2 / N_s  # Ohm
+    device.Adjust = None
+    device.gamma_r = None
+    device.Version = None
+    device.PTC = None
+
+    return device
+
+
+def sdm_make_gold_dataset(json_filepath=None):
```
wholmgren: recommend removing the option to write the file from within this function. I think it's better for this function to do it's one thing (return the data) and for another function to write (and maybe verify) the file.

wholmgren:
Oh, and regarding the larger reorganization proposal, I would say that we can solve that separately from this PR and move this file if needed.

pvlib/sdm_gold.py
```diff
@@ -0,0 +1,670 @@
...
+                         interval(g_sh)*(interval(vi_pair[0]) + interval(vi_pair[1])*interval(r_s)) -
+                         interval(vi_pair[1]) for vi_pair in zip(v_gold, i_gold)]
+
+                    # Make sure the computed interval is within the specified tolerance
+                    for v, i, current_sum_residual_interval in zip(v_gold, i_gold, current_sum_residual_interval_list):
+                        if current_sum_residual_interval not in TOL_CURRENT_SUM_RESIDUAL_GOLD_INTERVAL:
+                            # print([float.fromhex(v)
+                            #        for v in device_dict["iv_curves"][-1]["v_gold"]])
+                            # print([float.fromhex(i)
+                            #        for i in device_dict["iv_curves"][-1]["i_gold"]])
```
thunderfish24: Note to self: Remove these.

thunderfish24:
@wholmgren I noticed that .csv files are specified in the .gitignore. Should I override that here and commit the .csv file I generate in this PR? Furthermore, should .json files be added to .gitignore and treated in the same vein?

wholmgren:
> @wholmgren I noticed that .csv files are specified in the .gitignore. Should I override that here and commit the .csv file I generate in this PR? Furthermore, should .json files be added to .gitignore and treated in the same vein?

You may override it. It makes sense to treat json the same way.

thunderfish24:
@wholmgren Do you have any thoughts on how to `pip install interval` in CI (or equivalent) for the unit tests?

wholmgren:
@thunderfish24 for Travis, you'll need to add it to the yml files in the [ci folder](https://github.com/pvlib/pvlib-python/tree/master/ci). The appveyor config does not currently support pip installs. You'll need to define a ``requires_interval`` marker in [``conftest.py``](https://github.com/pvlib/pvlib-python/blob/master/pvlib/test/conftest.py), import it into your test module, and use it to decorate whichever tests need ``interval``.

thunderfish24:
@wholmgren Thanks for the explanation on the test setups :). It has mostly worked, however, I see numerous errors related to importing `tools` for the `ci/requirements-py27-min.yml` build, which seem to be the same issue that my new code is hitting. For example,
```console
ImportError while importing test module '/home/travis/build/pvlib/pvlib-python/pvlib/test/test_atmosphere.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
pvlib/test/test_atmosphere.py:9: in <module>
    from pvlib import atmosphere
pvlib/__init__.py:2: in <module>
    from pvlib import tools
E   ImportError: cannot import name tools
```
Your thoughts?

thunderfish24:
@wholmgren K. I removed all the scipy dependencies. Hopefully we get a passing build now. This PR also needs some small updates now that #471 is in master, and I STILL need to write some documentation, but it is close to ready!

 FYI, I spotted these items commented out in the `dependencies:` section of `ci/requirements-py36.yml`:
```python
    #- ephem
    #- numba
    #- siphon
```

thunderfish24:
@wholmgren Looks like a transitive scipy dependency is now an issue (in `lambertw()`). It seems really strange that a user could not call, e.g., `i_from_v()` with the `base` installation of pvlib.

wholmgren:
@thunderfish24 looks like your test function will also need the ``requires_scipy`` decorator. I am pretty sure they will stack. 

I forgot about the commented out dependencies. I think that was from when python 3.6 was very new and some packages were not yet available. Let's leave it alone for now.

I suggest a separate issue for discussing if scipy should be a required dependency.

wholmgren:
The other test environments are failing because of the API change in recently merged #471.

thunderfish24:
@wholmgren @cwhanse @adriesse @mikofski This is ready for review/testing, if anyone would like to volunteer. There are no functional changes or refactors to the user-facing code, but it does set up an initial pattern for gold dataset testing/benchmarking.

I locally tried a rebase over master, but there were some conflicts. I'm willing to try to work through them if that would actually be worthwhile.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+        pvsystem.retrieve_sam('cecmod').First_Solar_FS_495
+    ]
+
+    # Initialize the output, recording metadata
+    gold_dataset = {
+        "timestamp_utc_iso": datetime.datetime.utcnow().isoformat(),
+        "dataset_version": '1.0.0',  # dataset follows SemVer
+        "python_version": sys.version,
+        "numpy_version": numpy.__version__,
+        "pvlib_version": pvlib.version.__version__,
```
wholmgren: ``pvlib.__version__`` should work here

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+    ----------
+    gold_dataset : dict
+
+    test_func : dict
+
+    Returns
+    -------
+    results : list
+        A per-device list of results with each function tested/benchmarked. The
+worst computed I-V curve is returned along with performance timing information
```
wholmgren: These lines should be indented as well, at least for consistency with the rest of the library. I am surprised ``flake8`` does not complain on my checkout of the pr.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+    #  results returned for each device
+    return results
+
+
+# Command line option
+if __name__ == '__main__':
+
+    gold_dataset = make_gold_dataset()
+
+    # File locations are robust to running script from different directories
```
wholmgren: This makes sense for a local, development installation, but I'm not sure what the right behavior should be for a standard installation. Perhaps the default should be the current directory? ``argparse`` is another option.

pvlib/test/data/gold/test_sdm.py
```diff
@@ -0,0 +1,53 @@
...
+            os.path.dirname(sdm.__file__), "sdm.json")))
+
+    for device in gold_dataset["devices"]:
+        for iv_curve in device["iv_curves"]:
+            current_sum_res = sdm.sum_current(
+                iv_curve["r_sh"], iv_curve["r_s"], iv_curve["nNsVth"],
+                iv_curve["i_gold"], iv_curve["v_gold"], iv_curve["i_0"],
+                iv_curve["i_l"])
+
+            # atol typically smaller that the interval tolerance used for gold
```
wholmgren: Maybe I am confused but I thought ``atol`` would need to be larger than the interval tolerance. Won't the test fail if ``atol`` is smaller than the interval tolerance? Also not sure sure quite what you mean by "typically" -- is there another failure mode to be concerned about?

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+    gold_dataset = convert_gold_dataset(gold_dataset)
+
+    # Follow SemVer and maintain backwards dataset compatibility whenever
+    #  possible. Future code may require logic based on version loaded.
+    print("Gold dataset version {}.".format(gold_dataset["dataset_version"]))
+
+    # Initialize output
+    results = []
+
+    for device in gold_dataset["devices"]:
```
wholmgren: The nested ``for`` and ``if`` statements are hard for me to follow. Perhaps some utility functions could make that easier. I'd consider a handful of simple, private functions that have descriptive names and no real documentation.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+        # Reset the worst max abs residuals over all I-V curves per device
```
wholmgren: refactoring into functions also helps prevent issues with variable resetting on each loop iteration. I've made some embarrassingly bad errors due to this.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+                    iv_curve["r_sh"], iv_curve["r_s"], iv_curve["nNsVth"],
+                    i_test, iv_curve["v_gold"], iv_curve["i_0"],
+                    iv_curve["i_l"])
+                i_test_current_sum_max_abs_res = numpy.amax(
+                    numpy.abs(i_test_current_sum_res))
+                if worst_i_test_current_sum_max_abs_res < \
+                   i_test_current_sum_max_abs_res:
+                    worst_i_test_current_sum_max_abs_res = \
+                        i_test_current_sum_max_abs_res
+                    device_dict["i_from_v"]["current_sum"] = \
```
wholmgren: These ``device_dict`` assignments look similar enough to refactor into a function and improve readability. 

```python
def _create_device_result(iv_curve, test_name, test_result, residual, max_abs_residual):
    out = {
        "G": iv_curve["G"],
        "T": iv_curve["T"],
        "i_l": iv_curve["i_l"],
        "i_0": iv_curve["i_0"],
        "r_s": iv_curve["r_s"],
        "r_sh": iv_curve["r_sh"],
        "nNsVth": iv_curve["nNsVth"],
        "v_gold": iv_curve["v_gold"].tolist(),
        "i_gold": iv_curve["i_gold"].tolist(),
        test_name: test_result.tolist(),
        "residual": residual.tolist(),
        "max_abs_residual": max_abs_residual.tolist(),
    }
    return out

device_dict["i_from_v"]["current_sum"] = _create_device_result(iv_curve, "i_test", i_test, i_test_current_sum_res, i_test_current_sum_max_abs_res)
```

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+    data and summary statistics are also returned for the ensemble of
+    computations per function and per device.
+
+    This also helps verify that a given function meets the pvlib interface.
+
+    Parameters
+    ----------
+    gold_dataset : dict
+
+    test_func : dict
```
wholmgren: need some documentation for the expected dict structure.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+    Relative file paths are used, which are relative to the current directory.
+    Defaults are customizable. Set a file path to None to skip saving that
+    file.
+
+    Parameters
+    ----------
+    gold_dataset : dict
+
+    json_rel_path : str
+        Pass ``None`` to skip writing this file.
```
wholmgren: I'd probably go with a default ``None``. But I'll let it go if no one else agrees with me.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+            gold_dataset_csv = gold_dataset_csv + ", ".join(
+                map(str, iv_curve["v_gold"].tolist())) + "\n"
+            gold_dataset_csv = gold_dataset_csv + ", ".join(
+                map(str, iv_curve["i_gold"].tolist())) + "\n"
+
+    return gold_dataset_csv
+
+
+def save_gold_dataset(gold_dataset,
+                      json_rel_path="sdm.json",
```
wholmgren: ``os.path.abspath`` will work with either absolute or relative paths, so I think ``json_path`` is preferable. And similar for csv.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+        #  degraded device
+        r_s_scale_grid, g_sh_scale_grid = numpy.meshgrid([0., 1., 5.],
+                                                         [0., 1., 5.])
+
+        # For all "flavors" of this device, create I-V curves over (G,T) matrix
+        for r_s_scale, g_sh_scale in zip(r_s_scale_grid.flat,
+                                         g_sh_scale_grid.flat):
+            for G in G_list:
+                for T in T_list:
+                    # Compute the model parameters at each
```
wholmgren: This line through 373 should be its own function e.g. ``create_device_dict(device, G, T, r_s_scale, g_sh_scale)``. Either a public or private function is fine with me here.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,832 @@
...
+            "Technology": device.Technology,
+            "N_s": device.N_s,
+            "iv_curves": []
+        }
+
+        # Choose bandgap parameters based on device technology
+        # Add new technologies if/when needed, should match a CEC module
+        #  database type
+        # These are NOT saved, but are versioned with the dataset version
+        if device.Technology == 'Mono-c-Si' or \
```
wholmgren: Suggest moving this ``if`` block to the ``create_device_dict`` or similar function requested in the loop below. This block could also be a small function:

```python
def _determine_bandgap_parameters(device):
    if...elif...else...
    return bandgap_params  # a dict

# within "create_device_dict"
pvsystem.calcparams_desoto(..., **bandgap_params)
```

pvlib/constants.py
```diff
@@ -0,0 +1,20 @@
...
+elementary_charge_C = 1.6021766208e-19
+
+# Boltzmann constant, from scipy.constants.value('Boltzmann constant')
+boltzmann_J_per_K = 1.38064852e-23
+
+# Define standard test condition (STC) temperature in degrees Celsius
+T_stc_degC = 25.
+
+# Define standard test condition (STC) temperature in Kelvin
+T_stc_K = 273.15 + T_stc_degC
```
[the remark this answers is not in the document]
cwhanse: Recommend we use the common nomenclature here: q for elementary charge `q_C`, `k_J_per_K`, and the common variable for cell temperature `Tcell`: `Tcell_STC_degC`, `Tcell_STC_degK`.  I'm not a fan of the unit suffixes but I'm open to it, if the majority wants.

pvlib/data/gold/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+The I-V curve gold dataset for the SDM is generated by runnning the following
+python script in this directory
+
+```console
+$ python sdm.py
+```
+
+which will overwrite the existing `sdm.json` and `sdm.csv` files. `sdm.json` is
+the lossles, machine-interoperable format, and `sdm.csv` is formatted to be
```
cwhanse: lossless

pvlib/constants.py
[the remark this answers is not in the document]
thunderfish24: I can see that the shorter names would be better in the formulas. While I prefer the units included, I won't get hung up on units in the name as long as the units are in a comment next to the constant's definition, and if we have a centralized definition of the constants, then there is probably less opportunity for confusion.

pvlib/__init__.py
```diff
@@ -11,3 +11,4 @@
 from pvlib import pvsystem
 from pvlib import spa
 from pvlib import modelchain
+from pvlib import constants
```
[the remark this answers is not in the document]
thunderfish24: For consistency, I think users should be encouraged and able to query and use the same constants that we use internally.

pvlib/constants.py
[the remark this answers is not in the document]
adriesse: Am not a fan of the units either.  Why not actually import the values from scipy rather than copying?  (Not that they're likely to change, but still...)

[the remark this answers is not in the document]
thunderfish24: I reverted the scipy implementation because the "base" installation of pvlib doesn't demand scipy and changing that is a separate issue out of scope.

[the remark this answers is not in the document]
cwhanse: Let's open a new Issue to address having a common set of constants across all of pvlib.

thunderfish24:
I'm generally ok with the suggestions so far. Not sure of the timing to get back to this, and I'm wondering if the constants issue should get sorted first.

wholmgren:
I think we can move forward on this PR if the constants module is private or if they are defined in the ``sdm.py`` module where needed.

pvlib/__init__.py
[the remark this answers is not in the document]
thunderfish24: I removed `constants.py` and moved the constants themselves into the `sdm.py` module. They can be updated later when there is an official `constants.py` module.

pvlib/constants.py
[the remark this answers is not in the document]
thunderfish24: https://github.com/pvlib/pvlib-python/issues/483

I made these constants local to `sdm.py` for now. I didn't change any names, but they should be updated when the above issue is addressed.

pvlib/data/gold/sdm.py
thunderfish24: I now use `argparse` to make this more flexible (user can provide the directory and/or the extension-less filename), but I still have the default behavior create/overwrite files in the `pvlib/data/gold` directory.

pvlib/data/gold/sdm.py
thunderfish24: Updated.

pvlib/test/data/gold/test_sdm.py
thunderfish24: Interval computations are conservative, so the intervals should be no larger than any specific residual. In this case we have the same computed quantity (V + I*Rs) showing up twice in the current sum residual computation, and this in known to cause the intervals to be "pessimistic" because the two quantities are treated independently in the overall computation, which they are not. I updated the comment here to hopefully be less ambiguous.

pvlib/data/gold/sdm.py
thunderfish24: Will do, and note that the `gold_dataset` is actually a list of devices, each of which is a dict, which I will try to explain breifly.

thunderfish24: Correction: The gold dataset is indeed a dict, but a key part of it is a list of devices, each of which is a dict.

thunderfish24:
@wholmgren I have limited time now, and #478 is a priority for me and requires some larger changes to integrate into `ModelChain` than I was originally expecting because of `F` and `H` vs. `effective_irradiance` and `cell_temp`. (Time is of the essence here also because of other single-diode models being added presently.) I would say the code here is very well documented, and I humbly request that we wait on a larger refactor of this code until subsequent PRs that add additional gold data (i.e., Isc, Ix, Pmp, Ixx, Voc). Concerning that, I had originally intended to use issue #411 for a sequence of related "gold dataset" PRs. Are you OK with that? If so, then I will update that issue with the refactor items to implement in the next PR.

wholmgren:
@thunderfish24 I understand your perspective. The primary reason I requested some refactoring was to make the code easier to understand and review for correctness. I think the code is in good shape otherwise.

I am unlikely to use this feature, so I ask @cwhanse @adriesse @mikofski and any others for their opinion of how to proceed.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+boltzmann_J_per_K = 1.38064852e-23
+
+# Define standard test condition (STC) temperature in degrees Celsius
+T_stc_degC = 25.
+
+# Define standard test condition (STC) temperature in Kelvin
+T_stc_K = 273.15 + T_stc_degC
+
+
+def sum_current(resistance_shunt, resistance_series, nNsVth, current, voltage,
```
cwhanse: Minor : would be nice to have the positional arguments in the same order as in `pvsystem.singlediode`

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+    current_sum : numeric
+
+    References
+    ----------
+    [1] W. De Soto et al., "Improvement and validation of a model for
+    photovoltaic array performance", Solar Energy, vol 80, pp. 78-88, 2006.
+    """
+
+    # Compute voltage at diode node first
+    voltage_diode = voltage + resistance_series * current
```
cwhanse: If `voltage_diode` is used elsewhere, a function `voltage_diode(voltage, current, resistance_series)` would be useful.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+    return photocurrent - saturation_current * \
+        numpy.expm1(voltage_diode/nNsVth) - voltage_diode/resistance_shunt - \
+        current
+
+
+def bootstrap_si_device_bishop(cell_area_cm2=4., N_s=1, ideality_factor=1.5):
```
cwhanse: I'd prefer `create_si_device_bishop`

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+    device.Adjust = None
+    device.gamma_r = None
+    device.Version = None
+    device.PTC = None
+
+    return device
+
+
+def bootstrap_si_cell_large():
+    """
```
cwhanse: Prefer `create_si_cell_large`

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+    Create a single-large-area-cell c-Si device based on a SunPower module.
+    """
```
cwhanse: Add note that we are creating a De Soto model for the cell

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+
+    # Scale relevant module parameters to single cell values
+    device.name = 'Si_cell_large'
+    device.Date = '6/16/2018'
+    device.N_s = 1  # Single large cell
+    device.V_oc_ref = device.V_oc_ref / Ns_module
+    device.V_mp_ref = device.V_mp_ref / Ns_module
+    device.a_ref = device.a_ref / Ns_module
+    device.R_s = device.R_s / Ns_module
+    device.R_sh_ref = Ns_module * device.R_sh_ref
```
cwhanse: Should this be `Ns_module / device.R_sh_ref`  ?

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+
+    return device
+
+
+def make_gold_dataset():
+    """
+    Make a gold dataset of I-V curves that accurately solve single-diode model.
+
```
cwhanse: ` that accurately solve the single diode equation.`

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+    backwards compatibility whenever possible. Be sure to update all dependent
+    functions.
+
+    Parameters
+    ----------
+
+    Returns
+    -------
+    gold_dataset : dict
+        The gold data are returned in a dict with metadata (e.g., timestamp,
```
cwhanse: Is it practical to outline the `gold_dataset` dict here (i.e., list and describe each key), the device dict, and a list member dict for a device?

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+
+    The set of I-V curves are generated from a range of PV devices: small c-Si
+    cell, large c-Si cell, c-Si "mini-module", c-Si module, and CdTe module.
+    Additionally, each device is ideal, nominal, or degraded in terms of its
+    series and parallel resistances. I-V curves for each device are generated
+    over a matrix of irradiance and temperature. The I-V curve data are
+    returned in a python dict that is in a lossless, machine-interoperable
+    format, which is the expected input to most of the other functions in this
+    python module. (Other functions can convert this dict to a human-readable
+    format.) An explict method of Bishop [1] is used to compute (V, I) pairs
```
cwhanse: Would help to name the 'other functions'

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+                                ]))
+
+        # Append this I-V curve to the list for this device
+        gold_dataset["devices"].append(device_dict)
+
+    # Return the list of devices' gold data with floats in hex format
+    return gold_dataset
+
+
+def convert_gold_dataset(gold_dataset):
```
cwhanse: maybe add `_to_pvlib` to function name

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+            iv_curve["i_l"] = float.fromhex(iv_curve["i_l"])
+            iv_curve["v_gold"] = numpy.array(
+                [float.fromhex(v) for v in iv_curve["v_gold"]])
+            iv_curve["i_gold"] = numpy.array(
+                [float.fromhex(i) for i in iv_curve["i_gold"]])
+
+    return gold_dataset_converted
+
+
+def pretty_print_gold_dataset(gold_dataset):
```
cwhanse: `gold_dataset_to_csv`

pvlib/data/gold/sdm.py
cwhanse: On second thought, `gold_dataset_to_pvlib`

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+    gold_dataset_converted["units"]["r_sh"] = "Ohm"
+    del gold_dataset_converted["units"]["g_sh"]
+
+    # Convert the values
+    for device in gold_dataset_converted["devices"]:
+        for iv_curve in device["iv_curves"]:
+            # Load G, T, and 5 single-diode model parameters for each I-V curve
+            # Floats converted from lossless, machine-interoperable hex format
+
+            iv_curve["G"] = float.fromhex(iv_curve["G"])
```
cwhanse: should we use the pvlib keys?  `effective_irradiance` instead of `G`, `temp_cell` instead of `T`, etc.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+    This also helps verify that a given function meets the pvlib interface, but
+    note that no assertions are made about the quality of the gauged functions.
+
+    Parameters
+    ----------
+    gold_dataset : dict
+        See ``make_gold_dataset()``.
+
+    test_func : dict
+        A dictionary of functions to guage. Supported keys to functions are
```
cwhanse: 'gauge'

pvlib/data/gold/sdm.py
cwhanse: Belay that.  The arguments are in the same order as for `v_from_i` and `i_from_v`

adriesse:
Sorry, I'm not too clear on the question here.  There's been too much going on!

> @thunderfish24 I understand your perspective. The primary reason I requested some refactoring was to make the code easier to understand and review for correctness. I think the code is in good shape otherwise.

>I am unlikely to use this feature, so I ask @cwhanse @adriesse @mikofski and any others for their opinion of how to proceed.

cwhanse:
@adriesse Will is inviting you to review the pull request. OK if you don't have time, since both Will and I have done so.  The PR consists of two basic components: a function to precisely calculation sets of IV curves (a number of example devices are included), and a function to compare between two IV curve calculators (think of a function that runs and compares two functions).  I think Will's major interest is, do you think this capability is appropriate for and useful to pvlib?

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+
+                    # Solve the current sum at diode node residual as a
+                    #  numerically reliable interval
+                    current_sum_residual_interval_list = \
+                        [interval(i_l) - interval(i_0) * imath.expm1(
+                            (interval(v) + interval(i)*interval(r_s)) /
+                            interval(nNsVth)) - interval(g_sh) *
+                            (interval(v) + interval(i)*interval(r_s)) -
+                            interval(i) for v, i in zip(v_gold, i_gold)]
+
```
adriesse: What's the advantage of using intervals here, where all the intervals are single numbers?

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+                    # Make sure the computed interval is within the specified
+                    #  tolerance, report bad value in exception if not
+                    for v, i, current_sum_residual_interval in zip(
+                            v_gold, i_gold,
+                            current_sum_residual_interval_list):
+                        if current_sum_residual_interval not in tol_interval:
```
adriesse: What is the advantage of this over np.isclose()?

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+                             ["performance"]["time_s"]["data"])
+            device_dict["i_from_v"]["performance"]["time_s"]["max"] = \
+                max(device_dict["i_from_v"]["performance"]["time_s"]["data"])
+            device_dict["i_from_v"]["performance"]["time_s"]["avg"] = \
+                numpy.average(device_dict["i_from_v"]
+                              ["performance"]["time_s"]["data"])
+            device_dict["i_from_v"]["performance"]["time_s"]["std"] = \
+                numpy.std(device_dict["i_from_v"]
+                          ["performance"]["time_s"]["data"], ddof=1)
+
```
adriesse: This block is quite hard for me to parse.

pvlib/data/gold/sdm.py
```diff
@@ -0,0 +1,847 @@
...
+    # The "base" devices, having ideal, normal, and degraded versions
+    devices = [
+        bootstrap_si_device_bishop(),
+        bootstrap_si_device_bishop(N_s=8),
+        bootstrap_si_cell_large(),
+        pvlib.pvsystem.retrieve_sam('cecmod').SunPower_SPR_E20_327,
+        pvlib.pvsystem.retrieve_sam('cecmod').First_Solar_FS_495
+    ]
+
```
adriesse: What do you think about copying the parameters for all 5 devices into golden_devices.csv or something similar?  Then you just need a single read.

ci/requirements-py34.yml
```diff
@@ -17,5 +17,7 @@ dependencies:
     - nose
     - pip:
         - coveralls
+        - pvfactors==1.0.1
```
markcampanelli: @anomam Do you know if I'm fixing or breaking something here? (Is pvfactors 1.0.1 really not supported in Python 3.4?)

anomam: That's a good point @markcampanelli , starting with version 1.0.0 pvfactors should work with earlier versions of pandas, and therefore be compatible with Python 3.4.
Could you post an issue in pvfactors by any chance? A PR would be very much welcome as well :)

wholmgren: Python 3.4 hit "end of life" a few months ago. I don't plan to support it going forward.

markcampanelli:
It looks like https://github.com/scipy/scipy/pull/8028 may have fixed the significant issue with a `scipy.special.wlambert` calculation that I saw earlier on my machine (see way above). With `scipy 1.2.1,`, `scipy.special.lambertw(1.22045782051e+304)` now properly computes `693.6431367052406+0j`. I am now investigating the new computations and making sure that remaining ones with large residuals can be reasonably explained as being from numerics and not bugs. After this I will try to improve improve code naming and organization based on earlier suggestions before requesting another review.

ci/requirements-py34.yml
markcampanelli: https://github.com/SunPower/pvfactors/issues/49

markcampanelli:
Closing this in favor of https://github.com/pvlib/pvlib-python/pull/1573.

PR #427, opened 2018-02-13
add get_[relative_airmass,sky_diffuse,ground_diffuse,extra_radiation,total_irradiance] and pvlibDeprecationWarning

### LINKED ISSUE #422: irradiance.globalinplane appears redundant with total_irrad
Suggest removing globalinplane.

wholmgren:
- [x] Closes #422 

This PR implements the refactoring discussed in #422.

I added a deprecation warning to globalinplane. 

We could go ahead and change the name of total_irrad in this PR, but that would require changes to a few other parts of the library that I wasn't yet ready to commit to. If we do change it, we'd keep the old name until 0.6 and add a deprecation warning in the mean time (like I did for globalinplane).

Everything here is open for discussion.


docs/sphinx/source/api.rst
```diff
@@ -138,7 +138,7 @@ Decomposing and combining irradiance
    irradiance.aoi_projection
    irradiance.poa_horizontal_ratio
    irradiance.beam_component
-   irradiance.globalinplane
+   irradiance.poa_components
    irradiance.grounddiffuse
```
thunderfish24: Is this a good opportunity to make naming more consistent, such as `irradiance.extra_radiation` and `irradiance.ground_diffuse`?

docs/sphinx/source/api.rst
```diff
@@ -148,6 +148,7 @@ Transposition models
    :toctree: generated/
 
    irradiance.total_irrad
+   irradiance.get_sky_diffuse
    irradiance.isotropic
    irradiance.perez
    irradiance.haydavies
```
thunderfish24: Perhaps naming consistency is better here. Only one function has `get_` prefix.

pvlib/test/test_irradiance.py
```diff
@@ -260,8 +260,7 @@ def test_globalinplane():
         40, 180, irrad_data['dhi'], irrad_data['dni'], dni_et,
         ephem_data['apparent_zenith'], ephem_data['azimuth'], airmass)
     irradiance.globalinplane(
-        aoi=aoi, dni=irrad_data['dni'], poa_sky_diffuse=diff_perez,
-        poa_ground_diffuse=gr_sand)
+        aoi, irrad_data['dni'], diff_perez, gr_sand)
```
thunderfish24: Does this test essentially check that `poa_components()` executes without exception?

docs/sphinx/source/api.rst
wholmgren: Sure. Probably a good idea to make all API changes to a module at once.

docs/sphinx/source/api.rst
wholmgren: In this module, yes, but we use ``get_`` for a few other "wrapper" functions/methods in ``solarposition.py``, ``location.py``, and ``pvsystem.py``. I am torn on this.

cwhanse: I'm not in favor of 'get_isotropic', an exception to my preference for verb prefixes.  isotropic, perez, haydavies are named models.  If we want to extend the function name, I'd use a category prefix, i.e., sky_diffuse_isotropic, sky_diffuse_perez, etc., and have get_sky_diffuse wrap all of them with a kwarg.

pvlib/test/test_irradiance.py
wholmgren: Correct. I only changed this test because I changed the ``globalinplane`` function signature to ``*args`` so that it could be a direct pass through to ``poa_components``. No one ever bothered to add numeric assertions to this test.

docs/sphinx/source/api.rst
cwhanse: How about the pattern of get_extraterrestrial and get_ground_diffuse as the wrapper, and ground_diffuse_isotropic as our one and only wrapped function?  That's what I'm suggesting for sky_diffuse models. I can see the same pattern working for get_clearsky (kwarg to specify the clearsky model), get_airmass, get_iam, etc.

Many function names were simply carried over from Matlab where we wrote a library, not an object-organized package.

pvlib/test/test_irradiance.py
thunderfish24: Is there an issue for test improvement that we could put in a comment here?

wholmgren:
It might also make sense to change ``relativeairmass`` to ``get_relative_airmass`` at the same time. And ``absoluteairmass`` to ``absolute_airmass``

cwhanse:
> I implemented get_total_poa_irradiance, but get_poa_irradiance_total is arguably more similar to get_poa_sky_diffuse and get_poa_ground_diffuse.
get_poa_sky_diffuse or get_sky_diffuse?
get_poa_ground_diffuse or get_ground_diffuse?

I think the 'poa' is not necessary in these function names, since one can pass tilt=0.

> It might also make sense to change relativeairmass to get_relative_airmass at the same time. And absoluteairmass to absolute_airmass

Both make sense to me.

wholmgren:
Good point. I removed poa from the function names. I also changed the airmass function names as proposed.

I'm building the new API docs [here](http://wholmgren-pvlib-python-new.readthedocs.io/en/totalirrad/api.html)

To do:
- [ ] update the tests to use the new function names.
- [ ] update some doc strings that reference the wrong function.

wholmgren:
I've decided that we should wait on this for an 0.6 release. It seems a little rude to make pvlib spit out a bunch of new warnings when people make a minor update to the package. The old functions would then be removed in the 0.7 release. We can adjust other issues to 0.7 or 0.8 get the changes out faster.

cwhanse:
Concur

wholmgren:
I think this is finally ready, or at least very close to it.

I added a couple of new test functions that assert that the old functions raise ``PVLibDeprecationWarning``s. The test functions are decorated with a function that will force them to fail on pvlib versions >= 0.7. To ensure that the functions are not actually released in the 0.7 release, we should do the following:

1. Start an issue tracking deprecations for that release, and tag it with the milestone.
2. Make a release candidate tag and rerun the tests as part of every major release.

Question: ``PVLibDeprecationWarning`` or ``pvlibDeprecationWarning``?

cwhanse:
`pvlibDeprecationWarning` I know this probably breaks the capitalization convention, but `PVLib` is the name we use for the Matlab version.

wholmgren:
Updated to ``pvlibDeprecationWarning``. Turns out that, in the < version case, the ``fail_on_pvlib_version`` decorator was not actually returning the test to run in the code I pushed a few days ago. It's fixed now, and I added a ``test_conftest.py`` module to prove it (I think).

wholmgren:
@mikofski do you have suggestions/thoughts on deriving a pvlib warning from ``DeprecationWarning`` vs. ``FutureWarning``? See [comment above](https://github.com/pvlib/pvlib-python/pull/427#issuecomment-402922207). It's not clear to me what the best approach is for us. Maybe ``FutureWarning``?

mikofski:
@wholmgren I hadn't realized the distinction between future and deprecation or pending deprecation warnings until now. 

### TL;DR:
According to my read of the [PEP](https://www.python.org/dev/peps/pep-0565/) and [Python Warnings docs](https://docs.python.org/3/library/warnings.html#warning-categories), I think we should use `FutureWarning`.

### more detail
Since PVLib is more like a developer toolbox, and less likely to be consumed by another application, it seems appropriate to issue verbose warnings. I think until Python-3.7 becomes more widespread, most users with Python-3.6 won't see deprecation or pending deprecation warnings unless they have turned warning filters off. If PEP-565 is backported to Python-3.6 then we can reconsider.

I also think we need guidelines in the contributor documentation that establish our procedures and some terms such as:
* how many releases should a `FutureWarning` be used before an item is deprecated?
* shall a deprecated feature be kept in a deprecated state forever, to maintain backwards compatibility?
* should a list of future and deprecated features be kept somewhere?
* what annotations are used in the documentation like `.. versionadded:` and `.. deprecated:` and are ad hoc warnings also used?
* when if ever is `PendingDeprecationWarning` used?

Pandas seems like a similar toolbox to PVLib, so I looked at it to see their MO. AFAICT: they seem to take the view that when new features will deprecate old ones, there are 3 stages:
1. both exist and are accepted (`FutureWarning`)
2. the feature is about to be deprecated (`PendingDeprecationWarning`)
3. the feature is deprecated, no longer in use and could be removed at any time (`DeprecationWarning`)

#### Pandas
I searched through the [https:/github.com/pandas-dev](Pandas codebase) and [documents](http://pandas.pydata.org/pandas-docs/stable/) and here's my guess:

1. future warnings are used first, they indicate:

    - that in the future this command, argument, feature, api, syntax, etc. will change to something new
    - they are used when a new feature is added that would break backwards compatibility
    - sphinx documentation is annotated with [`.. versionadded:: x.y.z`](http://pandas.pydata.org/pandas-docs/stable/contributing.html?highlight=versionadded#documenting-your-code), for example see the [`join` argument in `concat`](http://pandas.pydata.org/pandas-docs/stable/text.html?highlight=futurewarning#concatenating-a-series-and-an-indexed-object-into-a-series-with-alignment)
    - sphinx documentation is annotated with [`.. deprecated: x.y.z`](http://pandas.pydata.org/pandas-docs/stable/contributing.html?highlight=deprecation#backwards-compatibility), for example see [`groupby.take` `convert` negative indices flag](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.core.groupby.DataFrameGroupBy.take.html?highlight=deprecated%20since)
    - ad hoc warnings and `.. warning: blah` annotations are used in the documentation
    - after X releases, the feature is deprecated
    - items are listed in [pandas issue 6581](https://github.com/pandas-dev/pandas/issues/6581)

2. `PendingDeprecationWarning` is used infrequently, perhaps because it is only used for 1 or 2 releases prior to actual deprecation

3. features that have actually been deprecated use deprecation warnings

    - sphinx documentation is annotated with `.. deprecated:: x.y.z`
    - warnings are used ad hoc if the feature may be removed in a future version
    - items are listed in [pandas issue 13777](https://github.com/pandas-dev/pandas/issues/13777)

Pandas provides [decorators to deprecated functions and kwargs](../../../pandas-dev/pandas/blob/master/pandas/util/_decorators.py#L9) which defaults to `FutureWarning`, and their [contributor docs also say to issue a future warning](http://pandas.pydata.org/pandas-docs/stable/contributing.html?highlight=deprecation#backwards-compatibility).

wholmgren:
Thank you for the detailed review @mikofski! This is very helpful. 

Let's go with a warning inherited from ``FutureWarning`` in this PR. I will also consider adding ``.. versionadded:: x.y.z`` and ``.. deprecated: x.y.z`` to the docs where appropriate.

Let's discuss the other questions in a separate issue and documentation PR.

wholmgren:
This PR has gone on for so long that I've forgotten where I started... It turns out that the deprecation code I copied from Matplotlib/Metpy uses ``UserWarning`` to avoid the ``DeprecationWarning`` issues discussed above. I could not find any discussion in Matplotlib about why ``UserWarning`` was chosen over ``FutureWarning``. I have a mild preference for continuing to follow the Matplotlib pattern. If this is ok with others, then I think the PR is ready to merge.

docs/sphinx/source/api.rst
```diff
@@ -102,8 +102,8 @@ Airmass and atmospheric models
    :toctree: generated/
 
    location.Location.get_airmass
-   atmosphere.absoluteairmass
-   atmosphere.relativeairmass
+   atmosphere.absolute_airmass
```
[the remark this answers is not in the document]
wholmgren: pvlib only includes one absolute airmass model so it didn't occur to me to add ``get_`` here. I'm fine with either approach at this point.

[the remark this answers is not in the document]
cwhanse: I see your point.  It may help when someone is trying to find the relevant function, to keep names parallel.  There are other models for absolute air mass which we haven't implement.

wholmgren:
I think this is finally ready to merge.

I updated the code in the sphinx documentation and the tutorials to use the new functions. Rendered documentation [here](https://wholmgren-pvlib-python-new.readthedocs.io/en/totalirrad/) and tutorials [here](https://nbviewer.jupyter.org/github/wholmgren/pvlib-python/tree/totalirrad/docs/tutorials/). Minor related documentation changes... added ``plt.close()`` calls to avoid matplotlib memory warnings when building the sphinx documentation. Also removed most traces of seaborn from the documentation and tutorials. No need for seaborn in our documentation now that matplotlib has better default styles.

The appveyor/windows python 2.7 test configurations were failing on a test of a new deprecation test decorator. The same test passed on travis python 2.7 and earlier in this PR it also passed on appveyor (with nothing relevant to it changed since then). I modified the test to allow it to fail on windows on python 2.7.

PR #429, opened 2018-02-14
avoid runtime warnings from nan comparisons and divide by zero

### LINKED ISSUE #428: avoid and/or silence common warnings
A handful of pvlib functions often cause spit out runtime warnings for things like dividing by 0 or comparing the size of a nan value to a float. We should avoid or silence the warnings that we know are ok. They also make the test output a pain to read.

Here's a possibly incomplete list:

- [x] irradiance.perez divide by 0
- [x] irradiance.perez comparison with nan
- [x] clearsky.simplified_solis comparison with nan
- [x] pvsystem.sapm divide by 0 in log
- [x] pvsystem.ashraeiam comparison with nan
- [x] pvsystem.physicaliam comparison with nan
- [x] pvsystem.adrinverter comparison with nan
- [x] pvsystem.v_from_i overflow

~I have some code ready for addressing the warnings in the perez model implementation. I'll post it soon.~

Checked items are addressed in #429.

wholmgren:
- [x] Closes #428 
 - [ ] ~Updates entries to `docs/sphinx/source/api.rst` for API changes.~

I'll look into adding a few more of the items in #428.

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -2318,10 +2318,17 @@ def adrinverter(v_dc, p_dc, inverter, vtol=0.10):
     ac_power = p_nom * (pdc-p_loss)
     p_nt = -1 * np.absolute(p_nt)
 
-    ac_power = np.where((v_lim_upper < v_dc) | (v_dc < v_lim_lower),
-                        np.nan, ac_power)
+    # set output to nan where input is outside of limits
+    # errstate silences case where input is nan
+    with np.errstate(invalid='ignore'):
+        invalid = (v_lim_upper < v_dc) | (v_dc < v_lim_lower)
+    ac_power = np.where(invalid, np.nan, ac_power)
+
+    # set night values
     ac_power = np.where(vdc == 0, p_nt, ac_power)
     ac_power = np.maximum(ac_power, p_nt)
+
+    # set max ac output
     ac_power = np.minimum(ac_power, pac_max)
 
     if isinstance(p_dc, pd.Series):
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -733,9 +733,9 @@ def ashraeiam(aoi, b=0.05):
     '''
 
     iam = 1 - b * ((1 / np.cos(np.radians(aoi)) - 1))
-    mask = np.full_like(aoi, False, dtype='bool')
-    np.greater_equal(np.abs(aoi), 90, where=~np.isnan(aoi), out=mask)
-    iam = np.where(mask, 0, iam)
+    aoi_gte_90 = np.full_like(aoi, False, dtype='bool')
+    np.greater_equal(np.abs(aoi), 90, where=~np.isnan(aoi), out=aoi_gte_90)
+    iam = np.where(aoi_gte_90, 0, iam)
     iam = np.maximum(0, iam)
 
     if isinstance(iam, pd.Series):
@@ -837,14 +837,14 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
     # after deducting the reflected portion of each
     iam = ((1 - (rho_para + rho_perp) / 2) / (1 - rho_zero) * tau / tau_zero)
 
-    # set up arrays for nan-tolerant masking operations
-    mask = np.full_like(aoi, False, dtype='bool')
+    # set up array for nan-tolerant masking operations
     notnan = ~np.isnan(aoi)
 
     # angles near zero produce nan, but iam is defined as one
     small_angle = 1e-06
-    np.less(np.abs(aoi), small_angle, where=notnan, out=mask)
-    iam = np.where(mask, 1.0, iam)
+    aoi_lt_small_angle = np.full_like(aoi, False, dtype='bool')
+    np.less(np.abs(aoi), small_angle, where=notnan, out=aoi_lt_small_angle)
+    iam = np.where(aoi_lt_small_angle, 1.0, iam)
 
     # angles at 90 degrees can produce tiny negative values,
     # which should be zero. this is a result of calculation precision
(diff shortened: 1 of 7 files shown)


wholmgren:
@adriesse pointed out that some of these changes have a negative impact on readability. I agree that the more conservative masking pattern is harder to read. 

As of now, I'm only ignoring warnings using ``with np.errstate`` in situations where doing otherwise would seriously hurt readability. But that's just my opinion of what constitutes serious readability issues. We could stick with more of the older patterns and make more use of ``with np.errstate``. Or maybe there's a better pattern that I don't know about.

For what it's worth, the process of hunting down these warnings has certainly improved the test suite. I think that says something to the importance of paying attention to the warnings rather than silencing them. 

So, overall, I think the changes are a net positive, but I would revert some of the changes if the community feels differently.

The one thing that I do not want to do is advise users to set their global error state to ignore warnings.

cwhanse:
Personally I like the changes, and don't see them as obscuring the calculations.

> The one thing that I do not want to do is advise users to set their global error state to ignore warnings.

^ That.

It would be nice to better handle the np.exp() overflow warnings in v_from_i if you get that far.

adriesse:
I think I was silencing warnings on a module, function and/or line number level at one point.  So that seems pretty conservative to me and I would have no hesitations about recommending that to anyone.

Readability is subjective and as long as you  have it on your radar screen that's fine with me.

wholmgren:
@adriesse I do appreciate the push back on the readability issue. I've changed some of the mask array names to improve readability of the new pattern. The ``np.where`` statements now read as, for example, ``iam = np.where(aoi_gte_90, 0, iam)`` instead of the more generic looking ``iam = np.where(mask, 0, iam)``. The mask creation code is still a bit clunky, but I hope this is helps.

I didn't know that you could silence a warning based on line number -- nice trick!

I think this PR is ready to merge if we can agree that the numpy version bump discussed in #430 is ok.

pvlib/pvsystem.py
```diff
@@ -1912,9 +1939,11 @@ def v_from_i(resistance_shunt, resistance_series, nNsVth, current,
...
     if np.any(idx_p):
         # LambertW argument, cannot be float128, may overflow to np.inf
-        argW = I0[idx_p] / (Gsh[idx_p]*a[idx_p]) * \
-            np.exp((-I[idx_p] + IL[idx_p] + I0[idx_p]) /
-                   (Gsh[idx_p]*a[idx_p]))
+        # overflow is explicitly handled below, so ignore warnings here
+        with np.errstate(over='ignore'):
+            argW = (I0[idx_p] / (Gsh[idx_p]*a[idx_p]) *
+                    np.exp((-I[idx_p] + IL[idx_p] + I0[idx_p]) /
+                           (Gsh[idx_p]*a[idx_p])))
```
wholmgren: @cwhanse (and maybe @thunderfish24): I added a ``with np.errstate(over='ignore'):`` context to ``v_from_i``. Perhaps this will be made obsolete before too long.

pvlib/pvsystem.py
```diff
@@ -2274,22 +2303,30 @@ def adrinverter(v_dc, p_dc, inverter, vtol=0.10):
...
-    vdc = v_dc/v_nom
-    poly = np.array([pdc**0, pdc, pdc**2, vdc-1, pdc*(vdc-1),
-                     pdc**2*(vdc-1), 1/vdc-1, pdc*(1./vdc-1),
-                     pdc**2*(1./vdc-1)])
+    v_lim_upper = np.nanmax([v_max, vdc_max, mppt_hi]) * (1 + vtol)
+    v_lim_lower = np.nanmax([v_min, mppt_low]) * (1 - vtol)
+
+    pdc = p_dc / p_nom
+    vdc = v_dc / v_nom
+    with np.errstate(invalid='ignore', divide='ignore'):
```
adriesse: # zero voltage will lead to division by zero, but since power is set to night time value later, errors can be ignore

pvlib/pvsystem.py
```diff
@@ -2274,22 +2303,30 @@ def adrinverter(v_dc, p_dc, inverter, vtol=0.10):
...
     ac_power = p_nom * (pdc-p_loss)
-    p_nt = -1*np.absolute(p_nt)
+    p_nt = -1 * np.absolute(p_nt)
 
     ac_power = np.where((v_lim_upper < v_dc) | (v_dc < v_lim_lower),
                         np.nan, ac_power)
-    ac_power = np.where((ac_power < p_nt) | (vdc == 0), p_nt, ac_power)
-    ac_power = np.where(ac_power > pac_max, pac_max, ac_power)
+    ac_power = np.where(vdc == 0, p_nt, ac_power)
+    ac_power = np.maximum(ac_power, p_nt)
```
adriesse: What do you think about using the `out` parameter here?

pvlib/pvsystem.py
```diff
@@ -836,16 +837,24 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
...
+    # angles at 90 degrees can produce tiny negative values,
+    # which should be zero. this is a result of calculation precision
+    # rather than the physical model
+    iam = np.maximum(0, iam)
 
     # for light coming from behind the plane, none can enter the module
-    iam = np.where(aoi > 90, 0, iam)
+    aoi_gt_90 = np.full_like(aoi, False, dtype='bool')
+    np.greater(aoi, 90, where=notnan, out=aoi_gt_90)
+    iam = np.where(aoi_gt_90, 0, iam)
```
adriesse: I would suggest a block ignore invalid on these (originally) three statements.  I think aoi nans propagate fine to iam, and then the logic is all correct the way it is.

pvlib/pvsystem.py
wholmgren: Either way is fine with me. ``maximum`` and ``minimum`` ignore ``nan`` values, and I figured the ``maximum`` function was easier to read that the ``where``/``out`` pattern.

adriesse: On my computer `np.maximum([1,2,3,np.nan], 2)` gives a runtime warning...

Also, `maximum` has an `out` parameter too.

wholmgren: Maybe a numpy version difference? I'm using 1.14.0 in my test environment.

pvlib/pvsystem.py
wholmgren: I agree that the errstate call is easier to read here and I don't think we lose anything by using it here. Updated accordingly.

pvlib/pvsystem.py
adriesse: I appear to be using 1.11.3.

wholmgren:
closes #430.

adriesse:
I only really aspire to nudge things a little one way or another, and I guess I've succeeded at that.  I learned a few new tricks too.  Still, I would have preferred fewer masks and more commented ignores in the end result.

PR #439, opened 2018-03-21
DOC: fix superscript in forecast.rst

mikofski:
* change `$W/m**2$` to `$W/m^2$` so 2 is superscripted
* also change legend location to `"best"`
* _n.b._: this image doesn't appear to be using seaborn, and it also appears to be different than the previous build, not sure why?
poa_irrad


wholmgren:
@mikofski thanks for the quick fix and sorry for overlooking this until now. 

Regarding the difference, it looks to me like you built the docs in an environment with matplotlib 2. RTD built the docs using 1.5 until recently (#446).

PR #444, opened 2018-03-30
Update SAM libraries

### LINKED ISSUE #440: update sam library files
We've fallen a few years behind the SAM releases, so I think it's time to update the module and inverter CEC files.

cwhanse:

 - [x] Closes #440 


wholmgren:
YES!!! 

This mostly looks good to me. It looks like you accidentally changed the line for looking up the adr inverter file instead of the cec inverter file.

We should add a comment to the whats new file. 

I think we should remove the old files, too, but that is up for debate.

```
git rm pvlib/data/sam-library-cec-inverters-2015-6-30.csv
git rm pvlib/data/sam-library-cec-modules-2015-6-30.csv
```

cwhanse:
We can remove the old files.  I checked that the new files are adding entries, there aren't any old ones that will disappear.  I checked that the pvlib function reads the new files.



And, it figures that once the commit got through it would have a mistake in it.





________________________________

cwhanse:
Fixed the inverter file lines

wholmgren:
It appears that the "Example Module" parameters are slightly different, so that explains a few of the failing tests. 

I think the others are failing because some of the fields are now read in as strings instead of as numbers. Maybe there is a slight change to the format or a new name that prevents pandas from assuming that the fields are numeric?

wholmgren:
I think the problem is line 2508 in the inverters file:

```
LeadSolar Energy: LS600T [208V] 208V [CEC 2018],208,500,75790.4,36,-nan(ind),1.22899e-07,-0.0251886,-nan(ind),0.0798754,1,45,2105.29,27,45
```

cwhanse:
Glad you tracked that line down.  I got as far are 'there's something different in format between these files'

wholmgren:
@cwhanse up to you on if we should hold off for the SAM team to update the files. In any case, you'll need to ``git rm`` the 2015 files, then commit the change.

cwhanse:
I talked with NREL, we should edit the SAM libraries and move ahead with our merge.  The alternative is to wait for the next SAM release, since that is when the SAM libraries are updated.

thunderfish24:
@cwhanse Thanks for working thru the issues with the new version. FWIW, while I was "scraping" the existing cec module dataset for single cells (i.e., Ns=1), my scraper code caught a few entries with missing/bad data.

wholmgren:
@cwhanse and @thunderfish24: do you want to remove those modules too or go ahead and merge? They don't seem to cause a problem for the import function, so my vote is to move on.

cwhanse:
I’d keep any module which has a full set of parameter values.  We don’t have a system to judge good from bad parameter sets, so we have no basis to remove a module. Even when a module has a questionable value (e.g. negative resistance, diode factor < 1) the calculation could still produce usable IV curves.

Sent from my iPhone

On Apr 6, 2018, at 10:17 AM, Will Holmgren <notifications@github.com<mailto:notifications@github.com>> wrote:


@cwhanse<https://github.com/cwhanse> and @thunderfish24<https://github.com/thunderfish24>: do you want to remove those modules too or go ahead and merge? They don't seem to cause a problem for the import function, so my vote is to move on.

wholmgren:
The only outstanding issue is to add an entry to the 0.5.2.rst whatsnew file. I don't think there should be any problem with merge commits because git is set to union those files by default.

cwhanse:
@mikofski thanks for that illustration.  I'm not opposed to adopting parts of pvfree into pvlib, as a replacement for the current method of reading csv files of model parameters.  But we'd want to agree that the change is needed and is an improvement.  I don't see that pvlib is ready for this change.  The problem is providing a traceable and stable input files.  The CEC databases contain module and inverter measurements, not model parameters.  The SAM libraries are output of processing the public CEC databases using tools that are now in the SAM SDK, but using those tools to produce the library files is currently somewhat ad hoc. Since we can't produce the model parameters natively in pvlib by processing the CEC databases, we rely on a particular SAM release to provide traceable and stable inputs. Because of this reliance on SAM, I think we're better using the files directly, than we would be with an additional layer of code between the files and pvlib.

I hope this makes sense.

cwhanse:
Note added for whatsnew

wholmgren:
Hm, I don't see the changes to whatsnew. Maybe missed a ``commit`` or ``push`` step?

wholmgren:
There it is. I don't understand why there is a merge conflict. I tested it locally and it worked fine on my machine. Irritating.

Can you click on the resolve conflicts button and remove the three marked lines?

wholmgren:
Great, thanks!

cedricleroy:
I think https://github.com/pvlib/pvlib-python/commit/f892664b255dc382b9e87f3340ff768ef31ba0ee did overwrite a couple of previous things in `v0.5.2.rst` 

There is a "0" here: https://github.com/pvlib/pvlib-python/commit/f892664b255dc382b9e87f3340ff768ef31ba0ee#diff-2df8cbc501f75a90135963ace8a28f63R1

And the text here has been removed: https://github.com/pvlib/pvlib-python/commit/f892664b255dc382b9e87f3340ff768ef31ba0ee#diff-2df8cbc501f75a90135963ace8a28f63L15

docs/sphinx/source/whatsnew/v0.5.2.rst
```diff
@@ -12,8 +12,10 @@ API Changes
 
 Enhancements
 ~~~~~~~~~~~~
-* Improve clearsky.lookup_linke_turbidity speed, changing .mat source
-  file to .h5 (:issue:`437`)
+* Updated libraries for CEC module parameters to SAM release 2017.9.5 
+(library dated 2017.6.30) and CEC inverter parameters to file posted to 
+www.github.com/NREL/SAM on 2018.3.18, with one entry removed due to a 
+missing parameter value. (:issue:'440')
```
mikofski: please change single quotes around #440 to backticks

PR #448, opened 2018-04-06
replace logging.warning with warnings.warn, remove logging.debug

### LINKED ISSUE #447: logging.basicConfig in __init__.py causes misbehavior after import
Since `basicConfig` is called as a import side-effect of pvlib, any software that imports anything from pvlib can not use it's own basicConfig (if basicConfig was already called, any further call just does nothing - not even giving a warning).

This leads to unexpected behavior (no more debug output) which is rather hard to track down.

I don't know whether it is bad to copy issues word-by-word, but this affects pvlib as well: https://github.com/circus-tent/circus/issues/826

The problem can be reproduced using the code form Python logging cookbook:
https://docs.python.org/2/howto/logging-cookbook.html#using-logging-in-multiple-modules
Adding `import pvlib` will cause all logging to be ouputted to terminal.

Maybe this is by design, but it's a bit difficult to get around.

wholmgren:
Will anyone miss these internal pvlib logging calls?

With these changes, ``grep logg pvlib/* -R`` returns nothing.

 - [x] Closes #447 

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.5.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.5.2.rst
@@ -6,6 +6,9 @@ v0.5.2 (__, 2017)
 API Changes
 ~~~~~~~~~~~
 * removed unused 'pressure' arg from irradiance.liujordan function (:issue:`386`)
+* replaced logging.warning calls with warnings.warn calls, and removed
+  logging.debug calls. We encourage users to explore tools such as pdb and
+  trackback in place of the logging.debug calls. Fixes (:issue:`447`).
 
 Enhancements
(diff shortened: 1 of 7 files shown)


mikofski:
No I won't miss them. Less to maintain if we cut them out. My vote is to approve this PR.

cwhanse:
I agree, we can remove the logging calls

Sent from my iPhone

On Apr 6, 2018, at 12:38 PM, Mark Mikofski <notifications@github.com<mailto:notifications@github.com>> wrote:


No I won't miss them. Less to maintain if we cut them out. My vote is to approve this PR.

Issue #436, opened 2018-03-04
Breaking out parts of pvsystem.py?


thunderfish24:
Sorry if this is a duplicate, but has anyone considered breaking out the different module performance models in pvsystem.py into their own modules for better manageability?

For example, put the desoto model and sapm into their own modules.

wholmgren:
@mikofski brought up something similar in #235.

I agree that the ``pvsystem.py`` module has grown to a difficult to mange length. Would you rather see the module broken up by model heritage (e.g. ``sapm.py``, ``pvwatts.py``, ``diode.py``) or would you rather see it broken up by model type (e.g. ``dcpower.py``, ``acpower.py``, ``iam.py``).

I've also considered moving ``PVSystem`` (and maybe ``SingleAxisTracker``) into a separate module. 

In any case, a subpackage might be preferable for API consistency and for keeping some kind of grouping for this related functionality. On the other hand, the Zen of Python says "flat is better than nested", so this too is up for debate.

jforbess:
I think it makes a little more sense to break them out by model heritage,
because I would tend to add more complicated modeling aspects in an sapm
model, whereas if I just wanted to apply a few quick and dirty losses I
might use pvwatts.

cwhanse:
I'm not in favor of breaking pvsystem into subpackages by model name (SAPM, Desoto, etc.).  I am in favor of restructuring pvsystem by modeling step, and writing wrapper functions where we don't currently use them (the iam functions, for example), then moving the wrapped functions into a subpackage.  As an example why I don't favor a 'sapm' subpackage, I believe that sapm_celltemp is the only cell temperature model in pvlib, currently.  We would be better off to have a celltemp function that wraps celltemp_sapm (and celltemp_faiman and others) and use kwargs, like we do with irradiance functions.

Some of the length in pvsystem can be reduced if I would ever get to writing pvlib.io.

wholmgren:
Good points @jforbess and @cwhanse.

Looking only at the "easier" ideas...

I think the only io related functions are ``retrieve_sam`` and ``_parse_raw_sam_df``. Moving those would be an improvement of 130 lines.

``systemdef`` could probably be removed entirely. It's not used anywhere else in the library and I don't know why anyone would use it on its own. That's another 80 lines.

``_golden_sect_DataFrame`` and ``_pwr_optfcn`` could be moved to a ``_optimize.py`` module and/or will be replaced. 80 lines.

[...]

Issue #394, opened 2017-11-02
Unit tests depend on functions not being tested


cwhanse:
`test_irradiance.py` uses 
```
ephem_data = solarposition.get_solarposition(
    times, tus.latitude, tus.longitude, method='nrel_numpy')

irrad_data = tus.get_clearsky(times, model='ineichen', linke_turbidity=3)

dni_et = irradiance.extraradiation(times.dayofyear)
```
to compute known values for unit tests.  This dependence on functions not under test could cause a unit test to fail when the unit's code is unchanged.

wholmgren:
@cwhanse I reviewed the tests and I believe we can close this issue now that the PRs referenced above have been merged.

``test_atmosphere.py`` uses a similar pattern as shown in your description, but the returned values are not explicitly tested. There is already another issue, #158, for improving that test module.

cwhanse:
I agree, let's close this issue.  If we find more tests with dependencies, we can reference this conversation, and fix those tests with their own issue and PR.

PR #466, opened 2018-05-23
Expose first_solar_spectral_correction

### LINKED ISSUE #359: Expose first_solar_spectral_correction in ModelChain and PVSystem
I suggest the following steps to expose the ``first_solar_spectral_correction`` function to ``ModelChain``:

1. Add a ``PVSystem.first_solar_spectral_correction(pw, airmass_absolute)`` method that wraps the ``atmosphere.first_solar_spectral_correction(pw, airmass_absolute, module_type=None, coefficients=None)`` function. This is straightforward except for deciding how to handle the keyword arguments. The Sandia module database has a ``Material`` key  and the CEC database has a ``Technology`` key. These keys could be mapped to the allowed ``module_type`` arguments. I'm not sure what to do about the ``coefficients`` option. Perhaps direct users to put their custom coefficients in ``module_parameters['fs_spectral_coefficients']``?

2. Implement the stub ``ModelChain.first_solar_spectral_loss`` method. Maybe something like:
```python
    def first_solar_spectral_loss(self):
        self.spectral_modifier = self.system.first_solar_spectral_loss(
            self.weather['precipitable_water'], self.airmass['airmass_absolute'])
        return self
``` 
I'm guessing that most users don't have PW in their weather dataframes, so this line is going to be a frequent source of errors. We might consider adding a ``try/except KeyError`` with a more helpful error message.

3. Improve ``ModelChain.infer_spectral_model``. This probably amounts to setting ``mc._spectral_model = mc.first_solar_spectral_loss`` if a CEC module or a ``module_parameters['fs_spectral_coefficients']`` key is detected on the ``PVSystem`` object.

4. Optional. Add a dewpoint temperature to precipitable water conversion to the ``ModelChain.prepare_inputs`` method.

5. Add tests.

6. Update documentation.

It's unfortunate that ``sapm_spectral_loss`` and ``first_solar_spectral_correction`` live in different modules. Fixing that could be step 0, or we could leave it alone.

I'm in no rush to implement this, but I'd help anyone with a pull request.

see also #358

cwhanse:

 - [x] Closes #359

#359 complete

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -444,7 +444,7 @@ class PVSystem(object):
             electrical current.
         """
 
-        if 'first_solar_spectral_coefficients' in 
+        if 'first_solar_spectral_coefficients' in \
                                self.module_parameters.keys():
             coefficients = \
                    self.module_parameters['first_solar_spectral_coefficients']
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -438,28 +438,24 @@ class PVSystem(object):
         Returns
         -------
         modifier: array-like
-            spectral mismatch factor (unitless) which is can be multiplied
+            spectral mismatch factor (unitless) which can be multiplied
             with broadband irradiance reaching a module's cells to estimate
             effective irradiance, i.e., the irradiance that is converted to
             electrical current.
         """        
 
-        if self.module_parameters['fs_spectral_coefficients']:
-            coefficients = self.module_parameters['fs_spectral_coefficients']
+        if self.module_parameters['first_solar_spectral_coefficients']:
+            coefficients = \
+                   self.module_parameters['first_solar_spectral_coefficients']
             module_type = None
         else:
             module_type = self._infer_cell_type()
             coefficients = None
 
-        if (module_type is not None and coefficients is None) or \
-           (module_type is None and coefficients is not None):
-               return atmosphere.first_solar_spectral_correction(pw, 
-                                                          airmass_absolute, 
-                                                          module_type,
-                                                          coefficients)
-        else:
-            # default to no_spectral_loss function
-            return 1
+        return atmosphere.first_solar_spectral_correction(pw, 
+                                                  airmass_absolute, 
+                                                  module_type,
+                                                  coefficients)
         
     def _infer_cell_type(self):
         
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -425,7 +425,14 @@ class PVSystem(object):
 
         """
         Use the :py:func:`first_solar_spectral_correction` function to
-        calculate the spectral loss modifier.
+        calculate the spectral loss modifier. The model coefficients are
+        specific to the module's cell type, and are determined by searching
+        for one of the following keys in self.module_parameters (in order):
+            'first_solar_spectral_coefficients (user-supplied coefficients)
+            'Technology' - a string describing the cell type, can be read from
+            the CEC module parameter database
+            'Material' - a string describing the cell type, can be read from
+            the Sandia module database.
 
         Parameters
         ----------
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -421,6 +421,91 @@ class PVSystem(object):
             poa_direct, poa_diffuse, airmass_absolute, aoi,
             self.module_parameters, reference_irradiance=reference_irradiance)
 
+    def first_solar_spectral_correction(self, pw, airmass_absolute):
+        
+        """
+        Use the :py:func:`first_solar_spectral_correction` function to 
+        calculate the spectral loss modifier.
+
+        Parameters
+        ----------
+        pw : array-like
+            atmospheric precipitable water (cm).
+    
+        airmass_absolute : array-like
+            absolute (pressure corrected) airmass.
+
+        Returns
+        -------
+        modifier: array-like
+            spectral mismatch factor (unitless) which is can be multiplied
+            with broadband irradiance reaching a module's cells to estimate
+            effective irradiance, i.e., the irradiance that is converted to
+            electrical current.
+        """        
+
+        if self.module_parameters['fs_spectral_coefficients']:
+            coefficients = self.module_parameters['fs_spectral_coefficients']
+            module_type = None
+        else:
+            module_type = self._infer_cell_type()
+            coefficients = None
+
+        if (module_type is not None and coefficients is None) or \
+           (module_type is None and coefficients is not None):
+               return atmosphere.first_solar_spectral_correction(pw, 
+                                                          airmass_absolute, 
+                                                          module_type,
+                                                          coefficients)
+        else:
+            # default to no_spectral_loss function
+            return 1
+        
+    def _infer_cell_type(self):
+        
+        """
+        Examines module_parameters and maps the Technology key for the CEC
+        database and the Material key for the Sandia database to a common
+        list of strings for cell type.
+        
+        Returns
+        -------
+        cell_type: str
+        
+        """
+
+        _cell_type_dict = {'Multi-c-Si': 'multisi',
+                           'Mono-c-Si': 'monosi',
+                           'Thin Film': 'cigs',
+                           'a-Si/nc': 'asi',
+                           'CIS': 'cigs',
+                           'CIGS': 'cigs',
+                           '1-a-Si': 'asi',
+                           'CdTe': 'cdte',
+                           'a-Si': 'asi',
+                           '2-a-Si': None,
+                           '3-a-Si': None,
+                           'HIT-Si': 'monosi',
+                           'mc-Si': 'multisi',
+                           'c-Si': 'multisi',
+                           'Si-Film': 'asi',
+                           'CdTe': 'cdte',
+                           'EFG mc-Si': 'multisi',
+                           'GaAs': None,
+                           'a-Si / mono-Si': 'monosi'}
+
+        if 'Technology' in self.module_parameters.keys():
+            # CEC module parameter set
+            cell_type = _cell_type_dict[self.module_parameters['Technology']]
+        elif 'Material' in self.module_parameters.keys():
+            # Sandia module parameter set
+            cell_type = _cell_type_dict[self.module_parameters['Material']]
+        else:
+            cell_type = None
+            
+        return cell_type
+        
+        
     def singlediode(self, photocurrent, saturation_current,
                     resistance_series, resistance_shunt, nNsVth,
                     ivcurve_pnts=None):
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -421,7 +421,7 @@ class PVSystem(object):
             poa_direct, poa_diffuse, airmass_absolute, aoi,
             self.module_parameters, reference_irradiance=reference_irradiance)
 
-    def first_solar_spectral_correction(self, pw, airmass_absolute):
+    def first_solar_spectral_loss(self, pw, airmass_absolute):
         
         """
         Use the :py:func:`first_solar_spectral_correction` function to 
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -428,7 +428,7 @@ class PVSystem(object):
         calculate the spectral loss modifier. The model coefficients are
         specific to the module's cell type, and are determined by searching
         for one of the following keys in self.module_parameters (in order):
-            'first_solar_spectral_coefficients (user-supplied coefficients)
+            'first_solar_spectral_coefficients' (user-supplied coefficients)
             'Technology' - a string describing the cell type, can be read from
             the CEC module parameter database
             'Material' - a string describing the cell type, can be read from
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -444,7 +444,7 @@ class PVSystem(object):
             electrical current.
         """        
 
-        if self.module_parameters['first_solar_spectral_coefficients']:
+        if 'first_solar_spectral_coefficients' in self.module_parameters.keys():
             coefficients = \
                    self.module_parameters['first_solar_spectral_coefficients']
             module_type = None
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -422,16 +422,16 @@ class PVSystem(object):
             self.module_parameters, reference_irradiance=reference_irradiance)
 
     def first_solar_spectral_loss(self, pw, airmass_absolute):
-        
+
         """
-        Use the :py:func:`first_solar_spectral_correction` function to 
+        Use the :py:func:`first_solar_spectral_correction` function to
         calculate the spectral loss modifier.
 
         Parameters
         ----------
         pw : array-like
             atmospheric precipitable water (cm).
-    
+
         airmass_absolute : array-like
             absolute (pressure corrected) airmass.
 
@@ -442,9 +442,10 @@ class PVSystem(object):
             with broadband irradiance reaching a module's cells to estimate
             effective irradiance, i.e., the irradiance that is converted to
             electrical current.
-        """        
+        """
 
-        if 'first_solar_spectral_coefficients' in self.module_parameters.keys():
+        if 'first_solar_spectral_coefficients' in 
+                               self.module_parameters.keys():
             coefficients = \
                    self.module_parameters['first_solar_spectral_coefficients']
             module_type = None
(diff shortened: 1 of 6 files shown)


pvlib/pvsystem.py
```diff
@@ -421,6 +421,91 @@ def sapm_effective_irradiance(self, poa_direct, poa_diffuse,
...
+            with broadband irradiance reaching a module's cells to estimate
+            effective irradiance, i.e., the irradiance that is converted to
+            electrical current.
+        """        
+
+        if self.module_parameters['fs_spectral_coefficients']:
+            coefficients = self.module_parameters['fs_spectral_coefficients']
+            module_type = None
+        else:
+            module_type = self._infer_cell_type()
```
wholmgren: module vs. cell is unfortunate here. No objection to what you've done, but we might need a separate discussion to try to clean up the terminology within the library.

pvlib/pvsystem.py
```diff
@@ -421,6 +421,91 @@ def sapm_effective_irradiance(self, poa_direct, poa_diffuse,
...
+            coefficients = None
+
+        if (module_type is not None and coefficients is None) or \
```
wholmgren: Recommend removing this if/else statement. If this method doesn't have the information necessary to successfully compute the spectral correction, then this method should say so rather than defaulting to 1. 

Consistent with Python's style, the ``atmosphere.first_solar_spectral_correction`` function throws a ``TypeError`` with some hint as to what the user needs to supply. We could (ideally/optionally) catch that error here and reraise it with a more specific message for how to modify ``module_parameters`` to make this work.

I am guessing that raising an exception here will cause issues in a ModelChain calculation for the few module/cell types for which ``_infer_cell_type`` mapping returns ``None``. One approach to avoid that may be to make the ``ModelChain.infer_spectral_model`` preemptively call ``pvsystem._infer_cell_type``.

pvlib/pvsystem.py
```diff
@@ -421,6 +421,91 @@ def sapm_effective_irradiance(self, poa_direct, poa_diffuse,
...
+        pw : array-like
+            atmospheric precipitable water (cm).
+    
+        airmass_absolute : array-like
+            absolute (pressure corrected) airmass.
+
+        Returns
+        -------
+        modifier: array-like
+            spectral mismatch factor (unitless) which is can be multiplied
```
wholmgren: typo (is can) that is carried over from the original function's doc string.

adriesse:
I looked it over and didn't find anything worth bothering you about.  :)

pvlib/pvsystem.py
cwhanse: Here I'm inclined to break the API and change the keyword argument from module_type to cell_type.  module_type is only used by the first_solar function.

pvlib/pvsystem.py
cwhanse: Good idea.  I'll move the test to ModelChain.infer_spectral_model and put a warning in place of that check in the actual functions.

cwhanse:
I'm tripping over function name conventions again.  Related to this PR, but maybe it merits its own Issue and PRs.  For spectral modifier models, we have `first_solar_spectral_loss` and `sapm_spectral_loss`, where the model name is the prefix to the model purpose.  We did likewise for `physicaliam` and `ashraeiam` .  For clear-sky models we conveniently bundled the models `haurwitz` etc. into a module (is that the right term?) so that the path provides the prefix, e.g., `clearsky.haurwitz`

I have a strong preference to have the model purpose as the prefix, followed by the model name, e.g., `spectral_loss_first_solar` and `iam_physical`.  Or we could create a separate module for each model purpose, e.g., `spectral_loss.first_solar`. 

Thoughts?

wholmgren:
Good point. I also prefer the naming schemes you describe. 

Perhaps the easiest way forward for this PR is to continue with ``PVSystem.first_solar_spectral_loss`` for consistency with the related function and with the other ``PVSystem`` methods. It's easy to change the name in a future PR within the same release cycle.

This is also related to the discussion in #436. And #427 adds machinery for deprecating function names without immediately breaking user code.

cwhanse:
I remember that discussion.  I'll complete this PR with function names as is, let's tackle the naming issue in a separate PR.

@wholmgren I'm going to need some guidance on implementing the tests.  Looks like we need a new `test_first_solar_spectral_loss` in `test_pvsystem`.  I can generate the correct value for different cell types in MATLAB. What's a reasonable value for `pw`?

I do not understand the code in the current `test_modelchain.test_spectral_models`  It looks like we are passing arguments from the aoi functions.

wholmgren:
If you want to compare to matlab, I recommend ensuring that the ``first_solar_spectral_loss`` function tests in ``test_atmosphere.py``  produce the same results. 

For ``PVSystem.first_solar_spectral_loss`` method tests I am more concerned with API correctness. One good approach may be to set up your test ``PVSystem`` object with the ``module_parameters`` you want, then ensure that the ``PVSystem.first_solar_spectral_correction`` produces the same result as an explicit call to ``atmosphere.first_solar_spectral_correction``. Then repeat the process with different configurations. Repeating the process is best done using the ``@pytest.mark.parametrize`` decorator, though I recommend starting simple. It's ok with me if you want to copy/paste 3 or 4 similar tests to get started.

cwhanse:
@wholmgren when you can, a little help to understand why these tests are failing?  I have the decorator wrong but I don't understand why.

wholmgren:
Traveling this weekend. Will check Tuesday.

cwhanse:
Progress.  I grok the test now.  The code I submitted is following the pattern for `test_aoi_models`. It defines a `ModelChain` object, uses the `run_model` method and inspects the output.  I had an error that I was examining the `ac` output instead of the `spectral_modifier` output. Now I'm up against a design decision for the test.

When we test individual functions like the spectral models we have to know the expected output. So I made up values for `precipitable_water` and `airmass_absolute` and computed the values each spectral model should produce. However, we are testing these functions as part of a model chain, where the output of the function being tested depends on the output of upstream functions. In this case, the output of `sapm_spectral_loss` depends on `airmass` which is calculated using the `times` input, ignoring the made-up values for `precipitable_water` and `airmass_absolute` I supplied when the model chain is created in `test_spectral_models`.

I see three ways to resolve the problem, neither of which are appealing:
1) I recalculate the expected values by running each model in the chain.  I don't like this approach, because a future change to one of the upstream functions could cause this test of the spectral models to fail.
2) Change the test pattern to not use a `ModelChain` object.  This is a problem because it wouldn't use the `ModelChain` methods which assign the spectral model method to the `ModelChain` object.
3) Add code to `ModelChain.__init__` to handle `airmass` as a kwarg, and impose user-supplied values in place of the calculation using `times` and `location`. This is my least liked option because I don't see a use of the additional code beyond this made-up test.

What is preferred here?  I'll proceed with the first option, but perhaps there's a fourth option I don't see.

adriesse:
Perhaps it is not necessary to check the precise values that are returned using the ModelChain.  If the spectral calculation function is already verified to be correct through its own tests, then you really just need to test/exercise the additional code related to the ModelChain integration.  It might be enough to check that you get a Series of floats or something like that, I'm not sure.  

I suppose the inference code would need to be tested somehow too.  That might be easier if the result of any inferences were explicitly stored in the ModelChain, so a test (or a user) could check what happened.

wholmgren:
Let me first try to state my current thinking about how to structure unit tests in pvlib in the context of this pull request.

1. The tests in ``test_atmosphere.py`` should ensure that the ``atmosphere.first_solar_spectral_correction`` function returns the desired output for a variety of function inputs. The tests should be independent of other pvlib functions (see #394). The tests should ensure that all reasonable combinations of input types (floats, nans, arrays, series, scalars, etc) work as expected. I think the existing tests for this function are reasonably good and no new work is needed in this pull request.

2. The tests in ``test_pvsystem.py`` should ensure that the ``PVSystem.first_solar_spectral_correction`` method calls the ``atmosphere.first_solar_spectral_correction`` function with the desired inputs determined from all relevant configurations of a ``PVSystem`` object. A test should also ensure that the method does return some reasonable data, though the precise values of the data should be covered by tests in item 1.

3. The tests in ``test_modelchain.py`` should ensure that ``ModelChain.__init__`` correctly configures the ``ModelChain`` object to eventually run ``PVSystem.first_solar_spectral_correction``. A test should ensure that method is eventually run in the course of ``Model_chain.run_model``. A test should ensure that the model selection does have a reasonable effect on the subsequent calculations, though the precise values of the data should be covered by tests in item 1.

That thinking has evolved from the many mistakes that I have made in writing pvlib tests, too many of which are still with us.

The function-level tests in item 1 are relatively straight forward so long as we avoid the dependency issue @cwhanse pointed out in #394. I used to be bad and lazy about that, so there are some older tests that have dependencies that they shouldn't have.

The ``PVSystem`` and ``ModelChain`` tests are less straight forward to implement. I often thought that I could reasonably check items 2 and 3 by testing that a subset of user-level numeric inputs produced the desired numeric outputs. Of course, these higher-level objects effectively inject a bunch of dependencies into the tests, and we're back at issue #394. The plus side is that they are simple to write the first time.

Moving forward to comments from @cwhanse and @adriesse 

> I recalculate the expected values by running each model in the chain. I don't like this approach, because a future change to one of the upstream functions could cause this test of the spectral models to fail.

Agreed. 

> Change the test pattern to not use a ModelChain object. This is a problem because it wouldn't use the ModelChain methods which assign the spectral model method to the ModelChain object.

We will want tests that only use a ``PVSystem`` object (item 2) and tests that do use a ``ModelChain`` object (item 3). 

> Add code to ModelChain.__init__ to handle airmass as a kwarg, and impose user-supplied values in place of the calculation using times and location. This is my least liked option because I don't see a use of the additional code beyond this made-up test.

Agreed -- I think we can come up with a better way.

> Perhaps it is not necessary to check the precise values that are returned using the ModelChain. If the spectral calculation function is already verified to be correct through its own tests, then you really just need to test/exercise the additional code related to the ModelChain integration. 

Agreed, though I do want to emphasize the importance (in my mind) of separating tests at the level of ``PVSystem`` (item 2) and ``ModelChain`` (item 3). I don't know how to write this kind of test, though, and welcome suggestions. I know it's possible to assert that a function is called (maybe even with specific arguments), but I don't have practice doing it. I think pytest has some support for this kind of testing. The ``mock`` library is also useful for related testing problems.

> I suppose the inference code would need to be tested somehow too. That might be easier if the result of any inferences were explicitly stored in the ModelChain, so a test (or a user) could check what happened.

The inferences are stored in the sense that a specific model's corresponding method e.g. ``ModelChain.first_solar_spectral_loss`` is assigned to the corresponding generic attribute e.g. ``ModelChain._spectral_model``. A test could ensure that this happens given some input ``PVSystem`` configurations. Also, the string representation of a ``ModelChain`` allows a user to easily check what happened with the inference using ``print(my_model_object)``. 

How to move forward? I will start by learning more about pytest and mock in an effort to design test templates that let us address items 2 and 3 while avoiding unintended dependencies. I'll look into that this week, though it may take longer to come up with a good solution. I am also ok moving forward with something like @cwhanse's option 1, but it's complicated by the fact that some of the existing model chain tests break due to the corrections applied by this new feature. 

Finally, we should add some text and examples to the contributing section of the documentation if we can come to some agreement about testing principles and practice.

wholmgren:
This might be easier than I thought... See this commit for an example of using pytest-mock with the ``PVSystem`` incidence angle modifiers: https://github.com/wholmgren/pvlib-python/commit/783ddf31d2326ab9c03b2d77cb9eb588be14a310

I don't fully understand the pros/cons of using [pytest-mock](https://github.com/pytest-dev/pytest-mock) vs. [pure mock](https://docs.python.org/3/library/unittest.mock.html#module-unittest.mock), but this seems doable either way.

pvlib/test/test_pvsystem.py
```diff
@@ -266,6 +266,17 @@ def test_PVSystem_sapm_spectral_loss(sapm_module_params):
     out = system.sapm_spectral_loss(airmass)
 
 
+@pytest.mark.parametrize("expected", [1.03173953])
+
```
wholmgren: python convention is no blank lines between decorators and the function.

pvlib/test/test_modelchain.py
```diff
@@ -263,21 +263,16 @@ def constant_spectral_loss(mc):
...
+    weather = pd.DataFrame(data=[0.3, 0.5, 1.0], index=times, columns=['precipitable_water'])
     mc = ModelChain(system, location, dc_model='sapm',
                     aoi_model='no_loss', spectral_model=spectral_model)
-    times = pd.date_range('20160101 1200-0700', periods=2, freq='6H')
-    ac = mc.run_model(times).ac
-
-    expected = pd.Series(np.array(expected), index=times)
-    assert_series_equal(ac, expected, check_less_precise=2)
+    spectral_modifier = mc.run_model(times=times, weather=weather).spectral_modifier
+    print(spectral_modifier)
```
wholmgren: I try to avoid putting print functions in the final tests. In case you are not already aware of it, the pytest ``--pdb`` option can be useful for avoiding print in development e.g. ``pytest pvlib/test/test_modelchain.py::test_spectral_models --pdb``

pvlib/test/test_modelchain.py
```diff
@@ -263,21 +263,16 @@ def constant_spectral_loss(mc):
...
 @requires_scipy
-@pytest.mark.parametrize('spectral_model, expected', [
-    ('sapm', [182.338436597, -2.00000000e-02]),
-    pytest.mark.xfail(raises=NotImplementedError)
-    (('first_solar', [179.371460714, -2.00000000e-02])),
-    ('no_loss', [181.604438144, -2.00000000e-02]),
-    (constant_spectral_loss, [163.061569511, -2e-2])
-])
-def test_spectral_models(system, location, spectral_model, expected):
+@pytest.mark.parametrize('spectral_model', ['sapm', 'first_solar', 'no_loss'])
```
wholmgren: please keep ``constant_spectral_loss`` in the list of parameters. This tests that a user-supplied function can be passed in as an argument. To be clear, the list should be: ``['sapm', 'first_solar', 'no_loss', constant_spectral_loss]``

pvlib/pvsystem.py
```diff
@@ -421,6 +421,87 @@ def sapm_effective_irradiance(self, poa_direct, poa_diffuse,
             poa_direct, poa_diffuse, airmass_absolute, aoi,
             self.module_parameters, reference_irradiance=reference_irradiance)
 
+    def first_solar_spectral_loss(self, pw, airmass_absolute):
+        
+        """
+        Use the :py:func:`first_solar_spectral_correction` function to 
+        calculate the spectral loss modifier.
```
wholmgren: Doc string should have some kind of reference to ``self.module_parameters``. Most of the other ``PVSystem`` method doc strings have some examples. This method is a little more complicated though, so might be worth being more explicit about what is looked for. For example...

Use the :py:func:`first_solar_spectral_correction` function to calculate the spectral loss modifier. The spectral loss modifier is determined by searching for one of the following keys in ``self.module_parameters`` (in order): ``'first_solar_spectral_coefficients'``, ``'Technology'``, ``'Material'``.

pvlib/test/test_modelchain.py
adriesse: constant_spectral_loss --> constant_loss ?

adriesse:
While browsing the code, it struck me that the word "loss" is used quite a lot.  In contrast to many other loss types, this one can be a gain also and the code would benefit from a more neutral word (or "gain/loss") where it is convenient to do so.

pvlib/test/test_pvsystem.py
cwhanse: Fixed

pvlib/test/test_modelchain.py
cwhanse: Removed, I had that print to debug the test.

pvlib/test/test_modelchain.py
cwhanse: I put the keyword back.  I tend to agree with @adriesse about 'loss' but the terminology is common.

pvlib/pvsystem.py
cwhanse: Use the :py:func:`first_solar_spectral_correction` function to
        calculate the spectral loss modifier. The model coefficients are
        specific to the module's cell type, and are determined by searching
        for one of the following keys in self.module_parameters (in order):
            'first_solar_spectral_coefficients (user-supplied coefficients)
            'Technology' - a string describing the cell type, can be read from
            the CEC module parameter database
            'Material' - a string describing the cell type, can be read from
            the Sandia module database.

pvlib/pvsystem.py
```diff
@@ -425,7 +425,14 @@ def first_solar_spectral_loss(self, pw, airmass_absolute):
 
         """
         Use the :py:func:`first_solar_spectral_correction` function to
-        calculate the spectral loss modifier.
+        calculate the spectral loss modifier. The model coefficients are
+        specific to the module's cell type, and are determined by searching
+        for one of the following keys in self.module_parameters (in order):
+            'first_solar_spectral_coefficients (user-supplied coefficients)
```
wholmgren: missing ' -

wholmgren:
@adriesse I am open to discussing and changing the "loss" terminology as part of a separate issue.

PR #468, opened 2018-05-30
use pytest-mock to simplify and improve tests, improve test docs

wholmgren:
Rewrites PVSystem tests using ``pytest-mock`` and adds a new ModelChain test. Please see the whatsnew diff and the new documentation for more details on how and why. 

The changes to ``test_pvsystem.py`` are complete (I think). I don't currently have time to go through the entire ``test_modelchain.py`` module and improve the tests to use the new methods. I suggest calling it good enough for this pull request. It at least provides an example for people working with the modelchain code in the future.

See discussion in #466 in [this comment](https://github.com/pvlib/pvlib-python/pull/466#issuecomment-392558404) and below for more context.

 - [ ] Closes issue #xxxx

### DIFF
--- a/docs/sphinx/source/contributing.rst
+++ b/docs/sphinx/source/contributing.rst
@@ -150,7 +150,7 @@ reasonable.
 .. code-block:: python
     def test_PVSystem_ashraeiam(mocker):
         # mocker is a pytest-mock object.
-        # mocker.spy adds code to a function to keep track of how its called
+        # mocker.spy adds code to a function to keep track of how it is called
         mocker.spy(pvsystem, 'ashraeiam')
 
         # set up inputs
@@ -180,10 +180,37 @@ The tests in test_modelchain.py should ensure that
 ``ModelChain.__init__`` correctly configures the ModelChain object to
 eventually run the selected models. A test should ensure that the
 appropriate method is actually called in the course of
-``Model_chain.run_model``. A test should ensure that the model selection
+``ModelChain.run_model``. A test should ensure that the model selection
 does have a reasonable effect on the subsequent calculations, though the
 precise values of the data should be covered by the function tests
 discussed above. ``pytest-mock`` can also be used for testing ``ModelChain``.
+
+The example below shows how mock can be used to assert that the correct
+PVSystem method is called through ``ModelChain.run_model``.
+
+.. code-block:: python
+    def test_modelchain_dc_model(mocker):
+        # set up location and system for model chain
+        location = location.Location(32, -111)
+        system = pvsystem.PVSystem(module_parameters=some_sandia_mod_params,
+                                   inverter_parameters=some_cecinverter_params)
+
+        # mocker.spy adds code to the system.sapm method to keep track of how
+        # it is called. use returned mock object m to make assertion later,
+        # but see example above for alternative
+        m = mocker.spy(system, 'sapm')
+
+        # make and run the model chain
+        mc = ModelChain(system, location,
+                        aoi_model='no_loss', spectral_model='no_loss')
+        times = pd.date_range('20160101 1200-0700', periods=2, freq='6H')
+        mc.run_model(times)
+
+        # assertion fails if PVSystem.sapm is not called once
+        m.assert_called_once()
+
+        # ensure that dc attribute now exists and is correct type
+        assert isinstance(mc.dc, (pd.Series, pd.DataFrame))
 
 
 This documentation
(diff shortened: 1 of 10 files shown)


docs/sphinx/source/contributing.rst
```diff
@@ -81,29 +81,113 @@ changes, such as fixing documentation typos.
...
+        module_parameters = pd.Series({'b': 0.05})
+        system = pvsystem.PVSystem(module_parameters=module_parameters)
+        thetas = 1
+
+        # call the method
+        iam = system.ashraeiam(thetas)
+
+        # did the method call the function as we expected?
+        # mocker.spy added assert_called_once_with to the function
+        pvsystem.ashraeiam.assert_called_once_with(thetas, b=0.05)
```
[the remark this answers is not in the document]
wholmgren: In the example above, either ``b=0.05`` or ``b=module_parameters['b']`` would work. What is important is that something dict-like is passed to ``PVSystem``. For example, this would also work:  
``system = pvsystem.PVSystem(module_parameters={'b': 0.05})`` followed by the assert statement as shown. Or this would work:

```
b = 0.05
system = pvsystem.PVSystem(module_parameters=dict(b=b))
# ...
pvsystem.ashraeiam.assert_called_once_with(thetas, b=b)
```

Would one of the above be a better example? Or more comments?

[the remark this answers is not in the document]
cwhanse: I think it would appear more clear to me if we change line 157 to `module_parameters = {'b' : 0.05}` which avoids recasting the dict as a pandas object.  Also, to me it is more consistent to use the dict in the assertion,  `assert_called_once_with(thetas, module_parameters['b'])` instead of `assert_called_once_with(thetas, b=0.05)`

pvlib/test/test_pvsystem.py
```diff
@@ -257,23 +242,14 @@ def test_sapm_spectral_loss(sapm_module_params, airmass, expected):
...
     system = pvsystem.PVSystem(module_parameters=sapm_module_params)
-
-    times = pd.DatetimeIndex(start='2015-01-01', periods=2, freq='12H')
-    airmass = pd.Series([1, 10], index=times)
-
+    airmass = 2
     out = system.sapm_spectral_loss(airmass)
-
-
-@pytest.mark.parametrize("expected", [1.03173953])
```
wholmgren: I accidentally deleted this test in a merge. I'll fix next.

pvlib/test/test_pvsystem.py
```diff
@@ -342,18 +318,23 @@ def test_sapm_effective_irradiance(sapm_module_params, test_input, expected):
...
+    aoi = 10
+    reference_irradiance = 1100
 
     out = system.sapm_effective_irradiance(
         poa_direct, poa_diffuse, airmass_absolute,
         aoi, reference_irradiance=reference_irradiance)
+    pvsystem.sapm_effective_irradiance.assert_called_once_with(
+        poa_direct, poa_diffuse, airmass_absolute, aoi, sapm_module_params,
+        reference_irradiance=reference_irradiance)
+    assert_allclose(out, 1, atol=0.1)
```
wholmgren: This low-precision test will fail if we change the units of effective irradiance. That could be an argument for or against having any low-precision tests at all, depending on your perspective.

pvlib/test/test_pvsystem.py
```diff
@@ -770,17 +752,13 @@ def test_scale_voltage_current_power(sam_data):
...
-        index=[0])
-    expected = pd.DataFrame(
-        np.array([[6, 4.5, 20, 16, 72, 1.5, 4.5]]),
-        columns=['i_sc', 'i_mp', 'v_oc', 'v_mp', 'p_mp', 'i_x', 'i_xx'],
-        index=[0])
+def test_PVSystem_scale_voltage_current_power(mocker):
+    data = None
     system = pvsystem.PVSystem(modules_per_string=2, strings_per_inverter=3)
+    m = mocker.patch(
+        'pvlib.pvsystem.scale_voltage_current_power', autospec=True)
```
wholmgren: This test function uses a different pattern than the others. Unlike ``mocker.spy``, ``mocker.patch`` can only be used to test that something is called because it completely replaces the function/method under consideration. I chose this pattern here because I think the input data is a little more complicated than is worth maintaining in the method test. Could be changed, though.

pvlib/test/test_pvsystem.py
wholmgren: Added below.

pvlib/test/test_pvsystem.py
cwhanse: Here, we are testing that `pvsystem.sapm_effective_irradiance` is called and returns a reasonable value, correct?  If so, then I would change `airmass_absolute = 1.5`, `aoi = 0` and `poa_direct = 900`.  We should keep `reference_irradiance = 1000`. With these inputs the method call should return 1.0 exactly.  At least until we act on #472

wholmgren: I updated the input parameters. 

>Here, we are testing that pvsystem.sapm_effective_irradiance is called and returns a reasonable value, correct? 

Yes, so I kept ``atol=0.1``.

Issue #470, opened 2018-05-31
Implement PVsyst single diode model


cwhanse:
I've heard a number of requests for this capability, which we have in MATLAB but not here.  Requires calcparams_PVsyst and an example set of module_parameters.

Jawairia123:
It would be great if you include PVsyst loss model!

Issue #469, opened 2018-05-31
Data structure(s) for PV measurement datasets such as I-V curves


thunderfish24:
I'm looking for input on using/creating a "standard" data structure to store PV measurement datasets, such as I-V curves supporting IEC 61853-1. I'm thinking of something flexible/extensible and self-documenting (esp. w.r.t. units). This space also seems to intersect with time-series of I-V curves and maybe PECOS workflows.

For example, I have a collection of I-V-F-T curves, each with possibly varying numbers of points, that are each taken at a "nominal" matrix of effective irradiance F = 0.1, 0.2, 0.4, 0.6, 0.8, 1.0, 1.1 (unitless) and temperature T = 15, 25, 50 degC. Sticking to just python and numpy (pandas doesn't seem like the right fit here), I came up with this dict-based structure:
```python
data = {
('0.1', '15 degC'): {'v_V': numpy.array([v_1, v_2, ..., v_M]), 'i_A': numpy.array([i_1, i_2, ..., i_M]), 'f': numpy.array([f_1, f_2, ..., f_M]), 't_degC': numpy.array([t_1, t_2, ..., t_M])},
('0.2', '15 degC'): {'v_V': numpy.array([v_1, v_2, ..., v_N]), 'i_A': numpy.array([i_1, i_2, ..., i_N]), 'f': numpy.array([f_1, f_2, ..., f_N]), 't_degC': numpy.array([t_1, t_2, ..., t_N])},
...
}
```
In this case, I could retrieve the currents vector for a particular curve using `data[('0.2', '15 degC')]['i_A']`. I also need to concatenate (in a consistent order) all the currents, voltages, etc. from all the curves together. One could also imagine repeated I-V-F-T curve measurements at each nominal setting (with possibly a different number of points in each repetition).

The ordered-pair keys can also be sorted in various ways using `sorted()`, as long as the chosen strings don't cause ordering problems. Note that replacing the keys with timestamps would produce time-series I-V-F-T curve data.

cwhanse:
@thunderfish24 I support this effort. FYI - we're planning a more formal taxonomy development as part of Orange Button that has these data within it's scope, but that's a year out.  Some thoughts about python structure:

- instead of 2 numpy.arrays, what about an array of tuples (v_V, I_A)?  The voltage and current come in pairs.  It may be awkward to work with them as tuple type, though.

[...]

Issue #463, opened 2018-05-16
Implement CEC module performance model in place of desoto


cwhanse:
The Desoto module performance model and the CEC model differ in that the CEC model has one additional parameter (an adjustment to temperature coefficients for Isc and Voc).  pvlib-python implements only the Desoto model, but supplies parameters that are fit for the CEC model.  Replacing the Desoto model with the CEC model would remove this inconsistency.  The differences between the two model's results are quite minor.

adriesse:
Are you talking about the so-called "adjust" parameter, whose origins I had trouble tracing a year or more back?

cwhanse:
Yes

adriesse:
I retained the impression that it was used to influence the fitting process rather than being a model parameter...

cwhanse:
Adjust enters the IV curve calculation via the equation for light current: IL = G * ( IL0 + alphaprime (Tc - T0)), and alphaprime = alpha ( 1 - Adjust/100).  As far as I know this is only detailed in A. Dobos paper describing the 6 parameter model coefficient calculator. The procedure to determine the value for Adjust is still obscure to me despite Aron's paper.

adriesse:
The way I saw it is that your diode model parameters may end up with a new, adjusted alpha (compared to the datasheet value) but to me that doesn't really constitute an additional model parameter. 

There are a number of things that worry me about that paper, including the sanity check that allows alpha and beta to be adjusted by +/- 100%.

cwhanse:
We can (and I agree that we should) add the CEC model and keep the Desoto functions.  However, strictly speaking we have CEC model parameters, rather than Desoto parameters.  The values in SAM are determined simultaneously with the Adjust parameter (in the CEC bot not the Desoto model) and thus there is some (certainly minor) influence of the Adjust value on the other parameters.

thunderfish24:
FWIW there is https://www.nrel.gov/docs/fy18osti/67399.pdf, esp. section 10.4. There are many reasons why “the” reported Isc temperature coefficient could be need adjustment, such as its spectral dependence during measurement. Why the same adjustment would be used simultaneously on Isc and Voc (presumably in data fitting) I cannot say. I too would be suspect if the adjustment were large, and is it true then that an IV curve’s Voc could be very different from the “adjusted” value? If yes, then what Voc value would pvlib report?

cwhanse:
[...]

PR #471, opened 2018-06-01
Clarify Egref and dEgdT values, changes calcparams_desoto api

### LINKED ISSUE #462: Clarify module_parameters['EgRef'] usage, and set a default value
The CEC module parameters database does not contain the parameters EgRef or dEgdT. In the CEC model, and in the process NREL follows to generate the database from the CEC module data list, values EgRef = 1.121 and dEgdT = -0.0002677 are implicit.  PVLib for Matlab exposed EgRef and dEgdT as a model parameter (which it is) and let the user determine the appropriate value.  pvlib-python does likewise, however, readthedocs.io examples do not inform the user that s/he must specify EgRef and dEgdT nor do we provide default values when they are not specified.

I propose that we:
1) assign default values within `pvsystem.retrieve_sam('cecmod')` and
2) add to the example at [http://pvlib-python.readthedocs.io/en/latest/modelchain.html](url) to illustrate use of the CEC module model, and the opportunity to specify EgRef and dEgdT, if desired.

cwhanse:

 - [x] Closes #462 
 - [X] Code quality and style is sufficient. Passes ``git diff upstream/master -u -- "*.py" | flake8 --diff`` and/or landscape.io linting service.

Sets default values for CEC module parameters `Egref` and `dEgdT` in `pvsystem.calcparams_desoto`.  Changes API for `pvsystem.calcparams_desoto` to individual keyword arguments, removing the `module_parameters` dict).  Adds `_build_kwargs` to `PVSystem.calcparams_desoto`

I believe this is ready to merge.  I could not fully check out the changes in the tutorial file, since the Jupyter notebook is calling `calcparams_desoto`


wholmgren:
Are the changes to the SAM file and irradiance.py due to some git hassles? If so, recommend

```
git fetch upstream
git checkout upstream/master -- pvlib/data/sam-library-cec-inverters-2018-3-18.csv pvlib/irradiance.py
git commit pvlib/data/sam-library-cec-inverters-2018-3-18.csv pvlib/irradiance.py -m 'checkout master sam and irradiance files'
```

I'm not 100% sure, but I think this will remove the proposed changes to these files.

cwhanse:
@wholmgren yes the changes to the SAM file and irradiance.py are attempts to resolve git conflicts.  The checkout/commit sequence you suggested fixed the issue.

pvlib/pvsystem.py
```diff
@@ -944,63 +945,61 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
     return iam
 
 
-def calcparams_desoto(poa_global, temp_cell, alpha_isc, module_parameters,
-                      EgRef, dEgdT, M=1, irrad_ref=1000, temp_ref=25):
+def calcparams_desoto(effective_irradiance, temp_cell,
+                      alpha_sc, a_ref, I_L_ref, I_o_ref, R_sh_ref, R_s, 
+                      EgRef=1.121, dEgdT=-0.0002677,
+                      irrad_ref=1000, temp_ref=25):
```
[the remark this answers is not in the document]
cwhanse: @thunderfish24 here we have a seeming conflict between adhering to the published reference (input would be absorbed irradiance, after reflections and soiling losses but without an adjustment from broadband to effective irradiance) and consistency within pvlib-python.  SAPM (and PVsyst, when it is implemented #470) accept effective irradiance as input.  I would prefer to see Desoto (and CEC, when it is implemented #463) use the same input quantity, with function signature and inline comments documenting the change from literature.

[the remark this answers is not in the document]
thunderfish24: From the standpoint of being able to implement and compare different models easily from the same time-dependent input data (say, global irradiance from pyranometer, albedo, ambient dry bulb temperature, and wind speed), I think having a consistent API would be preferable over complete fidelity to the original model formulation (and documenting the modifications in the code sounds acceptable). It sounds like the definition of "effective irradiance" that you are talking about here is [the last one here](https://pvpmc.sandia.gov/modeling-steps/2-dc-module-iv/effective-irradiance/) with SF=1. Can you confirm?

I also realize that my thoughts are muddy on POA irradiance. In particular, the above-mentioned irradiance ratio at the POA doesn't seem to include any of the losses that give DeSoto's S or M factors (or an SF for that matter). It sounds like we plan on moving those factors into a separate function (or functions)?

Computation graphs might be helpful here for the various models taking inputs to outputs.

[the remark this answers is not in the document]
cwhanse: SAPM documentation strikes again.  The explanation of effective irradiance in SAPM is rather opaque and is the question we are asked most frequently.  A better explanation would be:
- Ee is _measured_ as a ratio of Isc to Isc at reference conditions adjusted for temperature (this is consistent with @thunderfish24 F quantity
- For prediction of power from broadband POA irradiance, Ee is calculated by reducing the POA by surface reflection, soiling and adjusting for spectrum.  This calculation is implicit in Eq. 1 on this page : https://pvpmc.sandia.gov/modeling-steps/2-dc-module-iv/point-value-models/sandia-pv-array-performance-model/

From our attempts to bring order and consistency to PV modeling we arrived at this perspective:
- POA irradiance refers to broadband irradiance at the module's face.  This is the quantity produced by all of the popular transposition models
- effective irradiance is the irradiance converted to electric current.  Not everyone uses this term, but enough papers use it that we decided it made sense to emphasize it.
- translation of POA irradiance to effective irradiance accounts for reflections, soiling, spectral irradiance adjustment, and perhaps other mechanisms.
- De soto (and by inheritance CEC) is the only paper that uses a term for 'adsorbed irradiance', which is POA irradiance less reflections and soiling.

cwhanse:
Ready to merge I think.

pvlib/pvsystem.py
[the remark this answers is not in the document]
thunderfish24: Thanks for the further detailed explanation @cwhanse, which is making better sense to me now. I need a bit more time to think about the "swapping" of this effective irradiance into DeSoto, but it seems reasonable.

Also, it looks like the SAPM's effective irradiance does not try to treat the various angles of incidence of diffuse reflectance on the POA (which seems like a challenging prospect anyhow, and there is an f_D factor).

Note that the "effective irradiance" F in [my definition here](https://onlinelibrary.wiley.com/doi/full/10.1002/ese3.190) does _not_ include a temperature correction to Isc0. This is not an approximation, but falls out of the derivation.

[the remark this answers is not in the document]
thunderfish24: One can derive an alternative auxiliary equation for the photocurrent I_L at operating conditions in the Desoto model by setting V=0 in the SDM and rearranging:
```
I_L = I_sc + I_D * (exp(I_sc * R_s / (N_S * n * V_th(T)) - 1) + I_sc * R_s / R_sh
      = I_sc/I_sc_0 * I_sc_0 + I_D * (exp(I_sc/I_sc_0 * I_sc_0 * R_s / (N_S * n * V_th(T)) - 1) + I_sc/I_sc_0 * I_sc_0 * R_s / R_sh
      = F * I_sc_0 + I_D * (F * I_sc_0 * R_s / (N_S * n * V_th(T)) - 1) + F * I_sc_0 * R_s / R_sh
```
where some of these other parameters are also computed at operating conditions (e.g., I_D).

One can then use equation (1) as the model for I_sc at operating conditions [in the SAPM](https://pvpmc.sandia.gov/modeling-steps/2-dc-module-iv/point-value-models/sandia-pv-array-performance-model/) (possibly also with a SF) to compute F at operating conditions :).

I humbly request that we implement this auxiliary equation for I_L for the SDM in terms of F, which then would then allow us to use either the SAPM _or_ a matched (or unmatched) reference device measurement for the determination of the control variable F.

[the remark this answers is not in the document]
cwhanse: @thunderfish24 I could see your request being part of a new function calcparams_(name here).  That would be consistent with the API: calcparams_xxxxx returns the 5 values for the single diode equation at irradiance and temperature, etc. My concerns are 1) a source of model parameters (the module-specific constants used by calcparams_xxxxx and 2) ensuring that users understand the input F.  Most use cases start with measured broadband irradiance, although pvlib doesn't force the user to consider the instrument behind the measurement.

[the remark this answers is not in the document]
thunderfish24: K. I am happy to implement this after I work out a usage example with the SAPM I_sc in more detail. It seems to me that if someone is already using measured broadband irradiance as a starting point with SAPM, then that workflow should readily adapt to this implementation of SDM with existing calibration parameters (with only a translation of I_L0 to I_sc0, I think).

I also have to first wrap up #426. (I have most of the documentation done there except for some function parameter descriptions and two more tests that I really should write.)

docs/sphinx/source/whatsnew/v0.6.0.rst
```diff
@@ -5,13 +5,14 @@ v0.6.0 (___, 2018)
 
 API Changes
 ~~~~~~~~~~~
-
+* pvsystem.calcparams_desoto now accepts only keyword arguments for each module model parameter. Prior to v0.6.0, pvsystem.calcparams_desoto required some module model parameters as keyword argument and accpted a dict, module_parameters, as an input providing the remaining module model parameters.
```
wholmgren: Recommend something more like:

pvsystem.calcparams_desoto now requires arguments for each module model parameter.

The arguments you've added are positional, not [keyword-only](https://www.python.org/dev/peps/pep-3102/).

wholmgren: The whatsnew files generally use an 80 character line limit, though this is not as strict as for python modules.

pvlib/pvsystem.py
```diff
@@ -944,63 +945,61 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
...
-        The temperature dependence of the energy bandgap at SRC (in
-        1/C). May be either a scalar value (e.g. -0.0002677 as in [1])
+        The temperature dependence of the energy bandgap at reference
+        conditions in units of 1/C. May be either a scalar value (e.g. -0.0002677 as in [1])
         or a DataFrame of dEgdT values corresponding to each input
         condition (this may be useful if dEgdT is a function of
-        temperature).
+        temperature). For parameters read from
+        the SAM CEC module database, dEgdT=-0.0002677 is imposed by the 
+        parameter estimation algorithm used by NREL.
```
wholmgren: remove ``M`` documentation below here.

pvlib/pvsystem.py
```diff
@@ -1134,26 +1133,29 @@ def calcparams_desoto(poa_global, temp_cell, alpha_isc, module_parameters,
...
+    IL = effective_irradiance / irrad_ref * \
+              (I_L_ref + alpha_sc * (Tcell_K - Tref_K))
+    I0 = (I_o_ref * ((Tcell_K / Tref_K) ** 3) *
           (np.exp(EgRef / (k*(Tref_K)) - (E_g / (k*(Tcell_K))))))
-    Rsh = Rsh_ref * (irrad_ref / poa_global)
-    Rs = Rs_ref
+    # Note that the equation for Rsh differs from [1]. In [1] Rsh is given as
+    # Rsh = Rsh_ref * (S_ref / S) where S is broadband irradiance reaching
+    # the module's cells. If desired this model behavior can be duplicated
+    # by applying reflection and soiing losses to broadband plane of array
```
wholmgren: soiing --> soiling

pvlib/pvsystem.py
```diff
@@ -944,63 +945,61 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
     return iam
 
 
-def calcparams_desoto(poa_global, temp_cell, alpha_isc, module_parameters,
-                      EgRef, dEgdT, M=1, irrad_ref=1000, temp_ref=25):
+def calcparams_desoto(effective_irradiance, temp_cell,
+                      alpha_sc, a_ref, I_L_ref, I_o_ref, R_sh_ref, R_s,
```
wholmgren: I will [repeat](https://github.com/pvlib/pvlib-python/issues/469#issuecomment-394407197) my preference for spelling out most of these arguments in the function signature as in [i_from_v](http://pvlib-python.readthedocs.io/en/latest/generated/pvlib.pvsystem.i_from_v.html#pvlib.pvsystem.i_from_v). It would require adding some mapping function within ``PVSystem.calcparams_desoto`` but I think that's worth it for more descriptive argument names. The descriptive parameter names could be reassigned to the short names to keep the equations more readable.

The abbreviated names are still an improvement to the existing API, so I do not object to moving forward as is after some consideration.

docs/sphinx/source/whatsnew/v0.6.0.rst
wholmgren: accpted --> accepted

cwhanse: shortened to `pvsystem.calcparams_desoto now requires arguments for each module model parameter.` to stay within character limits.

pvlib/pvsystem.py
cwhanse: done

pvlib/pvsystem.py
cwhanse: fixed

docs/sphinx/source/whatsnew/v0.6.0.rst
wholmgren: Ok. To be clear and for future reference, you can also continue on to additional lines if you indent them by 2 spaces so that it looks like

```
* this is a long item....
  that continues here...
```

pvlib/pvsystem.py
wholmgren: Works for me.

cwhanse:
@wholmgren thanks for the careful checking and proofreading. I have added a less-clumsy exit from calcparams_desoto if the 4th positional argument looks like the old `module_parameters`.  I used `warnings.warn` - let me know if some other way to convey messages is preferred.

pvlib/pvsystem.py
```diff
@@ -1133,6 +1121,28 @@ def calcparams_desoto(effective_irradiance, temp_cell,
          Source: [4]
     '''
 
+    # test for use of function pre-v0.6.0 API change
+    if isinstance(a_ref, dict) or \
+       (isinstance(a_ref, pd.Series) and ('a_ref' in a_ref.keys())):
```
wholmgren: I don't understand the grouping of statements here. I would have thought you'd group on the ``or`` rather than the ``and``. In any case, I think ``isinstance(a_ref, (dict, pd.Series))`` alone may be what we want. A dict or Series that does not have 'a_ref' will fail in the equations that follow. Better to have it fail in the key look up. But maybe I am just misunderstanding.

pvlib/pvsystem.py
```diff
@@ -1133,6 +1121,28 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+        import warnings
+        warnings.warn('module_parameters detected as fourth positional'
```
wholmgren: Recommend using ``warnings.warn(message, DeprecationWarning)`` where message is both of your statements as one.

pvlib/pvsystem.py
```diff
@@ -1133,6 +1121,28 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+        warnings.warn('calcparams_desoto will require one argument for' \
+                      + ' each module model parameter in v0.7.0 and later')
+        try:
+            module_parameters = a_ref
+            a_ref = module_parameters['a_ref']
+            I_L_ref = module_parameters['I_L_ref']
+            I_o_ref = module_parameters['I_o_ref']
+            R_sh_ref = module_parameters['R_sh_ref']
+            R_s = module_parameters['R_s']
+            warnings.warn('module parameters extracted from fourth positional'\
```
wholmgren: No need for this warning. We already told them once above. Safe to assume success unless exception raised below.

pvlib/pvsystem.py
```diff
@@ -1133,6 +1121,28 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+                          + ' argument')
+        except:
```
wholmgren: Recommend something like

```
except Exception as e:
    raise e('Module parameters could not be extracted from '
            '4th positional argument. Check that parameters are '
            'from CEC database and/or update your code for new API.')
```

and then no need for the warning or return below

pvlib/pvsystem.py
cwhanse: Tested the `if` block on my own.

I added that block because I am assuming a use case where multiple sets of module parameters are passed into the new function as `pandas.Series` objects, and I didn't want the new function to fail in that case.

pvlib/pvsystem.py
cwhanse: Done

pvlib/pvsystem.py
cwhanse: OK

wholmgren:
I think this is ready to go. Does anyone object?

cwhanse:
Is there a way to rerun the test?  Looks like an unrelated failure in forecast.py

thunderfish24:
We used to close and reopen PR at my last job, but that was for a different build system.

mikofski:
@wholmgren or any pvlib admin, can do it from the [pvlib Travis ci page](https://travis-ci.org/pvlib/pvlib-python/pull_requests) there's usually a button that says rebuild. You can also use the Travis API but looks like this build already passed?

pvlib/pvsystem.py
thunderfish24: The R_s instead of R_s_ref in SAM has always puzzled me.

pvlib/pvsystem.py
```diff
@@ -944,75 +945,61 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
...
-              number of cells in series (Ns),
-              and cell temperature (Tcell) per equation (2) in [1].
-            * I_L_ref - Light-generated current (or photocurrent)
-              in amperes at reference conditions. This value is referred to
-              as Iph in some literature.
-            * I_o_ref - diode reverse saturation current in amperes,
-              under reference conditions.
-            * R_sh_ref - shunt resistance under reference conditions (ohms).
-            * R_s - series resistance under reference conditions (ohms).
+        module in units of A/C.
```
thunderfish24: C could be confused with Coulombs. Perhaps degC is preferable.

pvlib/pvsystem.py
```diff
@@ -944,75 +945,61 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
...
+        The shunt resistance at reference conditions, in ohms.
+        
+    R_s : float
+        The series resistance at reference conditions, in ohms.
 
     EgRef : float
-        The energy bandgap at reference temperature (in eV).
-        1.121 eV for silicon. EgRef must be >0.
+        The energy bandgap at reference temperature in units of eV.
+        1.121 eV for silicon. EgRef must be >0.  For parameters read from
```
thunderfish24: nit: Might it be better to say crystalline Silicon here (as opposed to amorphous, etc.)?

pvlib/pvsystem.py
```diff
@@ -944,75 +945,61 @@ def physicaliam(aoi, n=1.526, K=4., L=0.002):
...
+        the SAM CEC module database, EgRef=1.121 is imposed by the 
+        parameter estimation algorithm used by NREL.
 
     dEgdT : float
-        The temperature dependence of the energy bandgap at SRC (in
-        1/C). May be either a scalar value (e.g. -0.0002677 as in [1])
+        The temperature dependence of the energy bandgap at reference
+        conditions in units of 1/C. May be either a scalar value (e.g. -0.0002677 as in [1])
```
thunderfish24: C could be confused with Coulombs. Perhaps degC is preferable.

wholmgren:
I restarted the Appveyor build. I don't know if that requires admin access or not.

pvlib/pvsystem.py
wholmgren: I think the unit is clear in the context of the docstring.

adriesse: Amps/Coulomb = 1/second

pvlib/pvsystem.py
cwhanse: Good point, 1.121eV is for crystalline silicon.

pvlib/pvsystem.py
cwhanse: I changed this one to 1/K, because we use Kelvin in the calculation.

pvlib/pvsystem.py
cwhanse: I'd like to leave the temperature unit on the temperature coefficients as C, because that's the unit used in IEC standards and datasheets.  I agree with @wholmgren about being clear in context.

pvlib/pvsystem.py
```diff
@@ -1134,26 +1121,50 @@ def calcparams_desoto(poa_global, temp_cell, alpha_isc, module_parameters,
...
     Tref_K = temp_ref + 273.15
     Tcell_K = temp_cell + 273.15
 
     E_g = EgRef * (1 + dEgdT*(Tcell_K - Tref_K))
 
     nNsVth = a_ref * (Tcell_K / Tref_K)
 
-    IL = (poa_global/irrad_ref) * M * (IL_ref + alpha_isc * (Tcell_K - Tref_K))
-    I0 = (I0_ref * ((Tcell_K / Tref_K) ** 3) *
+    IL = effective_irradiance / irrad_ref * \
```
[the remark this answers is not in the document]
cwhanse: Added

cwhanse:
I updated the notebook `docs/tutorials/pvsystem.ipynb` but I don't known how to point it at the updated code in this pull request, so it is not fully tested.

wholmgren:
I ran the pvsystem.ipynb tutorial. It looks good to me. Results [here](https://github.com/wholmgren/pvlib-python/blob/Egref/docs/tutorials/pvsystem.ipynb). The PR currently contains output with exceptions, but that is ok with me. If you do want to clean it up... I run notebooks against test code using the procedure outlined [here](http://pvlib-python.readthedocs.io/en/latest/installation.html#install-as-an-editable-library). It's simplest if you start the jupyter notebook server within the virtual environment you've created for pvlib development rather than using the jupyter notebook server that came with your base Anaconda installation (assuming that's what you're using). Again, ok with me if we skip it for now.

Thanks @thunderfish24 for your careful reviews.

Issue #475, opened 2018-06-13
Implement hybrid SAPM+Campanelli et al. model


thunderfish24:
I would like to implement a hybrid model that combines the single-diode model (SDM) developments of Campanelli et al. ([see this](https://onlinelibrary.wiley.com/doi/full/10.1002/ese3.190)) with the SAPM to compute the effective irradiance ratio F and effective temperature ratio H using the SAPM when necessary, i.e., when not using a matched reference device or cell-embedded junction temperature measurement.

I figure one should embrace Python's duck typing (_Quack!_), so below is my proposed SDM 5 coefficient calculation architecture that is a partial generalization of the existing function `calcparams_desoto`. Basically, this is a way of implementing a (somewhat inelegant) computational graph for all the coefficients (e.g., IL, I0, Rs, Rsh, nNsVth). I realize many folks aren't familiar with my model, but if you work through this, you'll see that something like this pattern may make it easier for users to implement different "flavors" of auxiliary equations for the SDM when they are trying to compare models, find a better model, and the like. (I may need to recast my model in terms of `effective_irradiance` and `cell_temp` to make it fit better with the current direction of the pvlib API, and this might make possible more reuse of auxiliary equations between different SDMs.)

There is a lot of metadata packed into `kwargs`, and this provided metadata would change with the particular auxiliary equations selected. (I apologize that this will be a bit hard to understand from just the code snippet below, but I am still working on a runnable example.) What's nice about this pattern, is that one can specify a scalar, an array, or a function-to-be-computed for F, H, and IL, I0, Rs, Rsh, nNsVth, and once a parameter is computed, it can then "flow through" to the remaining computations. (However, unlike [Dask](https://dask.pydata.org/en/latest/), the user has to specify the computation in the proper (serial) order to avoid missing computations or re-computations of the computational graph's nodes.)

```python
def calc_params_hybrid(**kwargs):
    """
    Calculate the coefficients of the 5-parameter SDM using a hybrid method.

    TODO Explain hybrid method and reference SAPM and Campanelli et al paper.
    """

    # Unfortunately, order matters here because some computations rely on
    #  previous ones
    # A computational graph framework might make this more elegant and handle
    #  more complicated cases, including parallel/distributed computations
    kwargs['F'] = get_arg(kwargs['F'], **kwargs)
    kwargs['H'] = get_arg(kwargs['H'], **kwargs)
    kwargs['I0'] = get_arg(kwargs['I0'], **kwargs)
[...]

PR #478, opened 2018-06-14
Initial Campanelli et al. SDM

markcampanelli:

 - [ ] Closes issue #475


thunderfish24:
TBH, I'm not 100% sold on using `kwargs` like have so far. I am looking into Dask's computational tasks to see if there is a more transparent way to do this that retains flexibility while adding parallelism.

pvlib/pvsystem.py
```diff
@@ -279,7 +326,8 @@ def physicaliam(self, aoi):
 
         return physicaliam(aoi, **kwargs)
 
-    def calcparams_desoto(self, effective_irradiance, temp_cell, **kwargs):
```
thunderfish24: Any kwargs passed here are overwritten.

pvlib/pvsystem.py
```diff
@@ -153,11 +158,53 @@ def __init__(self,
...
+        function.
+
+        TODO Details
+        """
+
+        if self.model_dc == 'campanelli':
+            return sdm(get_coeffs_campanelli(irr, temp,
+                       **self.module_parameters),
+                       modules_per_string=self.modules_per_string,
+                       strings_per_inverter=self.strings_per_inverter)
```
thunderfish24: Hmmm..., I think I should be more consistent here and make `sdm` have the same API/signature as `sapm` and `pvwatts_dc`. This might mean that the `get_coeff_*` call is determined somehow from the `module_parameters` and done inside `sdm`.

pvlib/pvsystem.py
```diff
@@ -119,10 +123,11 @@ def __init__(self,
                  modules_per_string=1, strings_per_inverter=1,
                  inverter=None, inverter_parameters=None,
                  racking_model='open_rack_cell_glassback',
-                 name=None,
+                 name=None, model_dc='pvwatts',
                  **kwargs):
 
         self.name = name
+        self.model_dc = model_dc
```
thunderfish24: One possibility here is to specify `model_dc` directly as a callable in both the "built in" pvlib model cases and the user-defined model case, instead of a string or a callable, respectively. I like this idea, but it may seem more opaque to some users.

thunderfish24:
I should have some new code addressing my above comments coming soon, including a common API for pvwatts, sapm, and various single diode models, as well as assigning functions instead of strings to `self.model_dc` in `PVSystem`.

thunderfish24:
@wholmgren I have decided to try to implement my model in the existing `ModelChain` architecture as a "User-defined model" as described in `modelchain.ipynb#User-defined-models`. More precisely, there are essentially two flavors of implementation: (1) When matched reference device data are available for `F` and `H`, and (2) When the SAPM is used to compute `F` and `H` from `mc.effective_irradiance` and `mc.temps`. However, the `modelchain.ipynb` notebook doesn't seem to provide much detail on how `mc.run_model` actually works for a `ModelChain` using the SAPM, but instead manually assigns some "stub" data:
```
# manually assign data to the attributes that ModelChain.sapm will need.
# for standard workflows, run_model would assign these attributes.
mc.effective_irradiance = pd.Series(1000, index=[pd.Timestamp('20170401 1200-0700')])
mc.temps = pd.DataFrame({'temp_cell': 50, 'temp_module': 50}, index=[pd.Timestamp('20170401 1200-0700')])
```
Is there a "standard workflow" example of `mc.run_model` with with SAPM somewhere that I am missing?

thunderfish24:
@wholmgren Nevermind. I see that workflow example now earlier in `modelchain.ipynb`. My apologies.

thunderfish24:
@wholmgren I pushed up some code with an initial attempt at `ModelChain` integration of my model. In my example in `modelchain.ipynb#Model-wrapper-for-use-with-SAPM-effective-irradiance-and-cell-temperature`, I am getting the odd error: AttributeError: 'ModelChain' object has no attribute 'model_run'.

Would you mind taking a look, both at my integration approach and what might be causing this particular error?

thunderfish24:
@wholmgren Got that sorted (facepalm). Working too late the other night.

thunderfish24:
I now have a working example in `modelchain.ipynb#Model-wrapper-for-use-with-SAPM-effective-irradiance-and-cell-temperature` that uses SAPM to produce `F` and `H`. Working next on the matched reference cell version that uses `F` and `H` directly.

thunderfish24:
I have added working examples in of both types of "wrappers" for the Campanelli et al. model that work with `ModelChain`. Because PV reference devices provide a different description of "weather", I added a `pv_ref_device=False` keyword parameter to the argument list of  `ModelChain.run_model.

I also simplified the model implementation code for this first implementation, and I hit a problem running the SAPM in `pvsystem.ipynb`, which _might_ be related to a recent refactor merged into master (?):

```console
TypeError                                 Traceback (most recent call last)
<ipython-input-22-30449faf83de> in <module>()
      8     am_abs, aoi, module)
      9 
---> 10 sapm_1 = pvlib.pvsystem.sapm(effective_irradiance, temps['temp_cell'], **module)
     11 
     12 sapm_1.plot()

TypeError: sapm() got an unexpected keyword argument 'Vintage'
```

@wholmgren @cwhanse I am requesting a review of this code, because the changes might surprise some people and there might be better ideas out there on how to implement this new type of weather data.

cwhanse:
I could easily be wrong here.  You are passing `**module` to a position argument `module`. I believe that `**module` is expanded to a list of kwargs, which `pvsystem.sapm` interprets as positional arguments, and `Vintage` isn't among them.

Suggest simply passing `module` as the third argument.  `Vintage` is unused.

pvlib/modelchain.py
```diff
@@ -803,7 +803,7 @@ def prepare_inputs(self, times=None, weather=None):
             self.weather['temp_air'] = 20
         return self
 
-    def run_model(self, times=None, weather=None):
+    def run_model(self, times=None, weather=None, pv_ref_device=False):
```
cwhanse: I'm opposed to this new keyword argument.  See comments about `weather` below.

pvlib/pvsystem.py
```diff
@@ -279,7 +280,7 @@ def physicaliam(self, aoi):
 
         return physicaliam(aoi, **kwargs)
 
-    def calcparams_desoto(self, effective_irradiance, temp_cell, **kwargs):
+    def calcparams_desoto(self, effective_irradiance, temp_cell):
```
cwhanse: This is a problem. Method `calcparams_desoto` needs the `**kwargs` because it unpacks these arguments to pass them to function `pvsystem.calcparams_desoto` as positional arguments. 

I don't understand why you are removing `**kwargs` ?

pvlib/pvsystem.py
```diff
@@ -2461,13 +2500,14 @@ def scale_voltage_current_power(data, voltage=1, current=1):
...
 
-    # as written, only works with a DataFrame
-    # could make it work with a dict, but it would be more verbose
+    # Works with a DataFrame or (ordered) dict (e.g., output of singlediode)
     data = data.copy()
-    voltages = ['v_mp', 'v_oc']
-    currents = ['i_mp', 'i_x', 'i_xx', 'i_sc']
-    data[voltages] *= voltage
-    data[currents] *= current
+    data['v_mp'] *= voltage
```
cwhanse: Thank you for these improvements.

pvlib/pvsystem.py
thunderfish24: Line 302 below assigns directly to kwargs, which (as far as I can tell) clobbers any value for the kwargs dict passed to this method.

Actually, the situation appears similar in the sapm method too, but there the passed kwargs dict is simply never used.

pvlib/modelchain.py
thunderfish24: Understood. I'm looking into the alternatives.

thunderfish24:
I'm not sure what caused the `sapm` glitch in `pvsystem.ipynb`. It may be that the notebook was slow to update with my pull of master from upstream. In any event, it seems better now. Gremlins perhaps...

pvlib/pvsystem.py
thunderfish24: To followup further, would you be opposed to forgoing the `_build_kwargs()` call in the `calcparams_desoto` method and using the following (more direct) return statement instead?
```python
return calcparams_desoto(effective_irradiance, temp_cell, **self.module_parameters)
```

thunderfish24:
At the risk of a small bit of scope creep, I pushed up some simplifications to the `calcparams_desoto` and `sapm` method calls. It also looks like the docstring for `sapm` was a bit stale.

pvlib/modelchain.py
thunderfish24: I am open to suggestions on what to call `F`, as well as the symbol. Perhaps something more expressive (but wordy!) like "short-circuit current ratio". Likewise for `H`, but note that the formulation of this model in terms of `H` is more of a numerical convenience (a re-scaling that also obviates the need to explicitly specify the reference temperature), whereas `F` is fundamentally linked to an "observable" short-circuit current rather than a "non-observable" photocurrent.

thunderfish24:
So based on the test failures, the `_build_kwargs` appears to be necessary in `calcparams_desoto`. I have tried to write the new "campanelli" functions in the `pvsystem` module to be "less picky" about receiving arbitrary `kwargs`. I will need to make sure that I'm not missing something here.

pvlib/pvsystem.py
thunderfish24: Removing the `_build_kwargs(...)` step doesn't work given the current setup of `calcparams_desoto`.

thunderfish24:
Travis-CI 3.6 build passed. All other failures appear to be spurious and unrelated to these changes.

pvlib/modelchain.py
```diff
@@ -810,32 +810,101 @@ def run_model(self, times=None, weather=None):
...
+            self.effective_irradiance_model()
+            self.temp_model()
         self.dc_model()
         self.ac_model()
         self.losses_model()
 
         return self
+
+
+def sdm_campanelli(mc):
```
wholmgren: interesting idea to make a function here rather than a ModelChain method. It doesn't follow the pattern, but I could be convinced that all of the wrappers should look like this.

pvlib/modelchain.py
```diff
@@ -810,32 +810,101 @@ def run_model(self, times=None, weather=None):
...
+
+    TODO Details
+    """
+
+    # Compute the effective irradiance ratio F from SAPM effective irradiance
+    F = (1. + alpha_sc * (temps['temp_cell'].values - temp_ref)) * \
+        effective_irradiance.values / irrad_ref
+
+    # Model the effective temperature ratio H from SAPM temp_cell
+    H = (temps['temp_cell'].values + 273.15) / (temp_ref + 273.15)
```
wholmgren: why ``.values``? might need a comment if we decide to keep it.

pvlib/pvsystem.py
wholmgren: we need ``_build_kwargs`` for the function call. But I think we can get rid of the ``kwargs`` in the method signature above. Would need to be documented in API changes.

pvlib/pvsystem.py
```diff
@@ -279,64 +280,50 @@ def physicaliam(self, aoi):
...
-            The diffuse irradiance incident on module.
-
-        temp_cell : Series
-            The cell temperature (degrees C).
-
-        airmass_absolute : Series
-            Absolute airmass.
-
-        aoi : Series
-            Angle of incidence (degrees).
```
wholmgren: thanks for all this related clean up.

pvlib/pvsystem.py
```diff
@@ -574,7 +566,7 @@ def scale_voltage_current_power(self, data):
 
         Returns
         -------
-        scaled_data: DataFrame
+        scaled_data: DataFrame or (ordered) dict
```
wholmgren: good to add ``dict``, but no need to specify ordered. I think we'll be getting rid of OrderedDict as python 3.6+ becomes the standard.

thunderfish24:
@cwhanse Please review my updates. I removed the `pv_reference_device` flag from `ModelChain.run_model`. I now have a single wrapper function, `ModelChain.sdm_campanelli`, that covers all combinations of cases: (1) both `F` and `H` provided in `weather`, and (2)  `F` or `H` (but not both) provided in `weather`, (3) neither `F` nor `H` provided in `weather`. In (2) and (3), the SAPM is used to fill in the missing data. Your recommendations helped me simplify this a bit :)!

Note that there is still logic required in `run_model` to sort out the possible cases. I cannot think of a way around this given the existing architecture, but I'm happy to mark things as "provisional" in the documentation. If you approve, then I will complete the documentation and write tests.

I also improved the accuracy/consistency of some of the docstrings in the `pvsystem` module.

pvlib/pvsystem.py
wholmgren: we've used ``dict-like``  elsewhere and it applies here as well. so that's another option.

pvlib/pvsystem.py
```diff
@@ -1172,6 +1164,51 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+    """
+    Compute the Campanelli et al. model
+
+    TODO Details
+    """
+
+    return singlediode(*get_coeffs_campanelli(F, H, **kwargs))
+
+
+def get_coeffs_campanelli(F, H, **kwargs):
```
wholmgren: Why use ``**kwargs`` here? Would we not want to use positional arguments for everything that's required and does not have a default, and use named keyword arguments for everything that has a default?

pvlib/modelchain.py
```diff
@@ -810,32 +810,101 @@ def run_model(self, times=None, weather=None):
...
-        self.effective_irradiance_model()
-        self.temp_model()
+        if weather is not None and 'F' in weather:
+            # Use PV reference device weather information for F
+            self.F = weather['F']
+        else:
+            # Clear any stale value
+            if hasattr(self, 'F'):
+                print('F cleared')
+                del self.F
```
wholmgren: maybe this is debugging code, but we don't do this for any other modelchain attributes and I think we should avoid adding this layer of complexity in general.

thunderfish24: Yep. I forgot to remove the print statements, but in general these do need to be cleared to prevent subtle bugs that the user could run into when re-running a model with changing weather types, because I think it is generally valuable to "record" the `F` and `H` values used in the `ModelChain` object, which sometimes come from the "non-trivial" SAPM model calculations. (This is not too different than other "recorded" outputs, such as effective_irradiance.)

pvlib/pvsystem.py
thunderfish24: Unless someone can show that this is less performant, I think it is less fragile and thus preferable to `_build_kwargs`. Of course, the required `kwargs` must be well-documented in the "base" function. Given the considerable level of indirect inference and "kwargs packing" that occurs throughout the current architecture, I don't find this any more obscure than the status quo. I realize some might disagree here and prefer positional arguments, but I personally really prefer removing order dependence whenever possible in function arguments (esp. for model parameters like `IL` as opposed to control variables like `effective_irradiance`). On a related note, Python3 has [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/) that don't have to specify a default value :).

pvlib/modelchain.py
thunderfish24: IIRC, it is to ensure the "raw" numpy data structure in order to satisfy the (simplified) API of the low level function in the `pvsystem` module. (Of course, I haven't documented that yet.)

pvlib/modelchain.py
thunderfish24: Given the current architecture where the `PVSystem` wrappers are not ultimately responsible for determining the actual computations in a performance simulation, this is the "shortest path" I could come up with ;). It also pushes the idea that the "user" API could be the same as the "developer" API.

pvlib/modelchain.py
```diff
@@ -810,32 +810,101 @@ def run_model(self, times=None, weather=None):
...
+                                 mc.system.module_parameters['irrad_ref'],
+                                 mc.system.module_parameters['temp_ref'])
+        if not hasattr(mc, 'F'):
+            mc.F = F
+        if not hasattr(mc, 'H'):
+            mc.H = H
+
+    # Calculate the dc power and assign it to mc.dc
+    mc.dc = pvsystem.sdm_campanelli(mc.F, mc.H, **mc.system.module_parameters)
+
```
thunderfish24: Oops. I have forgotten to scale the output for `modules_per_string` and `strings_per_inverter`.

pvlib/pvsystem.py
mikofski: Python makes no distinction. The user can provide keyword arguments or positional arguments whenever they want unless the function signature contains splat operators like `*a, **kw`, which are only used to match inherited signatures, or to collect unknown user supplied extra arguments.

`import this` says explicit is better than implicit. IMHO unless you are collecting args for inherited signatures or slurring the user to supply their own keyword arguments, this usage should be avoided, and I believe most linters would warn

This

pvlib/modelchain.py
```diff
@@ -810,32 +810,102 @@ def run_model(self, times=None, weather=None):
...
+        if weather is not None and 'H' in weather:
+            # Use PV reference device weather information for H
+            self.H = weather['H']
+        else:
+            # Clear any stale value
+            if hasattr(self, 'H'):
+                del self.H
+        if not hasattr(self, 'F') and not hasattr(self, 'H') or \
+           hasattr(self, 'F') and not hasattr(self, 'H') or \
+           not hasattr(self, 'F') and hasattr(self, 'H'):
```
thunderfish24: Can anyone figure out a simplified equivalent logic?

pvlib/pvsystem.py
thunderfish24: K. So I am going to look up if there is a Python2 backport of keyword-only arguments that don't require a default value. I think this would be a fair compromise to my preference to avoid [the complexity of order](https://www.youtube.com/watch?v=rI8tNMsozo0) while also being uber-explicit in the low-level functions.

The only downside that I see here is potentially long keyword names, but again I think explicitness and "self-documenting" function readability trumps lengthiness, and think my names are short but sufficiently expressive and will of course be fully documented.

pvlib/modelchain.py
adriesse: Yes, something more wordy would be good.  Ideas:
effective_current_ratio, i_sc_ratio, i_sc_norm
temp_cell_abs_norm, temp_cell_norm

cwhanse: `effective_irradiance_from_isc` perhaps.  `temp_cell_normalized` is ok with me.

pvlib/pvsystem.py
cwhanse: I realize that @thunderfish24 and others will think this a step backwards, but I'd prefer `calcparams_campanelli` to `get_coeffs_campanelli` to keep the function name pattern parallel to the other version of single diode models. When we want to change this API pattern we can do it for all single diode model coefficient functions.

pvlib/modelchain.py
cwhanse: Can `sdm_campanellie_sapm` be built into `sdm_campanelli`?  The `_sapm` variation only fills in `F` and `H` when they are not supplied as weather data.

pvlib/modelchain.py
thunderfish24: A whole set of functions do not need to be run in some cases, and I think they fail in the case where there is no data for them to run.

pvlib/pvsystem.py
thunderfish24: No prob. Consistency is always an important consideration.

pvlib/modelchain.py
cwhanse: Fair points, but I think `prepare_inputs` should be protection against missing values - you can even pass `weather=None` and it will fill in with a clear-sky model. I realize that `sdm_campanelli` doesn't need anything upstream of `dc_model` but I'd prefer to see the model chain sequence maintained.

My instinct is that an `if...then` structure to skip model chain steps for a choice of a downstream model will be a problem to maintain, and will become difficult to manage as more models are added.

I would prefer a separate pull request to extend the `ModelChain` functions to handle measured `effective_irradiance` and `temp_module` as inputs in a general manner.  I think this pull request should conclude when `sdm_campanelli` works as a method of the current `ModelChain`, with supporting functions consistent with the current patterns.  I'm OK with less efficient code in favor of consistency.  Redesign of `ModelChain` and generalizing treatment of input options can be opened as their own issues.

pvlib/pvsystem.py
wholmgren: I agree with @cwhanse that ``calcparams_campanelli`` is preferable here. My intention (perhaps poorly communicated elsewhere) is for all ``get_*`` functions/methods to be wrappers for other functions/methods. See #427 for more on that pattern. This PR is big enough already, so let's not add a new wrapper here. 

Re keyword-only arguments... I am not aware of a way to do that in Python 2 without some miserable hacks. Users can still call positional arguments as keyword arguments in any order you'd like, though I think it's good to be consistent within pvlib itself.

pvlib/modelchain.py
```diff
@@ -810,32 +810,102 @@ def run_model(self, times=None, weather=None):
...
+
+    # Calculate the system-scaled dc power and assign it to mc.dc
+    mc.dc = mc.system.scale_voltage_current_power(
+        pvsystem.sdm_campanelli(mc.F, mc.H, **mc.system.module_parameters))
+
+    # Return ModelChain object to enable method chaining
+    return mc
+
+
+def get_F_H_from_sapm(effective_irradiance, temps, alpha_sc, irrad_ref,
```
wholmgren: same issue with ``get_`` prefix here as in comment on ``get_calcparams_campanelli``. Why did you put this function in ``modelchain.py`` rather than ``pvsystem.py``? Seems to me that it's more general purpose than the ``modelchain`` api.

pvlib/pvsystem.py
thunderfish24: I agree that Python2 hacks look really bad, so I'll probably have to give up and simply used named positional arguments :(.

pvlib/modelchain.py
wholmgren: I was thinking that it might make sense to move this function to the *module* ``pvsystem``, but not necessarily into the *class* ``pvsystem.PVSystem``. So, function name changes aside, we'd have ``pvsystem.get_F_H_from_sapm``. One could optionally implement ``pvsystem.PVSystem.get_F_H_from_sapm`` that dispatches work to ``pvsystem.get_F_H_from_sapm``, but that may or may not be of value.

pvlib/modelchain.py
wholmgren: Is this a reasonable compromise?

```python 
if self.dc_model == sdm_campanelli and 'F' in weather.columns and 'H' in weather.columns:
    mc.F = weather['F']
    mc.H = weather['H']
else:
    self.pepare_inputs ... self.temp_model()
self.dc_model()
...
```

cwhanse: This makes sense to me. At present the only model that uses the `F` and `H` inputs is Campanelli, but it is possible that `F` is more generally useful in place of `effective_irradiance` calculated from weather in, e.g., SAPM.

thunderfish24: @cwhanse I'm pretty sure that `F` (and `H`) may be generalized to "all" the single diode models, the only difference being how they propagate into the "auxiliary equations". The derivation is analogous for double-diode models as well. One of my (longer term) goals for pv-fit.com is to fit calibration data to multiple models at once and provide the user with a model "ranking" based on a goodness of fit metric.

Would you be able to refer me to where in the literature the SAPM definition of "effective irradiance" is derived, esp. with respect to the conventional meteorological datasets?

cwhanse: @thunderfish24 here: http://prod.sandia.gov/techlib/access-control.cgi/2004/043535.pdf

It's poorly explained. But, in a nutshell, `effective irradiance` is defined in the text as "the fraction of the total solar irradiance incident on the module to which the cells inside actually respond", i.e., define 'respond'. `Ee` is calculated from weather by Eq. 1 (to get Isc) then Eq. 7 (ratio of calculated Isc to a reference value). You can do the algebra and get an expression for `Ee` in terms of beam and diffuse on the array's surface. We often substitute measurements from an appropriate reference cell for the calculated `Ee` value.

AdamRJensen:
> I'm updating my `pvfit` codebase to finalize my fitting API so that it matches exactly with the symbols in this PR. I will be back here soon I hope.

@markcampanelli It seems that this PR has grown stale. Could we close this PR?

markcampanelli:
Closing due to inattention.

Issue #483, opened 2018-06-16
Add a public constants module and include scipy as base requirement for pvlib


markcampanelli:
pvlib should have a public constants python module so that all the different models share the same constants and so that users can see what constants pvlib is using and match them if/when needed.

For example, the data values in a gold dataset for the single-diode model depend on the choice of constants, see [this PR](https://github.com/pvlib/pvlib-python/pull/426). Note that `scipy.constants` contains many useful physical constants traceable to [CODATA](http://www.codata.org/committees-and-groups/fundamental-physical-constants), as well as temperature conversion functions and the like.

We also have to make sure we distinguish between the same constant with different units, or agree to use only one version of each constant throughout the codebase.

I think we should wrap the `scipy.constants` and make `scipy` an import requirement for the "base" installation of pvlib. (This would also remove a lot of the test decorations and import checks around `scipy`, and `scipy.special.lambertw` is already needed for `pvsystem.v_from_i` and `pvsystem.i_from_v`.) Simple tests should be written to catch and audit when constants have changed.

wholmgren:
Thanks for making this issue @thunderfish24 

We can make a more informed decision if we have a list of constants that we use in pvlib (or are likely to use in the future) and a list of the functions that require scipy. 

What is the point of a ``pvlib.constants`` module if we add scipy as a base dependency? Why not ``from scipy.constants import Boltzmann`` where needed? Direct imports from scipy seem more straightforward to me.

I am not inclined to keep track of constants values in scipy. I think that is the user's job when deciding what version of scipy to install.

Can anyone comment on easy of scipy installations on Windows and microcomputers in 2018?

mikofski:
[Wheels are available for Windows at the CheeseShop.](https://pypi.org/project/scipy/) I haven't installed them so perhaps I shouldn't comment. I hesitate to ask on the [SciPy Dev mail list](https://mail.python.org/mailman/listinfo/scipy-dev). But my guess is that most pvlib users are probably using Anaconda, and there's always [Christoph Gohlke](https://www.lfd.uci.edu/~gohlke/pythonlibs/#scipy).

wholmgren:
Thanks @mikofski. I see the wheels are only ~30 MB, which is not so bad (and comparable to pvlib itself), so probably ok for microcomputers. 

My biggest concern is making sure that a new user that knows some kind of integrated scientific programming environment (matlab, mathematica, igor, idl, etc) but little python and command line can install pvlib with as little trouble as possible.

[...]

PR #486, opened 2018-06-20
Implement calcparams_pvsyst and tests

### LINKED ISSUE #470: Implement PVsyst single diode model

cwhanse:

 - [x] Closes #470 

Implement `calcparams_pvsyst` for use with `singlediode`

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.6.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.6.0.rst
@@ -13,6 +13,7 @@ Enhancements
 * Add sea surface albedo in irradiance.py (:issue:`458`)
 * Implement first_solar_spectral_loss in modelchain.py (:issue:`359`)
 * Clarify arguments Egref and dEgdT for calcparams_desoto (:issue:`462`)
+* Add pvsystem.calcparams_pvsyst to compute values for the single diode equation using the PVsyst v6 model.
 
 
 Bug fixes
--- a/docs/sphinx/source/whatsnew/v0.6.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.6.0.rst
@@ -13,7 +13,7 @@ Enhancements
 * Add sea surface albedo in irradiance.py (:issue:`458`)
 * Implement first_solar_spectral_loss in modelchain.py (:issue:`359`)
 * Clarify arguments Egref and dEgdT for calcparams_desoto (:issue:`462`)
-* Add pvsystem.calcparams_pvsyst to compute values for the single diode equation using the PVsyst v6 model.
+* Add pvsystem.calcparams_pvsyst to compute values for the single diode equation using the PVsyst v6 model (:issue:'470')
 
 
 Bug fixes
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -307,6 +307,36 @@ class PVSystem(object):
         
         return calcparams_desoto(effective_irradiance, temp_cell, **kwargs)
 
+    def calcparams_pvsyst(self, effective_irradiance, temp_cell, **kwargs):
+        """
+        Use the :py:func:`calcparams_pvsyst` function, the input
+        parameters and ``self.module_parameters`` to calculate the
+        module currents and resistances.
+
+        Parameters
+        ----------
+        effective_irradiance : numeric
+            The irradiance (W/m2) that is converted to photocurrent.
+
+        temp_cell : float or Series
+            The average cell temperature of cells within a module in C.
+
+        **kwargs
+            See pvsystem.calcparams_desoto for details
+
+        Returns
+        -------
+        See pvsystem.calcparams_pvsyst for details
+        """
+
+        kwargs = _build_kwargs(['gamma_ref', 'mugamma', 'I_L_ref', 'I_o_ref', 
+                                'R_sh_ref', 'R_sh_0', 'R_sh_exp', 
+                                'R_s', 'alpha_sc', 'EgRef',
+                                'cells_in_series'],
+                                self.module_parameters)
+        
+        return calcparams_pvsyst(effective_irradiance, temp_cell, **kwargs)
+
     def sapm(self, effective_irradiance, temp_cell, **kwargs):
         """
         Use the :py:func:`sapm` function, the input parameters,
@@ -1167,6 +1197,147 @@ def calcparams_desoto(effective_irradiance, temp_cell,
     # irradiance and not applying a spectral loss modifier, i.e., 
     # spectral_modifier = 1.0.
     Rsh = R_sh_ref * (irrad_ref / effective_irradiance)
+    Rs = R_s
+
+    return IL, I0, Rs, Rsh, nNsVth
+
+
+def calcparams_pvsyst(effective_irradiance, temp_cell,
+                      alpha_sc, gamma_ref, mugamma,
+                      I_L_ref, I_o_ref, 
+                      R_sh_ref, R_sh_0, R_s, 
+                      cells_in_series,
+                      R_sh_exp=5.5,
+                      EgRef=1.121,
+                      irrad_ref=1000, temp_ref=25):
+    '''
+    Calculates five parameter values for the single diode equation at 
+    effective irradiance and cell temperature using the PVsyst v6 
+    model described in [1,2,3]. The five values returned by calcparams_pvsyst
+    can be used by singlediode to calculate an IV curve.
+
+    Parameters
+    ----------
+    effective_irradiance : numeric
+        The irradiance (W/m2) that is converted to photocurrent.
+
+    temp_cell : numeric
+        The average cell temperature of cells within a module in C.
+
+    alpha_sc : float
+        The short-circuit current temperature coefficient of the
+        module in units of A/C.
+
+    gamma_ref : float
+        The diode ideality factor
+
+    I_L_ref : float
+        The light-generated current (or photocurrent) at reference conditions,
+        in amperes.
+        
+    I_o_ref : float
+        The dark or diode reverse saturation current at reference conditions,
+        in amperes.
+        
+    R_sh_ref : float
+        The shunt resistance at reference conditions, in ohms.
+        
+    R_sh_0 : float
+        The shunt resistance at zero irradiance conditions, in ohms.
+        
+    R_s : float
+        The series resistance at reference conditions, in ohms.
+
+    cells_in_series : integer
+        The number of cells connected in series.
+
+    R_sh_exp : float
+        The exponent in the equation for shunt resistance, unitless. Defaults
+        to 5.5.
+
+    EgRef : float
+        The energy bandgap at reference temperature in units of eV.
+        1.121 eV for crystalline silicon. EgRef must be >0.
+
+    irrad_ref : float (optional, default=1000)
+        Reference irradiance in W/m^2.
+
+    temp_ref : float (optional, default=25)
+        Reference cell temperature in C.
+
+    Returns
+    -------
+    Tuple of the following results:
+
+    photocurrent : numeric
+        Light-generated current in amperes at irradiance=S and
+        cell temperature=Tcell.
+
+    saturation_current : numeric
+        Diode saturation curent in amperes at irradiance
+        S and cell temperature Tcell.
+
+    resistance_series : float
+        Series resistance in ohms at irradiance S and cell temperature
+        Tcell.
+
+    resistance_shunt : numeric
+        Shunt resistance in ohms at irradiance S and cell temperature
+        Tcell.
+
+    nNsVth : numeric
+        Modified diode ideality factor at irradiance S and cell
+        temperature Tcell. Note that in source [1] nNsVth = a (equation
+        2). nNsVth is the product of the diode ideality factor
+        (n), the number of series-connected cells in the module (Ns),
+        and the thermal voltage of a cell in the module (Vth) at a cell
+        temperature of Tcell.
+
+    References
+    ----------
+    [1] K. Sauer, T. Roessler, C. W. Hansen, Modeling the Irradiance and 
+     Temperature Dependence of Photovoltaic Modules in PVsyst, 
+     IEEE Journal of Photovoltaics v5(1), January 2015.
+
+    [2] A. Mermoud, PV modules modelling, Presentation at the 2nd PV
+     Performance Modeling Workshop, Santa Clara, CA, May 2013
+
+    [3] A. Mermoud, T. Lejeune, Performance Assessment of a Simulation Model
+     for PV modules of any available technology, 25th European Photovoltaic
+     Solar Energy Conference, Valencia, Spain, Sept. 2010
+
+    See Also
+    --------
+    calcparams_desoto
+    singlediode
+    
+    '''
+
+    # Boltzmann constant in J/K
+    k = 1.3806488e-23
+
+    # elementary charge in coulomb
+    q = 1.6021766e-19
+    
+    # reference temperature
+    Tref_K = temp_ref + 273.15
+    Tcell_K = temp_cell + 273.15
+
+    gamma = gamma_ref + mugamma * (Tcell_K - Tref_K)
+    nNsVth = gamma * k / q * cells_in_series
+
+    IL = effective_irradiance / irrad_ref * \
+              (I_L_ref + alpha_sc * (Tcell_K - Tref_K))
+
+    I0 = I_o_ref * ((Tcell_K / Tref_K) ** 3) * \
+          (np.exp((q * EgRef) / (k * gamma) * (1 / Tref_K - 1 / Tcell_K)))
+
+    Rsh_tmp = (R_sh_ref - R_sh_0 * np.exp(-R_sh_exp)) / (1.0 - np.exp(-R_sh_exp))
+    Rsh_base = np.maximum(0.0, Rsh_tmp)
+
+    Rsh = Rsh_base + (R_sh_0 - Rsh_base) * \
+              np.exp(-R_sh_exp * effective_irradiance / irrad_ref)
+    
     Rs = R_s
 
     return IL, I0, Rs, Rsh, nNsVth
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1041,28 +1041,21 @@ def calcparams_desoto(effective_irradiance, temp_cell,
     Tuple of the following results:
 
     photocurrent : numeric
-        Light-generated current in amperes at irradiance=S and
-        cell temperature=Tcell.
+        Light-generated current in amperes
 
     saturation_current : numeric
-        Diode saturation curent in amperes at irradiance
-        S and cell temperature Tcell.
+        Diode saturation curent in amperes
 
     resistance_series : float
-        Series resistance in ohms at irradiance S and cell temperature
-        Tcell.
+        Series resistance in ohms
 
     resistance_shunt : numeric
-        Shunt resistance in ohms at irradiance S and cell temperature
-        Tcell.
+        Shunt resistance in ohms
 
     nNsVth : numeric
-        Modified diode ideality factor at irradiance S and cell
-        temperature Tcell. Note that in source [1] nNsVth = a (equation
-        2). nNsVth is the product of the usual diode ideality factor
-        (n), the number of series-connected cells in the module (Ns),
-        and the thermal voltage of a cell in the module (Vth) at a cell
-        temperature of Tcell.
+        The product of the usual diode ideality factor (n, unitless), 
+        number of cells in series (Ns), and cell thermal voltage at 
+        specified effective irradiance and cell temperature.
 
     References
     ----------
@@ -1081,8 +1074,6 @@ def calcparams_desoto(effective_irradiance, temp_cell,
 
     See Also
     --------
-    sapm
-    sapm_celltemp
     singlediode
     retrieve_sam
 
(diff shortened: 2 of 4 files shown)


pvlib/pvsystem.py
```diff
@@ -1172,6 +1202,147 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+    temp_cell : numeric
+        The average cell temperature of cells within a module in C.
+
+    alpha_sc : float
+        The short-circuit current temperature coefficient of the
+        module in units of A/C.
+
+    gamma_ref : float
+        The diode ideality factor
+
```
wholmgren: missing mugamma documentation here. mugamma or mu_gamma?

pvlib/pvsystem.py
```diff
@@ -307,6 +307,36 @@ def calcparams_desoto(self, effective_irradiance, temp_cell, **kwargs):
...
+        Parameters
+        ----------
+        effective_irradiance : numeric
+            The irradiance (W/m2) that is converted to photocurrent.
+
+        temp_cell : float or Series
+            The average cell temperature of cells within a module in C.
+
+        **kwargs
+            See pvsystem.calcparams_desoto for details
```
wholmgren: pvsyst, not desoto.

cwhanse:
@mikofski would appreciate your review of `calcparams_pvsyst`

pvlib/pvsystem.py
```diff
@@ -1172,6 +1202,150 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+    R_sh_ref : float
+        The shunt resistance at reference conditions, in ohms.
+        
+    R_sh_0 : float
+        The shunt resistance at zero irradiance conditions, in ohms.
+        
+    R_s : float
+        The series resistance at reference conditions, in ohms.
+
+    cells_in_series : integer
```
thunderfish24: `Ns` would seem more consistent with the desoto model and the `nNsVth` variable.

pvlib/pvsystem.py
```diff
@@ -1172,6 +1202,150 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+        in amperes.
+        
+    I_o_ref : float
+        The dark or diode reverse saturation current at reference conditions,
+        in amperes.
+        
+    R_sh_ref : float
+        The shunt resistance at reference conditions, in ohms.
+        
+    R_sh_0 : float
```
thunderfish24: `R_sh_ref` seems more consistent with other variable names.

pvlib/pvsystem.py
```diff
@@ -1172,6 +1202,150 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+        The shunt resistance at zero irradiance conditions, in ohms.
+        
+    R_s : float
```
thunderfish24: R_s_ref seems more consistent with other variables at reference conditions.

pvlib/pvsystem.py
```diff
@@ -1172,6 +1202,150 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+        The average cell temperature of cells within a module in C.
+
+    alpha_sc : float
+        The short-circuit current temperature coefficient of the
+        module in units of A/C.
+
+    gamma_ref : float
+        The diode ideality factor
+
+    mugamma : float
```
thunderfish24: Perhaps `mu_gamma`?

pvlib/pvsystem.py
```diff
@@ -1172,6 +1202,150 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+
+    temp_ref : float (optional, default=25)
+        Reference cell temperature in C.
+
+    Returns
+    -------
+    Tuple of the following results:
+
+    photocurrent : numeric
+        Light-generated current in amperes at irradiance=S and
```
thunderfish24: Not sure about `irradiance=S` here and below.

thunderfish24:
@cwhanse Are you not planning to integrate the pvsyst model into `ModelChain` in this PR?

mikofski:
Baby steps @thunderfish24. Will this PR be incomplete without it?


* Haste makes waste.
* Bird in the hand is worth two in the bush
* Don't put all your eggs in one basket

thunderfish24:
@mikofski This was by no means a criticism. Rather, I was just wondering if complete integration was expected right away. LIkewise, #478 is much more easily implemented if I can worry about the `ModelChain` integration as a followup step. 

@cwhanse I do propose this PR adds a usage example in a subsection of `docs/tutorials/pvsystem.ipynb`.

mikofski:
I'm so sorry @thunderfish24 my comment was snarky. It was a fair question.

thunderfish24:
@mikofski No worries mate! My question could have been phrased better, and it's often easy to misinterpret typings when we are not face-to-face.

cwhanse:
@mikofski @thunderfish24 yes I'll open a new Issue and PR for extending `ModelChain` for new single diode models.

pvlib/pvsystem.py
cwhanse: added.  changed to `mu_gamma`

pvlib/pvsystem.py
```diff
@@ -1172,6 +1202,147 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+        cell temperature=Tcell.
+
```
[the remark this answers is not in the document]
cwhanse: I edited the output description (and in `pvsystem.calcparams_desoto`) to remove the phrase 'at irradiance....=Tcell' I think the context is clear.

pvlib/pvsystem.py
```diff
@@ -1172,6 +1202,147 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+    # Boltzmann constant in J/K
+    k = 1.3806488e-23
+
+    # elementary charge in coulomb
+    q = 1.6021766e-19
+    
+    # reference temperature
+    Tref_K = temp_ref + 273.15
+    Tcell_K = temp_cell + 273.15
+
```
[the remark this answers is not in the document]
cwhanse: these are local variables, so I think its OK that they are different in style.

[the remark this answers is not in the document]
cwhanse: So I changed them as you suggested :)

pvlib/pvsystem.py
cwhanse: I agree but we use `cells_in_series` as the key for the SAPM parameters, and some have a preference for descriptive variable names.

pvlib/pvsystem.py
cwhanse: I agree. But here we run into an unfortunate linkage to the SAM database for the CEC model parameters.  `R_s` is the heading in that file, which is used to define the keys for the module parameter dict. Although in my opinion `R_s` for PVsyst is a different parameter than `R_s` for De Soto or CEC (because parameter values are determined jointly, they are intrinsically linked and can't be regarded as correct for a different model) I do think it's useful to maintain a common nomenclature.

I think changing `R_s` to `R_s_ref` will involve some work in `pvsystem.retrieve_sam` so I'm not going to take that on in this PR.

pvlib/pvsystem.py
cwhanse: yes

pvlib/pvsystem.py
cwhanse: nuked

docs/sphinx/source/whatsnew/v0.6.0.rst
```diff
@@ -13,6 +13,7 @@ Enhancements
 * Add sea surface albedo in irradiance.py (:issue:`458`)
 * Implement first_solar_spectral_loss in modelchain.py (:issue:`359`)
 * Clarify arguments Egref and dEgdT for calcparams_desoto (:issue:`462`)
+* Add pvsystem.calcparams_pvsyst to compute values for the single diode equation using the PVsyst v6 model.
```
wholmgren: (:issue:`470`)

pvlib/pvsystem.py
wholmgren: it appears that the change was not applied here.

pvlib/test/test_pvsystem.py
```diff
@@ -174,6 +174,23 @@ def cec_module_params(sam_data):
     return module_parameters
 
 
+@pytest.fixture(scope="session")
+def pvsyst_module_params():
+    module_parameters = {}
+    module_parameters['gamma_ref'] = 1.05
+    module_parameters['mugamma'] = 0.001
```
wholmgren: test fixture and the two tests below need mugamma --> mu_gamma

cwhanse:
I changed a setting in Spyder, hopefully that improves the white space issue.  Test failures are for `read_tmy3`

pvlib/pvsystem.py
```diff
@@ -307,6 +307,36 @@ def calcparams_desoto(self, effective_irradiance, temp_cell, **kwargs):
...
+        Parameters
+        ----------
+        effective_irradiance : numeric
+            The irradiance (W/m2) that is converted to photocurrent.
+
+        temp_cell : float or Series
+            The average cell temperature of cells within a module in C.
+
+        **kwargs
+            See pvsystem.calcparams_pvsyst for details
```
wholmgren: and delete ``**kwargs`` from the docstring.

wholmgren:
I think this should be merged after resolving the minor issues around ``**kwargs``

pvlib/pvsystem.py
```diff
@@ -307,6 +307,36 @@ def calcparams_desoto(self, effective_irradiance, temp_cell, **kwargs):
         
         return calcparams_desoto(effective_irradiance, temp_cell, **kwargs)
 
+    def calcparams_pvsyst(self, effective_irradiance, temp_cell, **kwargs):
```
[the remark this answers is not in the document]
cwhanse: done

pvlib/pvsystem.py
cwhanse: done

pvlib/pvsystem.py
```diff
@@ -1011,28 +1040,21 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
 
     saturation_current : numeric
-        Diode saturation curent in amperes at irradiance
-        S and cell temperature Tcell.
+        Diode saturation curent in amperes
 
     resistance_series : float
-        Series resistance in ohms at irradiance S and cell temperature
-        Tcell.
+        Series resistance in ohms
```
thunderfish24: While passing through this territory, seems like this should be changed to "numeric" from "float".

pvlib/pvsystem.py
```diff
@@ -302,11 +302,40 @@ def calcparams_desoto(self, effective_irradiance, temp_cell, **kwargs):
...
+        """
+
+        kwargs = _build_kwargs(['gamma_ref', 'mu_gamma', 'I_L_ref', 'I_o_ref',
+                                'R_sh_ref', 'R_sh_0', 'R_sh_exp',
+                                'R_s', 'alpha_sc', 'EgRef',
+                                'irrad_ref', 'temp_ref',
+                                'cells_in_series'],
+                                self.module_parameters)
+
+        return calcparams_pvsyst(effective_irradiance, temp_cell, **kwargs)
```
thunderfish24: This could be done in a separate issue (see #476), but I think the `PVSystem` methods should scale the outputs to the `modules_per_string` and `strings_per_inverter`.

pvlib/pvsystem.py
```diff
@@ -1164,14 +1184,151 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+
+    See Also
+    --------
+    calcparams_desoto
+    singlediode
+
+    '''
+
+    # Boltzmann constant in J/K
+    k = 1.3806488e-23
```
thunderfish24: Appears to differ a bit from `scipy.constants.Boltzmann`, which is `1.38064852e-23`

pvlib/pvsystem.py
```diff
@@ -1164,14 +1184,151 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+    nNsVth = gamma * k / q * cells_in_series * Tcell_K
+
+    IL = effective_irradiance / irrad_ref * \
+              (I_L_ref + alpha_sc * (Tcell_K - Tref_K))
+
+    I0 = I_o_ref * ((Tcell_K / Tref_K) ** 3) * \
+          (np.exp((q * EgRef) / (k * gamma) * (1 / Tref_K - 1 / Tcell_K)))
+
+    Rsh_tmp = (R_sh_ref - R_sh_0 * np.exp(-R_sh_exp)) / (1.0 - np.exp(-R_sh_exp))
+    Rsh_base = np.maximum(0.0, Rsh_tmp)
```
thunderfish24: In looking over http://files.pvsyst.com/help/pvmodule_rshexp.htm, it is not clear to me why `Rsh_base` is not simply `R_sh_ref`. What am I missing?

pvlib/pvsystem.py
```diff
@@ -1164,14 +1184,151 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+
+    # reference temperature
+    Tref_K = temp_ref + 273.15
+    Tcell_K = temp_cell + 273.15
+
+    gamma = gamma_ref + mu_gamma * (Tcell_K - Tref_K)
+    nNsVth = gamma * k / q * cells_in_series * Tcell_K
+
+    IL = effective_irradiance / irrad_ref * \
+              (I_L_ref + alpha_sc * (Tcell_K - Tref_K))
```
thunderfish24: Optimization possibility: `Tcell_K - Tref_K` is computed twice.

pvlib/pvsystem.py
```diff
@@ -1164,14 +1184,151 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+
+    I0 = I_o_ref * ((Tcell_K / Tref_K) ** 3) * \
+          (np.exp((q * EgRef) / (k * gamma) * (1 / Tref_K - 1 / Tcell_K)))
+
+    Rsh_tmp = (R_sh_ref - R_sh_0 * np.exp(-R_sh_exp)) / (1.0 - np.exp(-R_sh_exp))
+    Rsh_base = np.maximum(0.0, Rsh_tmp)
+
+    Rsh = Rsh_base + (R_sh_0 - Rsh_base) * \
+              np.exp(-R_sh_exp * effective_irradiance / irrad_ref)
```
thunderfish24: Optimization possibility: `effective_irradiance / irrad_ref` is computed twice.

pvlib/test/test_pvsystem.py
```diff
@@ -386,6 +403,35 @@ def test_calcparams_desoto(cec_module_params):
...
+                                  EgRef=pvsyst_module_params['EgRef'])
+
+    assert_series_equal(np.round(IL, 3), pd.Series([0.0, 4.8200], index=times))
+    assert_series_equal(np.round(I0, 3),
+                        pd.Series([0.0, 1.47e-7], index=times))
+    assert_allclose(Rs, 0.500)
+    assert_series_equal(np.round(Rsh, 3),
+                        pd.Series([1000.0, 305.757], index=times))
+    assert_series_equal(np.round(nNsVth, 4),
+                        pd.Series([1.6186, 1.7961], index=times))
```
thunderfish24: My preference is to compute the expected values directly in the test. I'm curious if you have particular reasons not to do that?

pvlib/test/test_pvsystem.py
```diff
@@ -386,6 +403,35 @@ def test_calcparams_desoto(cec_module_params):
     assert_allclose(nNsVth, 0.473)
 
 
+def test_calcparams_pvsyst(pvsyst_module_params):
```
thunderfish24: Not critical, but you might consider a simple "API" test that passes the five computed coefficients into `singlediode` just to make sure the computation runs. This makes sure that pvsyst gets updated if the API changes.

pvlib/pvsystem.py
cwhanse: We've chosen not to scale inside `dc_model` functions to offer flexibility, e.g., model a system with strings of different lengths, to implement module-to-module mismatch models, etc. Granted some of these capabilities are not presently in pvlib.

pvlib/pvsystem.py
cwhanse: Cut and paste inheritance?  I don't think there is a difference here. `numeric` would admit `long`, `complex` and `int` types as well as `float`.  Probably should change to `float` unless there's a reason otherwise.

pvlib/pvsystem.py
cwhanse: Good catch.  That's a typo.

pvlib/pvsystem.py
cwhanse: You're not, PVsyst documentation is missing that definition of `Rsh_base`. Ken Sauer and Thomas Roessler are my sources, they got it verbally from PVsyst.

pvlib/test/test_pvsystem.py
cwhanse: I don't other than it was easy (for me) to compute the values in MATLAB and hardcode them here.  My way is certainly less transparent.

pvlib/pvsystem.py
thunderfish24: I'm good with float instead of numeric, because that seems to be most accurate.

pvlib/pvsystem.py
```diff
@@ -1164,14 +1184,151 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+                      R_sh_exp=5.5,
+                      EgRef=1.121,
+                      irrad_ref=1000, temp_ref=25):
+    '''
+    Calculates five parameter values for the single diode equation at
+    effective irradiance and cell temperature using the PVsyst v6
+    model described in [1,2,3]. The five values returned by calcparams_pvsyst
+    can be used by singlediode to calculate an IV curve.
+
+    Parameters
```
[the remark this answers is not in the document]
cwhanse: I'm pretty sure the function will work on array or Series inputs. Is the issue the type description `float` implies a singleton?

We aren't using `numeric` or `float` consistently in the comments.  For me to fix that, requires that I learn the distinction.  New issue, I think.

pvlib/test/test_pvsystem.py
cwhanse: Fair point. I'm not going to put a new test in here because I don't clearly see how to separate issues being tested, and to avoid redundant tests. The current `test_calcparams_pvsyst` verifies correct output of `pvsystem.calcparams_pvsyst`. Adding a call to `singlediode` within `test_calcparams_pvsyst` adds a new purpose (API check) and I prefer to keep tests to a single purpose. Adding a new test that calls `calcparams_pvsyst` and then `singlediode` would duplicate the test of `PVSystem.calcparams_pvsyst` method.

Do you have an example in the `campanelli` pull request?

pvlib/pvsystem.py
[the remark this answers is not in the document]
wholmgren: I've used ``numeric`` for input that can be scalar, ``np.array``, or ``pd.Series``. ``float`` is specific to scalar. Whether or not this is a good idea could be discussed in a separate issue, if necessary.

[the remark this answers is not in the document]
wholmgren: for more context, note that some functions require "array-like" (``np.array`` or ``pd.Series``) and others require ``pd.Series``.

pvlib/pvsystem.py
wholmgren: ref #483.

pvlib/pvsystem.py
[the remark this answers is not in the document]
cwhanse: So, `numeric` unless there's a reason to specify a scalar `float`?

pvlib/test/test_pvsystem.py
```diff
@@ -174,6 +174,23 @@ def cec_module_params(sam_data):
     return module_parameters
 
 
+@pytest.fixture(scope="session")
```
wholmgren: not a big deal, but if we're still making changes this should not have ``scope="session"``

pvlib/test/test_pvsystem.py
wholmgren: I'm confused. I don't understand how general test values can be computed inside a test without duplicating the function itself.

pvlib/pvsystem.py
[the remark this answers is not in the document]
wholmgren: or unless there's a reason to specify ``array-like`` or ``Series``.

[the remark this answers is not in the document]
cwhanse: I looked at both `calcparams_desoto` and `calcparams_pvsyst` and I'm going to leave the docstrings as is _for now_. The typical use of either function is to pass model parameters for one module and a list of conditions (irradiance and temperature). That use is reflected in the docstring's use of `numeric` and `float`, where each model parameter is a single value for a module.  The functions are more flexible that this use case.

If we want to overhaul the docstrings, let's open a new issue for it.

pvlib/test/test_pvsystem.py
cwhanse: I copied that.  `scope="module"`?

wholmgren: no scope argument. ``pytest.fixture`` and nothing else

wholmgren: The reasoning for this is that this function returns a mutable dictionary. In principle, one test could inadvertently modify the dictionary, and a second test could pass or fail because of that modification. ``@pytest.fixture`` will call the function and generate a new dictionary each time the data is needed. Adding ``scope="module"`` or ``scope="session"`` instructs ``pytest`` to recreate the data only once per test module (the fixture could be imported into a new module and data would be regenerated) or once per test session (this defeats a large part of the purpose of fixtures).

cwhanse: OK. I'll remove the other instances of `scope=session` from fixtures for CEC and SAPM module parameters, but leave the instance that uses `pvsystem.retrieve_sam`

cwhanse:
Looks like errors related to pulling tmy3 data from NREL's site.

pvlib/pvsystem.py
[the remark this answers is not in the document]
thunderfish24: I think an issue is wise in order to define what we mean to ourselves/users and make everything consistent.

pvlib/test/test_pvsystem.py
thunderfish24: No, I haven't written any tests yet and I admit that I am not familiar with this particular mocker setup. If a mock call can verify the API, then that would be great. I don't mean to hold up this PR and I'll look into it further with the Campanelli model.

pvlib/test/test_pvsystem.py
thunderfish24: Some view including a separate (re)computation as redundant, others view it as transparent. In this case I don't think any of the computations require an implicit solver, so where the values come from should not leave too much to question.

wholmgren:
@thunderfish24 thanks for your careful review -- it's a big help. 

@cwhanse ready to merge?

cwhanse:
Yes, ready to merge.

docs/sphinx/source/whatsnew/v0.6.0.rst
```diff
@@ -13,6 +13,7 @@ Enhancements
 * Add sea surface albedo in irradiance.py (:issue:`458`)
 * Implement first_solar_spectral_loss in modelchain.py (:issue:`359`)
 * Clarify arguments Egref and dEgdT for calcparams_desoto (:issue:`462`)
+* Add pvsystem.calcparams_pvsyst to compute values for the single diode equation using the PVsyst v6 model (:issue:'470')
```
mikofski: please change single spaces to backticks - @wholmgren or @cwhanse , can you do this directly in master, thx

PR #504, opened 2018-07-14
ENH: implementing pvsyst recombination loss current for CdTe and a:Si

### LINKED ISSUE #163: Adding New Single Diode Model For CdTe
Would like to add a new diode model(s) appropriate to non c-Si modules. There are some additional parameters in the model(s). Any thoughts on the correct structured way to do this?  Would this function or functions be added to pvsystem.py? Do modules need to be a class if they have different appropriate performance models?

mikofski:
* closes #163
* add constant VOLTAGE_BUILTIN = 0.9 (V) to singlediode_methods.py
* add arguments d2mutau=0, voltage_builtin=VOLTAGE_BUILTIN, and
cells_in_series to calculate i_recomb and v_recomb
* add check for d2mutau to calc i_recomb, v_recomb, and gradients
* add terms to current and gradients

 - [x] Closes issue #163

### DIFF
--- a/pvlib/singlediode_methods.py
+++ b/pvlib/singlediode_methods.py
@@ -102,7 +102,7 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
         :math:`\\frac{d^2 P}{dV dV_d}`
     """
     # check if need to calculate recombination loss current
-    i_recomb = v_recomb = 0
+    i_recomb, v_recomb = 0, np.inf
     if d2mutau > 0:
         v_recomb = voltage_builtin - diode_voltage / cells_in_series
         i_recomb = photocurrent * d2mutau / v_recomb
@@ -114,15 +114,18 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
     v = diode_voltage - i * resistance_series
     retval = (i, v, i*v)
     if gradients:
+        # check again if need to calculate recombination loss current gradients
+        grad_i_recomb = grad_2i_recomb = 0
         if d2mutau > 0:
-            grad_i_recomb = i_recomb/v_recomb
+            grad_i_recomb = i_recomb / v_recomb
+            grad_2i_recomb = 2 * grad_i_recomb / v_recomb
         g_diode = saturation_current * np.exp(v_star) / nNsVth  # conductance
         grad_i = -g_diode - g_sh - grad_i_recomb # di/dvd
         grad_v = 1.0 - grad_i * resistance_series  # dv/dvd
         # dp/dv = d(iv)/dv = v * di/dv + i
         grad = grad_i / grad_v  # di/dv
         grad_p = v * grad + i  # dp/dv
-        grad2i = -g_diode / nNsVth - 2*grad_i_recomb/v_recomb  # d2i/dvd
+        grad2i = -g_diode / nNsVth - grad_2i_recomb  # d2i/dvd
         grad2v = -grad2i * resistance_series  # d2v/dvd
         grad2p = (
             grad_v * grad + v * (grad2i/grad_v - grad_i*grad2v/grad_v**2)
--- a/pvlib/singlediode_methods.py
+++ b/pvlib/singlediode_methods.py
@@ -104,7 +104,7 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
     # check if need to calculate recombination loss current
     i_recomb, v_recomb = 0, np.inf
     if d2mutau > 0:
-        v_recomb = voltage_builtin - diode_voltage / cells_in_series
+        v_recomb = voltage_builtin * cells_in_series - diode_voltage
         i_recomb = photocurrent * d2mutau / v_recomb
     # calculate temporary values to simplify calculations
     v_star = diode_voltage / nNsVth  # non-dimensional diode voltage
--- a/pvlib/singlediode_methods.py
+++ b/pvlib/singlediode_methods.py
@@ -74,8 +74,8 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
 
     .. warning::
        * Do not use ``d2mutau`` with CEC coefficients.
-       * Usage of ``d2mutau`` with PVSyst coefficients is required for CdTe and
-         a:Si modules.
+       * Usage of ``d2mutau`` with PVSyst coefficients is required for cadmium-
+         telluride (CdTe) and amorphous-silicon (a:Si) PV modules only.
        * For PVSyst CdTe and a:Si modules, the ``cells_in_series`` parameter
          must only account for a single parallel sub-string if the module has
          cells in parallel greater than 1.
@@ -95,7 +95,7 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
     nNsVth : numeric
         product of thermal voltage ``Vth`` [V], diode ideality factor ``n``,
         and number of series cells ``Ns``
-    cells_in_series : int
+    cells_in_series : None or int
         number of cells in series per parallel module sub-string, only required
         for PVSyst thin-film recombination loss, if unset default is ``None``
         which raises ``TypeError`` if ``d2mutau`` is set.
@@ -123,9 +123,9 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
     The PVSyst thin-film recombination losses parameters ``d2mutau`` and
     ``voltage_builtin`` are only applied to cadmium-telluride (CdTe) and
     amorphous-silicon (a:Si) PV modules, [2]_, [3]_. The builtin voltage should
-    account for all junctions. *EG*: tandem and triple junction cell would have
-    builtin voltages of 1.8[V] and 2.7[V] respectively, based on the default of
-    0.9[V] for a single junction.
+    account for all junctions. For example: tandem and triple junction cell
+    would have builtin voltages of 1.8[V] and 2.7[V] respectively, based on the
+    default of 0.9[V] for a single junction.
 
     References
     ----------
@@ -143,11 +143,12 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
        2010
        :doi:`10.4229/25thEUPVSEC2010-4BV.1.114`
     """
-    # check if need to calculate recombination loss current
-    i_recomb, v_recomb = 0, np.inf
-    if d2mutau > 0:
-        v_recomb = voltage_builtin * cells_in_series - diode_voltage
-        i_recomb = photocurrent * d2mutau / v_recomb
+    # calculate recombination loss current where d2mutau > 0
+    is_recomb = d2mutau > 0  # True where there is thin-film recombination loss
+    v_recomb = np.where(is_recomb,
+                        voltage_builtin * cells_in_series - diode_voltage,
+                        np.inf)
+    i_recomb = np.where(is_recomb, photocurrent * d2mutau / v_recomb, 0)
     # calculate temporary values to simplify calculations
     v_star = diode_voltage / nNsVth  # non-dimensional diode voltage
     g_sh = 1.0 / resistance_shunt  # conductance
@@ -156,11 +157,9 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
     v = diode_voltage - i * resistance_series
     retval = (i, v, i*v)
     if gradients:
-        # check again if need to calculate recombination loss current gradients
-        grad_i_recomb = grad_2i_recomb = 0
-        if d2mutau > 0:
-            grad_i_recomb = i_recomb / v_recomb
-            grad_2i_recomb = 2 * grad_i_recomb / v_recomb
+        # calculate recombination loss current gradients where d2mutau > 0
+        grad_i_recomb = np.where(is_recomb, i_recomb / v_recomb, 0)
+        grad_2i_recomb = np.where(is_recomb, 2 * grad_i_recomb / v_recomb, 0)
         g_diode = saturation_current * np.exp(v_star) / nNsVth  # conductance
         grad_i = -g_diode - g_sh - grad_i_recomb # di/dvd
         grad_v = 1.0 - grad_i * resistance_series  # dv/dvd
--- a/pvlib/singlediode_methods.py
+++ b/pvlib/singlediode_methods.py
@@ -22,6 +22,8 @@ except ImportError:
 # rename newton and set keyword arguments
 newton = partial(_array_newton, tol=1e-6, maxiter=100, fprime2=None)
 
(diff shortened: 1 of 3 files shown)


mikofski:
@cwhanse the difference with and w/o PVSyst/Merten recombination losses are significant, does this look right? I used `d2mutau = 1.3  # [V]` and 108 series cells (or 216/2) for FS-495 from the PVSyst database for PV Modules.

fs495_recomb_power

fs495_recomb_iv

cwhanse:
Can you post the other PVsyst parameters for this module, and I'll check against a MATLAB implementation?

My quibble with changing `singlediode_methods.py` to `singlediode.py` is that `singlediode` is a function in `pvlib.pvsystem`.  We'd have both `pvlib.pvsystem.singlediode` and `pvlib.singlediode`.  Personally I find the re-use of names to be awkward.

mikofski:
The other values I took from `pvsystem.retrieveSAM('CECMod')`.

Ok, let's leave it as `singlediode_methods.py` then.

mikofski:
@cwhanse after talking with @jdnewmil I realized I should not have used the FS-495 parameters from CEC database - and also I used the wrong method, I should have used the new `calcparams_pvsyst`, so sorry for wasting your time, I'm revising the test and plots now, and I'll post them shortly

cwhanse:
@mikofski I'm having trouble interpreting your figures above.  Does DeltaI compare the Desoto model to the Pvsyst model?  I would like to verify that the results with the recombination term appear reasonable, and to do that it would help to compare two IV curves computed with the same model (Pvsyst) without and with the term activated.

mikofski:
OK, but AFAIK you can't use PVSyst this way, b/c the parameters would be different _ie_ using the PVSyst database parameters for FS-495 without the recombination loss will not fit the measured IV curve or nameplate output power.

Here are the deltas:

    delta Pmax = 5.797158
    delta Isc = 0.022876
    delta Voc = 0.544359

and plots:

pvsyst_fs495_recomb_power

pvsyst_fs495_recomb_iv

cwhanse:
thanks @mikofski I'm more interested to qualitatively verify the with/without recombination term calculation than to compare Desoto and Pvsyst for nominally the same module.  The results in the Pvsyst with/without figures look correct.

Comparing Desoto to Pvsyst brings in other assumptions in addition to the recombination current, not least of which is the method used to derive the model parameters.

I'm ready to merge this, pending concurrence from @wholmgren

wholmgren:
@mikofski let me know when the PR template boxes are checked and I will review.

mikofski:
Hi @wholmgren, I don't think this is ready yet.

I still need some direction on how many other methods these parameters should be added to, and what type of text we want to put in the documentation for users.

I think we would need to add `voltage_builtin` (_aka_ `vbi`), `d2mutau`, and `cells_in_series` to each of the following:
* `singlediode_methods.bishop88_i_from_v()`
* `singlediode_methods.bishop88_v_from_i()`
* `singlediode_methods.bishop88_mpp()`
* `singlediode_methods._lambertw_v_from_i()`
* `singlediode_methods._lambertw_i_from_v()`
* `singlediode_methods._lambertw()`
* `pvsystem.PVSystem.singlediode()`
* `pvsystem.PVSystem.i_from_v()`
* `pvsystem.singlediode()`
* `pvsystem.max_power_point()`
* `pvsystem.v_from_i()`
* `pvsystem.i_from_v()`

I think in addition to docstring API changes that describe the new arguments, we need some sternly worded messages that warn users:
> Do **not** use `d2mutau` with CEC coefficients.

... and ...
> Usage of `d2mutau` with PVSyst coefficients **is required** for CdTe and a:Si modules.

... and ...
> For PVSyst CdTe and a:Si modules, the `cells_in _series` parameter must only account for a single parallel sub-string if the module has cells in parallel greater than 1.

_EG_: FS-495 has two parallel sub-strings with 108 cells each, therefore `cells_in_series` should be 108.

Can someone comment to let me know if these changes are aligned with their thoughts? thx!

wholmgren:
@mikofski ah, I haven't followed closely and was only responding to @cwhanse's comment regarding ready to merge. Maybe there is a mismatch in anticipated scope?

Re ``singlediode.py`` vs ``singlediode_methods.py``... We've previously discussed breaking ``pvsystem.singlediode`` into a couple of functions. ``pvsystem.singlediode`` could disappear entirely if/when that happens. I rarely see module names as long as ``singlediode_methods`` or module names with underscores. So, my vote would be change the module name to ``singlediode``.

mikofski:
Ok, so then what should we do with this PR and do I need to open a few new issues to start discussion of these remaining tasks?

pvlib/singlediode_methods.py
```diff
@@ -62,7 +64,9 @@ def estimate_voc(photocurrent, saturation_current, nNsVth):
 
 
 def bishop88(diode_voltage, photocurrent, saturation_current,
-             resistance_series, resistance_shunt, nNsVth, gradients=False):
+             resistance_series, resistance_shunt, nNsVth, d2mutau=0,
+             cells_in_series=None, voltage_builtin=VOLTAGE_BUILTIN,
```
cwhanse: A nit - re-order the arguments to keep the non-recombination current terms first, followed by `d2mutau` and `voltage_builtin` - just move `cells_in_series` up one position.

pvlib/test/test_singlediode_methods.py
```diff
@@ -125,3 +126,135 @@ def test_brentq_fs_495():
                                 method='lambertw')
     assert np.isclose(pvs_ixx, ixx)
     return isc, voc, imp, vmp, pmp, i, v, pvs
+
+
+def pvsyst_fs_495():
+    """
+    PVSyst First Solar FS-495 parameters.
```
cwhanse: `PVsyst applied to First Solar FS-495 parameters.`

pvlib/test/test_singlediode_methods.py
```diff
@@ -125,3 +126,135 @@ def test_brentq_fs_495():
...
+    fs_495['I_L_ref'] = (
+        (fs_495['I_sc_ref'] + Id + Ish) / (1 - fs_495['d2mutau'] / dv)
+    )
+    return fs_495
+
+
+PVSYST_FS_495 = pvsyst_fs_495()  # PVSyst First Solar FS-495 parameters
+
+
+def test_pvsyst_fs_495_recombination_loss():
```
cwhanse: I think the test is more general than just the `fs_495` module - I'd drop that nomenclature from the function name.

pvlib/test/test_singlediode_methods.py
```diff
@@ -125,3 +126,135 @@ def test_brentq_fs_495():
...
+    pvsyst = bishop88(
+        diode_voltage=vd_pvsyst, photocurrent=il_pvsyst,
+        saturation_current=io_pvsyst, resistance_series=rs_pvsyst,
+        resistance_shunt=rsh_pvsyst, nNsVth=nnsvt_pvsyst,
+        d2mutau=pvsyst_fs_495['d2mutau'],
+        cells_in_series=pvsyst_fs_495['cells_in_series']
+    )
+
+    # test expected change in max power
+    assert np.isclose(max(desoto[2]) - max(pvsyst[2]), 0.01949420697212645)
```
cwhanse: This test passes if the Desoto model and the PVsyst model are 'close', meaning that the test would fail is the CEC parameter database entry for this module changes, if there is an error in the Desoto functions, or if the PVsyst function has an error.  I'd change the test condition to compare the PVsyst output to known values calculated by a separate method.

pvlib/test/test_singlediode_methods.py
```diff
@@ -125,3 +126,135 @@ def test_brentq_fs_495():
...
+    isc_pvsyst = np.interp(0, pvsyst[1], pvsyst[0])
+    assert np.isclose(isc_desoto - isc_pvsyst, -7.955827628380874e-05)
+    # test expected change in open circuit current
+    voc_desoto = np.interp(0, desoto[0][::-1], desoto[1][::-1])
+    voc_pvsyst = np.interp(0, pvsyst[0][::-1], pvsyst[1][::-1])
+    assert np.isclose(voc_desoto - voc_pvsyst, -0.04184247739671321)
+    return desoto, pvsyst
+
+
+if __name__ == '__main__':
```
cwhanse: Although great for development and testing, I believe we don't keep `if __name__ == '__main__':` blocks in modules.

mikofski: ✔️ agreed, removed

pvlib/test/test_singlediode_methods.py
mikofski: ✔️ agreed, changed to `test_pvsyst_recombination_loss`

pvlib/test/test_singlediode_methods.py
mikofski: ✔️ the initial data comes from the PVSyst v6.7.2 database, I don't think any of the values were altered but it was vetted internally.

I've removed the method, and hard coded the values with a comment re: initial source.

cwhanse: Not having a current PVsyst license, I'm guessing that it does NOT permit uploading that database to pvlib?

adriesse: Maybe not the whole database, but I'm guessing the relevant values come from pan files generated by FS.

pvlib/test/test_singlediode_methods.py
mikofski: ✔️ I've changed the test to be independent of DeSoto, now it only tests 2 conditions:
* reference condition, should agree with its own parameters
* another condition, (888[W/m^2], 55[degC]), for which I calculated the expected output using `calcparams_pvsyst` and `bishop88`


But I should note that the output at (888[W/m^2], 55[degC]) doesn't agree well with DeSoto at the same conditions:

```python
In [1]: from pvlib import pvsystem
In [2]: CECMOD = pvsystem.retrieve_sam('cecmod')
In [3]: fs_495 = CECMOD.First_Solar_FS_495
In [4]: il, io, rs, rsh, nnsvt= pvsystem.calcparams_desoto(888, 55,
   ...: alpha_sc=fs_495.alpha_sc, a_ref=fs_495.a_ref, I_L_ref=fs_495.I_L_ref,
   ...: I_o_ref=fs_495.I_o_ref, R_sh_ref=fs_495.R_sh_ref, R_s=fs_495.R_s,
   ...: EgRef=1.475, dEgdT=-0.0003)
In [5]: pvs = pvsystem.singlediode(il, io, rs, rsh, nnsvt, method='lambertw')
In [6]: pvs
Out[6]:
OrderedDict([('i_sc', 1.4020527838304147),  # PVSyst: 1.3868344548308347
             ('v_oc', 75.41704855182024),  # PVSyst: 79.29198489135703
             ('i_mp', 1.2599626119945482),
             ('v_mp', 57.70813302772382),
             ('p_mp', 72.71009002293975),  # PVSyst: 76.26211540956041 
             ('i_x', 1.3606253515830833),
             ('i_xx', 0.8340096657730403)])
```

According to @jdnewmil this is possibly due to a setting in PVSyst that adjusts `d2mutau` to a fixed max power temp coeff instead (or in addition???) to the adjustment on `gamma`. Even though I can see these check boxes in PVSyst, I'm not sure to what degree we want to copy or guess what PVSyst does here in PVLib (or in SAM either @timorichert)

pvlib/singlediode_methods.py
mikofski: ✔️

pvlib/test/test_singlediode_methods.py
mikofski: @cwhanse if we wanted to try to emulate this 2nd order `d2mutau` behavior, then IMO it should go into `calcparams_pvsyst`, and so those values: `d2mutau` (and `voltage_builtin` if multi junction?) should be outputs from there.

cwhanse: Agree. I don't know what PVsyst might be doing behind the interface when `d2mutau` is in action, however, so I can't help clarify here.

adriesse: If it isn't clear what PVsyst is doing, then I suggest mailing Bruno to get clarification.  I think they're quite open about the calculations.

mikofski:
okay, @wholmgren @cwhanse and anyone else, I think this is ready for review and merging if there are no more comments. thanks

mikofski:
Evidently Python-2 doesn't like "André Mermoud" because it had utf-8, let me as encoding to the file and see if that fixes it

mikofski:
_emphasis mine_:
>I'm not sure we need those 'sternly worded' warnings. We could test inside `calcparams_pvsyst` for cell type and non-zero `d2mutau` and warn the user when the recombination current term isn't set up right. But to me, **the real issue is abetting the view that the parameters in the CEC model are somehow generic for any single diode model**. I think a way to ameliorate that risk is tie the SAM parameter reader function to the CEC model, so that it is awkward to invoke the PVsyst model (or other single diode model) on a set of parameters read from the SAM file.

I think a possible solution to this situation is for `calcparams_pvsyst` to return both `d2mutau` and `NsVbi` which are then passed to `singlediode` and its delegates (`bishop88`) and proxies (`PVSystem.singlediode`. If this is what you are suggesting then I agree.

NOTE: having `calcparams_pvsyst` handle `NsVbi` also takes care of multi-junction cells, which is another caveat that users would have to deal with since the default 0.9-V is only for single junction, and although rare, tandem and triple a:Si cells would have a builtin voltage of 1.8-V and 2.7-V respectively.

And this could also avoid any confusion regarding CIS/CIGS which should _not_ use the thin-film recombination loss at all.

>I'm in favor of merging this pull request with its current scope (after approvals). We can open separate issues ...

I agree with everything in the second paragraph.

>OK, we can rename the module to singlediode. ...

Thank you. So shall this also be a separate issue out of scope for this PR?

mikofski:
failures are do to Siphon, not related to doc changes in 1e9434a

cwhanse:
> I think a possible solution to this situation is for calcparams_pvsyst to return both d2mutau and NsVbi which are then passed to singlediode and its delegates (bishop88) and proxies (PVSystem.singlediode. If this is what you are suggesting then I agree.

We agree.

> Thank you. So shall this also be a separate issue out of scope for this PR?

Yes, please.

mikofski:
Test failure for AppVeyor Python 3.6 due to `test_forecast` is unrelated. All other jobs pass and Travis all pass. Is there anything more anyone is expecting here? @cwhanse do you arrive this PR?

wholmgren:
@mikofski sorry I did not realize this was ready. Thanks for making the changes. Looks great. ``np.where`` is a good choice. Only catch is that your function will always return np.arrays for the gradients. I'm guessing that is no big deal.

Issue #506, opened 2018-07-18
Correct condition 5 in `clearsky.detect_clearsky`


cwhanse:
**Describe the bug**
Condition 5 in `clearsky.detect_clearsky` does not agree with Eq. 12 - 14 of the reference. See #505 

**Expected behavior**
`clearsky.detect_clearsky` currently calculates condition 5 as the difference in maximum absolute slopes, rather than the maximum of the absolute differences in slopes.

**Versions:**
 - ``pvlib.__version__``: 0.5

wholmgren:
If I understand correctly, the pvlib code is essentially:

```
X  = max(abs(meas_slope)) - max(abs(clear_slope))
c5 = X < slope_dev
```

but Reno and Hansen describe:

```
X  = max(abs(meas_slope - clear_slope))
   = max(abs(line_diff_slope))
c5 = X < slope_dev
```

It would be good to include an example/test where this changes the overall result.

cwhanse:
@wholmgren you understand correctly.

Do we need to test before/after when fixing a bug though?

adriesse:
Perhaps not in general, but in this case it sound like the published results may have been based on code with a bug, so technically you should perhaps run the corrected code, check the results and publish an erratum.

wholmgren:
@cwhanse my feeling is that the [detect_clearsky_data.csv](https://github.com/pvlib/pvlib-python/blob/master/pvlib/data/detect_clearsky_data.csv) test data should include values that produce a different test result when the code is changed.

cwhanse:
Of course, I agree that we should correct the test along with the code fix.  I can publish a comparison on pvpmc.sandia.gov, or post the results here.

Issue #507, opened 2018-07-18
Handle different time intervals in clearsky.detect_clearsky


cwhanse:
**Is your feature request related to a problem? Please describe.**
`clearsky.detect_clearsky` implicitly assumes data at 1 minute time intervals.  See #505 

**Describe the solution you'd like**
`clearsky.detect_clearsky` should be capable of accepting data at different, and possibly non-uniform, time intervals. PVLib MATLAB function `pvl_detect_clear_times` provides an example method.

wholmgren:
Wouldn't it be relatively straightforward to accept uniform intervals of any length by multiplying the supplied coefficients by a scaling factor?

I'm guessing that the non-uniform case should be implemented in a separate function that's called within ``detect_clearsky``. Probably move the uniform case to a separate function, too. Possibly private functions for simplicity, but perhaps there is value in exposing them to the public api.

adriesse:
I guess we never finished the discussion in #433, where I cited this function as an example where a warning might be useful.  Or perhaps more accurately, I never acted on the implementation suggestion!

If the model was developed, tested and published for and with 1-minute data, then modifying it for other sampling rates should get additional validation.

cwhanse:
@wholmgren up to a point, yes, but the threshold criteria should be re-considered if one is applying this method to data at much longer intervals.

@adriesse good point.  The paper looked only at 1 minute data.  I've used versions of that algorithm on _roughly_ one minute data (30s to 90s, not uniform intervals) and got reasonable results.  At 15 minute intervals or longer I'd question the outcome.  I'd be OK with a gentle warning about validation of the algorithm.  I'd be more than OK if someone wants to actually validate the algorithm, or show how it should change, to handle longer time intervals.

AndyMender:
I think it's alright to allow intervals other than 1 minute provided that the data points fit within the span of a single day (1440 minutes). Unevenly spaced data might be a major issue for the entire algorithm because of the static window size.

mdeceglie:
I suggest we plan to treat two issues/enhancements separately:

1. Eliminate the implicit assumption of 1 minute data
2. Allow for non-uniform timestamps.

Item 1 is relatively straight forward to address, and I think @bhellis725 has already addressed it here: https://github.com/bhellis725/pvlib-python/tree/clearsky Hopefully we'll see a pull request on that soon. Item 2 would be a great addition, but is a heavier lift, so I'd be in favor of at least advancing item 1 in the near term.

[...]

PR #510, opened 2018-07-26
Detect clearsky modifications

bhellis725:

The fifth condition (`c5`) of the `clearsky.detect_clearsky` function wasn't calculated as defined in the original reference (equations 12 - 14).  This pull request also addresses part of #507 where there was an implicit assumption of minutely data.  All derivatives are now calculated by explicitly dividing by the `sample_interval`.  This does not address non-uniform time intervals.

 - [ ] Closes issue #506 (addresses part of #507)
 - [ ] Code quality and style is sufficient. Passes ``git diff upstream/master -u -- "*.py" | flake8 --diff``

### DIFF
--- a/pvlib/clearsky.py
+++ b/pvlib/clearsky.py
@@ -688,9 +688,6 @@ def detect_clearsky(measured, clearsky, times, window_length,
                                   'times. consider resampling your data.')
 
     samples_per_window = int(window_length / sample_interval) + 1
-    # print(window_length)
-    # print(sample_interval)
-    # print(samples_per_window)
 
     # generate matrix of integers for creating windows with indexing
     from scipy.linalg import hankel
@@ -730,8 +727,7 @@ def detect_clearsky(measured, clearsky, times, window_length,
         c2 = np.abs(meas_max - alpha*clear_max) < max_diff
         c3 = (line_diff > lower_line_length) & (line_diff < upper_line_length)
         c4 = meas_slope_nstd < var_diff
-        #  c5 = (meas_slope_max - alpha*clear_slope_max) < slope_dev
-        c5 = np.max(meas_slope - alpha*clear_slope, axis=0) < slope_dev
+        c5 = np.max(np.abs(meas_slope - alpha*clear_slope), axis=0) < slope_dev
         c6 = (clear_mean != 0) & ~np.isnan(clear_mean)
         clear_windows = c1 & c2 & c3 & c4 & c5 & c6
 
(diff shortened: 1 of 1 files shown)


pvlib/clearsky.py
```diff
@@ -687,7 +687,7 @@ def detect_clearsky(measured, clearsky, times, window_length,
         raise NotImplementedError('algorithm does not yet support unequal ' \
                                   'times. consider resampling your data.')
 
-    samples_per_window = int(window_length / sample_interval)
+    samples_per_window = int(window_length / sample_interval) + 1
```
cwhanse: Revert this change please.  `samples_per_window` is counting intervals, not endpoints. Because it's an intermediate it could be renamed to be more clear.

pvlib/clearsky.py
```diff
@@ -697,25 +697,27 @@ def detect_clearsky(measured, clearsky, times, window_length,
     # calculate measurement statistics
     meas_mean = np.mean(measured[H], axis=0)
     meas_max = np.max(measured[H], axis=0)
-    meas_slope = np.diff(measured[H], n=1, axis=0)
+    meas_ghi_diff = np.diff(measured[H], n=1, axis=0)
+    meas_slope = np.diff(measured[H], n=1, axis=0) / sample_interval
```
cwhanse: The non-uniform time steps could be handled here by
- converting the time to UNIX timestamps
- `diff`ing the time to get time intervals
- element-wise division in the calculation of `meas_slope`
- using the `diff`ed time in the calculation of `meas_line_length` , 'clear_slope`, etc.

This could be the subject of a subsequent pull request. In hindsight, I could have separated #507 into two issues (different from 1 minute, and non-uniform time steps.)

pvlib/clearsky.py
bhellis725: I believe that not adding 1 gives the incorrect number of samples per window.  For example, if my data is 30-minute frequency (`sample_interval=30`) and I want 60 minute windows (`window_length=60`), the current implementation would only give 2 points per window (when the Hankel matrix is constructed in the following lines).  In this case, 2 points per window would only span a 30 minute window, not the intended 60.

cwhanse: Your example makes my point. The algorithm operates on intervals not on points in time. The value at a timestamp is considered as the value for the following (?) interval - I'd have to look carefully at the Hankel matrix and the diff to see if we adopted a left- or right- endpoint convention.

wholmgren:
completed in #596

PR #533, opened 2018-08-13
update installation instructions

### LINKED ISSUE #531: improve installation docs
Installation documentation should be improved to cover:

- [x] Only pandas and numpy are required for most functionality, but additional packages can be installed to support more features. Consider documenting which features require which packages.
- [x] pvlib is available on conda-forge as ``pvlib-python``. (we should also consider changing the conda-forge package to ``pvlib`` but that is a separate issue)

This is feedback from https://github.com/openjournals/joss-reviews/issues/884#issuecomment-412553989

wholmgren:
- [x] Closes #531

I built the documentation locally and it looked good to me.


docs/sphinx/source/installation.rst
```diff
@@ -58,8 +57,13 @@ non-editable way, use `conda <http://conda.pydata.org/docs/>`_
 (recommended if you use the Anaconda Python distribution) or `pip
 <https://pip.pypa.io>`_ (works with any Python distribution)::
 
+    # get the package from the pvlib conda channel
     conda install -c pvlib pvlib
 
+    # get the package from the conda-forge conda channel
+    conda install -c conda-forge pvlib-python
+
+    # get the package from the Python Package Index
```
mikofski: I recommend using `--no-deps` if you are using anaconda, otherwise you may get strange results. - In fact why not move this to an "Advanced User" section, b/c for the most part users who don't know what to do should only use Anaconda, and users who know how to use pip don't need installation instructions.

Also, a nit, I recommend creating a new conda env if using conda-forge, IMO you should not use conda-forge in your anaconda base env, so that leaves only the pvlib channel option which should work for most users. Maybe move conda-forge to advanced users section. If you use conda-forge in your base Anaconda, conda-forge has a tendency to take over, and so IMO it should probably be the default channel in any conda env it's used.

docs/sphinx/source/installation.rst
```diff
@@ -191,9 +195,33 @@ if you make changes to pvlib during an interactive Python
...
+
+A handful of pvlib-python features require additional packages that must
+be installed separately using pip or conda. These packages/features
+include:
+
+* scipy: single diode model, clear sky detection
+* pytables: Linke turbidity look up for clear sky models
+* numba: fastest solar position calculations
+* pyephem: solar positions calculations using an astronomical library
+* siphon: forecasting PV power using the pvlib.forecast module
```
mikofski: IMO having a `requirements.txt` file makes specifying requirements and letting users create non-conda virtual environments easier, here's my `requirements.txt` for pvlib in linux (not conda):
```
# build and run
pytz
numpy
scipy
pandas
numba
ephem
siphon
netcdf4
tables
# docs and testing
matplotlib
sphinx
sphinx_rtd_theme
ipython
jupyter
numpydoc
pytest
pytest-cov
pytest-mock
nose
nbsphinx
```
some of these are not available from either anaconda.org or conda-forge.

docs/sphinx/source/installation.rst
wholmgren: Thanks for the good feedback.

In my experience, ``pip install pvlib`` within an environment with sufficient numpy/pandas versions will not do anything but install pvlib itself. Is it safe to assume that people are using an Anaconda distribution that is recent enough (or updated enough) to meet the modest minimum numpy/pandas requirements? I would say yes, but perhaps you've seen otherwise. If it is safe for pvlib, then I don't think ``--no-deps`` helps and it makes it looks unnecessarily complicated. 

The "Install as an editable library" section is, I think, largely what an "Advanced User" section would look like. I think the existing name is preferable because a relatively large percentage of people that are new to pvlib are also people that would want to edit the source code. Maybe I'm wrong about that and/or maybe that was more true a few years ago than now.

What about adding a second comment line for each install option such as:

```
    # get the package from the pvlib conda channel
    # best option for installing in the base Anaconda distribution
    conda install -c pvlib pvlib

    # get the package from the conda-forge conda channel
    # best option if using forecast module. consider a separate conda env
    conda install -c conda-forge pvlib-python

    # get the package from the Python Package Index
    # best option if you know what you are doing
    pip install pvlib
```

docs/sphinx/source/installation.rst
wholmgren: I am open to 

1. adding a requirements file or files, and/or
2. adding support for e.g. ``pip install pvlib[all]``

in

1. this pull request, or
2. a separate pull request

Not sure what the best approach is, though, which makes me think defer each of those items to their own PRs. They are certainly related to this topic, though, especially if the result of the "Advanced User" or similar request involves a larger restructuring.

mikofski: Agreed let's consider `requirements.txt` in a separate PR, but I was just noting that there are more dependencies than just numpy and pandas.

docs/sphinx/source/installation.rst
mikofski: I do like the second comment you added, maybe a good compromise.

docs/sphinx/source/installation.rst
wholmgren: optional dependencies ;)

docs/sphinx/source/installation.rst
wholmgren: I modified the intro paragraph to push people to conda and I added the comments. Happy to continue iterating since this is important.

My typical workflow includes pip-installing pvlib in conda envs that already have numpy and pandas (or more). I've not had any issues with this and I would be surprised if any other users have. But it does require care to mix together pip and conda installs, so I agree that we should dissuade people from doing it.

wholmgren:
merging this one now, but @mikofski feel free open new issues/PRs with the additional changes you'd like to see.

PR #535, opened 2018-08-14
limit cos(aoi) in diffuse sky functions

### LINKED ISSUE #432: Dividing by cos(solar_zenith) in haydavies and reindl functions
A few issues with the pvlib package seem to be caused by dividing by cos(solar_zenith) since solar zenith can vary from 0-180 degrees. As the solar_zenith approaches 90 degrees, these results will approach infinity. The problem becomes very apparent when the value of solar_zenith is between 89 and 90 degrees which is rare unless working with minute interval (or smaller) because the denominator then becomes a large scaling factor. I have attached a quick sample code to illustrate the issue. 
[cos(solar_zenith)_errors.txt](https://github.com/pvlib/pvlib-python/files/1737381/cos.solar_zenith._errors.txt)

A few questions and comments to note: 

1. This seems to have been addressed in a similar context in the [DISC Model and Zenith Threshold](https://github.com/pvlib/pvlib-python/issues/311) issue but appear again in the haydavies algorithm implementation on line 729 of irradiance.py when calculating the projection_ratio and could potentially appear elsewhere in the package. 

2. The MATLAB version of the haydavies algorithm, pvl_haydavies1980.m, uses the formula "RB = max(COSTT, 0) ./ max(cosd(SunZen), 0.01745))" where 0.01745 corresponds to a zenith angle of 89 degrees to avoid the issue. 

3. [Sandia National Laboratories documentation](https://pvpmc.sandia.gov/modeling-steps/1-weather-design-inputs/plane-of-array-poa-irradiance/calculating-poa-irradiance/poa-sky-diffuse/hay-sky-diffuse-model/) seems to suggest that dividing by cos(solar_zenith) is not part of the Hay and Davies diffuse model.

Is there a documented reason for the differences between these formulas or is this a mistake in implementation?

### LINKED ISSUE #526: Nonzero direct irradiance contribution to sky diffuse when sun is behind the array
**Describe the bug**
In the `irradiance.py` module the `isotropic`, `klucher`, `haydavies`, `reindl`, and `king` functions have an `Rb` term that contributes an irradiance adjustment to the front-side diffuse even when the sun is behind the array or below the horizon. Under such conditions the only contributions to in-plane sky diffuse should be isotropic sky diffuse and (where modeled) horizon sky diffuse, and should not include the circumsolar component (particularly when this component is computed as a negative value).

**To Reproduce**
Run the code from the `test_total_irrad` function in `test_irradiance.py`. The second `ephem_data` row happens to represent this geometric condition as input... the test does not currently verify the values thus obtained, but if they are examined the behavior can be observed. For example, using the "haydavies" model the sky diffuse result is 20.4696793893283, but it should be 28.439461728635 because the `Rb` term is introducing a negative circumsolar irradiance contribution.

**Expected behavior**
Sky diffuse irradiance that does not vary significantly based on the sun apparent position when that position is behind the array or below the horizon.

I think that this can be fixed by setting `Rb=0` whenever `cos_tt` is negative or `cos_sun_zenith` is negative.

**Versions:**
 - ``pvlib.__version__``: '0.5.2+20.g09126e9'
 - ``pandas.__version__``: '0.23.1'
 - python: '3.6.6 |Anaconda, Inc.| (default, Jun 28 2018, 11:27:44) [MSC v.1900 64 bit (AMD64)]'

wholmgren:
Fixes two issues related to projection of the solar angle:

1. Dividing by near-0 cos(solar_zenith) when the sun is very close to the horizon.
2. non-0 contributions of the beam irradiance to the ``reindl``, ``haydavies``, and ``klucher`` functions.

Implementation is up for debate. Seemed easier to make a pull request with a concrete proposal.

I considered calculating the ``cos_solar_zenith`` as:

```python
cos_solar_zenith = aoi_projection(0, 0, solar_zenith, 0, projection_minimum=0.01745)
```

but I am worried this could lead to issues with small floating point arithmetic (worse, they could be platform and python/numpy version dependent).

 - [x] Closes #432, Closes #526

other:
- [x] Add test values for solar zenith between 89 and 90 deg.
- [x] Rerun irradiance tutorial


pvlib/irradiance.py
```diff
@@ -931,9 +955,10 @@ def reindl(surface_tilt, surface_azimuth, dhi, dni, ghi, dni_extra,
     '''
 
     cos_tt = aoi_projection(surface_tilt, surface_azimuth,
-                            solar_zenith, solar_azimuth)
+                            solar_zenith, solar_azimuth,
+                            projection_minimum=0)
 
-    cos_solar_zenith = tools.cosd(solar_zenith)
+    cos_solar_zenith = np.maximum(tools.cosd(solar_zenith), 0.01745)
```
adriesse: For Rb this limit might be good, but for HB further below the limit should be zero (already enforced).

pvlib/irradiance.py
```diff
@@ -169,6 +170,11 @@ def aoi_projection(surface_tilt, surface_azimuth, solar_zenith, solar_azimuth):
         Solar zenith angle.
     solar_azimuth : numeric
         Solar azimuth angle.
+    projection_minimum : None or numeric, default None
+        Minimum allowable value for the projection for the solar angle
+        into the plane of the array. For example, 0 is equivalent to a
+        90 degree limit, 0.01745 is equivalent to a 89 degree limit.
+        `None` applies no limit.
```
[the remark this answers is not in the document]
wholmgren: Yes, I also wondered this. I only went with the projection because it's the quantity that is hardcoded in the matlab library.

pvlib/irradiance.py
```diff
@@ -252,10 +269,15 @@ def poa_horizontal_ratio(surface_tilt, surface_azimuth,
     """
 
     cos_poa_zen = aoi_projection(surface_tilt, surface_azimuth,
-                                 solar_zenith, solar_azimuth)
+                                 solar_zenith, solar_azimuth,
+                                 projection_minimum=poa_projection_minimum)
 
     cos_solar_zenith = tools.cosd(solar_zenith)
 
+    if solar_zenith_projection_minimum is not None:
```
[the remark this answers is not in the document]
wholmgren: I suppose we could go a default value of -1 and always apply the limit.

pvlib/irradiance.py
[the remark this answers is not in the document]
wholmgren: After thinking through the response to your comment below, I am leaning towards keeping it as the dot product to avoid the complexity of handling angles > 90 degrees or less than 0.

pvlib/irradiance.py
wholmgren: Agreed. I'll fix this.

cwhanse: IMO it is more clear to leave the `cos_tt` and `cos_solar_zenith` calculations alone, and apply the filters to the value of `Rb`.

pvlib/irradiance.py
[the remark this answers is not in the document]
cwhanse: I'm having trouble seeing a use case which requires `projection_minimum`.  The function's description says it calculates a dot product so that's what I'd expect it to do, and to not do more.

pvlib/irradiance.py
```diff
@@ -243,6 +254,12 @@ def poa_horizontal_ratio(surface_tilt, surface_azimuth,
         Solar zenith angle.
     solar_azimuth : numeric
         Solar azimuth angle.
+    poa_projection_minimum : None or numeric, default None
+        Minimum allowable value for the projection for the solar angle
+        into the plane of the array. `None` applies no limit.
+    solar_zenith_projection_minimum : numeric, default None
+        Minimum allowable value for the projection of the solar angle
+        on to a horizontal surface. `None` applies no limit.
```
cwhanse: I'm struggling to see how the `xxx_minimum` kwargs are more clear than handling the behind-the-plane or nearly co-planar cases directly, in the calculation of the ratios in the diffuse sky models.

wholmgren: That's fair. I think this is a borderline case of "don't repeat yourself". I have no idea if people use these functions on their own.

pvlib/irradiance.py
wholmgren: @cwhanse I figured you'd be in favor of being consistent with the matlab approach. I like it because it addresses both #432 and #526.

pvlib/irradiance.py
```diff
@@ -252,10 +268,14 @@ def poa_horizontal_ratio(surface_tilt, surface_azimuth,
...
-                                 solar_zenith, solar_azimuth)
+                                 solar_zenith, solar_azimuth,
+                                 projection_minimum=poa_projection_minimum)
 
     cos_solar_zenith = tools.cosd(solar_zenith)
 
+    cos_solar_zenith = np.maximum(cos_solar_zenith,
+                                  solar_zenith_projection_minimum)
+
     # ratio of titled and horizontal beam irradiance
```
cwhanse: Nit: `tilted` not `titled`

pvlib/irradiance.py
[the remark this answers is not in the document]
adriesse: Actually it does *less* than a dot product since it assumes/requires the vectors are *unit* vectors.  It's really just calculating the cosine of the angle of incidence.  I think the name "projection" seems to suggest more, and may even be confusing to some.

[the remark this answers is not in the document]
wholmgren: I will remove the ``projection_minimum`` kwargs and put the limits in the diffuse models where necessary. We can address the function name in a separate issue.

wholmgren:
@cwhanse is it easy enough for you to run the matlab equivalents of klucher, haydavies, reindl, king with the same inputs as our test suite?

mikofski:
@cwhanse is pvlib-matlab all functions, no classes? If so it should run in Octave, so then if there's a test suite, you could run it on a CI server like Travis. Just use sudo install octave in the build script

cwhanse:
Yes, PVLib for Matlab is a collection of function. We currently have no test suite for PVLib for Matlab, which is a major deficiency.

wholmgren:
I'm working on some matlab/python tests here: https://github.com/wholmgren/MATLAB_PV_LIB/blob/skydifftest/tests/test_sky_diffuse.m I have little idea about what I'm doing with matlab testing, but it seems to get the job done for this pvlib python PR. @cwhanse if you're interested I can make a PR to the matlab repository, but I imagine you all might have a preferred way to do the testing.

cwhanse:
Klucher:

The expected result for `test_klucher_series` is the same as produced by PVLib for MATLAB. However, PVLib for MATLAB does not filter out for cos(AOI) < 0, which is why the expected result is different from the calculated results.  Bug in PVLib for MATLAB, in my opinion.

pvlib/irradiance.py
```diff
@@ -836,8 +837,9 @@ def haydavies(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
     if projection_ratio is None:
         cos_tt = aoi_projection(surface_tilt, surface_azimuth,
                                 solar_zenith, solar_azimuth)
+        cos_tt = np.minimum(cos_tt, 0)  # GH 526
```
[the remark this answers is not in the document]
cwhanse: Hay and Davies:

`test_haydavies` fails because of this error `np.minimum`. Otherwise I believe it should pass.

pvlib/irradiance.py
```diff
@@ -932,11 +934,13 @@ def reindl(surface_tilt, surface_azimuth, dhi, dni, ghi, dni_extra,
 
     cos_tt = aoi_projection(surface_tilt, surface_azimuth,
                             solar_zenith, solar_azimuth)
+    cos_tt = np.minimum(cos_tt, 0)  # GH 526
```
cwhanse: `np.maximum` here

PR #555, opened 2018-09-04
Add desoto and pvsyst as dc_model options to ModelChain

cwhanse:
pvlib python pull request guidelines
====================================

Thank you for your contribution to pvlib python! You may delete all of these instructions except for the list below.

You may submit a pull request with your code at any stage of completion.

The following items must be addressed before the code can be merged. Please don't hesitate to ask for help if you're unsure of how to accomplish any of the items below:

 - [x] Closes issue #487 
 - [x] I am familiar with the [contributing guidelines](http://pvlib-python.readthedocs.io/en/latest/contributing.html).
 - [x] Fully tested. Added and/or modified tests to ensure correct behavior for all reasonable inputs. Tests (usually) must pass on the TravisCI and Appveyor testing services.
 - [x] Updates entries to `docs/sphinx/source/api.rst` for API changes.
 - [x] Adds description and name entries in the appropriate `docs/sphinx/source/whatsnew` file for all changes.
 - [x] Code quality and style is sufficient. Passes ``git diff upstream/master -u -- "*.py" | flake8 --diff``
 - [x] New code is fully documented. Includes sphinx/numpydoc compliant docstrings and comments in the code where necessary.
 - [x] Pull request is nearly complete and ready for detailed review.

Brief description of the problem and proposed solution (if not already fully described in the issue linked to above):

PR #560, opened 2018-09-05
Implement CEC model

### LINKED ISSUE #463: Implement CEC module performance model in place of desoto

cwhanse:

 - [x] Closes  #463 
 - [ ] Code quality and style is sufficient. Passes ``git diff upstream/master -u -- "*.py" | flake8 --diff``

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.6.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.6.0.rst
@@ -60,8 +60,6 @@ API Changes
   the SAM CEC library sam-library-cec-modules-2017-6-5.csv include the Adjust parameter
   and ModelChain.infer_dc_model will now select the cec model rather than the desoto model.
   (:issue:`463`)
-* Add DC model methods desoto and pvsyst to ModelChain, and deprecates DC model method singlediode
-  (singlediode defaults to desoto until v0.7.0) (:issue:`487`)
 * The behavior of irradiance.perez(return_components=True) has changed.
   The function previously returned a tuple of total sky diffuse and
   an OrderedDict/DataFrame of components. The function now returns
--- a/docs/sphinx/source/whatsnew/v0.6.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.6.0.rst
@@ -53,8 +53,13 @@ API Changes
   dirindex functions. (:issue:`311`, :issue:`396`)
 * Method ModelChain.infer_dc_model now returns a tuple (function handle, model name string)
   instead of only the function handle (:issue:`417`)
-* Add DC model methods desoto and pvsyst to ModelChain, and deprecates DC model method singlediode
-  (singlediode defaults to desoto until v0.7.0) (:issue:`487`)
+* Add DC model methods desoto and pvsyst to ModelChain, and deprecates DC model method
+  singlediode (singlediode defaults to desoto until v0.7.0) (:issue:`487`)
+* Add the CEC module model in pvsystem.calcparams_cec and ModelChain.cec. The CEC model
+  differs from the desoto model by using the parameter Adjust. Modules selected from
+  the SAM CEC library sam-library-cec-modules-2017-6-5.csv include the Adjust parameter
+  and ModelChain.infer_dc_model will now select the cec model rather than the desoto model.
+  (:issue:`463`)
 
 
 Enhancements
@@ -106,7 +111,8 @@ Enhancements
 * Add irradiance.clearness_index function. (:issue:`396`)
 * Add irradiance.clearness_index_zenith_independent function. (:issue:`396`)
 * Add checking for consistency between module_parameters and dc_model. (:issue:`417`)
-* Add DC model methods desoto and pvsyst to ModelChain (:issue:`487`)
+* Add DC model methods ``'desoto'`` and ``'pvsyst'`` to ModelChain (:issue:`487`)
+* Add the CEC module model in `pvsystem.calcparams_cec` and `ModelChain.cec`. (:issue:`463`) 
 
 
 Bug fixes
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -441,7 +441,7 @@ class ModelChain(object):
         (photocurrent, saturation_current, resistance_series,
          resistance_shunt, nNsVth) = (
             self.system.calcparams_cec(self.effective_irradiance,
-                                          self.temps['temp_cell']))
+                                       self.temps['temp_cell']))
 
         self.diode_params = (photocurrent, saturation_current,
                              resistance_series,
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -252,7 +252,7 @@ class ModelChain(object):
     dc_model: None, str, or function, default None
         If None, the model will be inferred from the contents of
         system.module_parameters. Valid strings are 'sapm',
-        'desoto', 'pvsyst', 'pvwatts'. The ModelChain instance will
+        'desoto', 'cec', 'pvsyst', 'pvwatts'. The ModelChain instance will
         be passed as the first argument to a user-defined function.
 
     ac_model: None, str, or function, default None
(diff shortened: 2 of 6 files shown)


wholmgren:
@cwhanse I suggest you wait to resolve the merge conflict until #559 is merged (as soon as tests pass). One way to do it will be:

1. git fetch upstream
2. git merge upstream/master -X theirs
3. reapply and recommit your changes to ``DC_MODEL_PARAMS``

There are more elegant ways to do it but this is probably the most straightforward.

pvlib/pvsystem.py
```diff
@@ -1223,6 +1255,146 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
+    singlediode
+    retrieve_sam
+
+    Notes
+    -----
+    The CEC model differs from the De soto et al. model [3] only in the
+    parameter adjust.
+    '''
+
+    # test for use of function pre-v0.6.0 API change
```
wholmgren: Do we need to include this check? It's a new function, so I don't think we need to support an old API. Better to keep it simple if possible.

pvlib/pvsystem.py
```diff
@@ -1223,6 +1255,146 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
 
+def calcparams_cec(effective_irradiance, temp_cell,
+                   alpha_sc, a_ref, I_L_ref, I_o_ref, R_sh_ref, R_s,
+                   Adjust, EgRef=1.121, dEgdT=-0.0002677,
+                   irrad_ref=1000, temp_ref=25):
+    '''
+    Calculates five parameter values for the single diode equation at
+    effective irradiance and cell temperature using the CEC
+    model described in [1]. The five values returned by calcparams_cec
+    can be used by singlediode to calculate an IV curve.
```
wholmgren: Perhaps add a comment on about how this differs from ``calcparams_desoto``. Either here or a Notes section below the Returns section would work.

pvlib/modelchain.py
```diff
@@ -431,6 +437,22 @@ def desoto(self):
 
         self.dc = self.system.scale_voltage_current_power(self.dc).fillna(0)
 
+    def cec(self):
```
wholmgren: The copy/paste and diff rendering makes this look a little strange, but the net effect is that it seems that this PR currently removes ``return self`` from ``ModelChain.desoto``.

pvlib/modelchain.py
```diff
@@ -431,6 +437,22 @@ def desoto(self):
...
+
+        self.diode_params = (photocurrent, saturation_current,
+                             resistance_series,
+                             resistance_shunt, nNsVth)
+
+        self.dc = self.system.singlediode(
+            photocurrent, saturation_current, resistance_series,
+            resistance_shunt, nNsVth)
+
+        self.dc = self.system.scale_voltage_current_power(self.dc).fillna(0)
```
wholmgren: @mikofski's [_model_dispatcher](https://github.com/pvlib/pvlib-python/pull/555/files#r215348353) is looking better and better :/

wholmgren: Note that LGTM picked up on the duplication in desoto/pvsyst once that PR was merged. Not sure why it's not picking it up in this PR. I'm still ambivalent about the best way to proceed. https://lgtm.com/projects/g/pvlib/pvlib-python/snapshot/66ee982a45ef3799644f79ed73f7b9ac5805b799/files/pvlib/modelchain.py?sort=name&dir=ASC&mode=heatmap&showExcluded=false#xf10a995000ac1e72:1

pvlib/pvsystem.py
cwhanse: There was a comment in the Notes section below the References, but it's easy to overlook, so I moved the comment up and deleted the Notes.

pvlib/modelchain.py
cwhanse: Ah.  That's an inadvertent deletion.  Wonder how the tests passed without `return self`.  Maybe something worth looking into.

pvlib/modelchain.py
cwhanse: I agree about `_model_dispatcher`.  But I'd like a design that also admits alternatives to  `pvsystem.singlediode` in addition to `calcparams_xxxx`.  For a future PR.

cwhanse: Added `test_calcparams_cec`. I added a test condition that differentiates between `calcparams_desoto` and `calcparams_cec`.

pvlib/modelchain.py
wholmgren: Most of the the ModelChain methods mutate an object attribute, so returning self is not necessary for how we use it within pvlib. The advantage of return self is that it allows people to write their own custom code such as:

```
mc.dc().ac().losses()
```

Say ``mc.dc()`` returned the default ``None`` instead of ``self``. Then python would try to call ``None.ac()`` which obviously would fail.

We should add assertions such as ``assert mc == mc.cec()`` to many of the tests.

cwhanse:
How do I resolve the remaining stickler issues?  I'm not finding any open comments, and I've addressed the comments that I did find.

pvlib/pvsystem.py
```diff
@@ -1223,6 +1255,143 @@ def calcparams_desoto(effective_irradiance, temp_cell,
...
 
+def calcparams_cec(effective_irradiance, temp_cell,
+                   alpha_sc, a_ref, I_L_ref, I_o_ref, R_sh_ref, R_s,
+                   Adjust, EgRef=1.121, dEgdT=-0.0002677,
+                   irrad_ref=1000, temp_ref=25):
+    '''
+    Calculates five parameter values for the single diode equation at
+    effective irradiance and cell temperature using the CEC
+    model described in [1]. The CEC model differs from the De soto et al. 
+    model [3] by the parameter Adjust. The five values returned by
```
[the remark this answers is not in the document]
cwhanse: Removed

wholmgren:
@cwhanse I don't know. I have no problem merging pull requests that stickler hasn't passed. It's only a guide to help us.

docs/sphinx/source/whatsnew/v0.6.0.rst
```diff
@@ -111,6 +118,8 @@ Enhancements
 * Add irradiance.clearness_index function. (:issue:`396`)
 * Add irradiance.clearness_index_zenith_independent function. (:issue:`396`)
 * Add checking for consistency between module_parameters and dc_model. (:issue:`417`)
+* Add DC model methods ``'desoto'`` and ``'pvsyst'`` to ModelChain (:issue:`487`)
```
wholmgren: duplicate of a line below

pvlib/pvsystem.py
wholmgren: But the ``ModelChain`` path will go through a ``PVSystem.calcparams_cec/desoto`` method rather than calling the function directly. And we updated the method to use the correct signature. So, I don't see how we could break anyone's code.

cwhanse:
Ready to go, test failure is related to pulling data for the forecast unit tests.

wholmgren:
I still don't understand why we are maintaining compatibility with the old ``calcparams_desoto`` API in the new ``calcparams_cec``. What is the sequence of events that would lead to an issue? 

Within pvlib itself, the ``ModelChain`` path will go through a ``PVSystem.calcparams_cec/desoto`` method rather than calling the function directly, and we already updated those ``PVSystem`` methods to use the new function signature. 

User code that directly uses ``calcparams_desoto`` will emit the warning and will need to be updated. User code that directly uses ``calcparams_desoto`` but should be updated to ``calcparams_cec`` can adapt to the new API at that time. Now that I think about it, perhaps the warning in ``calcparams_desoto`` should also tell people to consider using the ``calcparams_cec`` function.

cwhanse:
I agree with your reason for removing it, I just forgot to do it.

wholmgren:
Great. Merging.

PR #554, opened 2018-09-04
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/pvlib/pvlib-python/ci/#badges) to your project. They will indicate a very high code quality to potential users and contributors.
You can also check the alerts discovered by [LGTM](https://lgtm.com/projects/g/pvlib/pvlib-python).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

PR #561, opened 2018-09-05
follow up LGTM integration

wholmgren:
Follow up to #554


wholmgren:
LGTM shows "no alert changes" which is a little odd because the new ``.lgtm.yml`` file should prevent it from warning about an issue in ``pvlib/_version.py``. I'm guessing it only checks that file on the master branch. I'll merge and we'll find out...

PR #562, opened 2018-09-05
STY: BUG: fix pvsystem imports pvlib twice

mikofski:
* from pvlib import singlediode as _singlediode # to differentiate from
local function pvsystem.singlediode
* replace pvlib.singlediode with _singlediode everywhere
* remove extra line and whitespace around FC_MODEL_PARAMS

this should fix [LGTM alert in `pvsystem.py`](https://lgtm.com/projects/g/pvlib/pvlib-python/alerts/)
>Module 'pvlib' is imported with both 'import' and 'import from'

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -19,31 +19,29 @@ import pandas as pd
 from pvlib import tools
 from pvlib.tools import _build_kwargs
 from pvlib.location import Location
-from pvlib import irradiance, atmosphere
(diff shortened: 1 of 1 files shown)


wholmgren:
Thanks. I thought about doing exactly this in #559 but decided it should probably be done separately.

mikofski:
Actually, you had recommended this change already in [this comment](../pull/525/commits/611f409082655f3d1312ce26320691dae73ea7a1#r208771127) on the original PR :smirk:

Issue #551, opened 2018-08-31
Add clearsky index function to pvlib.irradiance


kevinsa5:
The clearness index kt (the ratio of surface-level GHI to ET irradiance) can be calculated via `pvlib.irradiance.clearness_index` and is used in the DIRINT and DISC models.  Though other metrics in the same vein are not necessary for the irradiance models implemented in pvlib, including them in pvlib could be a convenience to users.  Is there interest in implementing other irradiance metrics for users to access?  Some examples:

- clear-sky index:  the ratio of surface GHI to modeled clear-sky GHI
- Variability Scores as described in this [Sandia paper](https://energy.sandia.gov/wp-content/uploads/dlm_uploads/2016/06/SAND2014-1638J-high_freq_solar_for_dist_studies_preprint.pdf) 
- Variability Index as described in this [Sandia paper, coauthored by Cliff Hansen](https://www.osti.gov/servlets/purl/1068417)

cwhanse:
I think `clearsky_index` is needed.  Would we pass the clear-sky GHI as an argument, or the name of a clear-sky model?  We'd have to agree how to deal with low irradiance conditions, e.g., an upper bound on `clearsky_index`. 

I think the variability calculations are interesting and would be worth adding as functions to the library.

wholmgren:
I am all for including these functions. Usually the best approach is to implement very simple functions that implement the core algorithm, then add convenience functions/methods around them. So, for clearsky index, I'd want a function that doesn't do any more than the division and limits (much like ``clearness_index``). That API would be straightforward and can be integrated into many modeling workflows. Consider something like ``Location.clearsky_index`` if you want the convenience of computing clear sky data within the function/method call.

Please do submit PRs!

cwhanse:
Closed via #571.  If we want to add the variability index, let's open a new issue.

Issue #433, opened 2018-02-20
Use of models beyond the limits for which they were validated or intended


adriesse:
It occurs to me that it could be an interesting feature of PVLIB, to produce a certain class of warnings whenever input parameters fall outside the range for which models were developed or validated.  Could be a latitude range in the case of tilted irradiance models, as a possible example.  Thoughts?

cwhanse:
A good idea, but determining the conditions for validation is probably not practical. It's not always clear what filtering was done on the data.  Also, some models were developed using hourly data, for example, would we warn if they were applied to minute data?

wholmgren:
I agree that could be interesting but challenging. The first step would probably be to add more and better documentation of model limitations to pvlib. Perhaps put these in a special box in the rendered documentation.

thunderfish24:
Wouldn’t it be nice to better know what parts of pvlib are used most? That’s a can of worms too, I suppose.

adriesse:
Good ideas!  I like the warning for mismatch on the time scale too.

Yes, it would be challenging and could probably never be done everywhere and documentation could come first.  But what would speak against doing it for a few easy-ish cases as examples for future coders to follow?  The time scale one could be appropriate for `detect_clearsky()` for example, where the limits are already in the documentation.

I must admit I don't know quite how to do this.  Can you create warning classes similar to exception classes or something?

cwhanse:
I think it needs to be more mild than a warning, more like a FYI





________________________________

wholmgren:
> Can you create warning classes similar to exception classes or something?

See ``PVLibDeprecationWarning`` in the examples in https://github.com/pvlib/pvlib-python/pull/427#issuecomment-366825480

The gist of the warnings in #427 is

```python
import warnings

class PVLibDeprecationWarning(UserWarning):
    pass

message = complicated_stuff_copied_from_matplotlib()
warnings.warn(message, PVLibDeprecationWarning)
```

You could imagine a ``pvlibModelValidationWarning`` defined and used in this simple way. The complicated stuff in #427 is to autofill information from the decorated function, but we'd skip that for this feature (at least initially). 

I think this strategy would be much better than ``print`` or ``logging`` because users that understand the risk can easily silence the warnings using the python warning machinery.

We'd also want to add tests to ensure that the warnings are issued and not issued as appropriate.

mikofski:
[...]

PR #571, opened 2018-09-11
Add new irradiance metrics

kevinsa5:
- [ ] Closes issue #551 

### DIFF
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -1182,6 +1182,39 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
         return sky_diffuse
 
 
+def clearsky_index(ghi, clearsky_ghi, max_clearsky_index=2.0):
+    """
+    Calculate the clearsky index.
+
+    The clearsky index is the ratio of global to clearsky global irradiance.
+
+    Parameters
+    ----------
+    ghi : numeric
+        Global horizontal irradiance in W/m^2.
+
+    clearsky_ghi : numeric
+        Modeled clearsky GHI
+
+    max_clearsky_index : numeric, default 2.0
+        Maximum value of the clearsky index. The default, 2.0, allows
+        for over-irradiance events typically seen in sub-hourly data.
+
+    Returns
+    -------
+    kt : numeric
+        Clearness index
+
+    References
+    ----------
+    .. [1] None
+    """
+    kt = ghi / clearsky_ghi
+    kt = np.maximum(kt, 0)
+    kt = np.minimum(kt, max_clearsky_index)
+    return kt
+
+
 def clearness_index(ghi, solar_zenith, extra_radiation, min_cos_zenith=0.065,
                     max_clearness_index=2.0):
     """


kevinsa5:
Two questions for the pvlib team:  
- Anyone have a citation for the definition of the clear-sky index?  My searching was unsuccessful. 
- I don't have much experience writing tests for numerical functions -- is the approach used in the tests for `irradiance.clearness_index` a good example to follow?  Hardcoding numbers strikes me as bad form, but the alternative would just be to reimplement the function in the test...

cwhanse:
@kevinsa5 also, thanks.  I'd recommend hard-coding some inputs you select, and not running a clear-sky model to generate inputs for the test.

kevinsa5:
Apologies for the long delay on this PR.  Two questions:
- I'm not very experienced in CI, but it seems like the travis failures are related to the test framework, rather than the pvlib code?  
- Also unfamiliar with this stickler bot.  Shall I add the .stickler.tml from the master branch to this PR?

wholmgren:
@kevinsa5 there are new test failures due to a new version of pytest and there is a known failure due to python/numba. Just make sure your tests are passing.

Can you merge the latest pvlib/master into your branch? That should allow stickler to check your code for style/formatting issues.

pvlib/test/test_irradiance.py
```diff
@@ -725,6 +725,43 @@ def test_kt_kt_prime_factor(airmass_kt):
...
 
 
+def test_clearsky_index():
+    ghi = np.array([-1., 0., 1., 500., 1000.])
+    ghi_measured, ghi_modeled = np.meshgrid(ghi, ghi)
+    # default max_clearsky_index
+    with np.errstate(invalid='ignore', divide='ignore'):
+        out = irradiance.clearsky_index(ghi_measured, ghi_modeled)
+    expected = np.array(
+        [[1.   , 0.   , 0.   , 0.   , 0.  ],
```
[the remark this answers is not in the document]
wholmgren: @kevinsa5 you can ignore comments about whitespace within copy/paste arrays.

pvlib/test/test_irradiance.py
```diff
@@ -725,6 +725,43 @@ def test_kt_kt_prime_factor(airmass_kt):
...
+         [0.   , 0.   , 0.001, 0.5  , 1.  ]])
+    assert_allclose(out, expected, atol=0.001)
+    # scalars
+    out = irradiance.clearsky_index(10, 1000)
+    expected = 0.01
+    assert_allclose(out, expected, atol=0.001)
+    # series
+    times = pd.DatetimeIndex(start='20180601', periods=2, freq='12H')
+    ghi_measured = pd.Series([100,  500], index=times)
+    ghi_modeled  = pd.Series([500, 1000], index=times)
```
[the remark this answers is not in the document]
kevinsa5:
```suggestion
    ghi_modeled = pd.Series([500, 1000], index=times)
```

pvlib/irradiance.py
```diff
@@ -1182,6 +1182,39 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    clearsky_ghi : numeric
+        Modeled clearsky GHI
+
+    max_clearsky_index : numeric, default 2.0
+        Maximum value of the clearsky index. The default, 2.0, allows
+        for over-irradiance events typically seen in sub-hourly data.
+
+    Returns
+    -------
+    kt : numeric
```
adriesse: This shouln't be called kt nor 'clearness index' as it creates potential confusion with the function below, which *does* calculate clearness index.

kevinsa5: @adriesse Good catch, thank you.  Careless copy/pasting.  I'm not aware of a shorthand symbol for clear-sky index, do you have any suggestions?

cwhanse: @kevinsa5 I'd spell it out `clearsky_index : numeric` there are some common notations, e.g., Kt and k'_t but not a consensus AFAIK.

pvlib/irradiance.py
```diff
@@ -1194,6 +1194,35 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    max_clearsky_index : numeric, default 2.0
+        Maximum value of the clearsky index. The default, 2.0, allows
+        for over-irradiance events typically seen in sub-hourly data.
+
+    Returns
+    -------
+    clearsky_index : numeric
+        Clearsky index
+    """
+    kt = ghi * 1. / clearsky_ghi
```
cwhanse: rename `kt` to `clearsky_index` in function statements.

kevinsa5: Thanks @cwhanse

cwhanse:
@kevinsa5 we'd like to merge this before v0.6.1 release.  Could you go through the pull request checklist?  I think you'll only need to add entries to api.rst and whatsnew.  Thanks again.

pvlib/irradiance.py
```diff
@@ -1194,6 +1194,35 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+        Maximum value of the clearsky index. The default, 2.0, allows
+        for over-irradiance events typically seen in sub-hourly data.
+
+    Returns
+    -------
+    clearsky_index : numeric
+        Clearsky index
+    """
+    clearsky_index = ghi * 1. / clearsky_ghi
+    clearsky_index = np.maximum(clearsky_index, 0)
```
wholmgren: Maybe the docstring should include a note about this limit. something like "negative input GHI will result in 0 clearsky index."

pvlib/irradiance.py
```diff
@@ -1194,6 +1194,35 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    max_clearsky_index : numeric, default 2.0
+        Maximum value of the clearsky index. The default, 2.0, allows
+        for over-irradiance events typically seen in sub-hourly data.
+
+    Returns
+    -------
+    clearsky_index : numeric
+        Clearsky index
+    """
+    clearsky_index = ghi * 1. / clearsky_ghi
```
wholmgren: How should we handle:

* 0 / 0
* 1 / 0
* -1 / 0

I suggest all return 0.

wholmgren: why `* 1.`? If this is important for something (types? broadcasting?) we need a comment to explain why, otherwise someone may change it when trying to clean up code in the future.

kevinsa5: I followed the lead of `irradiance.clearness_index`, which will return NaN for some inputs.  I'd be happy to add NaN correction here if you think it's appropriate.  

For the `* 1.`, I was worried about the special case of Python `2.x` and the user passing in two integers instead of an np array/pd series.  Is this corner case worth it?  Let me know if I should remove it or document it.

wholmgren: If I remember correctly, `clearness_index` returns 0 for night time values and should only return `nan` when `solar_zenith=90` and `min_cos_zenith` is changed from the default. 

We use `from __future__ import division` at the top of the module to account for the possibility of integer division, so I think the `*1.` is not necessary.

kevinsa5: I updated the behavior to set nan/inf to zero as you suggested, and removed the float conversion -- I had forgotten about the division import.

pvlib/irradiance.py
kevinsa5: Done.

kevinsa5:
I appreciate everyone's patience here.  If there are any other outstanding issues keeping this from being merge-ready, please let me know.

wholmgren:
Looks good to me so long as we change the PR and issue names to something about clearsky index instead of the more generic "irradiance metrics".

pvlib/irradiance.py
```diff
@@ -1198,6 +1198,44 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    max_clearsky_index : numeric, default 2.0
+        Maximum value of the clearsky index. The default, 2.0, allows
+        for over-irradiance events typically seen in sub-hourly data.
+
+    Returns
+    -------
+    clearsky_index : numeric
+        Clearsky index
+    """
+    clearsky_index = ghi / clearsky_ghi
```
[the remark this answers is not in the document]
wholmgren: Good catch thanks. I think in this function we can get by with testing if the input clearsky is small (< .001?).  I suggest we discuss the proposed division function elsewhere.

pvlib/test/test_irradiance.py
```diff
@@ -766,6 +766,43 @@ def test_kt_kt_prime_factor(airmass_kt):
     assert_allclose(out, expected, atol=1e-5)
 
 
+def test_clearsky_index():
+    ghi = np.array([-1., 0., 1., 500., 1000.])
```
wholmgren: Add nan here

pvlib/irradiance.py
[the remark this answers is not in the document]
markcampanelli: I agree that the cutoff exactly at zero might not be sufficient in some cases. I've probably missed some discussions on this, but for example it is unclear to me how the perez model is supposed to behave as both `dni` and `dhi` irradiance get very close to zero: https://github.com/pvlib/pvlib-python/blob/master/pvlib/irradiance.py#L1126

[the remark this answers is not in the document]
cwhanse: Reprise of #433 .  For now let's have the clearsky index function calculate the ratio of GHI to clearsky GHI, with polite treatment of Inf, 0 division and nan, and not assume how very low GHI values should be handled, especially output of clearsky models.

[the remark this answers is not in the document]
wholmgren: I don't think we should assume how very low values of `ghi` should be treated, but I think we can and should make some assumptions about low values of `clearsky_ghi`. For nighttime values, we can expect ratios such as `1 / 0` and I hope we agree that this should return 0 instead of `max_clearsky_index`. Near sunrise/sunset, we could occasionally see ratios such as `1 / 0.001`. I think this should return 0, but I won't object if this returns `max_clearsky_index`.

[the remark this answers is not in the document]
cwhanse: I agree with returning 0 for nighttime as defined by the clearsky GHI. Probably should infer nighttime based on solar angle rather than clearsky GHI but let's do that in a different PR.

I would prefer returning `max_clearsky_index` for large ratios to get this PR to closure. I'd prefer to return 0 for low values of clearsky GHI, but that is asking for another kwarg, e.g., `min_clearksy_ghi` for this function.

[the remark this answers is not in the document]
wholmgren: Ok. Then this PR only needs to be modified to preserve `nan` inputs. Everything else is kicked down the road.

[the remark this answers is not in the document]
kevinsa5: I've seen elsewhere the idiom of multiplying a final result by eg `(ghi / ghi)` to preserve `nan`s in the input.  Is that approach acceptable here as well?  If so, I will add that in and update the tests after work this evening.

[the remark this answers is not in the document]
mikofski: @markcampanelli and @anomam IMHO [`divide_without_warning()`](#discussion_r) seems like a lot of overhead for something that it might be possible to solve using with [Python `warnings`](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) and maybe `np.where()`. I agree this should be considered in a separate PR that handles these types of limits, `NaN` returns, warnings, etc. in a simple, stable, and efficient way that we mostly agree on.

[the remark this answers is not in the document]
wholmgren: > I've seen elsewhere the idiom of multiplying a final result by eg (ghi / ghi) to preserve nans in the input. Is that approach acceptable here as well? 

No objection if commented appropriately. It may or may not be more efficient that other approaches, but let's not worry too much about that unless someone complains.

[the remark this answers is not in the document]
kevinsa5: Multiplying by `(ghi/ghi)` actually introduced new nan values when ghi=0, so I opted to just fill `nan` back into the array where appropriate with the same type-preserving methods as before.

pvlib/irradiance.py
```diff
@@ -1198,6 +1198,48 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    # set +inf, -inf, and nans to zero, while preserving input type
+    # but preserve nans in the input arrays as well
+    input_is_nan = ~np.isfinite(ghi) | ~np.isfinite(clearsky_ghi)
+    if isinstance(clearsky_index, pd.Series):
+        clearsky_index[~np.isfinite(clearsky_index)] = 0
+        clearsky_index[input_is_nan] = np.nan
+    else:
+        clearsky_index = np.where(~np.isfinite(clearsky_index), 0,
+                                  clearsky_index)
+        clearsky_index = np.where(input_is_nan, np.nan, clearsky_index)
```
wholmgren: since the logic is a little more complicated now, I suggest we remove the `if` statement here and only use `np.where`. Then, right before the return statement, add 

```python
if isinstance(ghi, pd.Series): 
    clearsky_index = pd.Series(clearsky_index, index=ghi.index)
```

This way we separate the type handling from the function logic

pvlib/irradiance.py
```diff
@@ -1198,6 +1198,48 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+
+    Returns
+    -------
+    clearsky_index : numeric
+        Clearsky index
+    """
+    clearsky_index = ghi / clearsky_ghi
+    # set +inf, -inf, and nans to zero, while preserving input type
+    # but preserve nans in the input arrays as well
+    input_is_nan = ~np.isfinite(ghi) | ~np.isfinite(clearsky_ghi)
```
wholmgren: why use `isfinite` instead of `isnan` here? To set input `inf` to `nan`? Seems like that is unnecessary and may be harmful in some weird situations.

kevinsa5: Using `isnan` will map input `inf` to a clearsky index of `0`.  It's not clear to me that `0` would be preferable to `nan` for `inf` inputs, but if you think that returning zero is better then that's fine by me.

wholmgren: Ah, I see how that works now. It's fine to leave it as is. It's documented correctly. I suggest that if someone doesn't like it they can make an issue and fix it.

pvlib/irradiance.py
kevinsa5: Good idea, done.

adriesse:
Please let me know whether I understand the code correctly:

 - `clearsky_index = ghi / clearsky_ghi` propagates nans without creating new ones.
 - `clearsky_index = np.maximum(clearsky_index, 0)` pulls -inf to zero
 - `clearsky_index = np.minimum(clearsky_index, max_clearsky_index)` pulls +inf to `max_clearsky_index`

So the extra steps basically serve to pull +inf down to zero rather than to `max_clearsky_index`?

kevinsa5:
`clearsky_index = ghi / clearsky_ghi` will both propagate nans and create new nans -- consider `ghi = clearsky_ghi = 0.`  The extra steps are so input nan/inf are mapped to nan, but any nan/inf created internally result in zero, eg:

    In [9]: clearsky_index(np.array([1,1,np.nan]), np.array([np.nan, 0, 1]))
    Out[9]: array([ nan,   0.,  nan])

PR #573, opened 2018-09-11
fix incorrect nan assignments in singleaxis, allow scalar and 1d array input

### LINKED ISSUE #569: incorrect horizon test in tracker.singleaxis
**Describe the bug**
`    wid[zp <= 0] = np.nan` sets the tracker rotation angle to `np.nan` when the sun is behind the array plane, but the array plane is not yet defined. At this line, the array plane is defined only by the tracker tilt and azimuth, but not its rotation.

**To Reproduce**
```
import pvlib
import numpy as np
import pandas as pd

loc = pvlib.location.Location(latitude=30, longitude=110)
sat = pvlib.tracking.SingleAxisTracker(axis_tilt=90, axis_azimuth=180, backtrack=False, max_angle=180)
sat.localize(location=loc)

solar_azimuth = pd.Series(data=np.arange(0.0, 359.0))
solar_zenith = 45.0 * solar_azimuth / solar_azimuth

df = sat.singleaxis(apparent_zenith=solar_zenith, apparent_azimuth=solar_azimuth)

p1 = plt.plot(df)
plt.legend(df.columns)
```
The plot shows that the tracker rotation `tracker_theta` is limited to between -90 and 90.

**Expected behavior**
In the above example, the sun rotates in azimuth at 45 zenith. The vertical tracker without rotation limit should follow the sun around the complete circle.

**Screenshots**
test_sat

**Versions:**
 - ``pvlib.__version__``: 0.5.2 

**Additional context**

One fix would be to substitute a test for solar zenith, so that the tracker rotation is not set when the sun is below the visible horizon. I also wonder if we should set the rotation to 0 (stow position)  instead of `np.nan` in this case.

`    wid[z <= 0] = np.nan`
`    wid[z <= 0] = 0.0`

wholmgren:
- [x] Closes #569

I punted on the question of nan/0. 

~Would also be nice to remove the requirement that input data be series. Yuck.~ Now accepts scalar/1d input

### DIFF
--- a/pvlib/tracking.py
+++ b/pvlib/tracking.py
@@ -427,14 +427,11 @@ def singleaxis(apparent_zenith, apparent_azimuth,
         # (always positive b/c acosd returns values between 0 and 180)
         wc = np.degrees(np.arccos(temp))
 
-        v = wid < 0
-        widc = np.full_like(wid, np.nan)
-        widc[~v] = wid[~v] - wc[~v]  # Eq 4 applied when wid in QI
-        widc[v] = wid[v] + wc[v]     # Eq 4 applied when wid in QIV
+        # Eq 4 applied when wid in QIV (wid < 0 evalulates True), QI
+        tracker_theta = np.where(wid < 0, wid + wc, wid - wc)
     else:
-        widc = wid
+        tracker_theta = wid
 
-    tracker_theta = widc.copy()
     tracker_theta[tracker_theta > max_angle] = max_angle
     tracker_theta[tracker_theta < -max_angle] = -max_angle
 
--- a/pvlib/tracking.py
+++ b/pvlib/tracking.py
@@ -411,7 +411,7 @@ def singleaxis(apparent_zenith, apparent_azimuth,
     wid = pd.Series(90 - np.degrees(np.arctan2(zp, xp)), index=times)
 
     # filter for sun above panel horizon
-    wid[zp <= 0] = np.nan
+    wid[apparent_zenith > 90] = np.nan
 
     # Account for backtracking; modified from [1] to account for rotation
     # angle convention being used here.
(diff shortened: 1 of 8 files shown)


pvlib/tracking.py
```diff
@@ -545,18 +544,17 @@ def singleaxis(apparent_zenith, apparent_azimuth,
     surface_azimuth[surface_azimuth >= 360] -= 360
 
     # Calculate surface_tilt
-    # Use pandas to calculate the sum because it handles nan values better.
```
wholmgren: The issue referenced in the comment above is that in earlier versions of pandas the sum of a Series (or along one axis of a DataFrame) of all NaNs returned NaN (so long as bottleneck was not installed). That behavior changed a year or two ago, and now ``pd.Series([NaN, NaN]).sum() = 0``. So there is no longer a reason to prefer pandas. I think my original motivation was to preserve input nan values, and that no longer happens.

wholmgren: new ``test_nans`` ensures that input nans are properly handled.

pvlib/tracking.py
```diff
@@ -408,10 +411,11 @@ def singleaxis(apparent_zenith, apparent_azimuth,
 
     # Calculate angle from x-y plane to projection of sun vector onto x-z plane
     # and then obtain wid by translating tmp to convention for rotation angles.
-    wid = pd.Series(90 - np.degrees(np.arctan2(zp, xp)), index=times)
+    wid = 90 - np.degrees(np.arctan2(zp, xp))
 
     # filter for sun above panel horizon
-    wid[zp <= 0] = np.nan
+    zen_gt_90 = apparent_zenith > 90
+    wid[apparent_zenith > 90] = np.nan
```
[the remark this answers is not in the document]
wholmgren: yes, thanks

Issue #114, opened 2016-01-21
add sunrise / sunset to Location


dacoex:
In reference to changes in #93:

add the following functions to `:py:class:`Location``:

`get_sunrise` and `get_sunset` and deferred from these `get_dayduration`.

This helps filtering values, e.g. set night values to `np.nan`.

wholmgren:
Probably a good idea, though it's not likely to happen as part of that PR. Are you aware of the [get_sun_rise_set_transit](https://github.com/pvlib/pvlib-python/blob/master/pvlib/solarposition.py#L312) function? It returns both sunrise and sunset. I'm less interested in ``get_dayduration`` since I think it's less likely to be used and it can easily be calculated if needed (though not totally opposed to it).

All that being said, I've found that filtering using zenith or elevation angle is usually a better approach.

cwhanse:
I'm in need of this functionality, and am looking for opinions about implementation. I'd prefer methods attached to a `PVsystem` or `Location` object that mimic `pyephem`, e.g., `Location.get_sunrise(date)`, `Location.get_next_sunrise(datetime)`.  

The current `get_sun_rise_set_transit` runs `solarposition.spa` and returns only sunrise, sunset and transit.  #316 describes some odd behavior of `get_sun_rise_set_transit` and ideas about fixing it.  Alternatively, we could request sunrise etc. from `solarposition.spa` (which is what `get_sun_rise_set_transit` is doing) or use `solarposition.pyphem` by adding a call to the appropriate method of the Sun object.

I'd like to avoid a workflow that calculates ephemeris twice, which seems to be the likely result if we add an independent `get_sunrise` function. 

Thoughts?

mikofski:
couple of ideas

1. when I fixed the timezone bug in spa, I changed the [_SPA function code_](https://midcdmz.nrel.gov/spa/spa.h) to [`SPA_ALL`](../blob/master/pvlib/spa_c_files/spa_py.pyx#L26) which calculates sunrise, sun-transit, and sunset times, so if it's not implemented in the numpy version of spa, you could use the NREL C version

2. I think sunrise/set is a perfect use for the [analytical zenith function](../blob/master/pvlib/solarposition.py#L1129-L1171), just find where it equals zero, either using a solver, or by solving it analytically:

    ```
    0 = np.arccos(
        np.cos(declination) * np.cos(latitude) * np.cos(hour_angle) +
        np.sin(declination) * np.sin(latitude)
    )
[...]

PR #583, opened 2018-09-19
ENH: geometric calculations for sunrise, sunset, and transit

mikofski:
… and sun transit times

 - [x] Closes issue #114


pvlib/solarposition.py
```diff
@@ -1210,3 +1210,50 @@ def hour_angle(times, longitude, equation_of_time):
     )).total_seconds() / 3600. for t in times])
     timezone = times.tz.utcoffset(times).total_seconds() / 3600.
     return 15. * (hours - 12. - timezone) + longitude + equation_of_time / 4.
+
+
+def _hours(times, hour_angle, longitude, equation_of_time):
+    timezone = times.tz.utcoffset(times).total_seconds() / 3600.
+    return (hour_angle - longitude - equation_of_time/4.)/15. + 12. + timezone
```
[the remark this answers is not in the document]
mikofski: from [pep8: whitespace in expressions](http://pep8.org/#other-recommendations):
>If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator.

pvlib/test/test_solarposition.py
```diff
@@ -499,3 +503,35 @@ def test_analytical_azimuth():
...
+    lat, lon = 39.743, -105.178  # degrees
+    eot = solarposition.equation_of_time_pvcdrom(times.dayofyear)  # minutes
+    decl = solarposition.declination_spencer71(times.dayofyear)  # radians
+    sst = solarposition.sunrise_sunset_transit_analytical(
+        times, latitude=lat, longitude=lon, declination=decl,
+        equation_of_time=eot)
+    sunrise = (sr.time() for sr in sst[0])
+    sunset = (ss.time() for ss in sst[1])
+    transit = (tr.time() for tr in sst[2])
+    sunrise_hours = [sr.hour + (sr.minute + sr.second/60.)/60.
```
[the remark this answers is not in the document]
mikofski: from [pep8: whitespace in expressions](http://pep8.org/#other-recommendations):
>If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator.

pvlib/test/test_solarposition.py
```diff
@@ -499,3 +503,35 @@ def test_analytical_azimuth():
...
+                     for sr in sunrise]
+    sunset_hours = [ss.hour + (ss.minute + ss.second/60.)/60. for ss in sunset]
```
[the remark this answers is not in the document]
mikofski: from [pep8: whitespace in expressions](http://pep8.org/#other-recommendations):
>If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator.

pvlib/test/test_solarposition.py
```diff
@@ -499,3 +503,35 @@ def test_analytical_azimuth():
...
+    transit_hours = [tr.hour + (tr.minute + tr.second/60.)/60.
```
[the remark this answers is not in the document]
mikofski: from [pep8: whitespace in expressions](http://pep8.org/#other-recommendations):
>If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator.

mikofski:
@thunderfish24 ask an ye shall receive

mikofski:
note: [test failure for py3.6](https://travis-ci.org/pvlib/pvlib-python/jobs/430794462)  is unrelated

adriesse:
By the way, the same formula is in Duffie and Beckman and in probably many other solar textbokks!

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,50 @@ def hour_angle(times, longitude, equation_of_time):
...
+    .. warning:: The analytic form neglects the effect of atmospheric
+        refraction.
+
+    Parameters
+    ----------
+    times : :class:`pandas.DatetimeIndex`
+        Corresponding timestamps, must be timezone aware.
+    latitude : float
+        Latitude in degrees
+    longitude : float
```
cwhanse: Please add a note clarifying the longitude sign convention (I'll guess negative is west of prime meridian, but this convention isn't always followed)

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,50 @@ def hour_angle(times, longitude, equation_of_time):
     )).total_seconds() / 3600. for t in times])
     timezone = times.tz.utcoffset(times).total_seconds() / 3600.
     return 15. * (hours - 12. - timezone) + longitude + equation_of_time / 4.
+
+
+def _hours(times, hour_angle, longitude, equation_of_time):
```
wholmgren: I'd like a more descriptive function name or a one line documentation for `_hours` and `_times`.

pvlib/solarposition.py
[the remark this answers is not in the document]
wholmgren: fine with me here and below.

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,50 @@ def hour_angle(times, longitude, equation_of_time):
...
+    timezone = times.tz.utcoffset(times).total_seconds() / 3600.
+    return (hour_angle - longitude - equation_of_time/4.)/15. + 12. + timezone
+
+
+def _times(times, hours):
+    return np.array([(dt.timedelta(hours=h) + t.tz.localize(
+        dt.datetime(t.year, t.month, t.day)
+    )) for h, t in zip(hours, times)])
```
wholmgren: This loop using pure python is likely to be painfully slow and defeat the purpose of an analytical approach.

pvlib/test/test_solarposition.py
```diff
@@ -15,6 +16,9 @@
 from conftest import (requires_ephem, needs_pandas_0_17,
                       requires_spa_c, requires_numba)
 
+DIRNAME = os.path.dirname(__file__)
+PROJDIR = os.path.dirname(DIRNAME)
+DATADIR = os.path.join(PROJDIR, 'data')
```
wholmgren: probably better to put these in the function if they're not going to be reused elsewhere.

pvlib/test/test_solarposition.py
```diff
@@ -499,3 +503,35 @@ def test_analytical_azimuth():
...
+    sunrise = (sr.time() for sr in sst[0])
+    sunset = (ss.time() for ss in sst[1])
+    transit = (tr.time() for tr in sst[2])
+    sunrise_hours = [sr.hour + (sr.minute + sr.second/60.)/60.
+                     for sr in sunrise]
+    sunset_hours = [ss.hour + (ss.minute + ss.second/60.)/60. for ss in sunset]
+    transit_hours = [tr.hour + (tr.minute + tr.second/60.)/60.
+                     for tr in transit]
+    test_data_file = os.path.join(
+        DATADIR, 'sunrise_sunset_transit.txt')
```
wholmgren: do we need such a large data file to comprehensively test the algorithm? I am guessing 10-20 well chosen points is sufficient.

pvlib/test/test_solarposition.py
```diff
@@ -499,3 +503,35 @@ def test_analytical_azimuth():
...
+    test_data_type = np.dtype(
+        [('sunrise', float), ('sunset', float), ('transit', float)])
+    test_data = np.loadtxt(test_data_file, dtype=test_data_type)
+    # test data created using SPA algorithm at following conditions:
```
wholmgren: Can you be more specific? NREL's spa.c file?

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,50 @@ def hour_angle(times, longitude, equation_of_time):
...
+    # solar noon is at hour angle zero
+    # so sunrise is just negative of sunset
+    sunrise_angle = -sunset_angle
+    sunrise_hour = _hours(times, sunrise_angle, longitude, equation_of_time)
+    sunset_hour = _hours(times, sunset_angle, longitude, equation_of_time)
+    transit_hour = _hours(times, 0, longitude, equation_of_time)
+    sunrise = _times(times, sunrise_hour)
+    sunset = _times(times, sunset_hour)
+    transit = _times(times, transit_hour)
+    return (sunrise, sunset, transit)
```
wholmgren: We should try to make the API consistent with the existing function. So this should be changed to support dict or DataFrame output or we can discuss changing the API of the other function elsewhere.

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,50 @@ def hour_angle(times, longitude, equation_of_time):
     )).total_seconds() / 3600. for t in times])
     timezone = times.tz.utcoffset(times).total_seconds() / 3600.
     return 15. * (hours - 12. - timezone) + longitude + equation_of_time / 4.
+
+
+def _hours(times, hour_angle, longitude, equation_of_time):
+    timezone = times.tz.utcoffset(times).total_seconds() / 3600.
```
wholmgren: hmm odd to see `times` as both the object and explicit argument of its own method.

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,50 @@ def hour_angle(times, longitude, equation_of_time):
...
+        Corresponding timestamps, must be timezone aware.
+    latitude : float
+        Latitude in degrees
+    longitude : float
+        Longitude in degrees
+    declination : numeric
+        declination angle in radians at ``times``
+    equation_of_time : numeric
+        difference in time between solar time and mean solar time in minutes
+    """
```
wholmgren: returns?

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,50 @@ def hour_angle(times, longitude, equation_of_time):
...
+                                      equation_of_time):
+    """
+    Analytical calculation of solar sunrise, sunset, and transit.
+
+    .. warning:: The analytic form neglects the effect of atmospheric
+        refraction.
+
+    Parameters
+    ----------
+    times : :class:`pandas.DatetimeIndex`
```
wholmgren: please remove the rst markup here. it detracts from readability when inspecting the function signature outside of the rendered docs.

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,50 @@ def hour_angle(times, longitude, equation_of_time):
...
+def _times(times, hours):
+    return np.array([(dt.timedelta(hours=h) + t.tz.localize(
+        dt.datetime(t.year, t.month, t.day)
+    )) for h, t in zip(hours, times)])
+
+
+def sunrise_sunset_transit_analytical(times, latitude, longitude, declination,
+                                      equation_of_time):
+    """
+    Analytical calculation of solar sunrise, sunset, and transit.
```
wholmgren: I don't care for the term `analytical` but I don't have a better one to suggest. Perhaps the references mentioned have a better name for this?

pvlib/solarposition.py
cwhanse: Also, please add the reference for these calculations.

pvlib/solarposition.py
cwhanse: We may want to clean up the naming convention in this module.  We have prefixes 'spa' and 'nrel' for functions using SPA, we sometimes have 'pyephem' but not always, and use the suffix 'analytical' (I'm Ok with this) for some purely geometric methods.

I prefer <quantity>_<algorithm>, e.g., `declination_cooper69`.  where there are multiple options to calculate quantity.

adriesse: "geometric" might be an alternative to  "analytical"

mikofski: I really like "geometric" but would we change zenith and azimuth to match?

pvlib/solarposition.py
mikofski: it is weird, but I think it's the only way, I broke it up into two lines that explains a bit more:
```python
tz_info = times.tz  # this is the pytz timezone info object that has the utcoffset method
tz = tz_info.utcoffset(times)  # this method requires a datetime-like argument to operate on
```

pvlib/solarposition.py
mikofski: I was thinking that this was a lower level helper function, and that a higher level api function like `get_sunrise` would call it and based on the `how` argument, and then package up the return values into a DataFrame as needed. But happy to change this to output a DataFrame for now if that's what the consensus is.

pvlib/test/test_solarposition.py
mikofski: I guess not, but I thought others might want to use this, it's a full year of hourly sunrise/sunset/transit times from NREL SPA. It's about 600kb, is that too much? Whatever you think is best, maybe ditch it completely, and just test 5-10 values hardcoded in the test?

pvlib/test/test_solarposition.py
mikofski: yes, that's correct

pvlib/solarposition.py
mikofski: I think I can use numpy arrays with `timedelta64[s]` here instead

mikofski: Nice catch @wholmgren , I get about a 70x speed improvement (from 43.5 ms to 605us) to  by using numpy `datetime64` and `timedelta64[ns]`. I wonder where else we could use these? There is an oddity that `hours` is a `Float64Index` from pandas, not just floats, which bothers me, but ok for now.

See this script that shows improvements with numpy over looping.

```python
In [1]: import pandas as pd

In [2]: import datetime as dt

In [3]: import numpy as np

In [4]: d = pd.DatetimeIndex(start='2013-01-01 00:00:00', end='2013-12-31 23:59:59', freq='H', tz='Etc/GMT+8')

In [5]: h = np.loadtxt('../sunrise_sunset_transit.txt', dtype=np.dtype([('sunrise', float), ('sunset', float), ('transit', float)]))

In [6]: sr = pd.Float64Index(h['sunrise'])

In [7]: %paste
def _times(times, hours):
    return np.array([(dt.timedelta(hours=h) + t.tz.localize(
        dt.datetime(t.year, t.month, t.day)
    )) for h, t in zip(hours, times)])

## -- End pasted text --

In [8]: %timeit _times(d, sr)
43.5 ms ± 354 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [9]: del _times

In [10]: %paste
def _times(times, hours):
    """helper converts hours from an array of floats to localized times"""
    tz_info = times.tz
    return pd.DatetimeIndex(
        times.tz_localize(None).values.astype('datetime64[D]')
        + (hours.values * 3600. * 1.e9).astype('timedelta64[ns]'), tz=tz_info)

## -- End pasted text --

In [11]: %timeit _times(d, sr)
605 µs ± 5.07 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,72 @@ def hour_angle(times, longitude, equation_of_time):
...
+    tz = tz_info.utcoffset(times).total_seconds() / 3600.
+    return (hour_angle - longitude - equation_of_time / 4.) / 15. + 12. + tz
+
+
+def _times(times, hours):
+    """helper converts hours from an array of floats to localized times"""
+    tz_info = times.tz
+    return pd.DatetimeIndex(
+        times.tz_localize(None).values.astype('datetime64[D]')
+        + (hours.values * 3600. * 1.e9).astype('timedelta64[ns]'), tz=tz_info)
```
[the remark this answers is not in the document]
mikofski: [pep8: Should a line break before or after a binary operator?](http://pep8.org/#break-before-or-after-binary-operator)
>To solve this readability problem, mathematicians and their publishers follow the opposite convention. Donald Knuth explains the traditional rule in his Computers and Typesetting series:
>>“Although formulas within a paragraph always break after binary operations and relations, displayed formulas always break before binary operations”
>
>**In Python code, it is permissible to break before or after a binary operator, as long as the convention is consistent locally. For new code Knuth's style is suggested.**

_emphasis mine_

mikofski:
Notes:
* travis py35 failure was due to bad http request that timed out
* the [py27-min fails](https://travis-ci.org/pvlib/pvlib-python/jobs/432341775#L960-L986) because in pandas-0.14, the minimum requirement, you can't reset a tz-aware datetime

do new features have to support the minimum requirements? I can see that patching old code should be backwards compatible, but if you are installing pvlib now to use an analytical sunrise/sunset equation, then you will have the newest pandas

pvlib/solarposition.py
cwhanse: I'm in favor of returning a pandas object rather than the tuple, and also, the intermediate function `get_next_sunrise` with a `how` argument that wraps the geometric, pyephem and spa methods.  The input is a tz-aware datetime and location.  `get_sunrise` is somewhat ambiguous: assume the datetime is solar noon at the location.  Is the desired output the sunrise of the current day, or the next sunrise (the following day)? `get_sunrise(date)` clarifies things but requires the function to localize date (for pyephem) or determine the corresponding date in UTC time (for spa).  For these reasons I'm suggesting the `next` approach.

To complete the function set we'd add `get_previous_sunrise`, `get_next_sunset`, etc.

mikofski:
@wholmgren I am seeing some backwards compatibility issues on Travis with numpy and pandas minimum requirements
* [pandas-014.0, `TypeError: Already tz-aware, use tz_convert to convert.`](#issuecomment-423892716)
* [numpy-1.10.1](https://travis-ci.org/pvlib/pvlib-python/jobs/432355251), `TypeError: Cannot cast ufunc add input from dtype('<M8[D]') to dtype('<M8[ns]') with casting rule 'same_kind'`
* [numpy-1.11.3, pandas-0.20.3](https://travis-ci.org/pvlib/pvlib-python/jobs/432355253), `TypeError: Setting <class 'pandas.core.indexes.numeric.Float64Index'> dtype to anything other than float64 or object is not supported`

Since there are so many API changes in v0.6 can we consider maybe updating some of the minimum requirements now? Pandas is currently v0.23.4 and NumPy is at 1.15, keeping really old requirements seems limiting to me a little. Thanks.

mikofski:
Also, can we maybe consider dropping support for Python-3.4?

wholmgren:
> Since there are so many API changes in v0.6 can we consider maybe updating some of the minimum requirements now? Pandas is currently v0.23.4 and NumPy is at 1.15, keeping really old requirements seems limiting to me a little.

I am all in favor of raising the minimums if there is a reason to do so. But I take these incompatibilities as sign that the approach might be too complicated and will be difficult to maintain in the future. Let's think about if this is really the implementation that we want.

Somewhere you commented something to the effect of the spec'ed minimum numpy/pandas versions are incompatible -- is it that they are incompatible or that Continuum doesn't supply pre-built binaries for them?

> Also, can we maybe consider dropping support for Python-3.4?

We can consider this too but not in this pull request.

mikofski:
it seems to be working now, only had to make 2 minor changes to satisfy the CI
* ducktype `hours` for different versions of pandas that return either float or `Float64index`
* cast the date back to nano-seconds for different versions of numpy that require safe casting of inputs to ufunc add

>Somewhere you commented something to the effect of the spec'ed minimum numpy/pandas versions are incompatible -- is it that they are incompatible or that Continuum doesn't supply pre-built binaries for them?

I was wrong, you were right, [according to `setup.py` pandas-0.14.0 min npy ver is 1.6.1 for py2 or 1.7.0 for py3](https://github.com/pandas-dev/pandas/blob/v0.14.0/setup.py#L38)

>assume the input time is localized or in UTC

I am already assuming that the times must be tz-aware and that they have a tz or tz_info field

>convert to ints instead of datetime64.

good idea, like Linux time, can we pursue this as a course of last resort? The calculations seem to be working now, I believe that internally that is what `timedelta64` already does, since it is `int` already.

>utilize the same helper functions that `irradiance.get_extra_radiation` uses.

Do you mean `_handle_extra_radiation_types`, I think this could be useful for handling the input if it's not a tz-aware pandas `DatetimeIndex`, but my view of the API is that would be handled by the `get_*` function for all `how` args, assuming that `sunrise_sunset_transit_analytical` is a low level function called by a `get_sunrise...` high level function. ok?

otherwise, (minus what's new) I think this is ready for review

mikofski:
>I am all in favor of raising the minimums if there is a reason to do so.

BTW: pandas-0.14.0 is from 2014, IMO we should not support this because it's not a v1.0 release, so a version that is 4 years old may have more issues than just incompatibility.

>But I take these incompatibilities as sign that the approach might be too complicated and will be difficult to maintain in the future. Let's think about if this is really the implementation that we want.

did you see [this comment](#discussion_r219738771), the simpler approach was 70x slower, I don't want to create a maintenance headache, but I don't this approach is too complicated.

pvlib/solarposition.py
mikofski: OK, I like it! Everyone else in agreement? It'll have to wait until after SPI, but I'll try to get back to this asap. Thanks!

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,84 @@ def hour_angle(times, longitude, equation_of_time):
...
+        tz=tz_info)
+
+
+def sunrise_sunset_transit_analytical(times, latitude, longitude, declination,
+                                      equation_of_time):
+    """
+    Analytical calculation of solar sunrise, sunset, and transit.
+
+    .. warning:: The analytic form neglects the effect of atmospheric
+        refraction.
```
wholmgren: Probably "geometric" instead of analytic. Also guessing it neglects more than refraction, but haven't looked at the sources.

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,84 @@ def hour_angle(times, longitude, equation_of_time):
...
+        times.tz = None
+    except AttributeError:
+        # NEW pandas, "AttributeError: Cannot directly set timezone. Use
+        # tz_localize() or tz_convert() as appropriate"
+        times = times.tz_localize(None)
+    return pd.DatetimeIndex(
+        times.values.astype('datetime64[D]').astype('datetime64[ns]') +
+        (hours * 3600. * 1.e9).astype('timedelta64[ns]'),
+        tz=tz_info)
+
```
cwhanse: Maybe naive, but why not

`return times + pd.to_timedelta(hours * 3600. * 1.e9)`

pvlib/solarposition.py
wholmgren: I like "geometric". We should use "geometric" for the new function in this PR, but let's take on the deprecation of the existing functions in a separate PR.

pvlib/solarposition.py
cwhanse: and if that works, then I don't see the value of `_hours`, just do the addition in `sunrise_sunset_transit_analytical`

pvlib/solarposition.py
cwhanse: Also spherical earth, sun is represented by a point at its center, ...

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,84 @@ def hour_angle(times, longitude, equation_of_time):
...
+    tz = tz_info.utcoffset(times).total_seconds() / 3600.
+    return (hour_angle - longitude - equation_of_time / 4.) / 15. + 12. + tz
+
+
+def _times(times, hours):
+    """helper converts hours from an array of floats to localized times"""
+    tz_info = times.tz
+    try:
+        hours = hours.values  # make a new reference
+    except AttributeError:
```
wholmgren: I don't understand the point of this try/except. Is `hours` not always the same type? If not, suggest we change `_hours()` to do the right thing with something like 

```python
hours = # your eqn
# enforce Series
hours = pd.Series(hours, index=times)
# OR enforce array
hours = np.array(hours)
return hours
```

Then we can remove this try except.

pvlib/solarposition.py
mikofski: The times have to be reduced to dates first, while preserving their `tz_utcoffset`. There are different ways to do that, my approach here is to set `tz` to `None`, which effectively makes is naive local time, then cast to numpy `datetime64[D]` which is a date.

I'm sure there are a plethora of ways, this one is working and faster than the looping method, so is it okay?

pvlib/solarposition.py
mikofski: no apparently depending on the pandas version it changes, sometimes it's a numpy array of floats and sometimes it's a pandas series of `pandas.Float64Index`, I didn't bother to track down which versions are which, but it seems to be in the 0.teen versions, I think >0.20 yields series. I'm sure I could enforce the type somewhere else, as an output of some calculation, but I didn't bother to

pvlib/solarposition.py
wholmgren: Here and in the other "analytical"/"geometrical" functions should we give an estimate of the typical magnitude of the errors? Something like "The error depends on location and time of year but is of order 10 minutes."

pvlib/solarposition.py
```diff
@@ -1210,3 +1210,84 @@ def hour_angle(times, longitude, equation_of_time):
...
+        pass
+    try:
+        # OLD pandas, "TypeError: Already tz-aware, use tz_convert to convert."
+        times.tz = None
+    except AttributeError:
```
wholmgren: I also don't understand this try/except. Your API declaration requires a localized DatetimeIndex, so you can use `times.tz_convert` is guaranteed to work. This is true for pandas 0.14.

mikofski: `tz_convert` changes the timezone from one to another, ie: it applies the `tz_utcoffset` so eg: from `Etc/GMT+8` to `UTC` would add 8 hours to all of the time.

But that's not what we want. We just want to make the tz-aware time a local naive time, so we want to effectively replace `tz_info` with `None`.

Again in some older version of pandas, you could just do this by setting `tz=None`, and in newer ones you can call `tz_localize(None)` to convert a tz-aware time to a naive one, *without* applying a utc offset.

EG: `2017-01-01T12:30:00-0800` just becomes `2017-01-01T12:30:00` same time, but no more timezone

pvlib/solarposition.py
mikofski: can you figure out where hours becomes a series? I'm not sure where it happens?

pvlib/solarposition.py
cwhanse: Since `times.values.astype('datetime64[D]') extracts the date from `times`, then it is implicit that `times` is localized to the timezone of the desired location.  Otherwise, the algorithm can mix the hour with a different date.  E.g., 

```
ts = pd.DatetimeIndex(['2016-9-20 12:00:00', '2016-9-20 21:00:00']).tz_localize('MST')
tsutc = ts.tz_convert('UTC')
print(ts.values.astype('datetime64[D]'))
['2016-09-20' '2016-09-21']
print(tsutc.values.astype('datetime64[D]'))
['2016-09-20' '2016-09-21']
```

and `_hours` adds the hour value to these dates.

This is also the problem with `get_sun_rise_set_transit`

pvlib/solarposition.py
mikofski: from the [old test](../../../mikofski/pvlib-python/blob/3b0e59757a8f95e766b1d99dfa77509e113f61d2/pvlib/test/test_solarposition.py#L535-L537) the max differences are

* sunrise 0.11 hours
* sunset 0.11 hours
* transit 0.02 hours

this is a rough estimate, only to 2 decimals, I manually iterated the 2nd decimal to see what tolerance I could get.

Also the errors are one sided, ie: for sunrise the geometric estimate is always *earlier* up to 0.11 hours, and down to some positive, non-zero number, so there's always some sunrise error that is a minute or so earlier than the apparent sunrise, and ie: for sunset it's always *later* up to 0.11 hours, and down to a positive, non-zero number, so there's always some sunset error that is a minute or so later than the apparent sunset.

pvlib/solarposition.py
mikofski: >If I'm correct then the input description of sunrise_sunset_transit_analytical must clarify the restriction on the timezone.

Yes sorry, I should clarify that in the docstring, this function does assume that the times are tz-aware. I mentioned that in [this comment](#issuecomment-424080433).

I guess I thought that this was already a requirement for all pvlib times, not sure why I thought that.

I could relax this requirement though. If the times are naive, and local, then that actually makes the calculations easier, b/c we just use the dates as is, no need to replace `tz` with `None` to convert it to naive if it already is, right?

cwhanse: Please disregard the comment I made earlier and then edited while you were replying.  Not a good GitHub workflow on my part.

I think we must require that `times` is localized for this function.  It appears that numpy.datetime64[D] handles the timezone adjustment.  I don't think it is worth the trouble to extend this function to naïve datetimes.

mikofski: it does say it in the docstring already [here](../../../mikofski/pvlib-python/blob/analytical_sunrise_sunset_transit/pvlib/solarposition.py#L1252-L1253)

    times : pandas.DatetimeIndex
        Corresponding timestamps, must be timezone aware.

pvlib/solarposition.py
wholmgren: I understand now, thanks. 

> newer ones you can call tz_localize(None) to convert a tz-aware time to a naive one, without applying a utc offset.

Is this a documented behavior of `tz_localize`? I can't find official support for it. If not I recommend a different implementation.

mikofski:
[thanks for feedback!](#discussion_r219975348), tests are passing! all good now?

pvlib/solarposition.py
mikofski: @wholmgren it's in the pandas documentation for [`DatetimeIndex.tz_localize`](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DatetimeIndex.tz_localize.html#pandas.DatetimeIndex.tz_localize)
>Passing `None` will remove the time zone information preserving local time.

It's and also [here](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.dt.tz_localize.html#pandas.Series.dt.tz_localize):
>Passing `None` will remove the time zone information preserving local time.

and [there](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Timestamp.tz_localize.html)
>None will remove timezone holding local time.

but neither [here](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.tz_localize.html#pandas.Series.tz_localize)

nor [there](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.tz_localize.html#pandas.DataFrame.tz_localize)

@cwhanse [`normalize`](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.dt.normalize.html#pandas.Series.dt.normalize) seems very interesting, and perhaps what we're both looking for? [Since v0.16](https://github.com/pandas-dev/pandas/pull/10047), see [what's new](https://pandas.pydata.org/pandas-docs/version/0.16/whatsnew.html).
>Convert times to midnight.
>The time component of the date-timeise converted to midnight i.e. 00:00:00. This is useful in cases, when the time does not matter. Length is unaltered. The timezones are unaffected.

wholmgren:
where is this conversation in github??? these line by line conversations
are great until they're horrible.

I guess I was looking at the neither/nor docs. Normalize seems useful, but
does it do something different than .floor('1D')? I'm ok bumping the pandas
version to support this.

wholmgren:
If I understand correctly, pandas 0.16 adds normalize for Timestamps and Series. 0.14 does have it for DatetimeIndex:

```python
>>> times = pd.DatetimeIndex(start='20180924 1600', end='20180924 1700', freq='1H', tz='America/Phoenix')
>>> times
<class 'pandas.tseries.index.DatetimeIndex'>
[2018-09-24 16:00:00-07:00, 2018-09-24 17:00:00-07:00]
Length: 2, Freq: H, Timezone: America/Phoenix
>>> times.normalize()
<class 'pandas.tseries.index.DatetimeIndex'>
[2018-09-24 00:00:00-07:00, 2018-09-24 00:00:00-07:00]
Length: 2, Freq: None, Timezone: America/Phoenix
>>> pd.__version__
'0.14.0'
```

mikofski:
Just wondering if this okay as is or do you want more changes? Thanks!

cwhanse:
@mikofski @wholmgren `normalize` has the same behavior as `DatetimeIndex.date`, it sets hour, minute, second to zero without adjusting the date for the timezone offset. That's a problem for the sunrise/sunset calculation.

cwhanse:
@mikofski I would like to see a more descriptive name for `_hours` and `_times`, perhaps `_hour_angle_to_hour` and `_times_to_midnight_utc`

wholmgren:
> normalize has the same behavior as DatetimeIndex.date, it sets hour, minute, second to zero without adjusting the date for the timezone offset. That's a problem for the sunrise/sunset calculation.

Right, but maybe more efficient/straightforward than going through `DatetimeIndex.date` which generates a bunch of python date objects. My understanding was the pandas `tz = None`/`tz_localize(None)` trick solves the time zone offset issue but that's not actually clear from the test. I think we want the same set of tests that are in progress in #588. We want to quantify the  set of function behavior for localized/unlocalized inputs and before/after the UTC offset for the next day. I would hope some of these tz handling functions in `solarposition.py` and test functions in `test_solarposition.py` could share code.

mikofski:
I did test that numpy `datetime[D]` and `tz=None`, keeps the dates, without shifting them by UTC offset. Look at part of this example from [this comment](#discussion_r219738771)

```python
In [1]: import pandas as pd

In [2]: import datetime as dt

In [3]: import numpy as np

In [4]: d = pd.DatetimeIndex(start='2013-01-01 00:00:00', end='2013-12-31 23:59:59', freq='H', tz='Etc/GMT+8')

In [5]: dt = d.tz_localize(None)

In [6]: t = dt.values.astype('datetime64[D]')

In [7]: d[:24]  # the original localized datetime indices with tz=Etc/GMT+8
Out[7]:
DatetimeIndex(['2013-01-01 00:00:00-08:00', '2013-01-01 01:00:00-08:00',
               '2013-01-01 02:00:00-08:00', '2013-01-01 03:00:00-08:00',
               '2013-01-01 04:00:00-08:00', '2013-01-01 05:00:00-08:00',
               '2013-01-01 06:00:00-08:00', '2013-01-01 07:00:00-08:00',
               '2013-01-01 08:00:00-08:00', '2013-01-01 09:00:00-08:00',
               '2013-01-01 10:00:00-08:00', '2013-01-01 11:00:00-08:00',
               '2013-01-01 12:00:00-08:00', '2013-01-01 13:00:00-08:00',
               '2013-01-01 14:00:00-08:00', '2013-01-01 15:00:00-08:00',
               '2013-01-01 16:00:00-08:00', '2013-01-01 17:00:00-08:00',
               '2013-01-01 18:00:00-08:00', '2013-01-01 19:00:00-08:00',
               '2013-01-01 20:00:00-08:00', '2013-01-01 21:00:00-08:00',
               '2013-01-01 22:00:00-08:00', '2013-01-01 23:00:00-08:00'],
              dtype='datetime64[ns, Etc/GMT+8]', freq='H')

In [8]: dt[:24]  # with tz=None, NOTE: same times, same dates, but no timezone
Out[8]:
DatetimeIndex(['2013-01-01 00:00:00', '2013-01-01 01:00:00',
               '2013-01-01 02:00:00', '2013-01-01 03:00:00',
               '2013-01-01 04:00:00', '2013-01-01 05:00:00',
               '2013-01-01 06:00:00', '2013-01-01 07:00:00',
               '2013-01-01 08:00:00', '2013-01-01 09:00:00',
               '2013-01-01 10:00:00', '2013-01-01 11:00:00',
               '2013-01-01 12:00:00', '2013-01-01 13:00:00',
               '2013-01-01 14:00:00', '2013-01-01 15:00:00',
               '2013-01-01 16:00:00', '2013-01-01 17:00:00',
               '2013-01-01 18:00:00', '2013-01-01 19:00:00',
               '2013-01-01 20:00:00', '2013-01-01 21:00:00',
               '2013-01-01 22:00:00', '2013-01-01 23:00:00'],
              dtype='datetime64[ns]', freq='H')

In [9]: t[:24]  # same datetime indices cast as numpy datetime[D], NOTE: just dates, no times
Out[9]:
array(['2013-01-01', '2013-01-01', '2013-01-01', '2013-01-01',
       '2013-01-01', '2013-01-01', '2013-01-01', '2013-01-01',
       '2013-01-01', '2013-01-01', '2013-01-01', '2013-01-01',
       '2013-01-01', '2013-01-01', '2013-01-01', '2013-01-01',
       '2013-01-01', '2013-01-01', '2013-01-01', '2013-01-01',
       '2013-01-01', '2013-01-01', '2013-01-01', '2013-01-01'],
      dtype='datetime64[D]')

In [10]: (dt.values[:24] - t[:24]).astype(float) / 1.e9 / 3600.  # timedelta shows hours since midnite
Out[10]:
array([ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10., 11., 12.,
       13., 14., 15., 16., 17., 18., 19., 20., 21., 22., 23.])

In [11]: d[24:48]  # day 2
Out[11]:
DatetimeIndex(['2013-01-02 00:00:00-08:00', '2013-01-02 01:00:00-08:00',
               '2013-01-02 02:00:00-08:00', '2013-01-02 03:00:00-08:00',
               '2013-01-02 04:00:00-08:00', '2013-01-02 05:00:00-08:00',
               '2013-01-02 06:00:00-08:00', '2013-01-02 07:00:00-08:00',
               '2013-01-02 08:00:00-08:00', '2013-01-02 09:00:00-08:00',
               '2013-01-02 10:00:00-08:00', '2013-01-02 11:00:00-08:00',
               '2013-01-02 12:00:00-08:00', '2013-01-02 13:00:00-08:00',
               '2013-01-02 14:00:00-08:00', '2013-01-02 15:00:00-08:00',
               '2013-01-02 16:00:00-08:00', '2013-01-02 17:00:00-08:00',
               '2013-01-02 18:00:00-08:00', '2013-01-02 19:00:00-08:00',
               '2013-01-02 20:00:00-08:00', '2013-01-02 21:00:00-08:00',
               '2013-01-02 22:00:00-08:00', '2013-01-02 23:00:00-08:00'],
              dtype='datetime64[ns, Etc/GMT+8]', freq='H')

In [9]: dt[24:48]  # day 2
Out[9]:
DatetimeIndex(['2013-01-02 00:00:00', '2013-01-02 01:00:00',
               '2013-01-02 02:00:00', '2013-01-02 03:00:00',
               '2013-01-02 04:00:00', '2013-01-02 05:00:00',
               '2013-01-02 06:00:00', '2013-01-02 07:00:00',
               '2013-01-02 08:00:00', '2013-01-02 09:00:00',
               '2013-01-02 10:00:00', '2013-01-02 11:00:00',
               '2013-01-02 12:00:00', '2013-01-02 13:00:00',
               '2013-01-02 14:00:00', '2013-01-02 15:00:00',
               '2013-01-02 16:00:00', '2013-01-02 17:00:00',
               '2013-01-02 18:00:00', '2013-01-02 19:00:00',
               '2013-01-02 20:00:00', '2013-01-02 21:00:00',
               '2013-01-02 22:00:00', '2013-01-02 23:00:00'],
              dtype='datetime64[ns]', freq='H')

In [6]: t[24:48]  # day 2
Out[6]:
array(['2013-01-02', '2013-01-02', '2013-01-02', '2013-01-02',
       '2013-01-02', '2013-01-02', '2013-01-02', '2013-01-02',
       '2013-01-02', '2013-01-02', '2013-01-02', '2013-01-02',
       '2013-01-02', '2013-01-02', '2013-01-02', '2013-01-02',
       '2013-01-02', '2013-01-02', '2013-01-02', '2013-01-02',
       '2013-01-02', '2013-01-02', '2013-01-02', '2013-01-02'],
      dtype='datetime64[D]')

In [13]: (dt.values[24:48] - t[24:48]).astype(float) / 1.e9 / 3600.  # day 2
Out[13]:
array([ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10., 11., 12.,
       13., 14., 15., 16., 17., 18., 19., 20., 21., 22., 23.])
```

ok?

cwhanse:
Same hope. Let’s get one implementation that works, then we can tune it up and reuse it.

Sent from my iPhone

On Sep 24, 2018, at 7:40 PM, Will Holmgren <notifications@github.com<mailto:notifications@github.com>> wrote:


normalize has the same behavior as DatetimeIndex.date, it sets hour, minute, second to zero without adjusting the date for the timezone offset. That's a problem for the sunrise/sunset calculation.

Right, but maybe more efficient/straightforward than going through DatetimeIndex.date which generates a bunch of python date objects. My understanding was the pandas tz = None/tz_localize(None) trick solves the time zone offset issue but that's not actually clear from the test. I think we want the same set of tests that are in progress in #588<https://github.com/pvlib/pvlib-python/pull/588>. We want to quantify the set of function behavior for localized/unlocalized inputs and before/after the UTC offset for the next day. I would hope some of these tz handling functions in solarposition.py and test functions in test_solarposition.py could share code.

mikofski:
>This PR and #588 should coordinate on what the tests look like.

@cwhanse how do you want to proceed? I'll follow your lead

cwhanse:
@mikofski about testing the sunrise, sunset and transit functions.  First we need to be crystal clear what behavior is expected.  For the geometric and spa models, the behavior (I believe) is: given datetime and location, return sunrise and sunset for the given date at the given location.  The challenge is that datetime may not be localized to location, so simply parsing the date for the input datetime without adjusting for timezone offset does not work. Also, the spa function requires input in UTC only, so the pvlib wrapper for the spa function must handle the timezone issue. I don't know if the geometric model requires its datetime input to be localized to the location, or if it accepts input in any timezone.

The pyephem functions have different behavior.  From input datetime and location, they return the next (or previous) sunrise (sunset) time at the location, handling different timezones.

I have observed that sunrise time from spa and pyephem differ enough that even when rounded to whole minutes, they provide different sunrise/sunset times, even when I have tried the same definition of sunrise (edge of disc) and atmosphere parameters (pressure and air temperature). So in #588 there are different test fixtures for spa and pyephem sunrise/sunset tests.  This is OK because the fixtures are referencing independent calculations by NREL SPA online and the US Naval Observatory.

So what does this mean?  We need a known, correct solution for the geometric model.  If that comes from spa, that is OK with me.  I'm ok with a geometric model test fixture because the geometric model uses different assumptions than are typical for spa.

As far as test cases, here's my thinking:
- single location, test for sunrise/sunset given a time before sunrise, a time between sunrise and transit, a time between transit and sunset, and a time after sunset but before midnight the next day.
- same location, but localize all times to UTC.
- same location, but localize all times to an opposing longitude (e.g., east instead west of meridian).
- a single time test at a location at a location at an opposing longitude.
- a single time test at a location in the southern hemisphere.

mikofski:
>I don't know if the geometric model requires its datetime input to be localized to the location, or if it accepts input in any timezone.

currently [geometric calc](../../../mikofski/pvlib-python/blob/analytical_sunrise_sunset_transit/pvlib/solarposition.py#L1247-L1277) requires times to be localized to the latitude and longitude of the site, not any other timezone.

EG: if latitude and longitude are (39, -105) ) as in the [test example](../../../mikofski/pvlib-python/blob/analytical_sunrise_sunset_transit/pvlib/test/test_solarposition.py#L526-L531), then the timzone must have a UTC offset of -7 hours, corresponding to that location in the US/Mountain time zone.

>We need a known, correct solution for the geometric model. If that comes from spa, that is OK with me. I'm ok with a geometric model test fixture because the geometric model uses different assumptions than are typical for spa.

the current [test example](../../../mikofski/pvlib-python/blob/analytical_sunrise_sunset_transit/pvlib/test/test_solarposition.py#L526-L531) was validated against spa.c, from which the geometric calc which had a max difference of approx. 6 minutes at sunrise/sunset, and max difference of approx 1-min at transit time, but I [removed that data](../../../mikofski/pvlib-python/commit/af206f2dcff7291fff7b2c1451d21e2317ec485f) in lieu of static hard coded values from the current calculation output  I think that's okay, because it was vetted against spa.c originally.

- [x] single location, test for sunrise/sunset given a time before sunrise, a time between sunrise and transit, a time between transit and sunset, and a time after sunset but before midnight the next day. *this is what [the original test]((../../../mikofski/pvlib-python/commit/af206f2dcff7291fff7b2c1451d21e2317ec485f) did, since it just checked the entire year, all 8760 times*
- [ ] same location, but localize all times to UTC. *this wouldn't work for current geometric implementation b/c times must be localized to the timezone of the location*

cwhanse:
> currently geometric calc requires times to be localized to the latitude and longitude of the site, not any other timezone.

Same constraint as the spa sunrise function, then.  That's OK with me.  We should ensure this constraint is clear in the definition of inputs.

Tests 1, 4 and 5.  For #1, the test is more easily maintained with a short, explicit list of times, rather than selecting every 438th time from the year.  OK with me if you also hardwire the solution that has already been verified.

mikofski:
@cwhanse shall I just reuse your test fixtures?

mikofski:
note, failure is unrelated, the [Travis Python-2.7 job](https://travis-ci.org/pvlib/pvlib-python/jobs/435175311) stalled

pvlib/test/test_solarposition.py
```diff
@@ -499,3 +498,58 @@ def test_analytical_azimuth():
     azimuths = solarposition.solar_azimuth_analytical(*test_angles.T, zenith=zeniths)
 
     assert not np.isnan(azimuths).any()
+
+
+@pytest.fixture()
+def expected_rise_set_geometric():
+    # for Golden, CO, from USNO websites
```
cwhanse: Update source for solution, this isn't from USNO since that site only displays minutes

cwhanse:
> maybe let's wait to merge this until after #588, because I think it would be better to use the existing fixtures

OK with me.

mikofski:
@wholmgren , @cwhanse , ready to merge I think,
- I reused the SPA fixture, and set `atol` to be roughly 6-minutes for sunset/sunrise, and about 1-minute for transit,
- also moved `_times_to_hours` from `test_solarposition.py` to `solarposition.py` and changed to the approach Will suggested (pandas>=0.15, `normalize`, and `astype(np.int64)`) which seems to work best
- had to finagle the columns in `expected_rise_set_spa` from numpy datetimes at UTC to the pandas `DatetimeIndex` localized to `golden_MST`
- still seeing the build error in py27-min which is still set to use pandas==0.14

mikofski:
question: should we define

    NS_PER_HR = 1.e9 * 3600.

somewhere, to save time? it's calculated in a few places now

wholmgren:
Thanks @mikofski. Ok with me if you want to define `NS_PER_HR` somewhere in `solarposition.py`.

I'd like to see this PR tested against the pandas 0.15 changes in #595. Probably easiest to wait until that goes in. I want to test that one myself, probably on Wednesday. Sorry for the backlog.

midnite or midnight? Seems like we should prefer the standard spelling.

I'm less confident working with the `datetime64` types than with integers, but ok with me if you think it should work reliably. 

Can you revise the name for this PR? 

Also a few items on the checklist to be completed.

pvlib/solarposition.py
```diff
@@ -1329,9 +1331,96 @@ def hour_angle(times, longitude, equation_of_time):
...
+
+    .. warning:: The geometric calculation assumes a circular earth orbit with
+        the sun as a point source at its center, and neglects the effect of
+        atmospheric refraction on zenith. The error depends on location and
+        time of year but is of order 10 minutes.
+
+    Parameters
+    ----------
+    times : pandas.DatetimeIndex
+        Corresponding timestamps, must be timezone aware.
```
cwhanse: The description of `times` needs to specify localized to the timezone for `latitude` and `longitude`, it would help to state that sunrise, sunset and transit are calculated for the date part of `times`.

docs/sphinx/source/whatsnew/v0.6.1.rst
```diff
@@ -16,27 +16,35 @@ API Changes
...
-* :func:`~pvlib.solarposition.rise_set_transit_ephem` returns sunrise, sunset and transit times using pyephem
-* Created :py:func:`pvlib.iotools.read_srml` and :py:func:`pvlib.iotools.read_srml_month_from_solardat`
-  to read University of Oregon Solar Radiation Monitoring Laboratory data. (:issue:`589`)
-  
+* :func:`~pvlib.solarposition.rise_set_transit_ephem` returns sunrise, sunset
+  and transit times using pyephem (:issue:`114`)
+* Created :py:func:`pvlib.iotools.read_srml` and
+  :py:func:`pvlib.iotools.read_srml_month_from_solardat` to read University of
+  Oregon Solar Radiation Monitoring Laboratory data. (:issue:`589`)
+* Add geometric functions for sunrise, sunset, and sun transit times,
```
cwhanse: Wasn't going to pick this nit but since I have other comments, let's sort this list to group the sunrise etc. changes together and the `iotools` items together.

mikofski:
@cwhanse thanks for your feedback, I believe I've addressed your comments.

mikofski:
FYI: the [appveyor error is unrelated, caused by http error](https://ci.appveyor.com/project/wholmgren/pvlib-python-fv2to/builds/19377349/job/7kmg2jdal1gnmabd#L220)

>An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.

cwhanse:
@wholmgren OK to merge?

wholmgren:
Thanks Mark!

PR #584, opened 2018-09-20
add v0.6.1 whatsnew and update zenodo badge

wholmgren:
Adds a new whatsnew file for v0.6.1 and updates the zenodo badge to point to 0.6.0

Builds ok locally so will go ahead and merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ pvlib-python
 [![Coverage Status](https://img.shields.io/coveralls/pvlib/pvlib-python.svg)](https://coveralls.io/r/pvlib/pvlib-python)
 [![codecov](https://codecov.io/gh/pvlib/pvlib-python/branch/master/graph/badge.svg)](https://codecov.io/gh/pvlib/pvlib-python)
 [![Documentation Status](https://readthedocs.org/projects/pvlib-python/badge/?version=latest)](http://pvlib-python.readthedocs.org/en/latest/)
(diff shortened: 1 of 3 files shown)


mikofski:
Just curious, why did you create this PR? did you expect comments? Why not just make this commit straight in master?

wholmgren:
The main reason I went through the PR is so that people can view the changes in the same github PR interface that most things go through. Sometimes I catch problems seeing the diff here (in addition to viewing the diff locally). I did not expect comments. I'm not sure what the right approach is for changes like this. I always feel bad pushing directly to pvlib/master. Your thoughts?

mikofski:
>The main reason I went through the PR is so that people can view the changes in the same github PR interface that most things go through. Sometimes I catch problems seeing the diff here (in addition to viewing the diff locally).

That makes a lot of sense to me, thanks!

>I always feel bad pushing directly to pvlib/master. Your thoughts?

IMO it's okay for the maintainer/owner/bdfl to make changes directly in master, especially small routine ones like these, I think the only foul is forcefully rewriting master (or any branch) without discussion.

cwhanse:
I prefer to see a PR, even for these housekeeping tasks, because it helps all of us stay aware of the repository status and the work required to maintain it.

mikofski:
Also you could make an issue and then close it with your commit message, like `closes #999`, then the issue could also serve as a discussing point too.

PR #596, opened 2018-10-03
Correct condition 5 in detect_clearksy

### LINKED ISSUE #506: Correct condition 5 in `clearsky.detect_clearsky`

cwhanse:

 - [x] Closes #506 

Imports changes from @bhellis #510.


cwhanse:
Ready for merge. I did not change the tests. The condition with the error (condition 5) is not the unique determining factor for clear or not clear for the test we have.

Perhaps we should create a set of tests that fail at different times for each condition, one at a time.

wholmgren:
> Perhaps we should create a set of tests that fail at different times for each condition, one at a time.

That would be ideal. I prefer making that change as part of this PR to ensure that it's properly tested.

pvlib/clearsky.py
```diff
@@ -756,6 +760,14 @@ def rmse(alpha):
         components['slope_max'] = c5
         components['mean_nan'] = c6
         components['windows'] = clear_windows
+
+        components['mean_diff_array'] = np.abs(meas_mean - alpha * clear_mean)
+        components['max_diff_array'] = np.abs(meas_max - alpha * clear_max)
+        components['line_length_array'] = meas_line_length - clear_line_length
+        components['slope_nstd_array'] = meas_slope_nstd
+        components['slope_max_array'] = (np.max(
```
wholmgren: maybe "values" instead of "array"?

pvlib/clearsky.py
```diff
@@ -680,36 +680,39 @@ def detect_clearsky(measured, clearsky, times, window_length,
...
+        sample_interval * sample_interval), axis=0)
 
     # calculate clear sky statistics
     clear_mean = np.mean(clearsky[H], axis=0)
     clear_max = np.max(clearsky[H], axis=0)
-    clear_slope = np.diff(clearsky[H], n=1, axis=0)
-    clear_slope_max = np.max(np.abs(clear_slope), axis=0)
+    clear_ghi_diff = np.diff(clearsky[H], n=1, axis=0)
+    clear_slope = np.diff(clearsky[H], n=1, axis=0) / sample_interval
+    # clear_slope_max = np.max(np.abs(clear_slope), axis=0)
```
wholmgren: delete this line

pvlib/clearsky.py
```diff
@@ -680,36 +680,39 @@ def detect_clearsky(measured, clearsky, times, window_length,
...
     # calculate measurement statistics
     meas_mean = np.mean(measured[H], axis=0)
     meas_max = np.max(measured[H], axis=0)
-    meas_slope = np.diff(measured[H], n=1, axis=0)
+    meas_ghi_diff = np.diff(measured[H], n=1, axis=0)
+    meas_slope = np.diff(measured[H], n=1, axis=0) / sample_interval
     # matlab std function normalizes by N-1, so set ddof=1 here
     meas_slope_nstd = np.std(meas_slope, axis=0, ddof=1) / meas_mean
-    meas_slope_max = np.max(np.abs(meas_slope), axis=0)
+    # meas_slope_max = np.max(np.abs(meas_slope), axis=0)
```
wholmgren: delete this line

pvlib/clearsky.py
```diff
@@ -680,36 +680,39 @@ def detect_clearsky(measured, clearsky, times, window_length,
...
-    H = hankel(np.arange(samples_per_window),                   # noqa: N806
-               np.arange(samples_per_window-1, len(times)))
+    H = hankel(np.arange(intervals_per_window),                   # noqa: N806
+               np.arange(intervals_per_window - 1, len(times)))
 
     # calculate measurement statistics
     meas_mean = np.mean(measured[H], axis=0)
     meas_max = np.max(measured[H], axis=0)
-    meas_slope = np.diff(measured[H], n=1, axis=0)
+    meas_ghi_diff = np.diff(measured[H], n=1, axis=0)
```
wholmgren: I prefer `meas_diff` to `meas_ghi_diff`. None of the other code in this function specifically refers to ghi.

pvlib/clearsky.py
```diff
@@ -680,36 +680,39 @@ def detect_clearsky(measured, clearsky, times, window_length,
...
-        meas_slope*meas_slope + sample_interval*sample_interval), axis=0)
+        meas_ghi_diff * meas_ghi_diff +
+        sample_interval * sample_interval), axis=0)
 
     # calculate clear sky statistics
     clear_mean = np.mean(clearsky[H], axis=0)
     clear_max = np.max(clearsky[H], axis=0)
-    clear_slope = np.diff(clearsky[H], n=1, axis=0)
-    clear_slope_max = np.max(np.abs(clear_slope), axis=0)
+    clear_ghi_diff = np.diff(clearsky[H], n=1, axis=0)
```
wholmgren: same comment as above re "ghi"

mdeceglie:
Thanks for advancing this @cwhanse and @wholmgren! I believe this will also close #507, depending on whether #507 is interpreted to include the handling of **non-uniform** time intervals or not.

pvlib/clearsky.py
cwhanse: `_flag` on the previous set of keys, removed `_array` here

pvlib/clearsky.py
cwhanse: gone

pvlib/clearsky.py
cwhanse: gone

pvlib/clearsky.py
cwhanse: done

pvlib/clearsky.py
cwhanse: same action

pvlib/clearsky.py
wholmgren: I like the change, but it would break user code if people are calling `return_components`. Probably not likely to be a serious issue but we should think twice about doing this. In any case it should be listed as an API change.

wholmgren:
We can ignore the failing 2.7-min configuration as it's transitioned to pandas 0.15 in a few PRs.

pvlib/clearsky.py
cwhanse: what was meant by "values" instead of "array"?  I'm ok reverting this change, but I thought you were requesting renaming the keys.

wholmgren: Sorry for not being clear. I originally meant to suggest only slightly different names for the new keys that you've added. I was not suggesting renaming the existing keys. I thought "array" was an undesirable identifier because everything is an array here.

I do like the latest proposal, but we'd need to document it as an API change.

wholmgren:
> It is turning out to be rather challenging too cook up data that cause only one condition to fail but not others. How much do we care here about being less than ideal?

I don't care enough to hold up this PR. Thanks for trying.

cwhanse:
ready I think.  cause of py27 failures appear to be addressed in #583

PR #595, opened 2018-10-02
Surfrad data reader

lboeman:
pvlib python pull request guidelines
====================================

Thank you for your contribution to pvlib python! You may delete all of these instructions except for the list below.

You may submit a pull request with your code at any stage of completion.

The following items must be addressed before the code can be merged. Please don't hesitate to ask for help if you're unsure of how to accomplish any of the items below:

 - [x] Closes #590
 - [x] I am familiar with the [contributing guidelines](http://pvlib-python.readthedocs.io/en/latest/contributing.html).
 - [x] Fully tested. Added and/or modified tests to ensure correct behavior for all reasonable inputs. Tests (usually) must pass on the TravisCI and Appveyor testing services.
 - [x] Updates entries to `docs/sphinx/source/api.rst` for API changes.
 - [x] Adds description and name entries in the appropriate `docs/sphinx/source/whatsnew` file for all changes.
 - [x] Code quality and style is sufficient. Passes LGTM and SticklerCI checks.
 - [x] New code is fully documented. Includes sphinx/numpydoc compliant docstrings and comments in the code where necessary.
 - [x] Pull request is nearly complete and ready for detailed review.

Brief description of the problem and proposed solution (if not already fully described in the issue linked to above):

PR #588, opened 2018-09-21
sunrise and sunset from pyephem

cwhanse:
pvlib python pull request guidelines
====================================

Thank you for your contribution to pvlib python! You may delete all of these instructions except for the list below.

You may submit a pull request with your code at any stage of completion.

The following items must be addressed before the code can be merged. Please don't hesitate to ask for help if you're unsure of how to accomplish any of the items below:

 - [ ] Closes issue #114 
 - [ ] I am familiar with the [contributing guidelines](http://pvlib-python.readthedocs.io/en/latest/contributing.html).
 - [ ] Fully tested. Added and/or modified tests to ensure correct behavior for all reasonable inputs. Tests (usually) must pass on the TravisCI and Appveyor testing services.
 - [ ] Updates entries to `docs/sphinx/source/api.rst` for API changes.
 - [ ] Adds description and name entries in the appropriate `docs/sphinx/source/whatsnew` file for all changes.
 - [ ] Code quality and style is sufficient. Passes LGTM and SticklerCI checks.
 - [ ] New code is fully documented. Includes sphinx/numpydoc compliant docstrings and comments in the code where necessary.
 - [ ] Pull request is nearly complete and ready for detailed review.

Brief description of the problem and proposed solution (if not already fully described in the issue linked to above):

PR #599, opened 2018-10-05
add test for hour_angle, vectorize

### LINKED ISSUE #597: improve solarposition.hour_angle and other analytical test functions
Problems:

`solarposition.hour_angle` should have its own test in `test_solarposition`. Its output is tested in `test_bird` but that's not the right place for it. 

`solarposition.hour_angle` is used in `test_analytical_zenith` and `test_analytical_azimuth`. That's bad practice. The success of the zenith and azimuth functions should not depend on the hour_angle function. 

This situation is making it hard for me to track down an issue. There might be something wrong with `solarposition.hour_angle` timezone handling, but it's hard to say at the moment.

**Versions:**
 - ``pvlib.__version__``: 0.6

**Additional context**
discovered when addressing pandas upgrade issues in #595

### LINKED ISSUE #598: solarposition.hour_angle is inefficient and fails on some pandas versions
**Describe the bug**
For some reason related to timezone handling, the `solarposition.hour_angle` function fails on pandas versions 0.15-0.18. We test 0.14 and latest. Apparently we were on pandas 0.19 or greater when this function was created.

Also, the loop over datetime values is probably slow.

```python
    hours = np.array([(t - t.tz.localize(
        dt.datetime(t.year, t.month, t.day)
    )).total_seconds() / 3600. for t in times])
```

I think this gives the same result:

```python
hours = (times.astype(int) - times.normalize().astype(int)) / 1e9 / 3600
```

**Versions:**
 - ``pvlib.__version__``: 0.6.0
 - ``pandas.__version__``: 
 - python:

**Additional context**
probably should address #597 first

mikofski:
* closes #598 vectorize to make it more efficient
* closes #597 add test

Signed-off-by: Mark Mikofski <bwana.marko@yahoo.com>

 - [ ] Closes issue #xxxx

### DIFF
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -1327,17 +1327,8 @@ def hour_angle(times, longitude, equation_of_time):
     equation_of_time_Spencer71
     equation_of_time_pvcdrom
     """
-    # utcoffset() has unpredictable outcome when used with pandas DatetimeIndex
-    # change the localize tz-aware times to local, naive without converting tz
-    # by either replacing tz with None or using tz_localize(None)
-    try:
-        naive_times = times.tz_localize(None)
-    except TypeError:
-        # pandas <0.15 doesn't allow localization of tz-aware datetime index
-        naive_times = times.copy()
-        naive_times.tz = None
-    # combine some arithmetic to make calculation more efficient:
-    # hours - timezone = times - times.normalized - (naive_times - times)
+    naive_times = times.tz_localize(None)  # naive but still localized
+    # hours - timezone = (times - normalized_times) - (naive_times - times)
     hrs_minus_tzs = 1 / (3600. * 1.e9) * (
         2 * times.astype(np.int64) - times.normalize().astype(np.int64) -
         naive_times.astype(np.int64))
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -1327,8 +1327,17 @@ def hour_angle(times, longitude, equation_of_time):
     equation_of_time_Spencer71
     equation_of_time_pvcdrom
     """
-    tz_info = times.tz
-    timezone = tz_info.utcoffset(times).total_seconds() / 3600.
-    hours = 1 / (3600. * 1.e9) * (
-        times.astype(np.int64) - times.normalize().astype(np.int64))
-    return 15. * (hours - 12. - timezone) + longitude + equation_of_time / 4.
+    # utcoffset() has unpredictable outcome when used with pandas DatetimeIndex
+    # change the localize tz-aware times to local, naive without converting tz
+    # by either replacing tz with None or using tz_localize(None)
+    try:
+        naive_times = times.tz_localize(None)
+    except TypeError:
+        naive_times = times.copy()
+        naive_times.tz = None
+    ns2hr = 1 / (3600. * 1.e9)
+    timezones = np.array(ns2hr * (
+            naive_times.astype(np.int64) - times.astype(np.int64)))
+    hours = ns2hr * (
+            times.astype(np.int64) - times.normalize().astype(np.int64))
+    return 15. * (hours - 12. - timezones) + longitude + equation_of_time / 4.
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -1327,8 +1327,7 @@ def hour_angle(times, longitude, equation_of_time):
     equation_of_time_Spencer71
     equation_of_time_pvcdrom
     """
-    hours = np.array([(t - t.tz.localize(
-        dt.datetime(t.year, t.month, t.day)
-    )).total_seconds() / 3600. for t in times])
-    timezone = times.tz.utcoffset(times).total_seconds() / 3600.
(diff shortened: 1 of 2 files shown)


pvlib/solarposition.py
```diff
@@ -1327,8 +1327,7 @@ def hour_angle(times, longitude, equation_of_time):
     equation_of_time_Spencer71
     equation_of_time_pvcdrom
     """
-    hours = np.array([(t - t.tz.localize(
-        dt.datetime(t.year, t.month, t.day)
-    )).total_seconds() / 3600. for t in times])
-    timezone = times.tz.utcoffset(times).total_seconds() / 3600.
+    tz_info = times.tz
+    timezone = tz_info.utcoffset(times).total_seconds() / 3600.
```
wholmgren: is this the expected behavior?

```python
In [15]: times = pd.DatetimeIndex(start='20180101', end='20190101', freq='1s', tz='America/Phoenix')

In [16]: timezone = times.tz.utcoffset(times).total_seconds() / 3600.

In [17]: timezone
Out[17]: -7.466666666666667
```

I am surprised because 1. I expected -7 and 2. `times` is a vector so I expected to get a result for each time (offset could change for DST).

Is this preferable?

```python
In [27]: np.array((times.tz_localize(None).astype(int) - times.astype(int)) / 1e9 / 3600)
Out[27]: array([-7., -7., -7., ..., -7., -7., -7.])
```

The `np.array` wrapper ensures a known type instead of a version-dependent Index/array type.

mikofski: FYI: it's only used 2 places:
```
Searching 160 files for "utcoffset"

C:\Users\mikm\Projects\pvlib-python\pvlib\clearsky.py:
  664          * automatically determines sample_interval
  665          * requires a reference clear sky series instead calculating one
  666:           from a user supplied location and UTCoffset
  667          * parameters are controllable via keyword arguments
  668          * option to return individual test components and clearsky scaling

C:\Users\mikm\Projects\pvlib-python\pvlib\solarposition.py:
  547          # pyephem drops timezone when converting to its internal datetime
  548          # format, so handle timezone explicitly here
  549:         obs.date = ephem.Date(thetime - thetime.utcoffset())
  550          sunrise.append(_ephem_to_timezone(rising(sun), time.tz))
  551          sunset.append(_ephem_to_timezone(setting(sun), time.tz))
  ...
 1329      """
 1330      tz_info = times.tz
 1331:     timezone = tz_info.utcoffset(times).total_seconds() / 3600.
 1332      hours = 1 / (3600. * 1.e9) * (
 1333          times.astype(np.int64) - times.normalize().astype(np.int64))
```

So [`utcoffset()` is also a Python builtin](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcoffset) which is overloaded by PyTz. The difference is that if `utcoffset` is called on a Python datetime then it will return the utcoffset of it's tz info called on itself. So in our example above we could have called:

    >>> dt.utcoffset()
    datetime.timedelta(-1, 59520)

Also FYI: there is an [inefficient loop in `rise_set_transit_ephem`](https://github.com/pvlib/pvlib-python/blob/master/pvlib/solarposition.py#L545-L552) which uses slow `to_pydatetime()` and `utcoffset()`, but it's probably addressed in #588?

Other than that I think there's only 2 other instances of `utcoffset` this one in hour_angle and one in #583. I will fix both. Thanks for catching this!

mikofski: FYI: this is also an example of why we should *never* replace tzinfo manually, but **always** use the `localize()` method so that the correct timezone from the transition info table is used.

wholmgren: Thanks for looking into it. I've seen this kind of funny behavior with the AZ timezones before but it didn't occur to me that it could be a problem in this function. Some other common timezones have similar behavior (though I don't remember which).

wholmgren:
@mikofski I suggest that you take the upgrade to pandas 0.15 commits from #595 and drop your try/except block. You'll also need to make a minor change to pvsystem.py described [here](https://github.com/pvlib/pvlib-python/pull/595#issuecomment-427188120).

wholmgren:
or maybe we just leave master in a failing state until a few PRs go through.

pvlib/solarposition.py
mikofski: To be clear, I don't think it's a problem with the function `utcoffset()`. Honestly, I think this was my mistake for using `utcoffset()` with a sequence of times like `pandas.DatetimeIndex`, because there could potentially be a different timezone for each item in the sequence, so it's ambiguous which timezone should be used. Therefore, I think the problem is that the `tz` (aka `tzinfo`) that's set in the `DatetimeIndex` is unreliable and ambiguous, or the problem could be with `utcoffset` which should be overloaded to return a list of timezones localized to each item in the sequence.

mikofski:
okay, I removed the try-except block , which ha ha didn't work anyway, and we can just wait for pandas min-requirements to be updated, thanks!

wholmgren:
@mikofski thanks! I will go ahead and merge to get the PR logjam moving, but please update the whatsnew in #583 to include these commits.

Issue #601, opened 2018-10-10
add NREL MIDC reader to iotools


wholmgren:
It might be useful to add a `pvlib.iotools` parser for the NREL MIDC sites, particularly for the 1 minute data from sites that are currently active. 

I've copied/pasted code similar to this more times than I'd care to admit:

```python
url = 'http://midcdmz.nrel.gov/apps/plot.pl?site=UAT&start=20101103&edy=2&emo=4&eyr=2018&year=2016&month=1&day=1&endyear=2017&endmonth=12&endday=31&time=23:59&inst=4&inst=5&inst=6&inst=7&inst=9&type=data&first=3&math=0&second=-1&value=0.0&user=0&axis=1'
weather = pd.read_csv(url)
datetime = weather['DATE (MM/DD/YYYY)'] + weather['MST']
datetime = pd.to_datetime(datetime, format='%m/%d/%Y%H:%M')
weather = weather.set_index(datetime)
weather = weather.tz_localize('America/Phoenix')
weather = weather.rename(columns={'Global Horiz (platform) [W/m^2]': 'ghi', 
                                  'Direct Normal [W/m^2]': 'dni', 
                                  'Diffuse Horiz [W/m^2]': 'dhi',
                                  'Air Temperature [deg C]': 'temp_air',
                                  'Avg Wind Speed @ 3m [m/s]': 'wind_speed'})
weather = weather.drop(columns=['DATE (MM/DD/YYYY)', 'MST'])
```

Like the UO SRML network, the NREL network has different instruments at different sites, and sometimes the instruments change.

NREL has a fairly straightforward [API for daily data](https://midcdmz.nrel.gov/apps/data_api_doc.pl). I don't know if it's useful to try to wrap that in addition to the 1 minute data. It could help determine instrument changes.

There's also a nice site API here: https://midcdmz.nrel.gov/apps/data_api_doc.pl?_idtextlist_

**Describe alternatives you've considered**
Let users implement their own MIDC parsers as needed.

**Additional context**
Could be useful for Solar Forecasting 2. cc @cwhanse @lboeman @alorenzo175

cwhanse:
Perhaps #605 isn't the PR to do this, but, do we provide the pvlib variable names someplace?  We ask users to create a dict mapping MIDC to pvlib names, it would help to point users to the list of available pvlib names.

lboeman:
[...]

PR #605, opened 2018-10-16
Add MIDC reader

lboeman:

 - [x] Closes issue #601


lboeman:
Revisiting providing some default mapping, originally I was concerned about mappings making assumptions for the user. 
These are some of the GHI labels across different sites, with some sites providing measurements from multiple instruments.
  
Global PSP [W/m^2]
Global CMP22 [W/m^2]
Global LI-200 [W/m^2]
Global NRCS [W/m^2]

@wholmgren suggested that we either:
- Map variables by their type and append their instrument/unit information to their label.
  e.g. Global PSP [W/m^2] -> ghi_PSP_[W/m^2]
or
- Provide a preferred field to map directly to a pvlib variable name
  e.g. Global PSP [W/m^2] -> ghi

@cwhanse do you have any thoughts on these approaches or ideas on instruments to favor?

cwhanse:
My reaction is to favor #1, append an instrument type code to the measurement code, e.g., ghi_psp, ghi_cmp22, ghi_li200, ghi_rc (for reference cell).  I don't think we want to arbitrate any debates about the merits of various instruments.  Better for us to help to maintain clarity, and let the user decide which of the instruments is mapped to `ghi`.

lboeman:
Great, thanks for the feedback Cliff.

wholmgren:
Thanks @cwhanse. @lboeman Do we know if units are always consistent? For example, all irradiance measurements are W/m^2, all temperature measurements are deg C, etc. If so, let's drop them from the column names but put a note in the documentation.

lboeman:
@wholmgren There is at least one site that reports everything in Standard units: https://midcdmz.nrel.gov/apps/go2url.pl?site=SPMD 
So I was planning to leave units in the column names of the returned dataframe.

I did find one more issue, sometimes there appear to be two names for the same variable at a site, as is the case with "Dew Point Temp [deg C]" and "Dewpoint Temp [deg C]" listed in the field api here: https://midcdmz.nrel.gov/apps/field_api.pl?NWTC
It looks like these two tags shouldn't appear in the same file as I can only find one prior to 2001/08/24 and the other since then. It might be safest to add a second, suffixed mapping to some of these to be sure that if two fields like this do appear together one isn't clobbered with the other.
Something like:
"Dew Point Temp": 'temp_dew'
"Dewpoint Temp": 'temp_dew_1'
Is that something we should be concerned with or should we map the newer values and leave the user to pass in a different mapping if they need to map older data?

wholmgren:
Too bad about the units inconsistency. I suppose that everyone that wants to use this function is going to need to inspect the sites on the MIDC page, though, so maybe it's reasonable to say check the MIDC page for units and be sure to convert before using with other pvlib functions. Either way is fine with me.

> or should we map the newer values and leave the user to pass in a different mapping if they need to map older data?

this is ok with me

pvlib/iotools/midc.py
```diff
@@ -0,0 +1,105 @@
...
+}
+
+
+def map_midc_to_pvlib(variable_map, field_name):
+    """A mapper function to rename Dataframe columns to their pvlib counterparts.
+
+    Parameters
+    ----------
+    variable_map: Dictionary
+        A dictionary for mapping MIDC field nameto pvlib name. See VARIABLE_MAP
```
wholmgren: nameto -> name to

pvlib/iotools/midc.py
```diff
@@ -0,0 +1,105 @@
...
+    label: string
+        The pvlib variable name associated with the MIDC field or the input if
+        a mapping does not exist.
+    """
+    new_field_name = field_name
+    for midc_name, pvlib_name in variable_map.items():
+        if field_name.startswith(midc_name):
+            # extract the instument and units field and then remove units
+            instrument_units = field_name[len(midc_name):]
+            instrument = instrument_units[:instrument_units.find('[') - 1]
```
wholmgren: I think this logic assumes that units are always specified in this format and will produce the wrong result otherwise. Probably a safe assumption, but maybe add a comment.

lboeman: This is something I had considered and forgot to include in implementation. Would it be preferable to handle the case where no units are provided? e.g. instrument_units.find('[') returns -1?

lboeman:
@wholmgren Building urls programatically for the plot.pl endpoint for more than one site seems problematic. Variables are requested with the `inst` parameter using inconsistent integer identifiers across sites, and there doesn't seem to be a simple way of requesting all variables at a given site. Users would have to use the web UI to get the 'inst' values of the variables for the site they are requesting.

We could request data from the raw data api which has a simpler url format that returns all variables. However, the raw data is not quality checked, it is just raw data from the data loggers.

pvlib/test/test_midc.py
```diff
@@ -0,0 +1,62 @@
...
+import pandas as pd
+from pandas.util.testing import network
+import pytest
+
+from pvlib.iotools import midc
+
+
+test_dir = os.path.dirname(
+    os.path.abspath(inspect.getfile(inspect.currentframe())))
+midc_testfile = os.path.join(test_dir, '../data/midc_20181014.txt')
```
wholmgren: perhaps we should add a raw data file as well to decouple the format test from the network test

pvlib/test/test_midc.py
```diff
@@ -0,0 +1,62 @@
...
+    start = pd.Timestamp("20181014 00:00")
+    start = start.tz_localize("MST")
+    end = pd.Timestamp("20181014 23:59")
+    end = end.tz_localize("MST")
+    assert type(data.index) == pd.DatetimeIndex
+    assert data.index[0] == start
+    assert data.index[-1] == end
+
+
+def test_midc_format_index_raw():
```
wholmgren: doesn't this also need `@network`?

lboeman:
I just ran into an issue with this where the reported Timezone throws an error when trying to call DataFrame.tz_localize(), particularly with PST. I'll try to find a reasonable solution to this before it gets merged.

wholmgren:
See bottom of this sub-section:
https://pvlib-python.readthedocs.io/en/latest/timetimezones.html#finding-a-time-zone

We could add a dict that maps problematic time zones to corresponding
Etc/GMT. Could be general purpose and public, or we could keep it simple
and private.

wholmgren:
I think this is ready to merge. Any final comments/concerns?

cwhanse:
Merge it

PR #610, opened 2018-10-23
polite handling of nan in irradiance.perez

cwhanse:

 - [x ] I am familiar with the [contributing guidelines](http://pvlib-python.readthedocs.io/en/latest/contributing.html).
 - [x ] Fully tested. Added and/or modified tests to ensure correct behavior for all reasonable inputs. Tests (usually) must pass on the TravisCI and Appveyor testing services.

### DIFF
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -1147,17 +1147,17 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
     F2c = np.vstack((F2c, nans))
 
     F1 = (F1c[ebin, 0] + F1c[ebin, 1] * delta + F1c[ebin, 2] * z)
-    F1 = np.maximum(F1, 0)
+    F1 = np.where(np.isnan(F1), np.nan, np.maximum(F1, 0))
 
     F2 = (F2c[ebin, 0] + F2c[ebin, 1] * delta + F2c[ebin, 2] * z)
-    F2 = np.maximum(F2, 0)
+    F2 = np.where(np.isnan(F2), np.nan, np.maximum(F2, 0))
 
     A = aoi_projection(surface_tilt, surface_azimuth,
                        solar_zenith, solar_azimuth)
     A = np.maximum(A, 0)
 
     B = tools.cosd(solar_zenith)
-    B = np.maximum(B, tools.cosd(85))
+    B = np.where(np.isnan(B), np.nan, np.maximum(B, tools.cosd(85)))
 
     # Calculate Diffuse POA from sky dome
     term1 = 0.5 * (1 - F1) * (1 + tools.cosd(surface_tilt))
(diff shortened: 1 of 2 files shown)


pvlib/irradiance.py
```diff
@@ -1147,17 +1147,17 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
     F2c = np.vstack((F2c, nans))
 
     F1 = (F1c[ebin, 0] + F1c[ebin, 1] * delta + F1c[ebin, 2] * z)
-    F1 = np.maximum(F1, 0)
+    F1 = np.where(np.isnan(F1), np.nan, np.maximum(F1, 0))
 
     F2 = (F2c[ebin, 0] + F2c[ebin, 1] * delta + F2c[ebin, 2] * z)
-    F2 = np.maximum(F2, 0)
+    F2 = np.where(np.isnan(F2), np.nan, np.maximum(F2, 0))
```
mikofski: if this polite handling is going to become a pattern, it might be worth putting it into `tools` as a dedicated set of functions like

```python
def polite_handling_of_nan(x, f):
    """return nan where x is nan otherwise return f(x)"""
    return np.where(np.isnan(x), np.nan, f(x))
```

Also isn't another option to turn off NumPy warnings? If that's an alternate, I'm curious if there is any penalty for calling `np.where()` all the time, instead of just ignoring the warnings.

mikofski: Also isn't another option to turn off NumPy warnings? If that's an alternate, I'm curious if there is any penalty for calling `np.where()` all the time, instead of just ignoring the warnings.

cwhanse:
Actually that earlier commit did nothing, since np.where evaluates both cases before constructing its output.  See the latest commit, adding `nanmaximum` to `pvlib.tools`.

Having `np.nan` in series inputs like `dni` and `airmass` should be handled gracefully, where we can.

wholmgren:
[np.fmax documentation](https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.fmax.html#numpy.fmax) says "See Also... [maximum](https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.maximum.html) Element-wise maximum of two arrays, propagates NaNs."

It is ridiculous that a function specifically advertised to propagate NaNs would also complain that it was provided a NaN! There's got to be a better way than trying to maintain our own nan-friendly maximum function.

cwhanse:
I agree.  We could open an issue for `numpy`...

wholmgren:
I'm working on a comment here.... https://github.com/numpy/numpy/issues/12038

cwhanse:
OK, let's sit on this until numpy makes up its mind.  The custom function is messing with the output typing in `irradiance.perez`

mikofski:
Also I'm still concerned about penalty from all these extra `np.where` calls if we can just turn this warning off as I asked in [this comment](#discussion_r227577179)

wholmgren:
Looks like numpy will stop emitting the warnings. I suggest we live with them for now and close this PR.

PR #626, opened 2018-11-26
add max_airmass=12 kwarg to disc

### LINKED ISSUE #450: DISC Knc out of range
The expression for the DNI clearness index for clear conditions (Knc) is a polynomial fit over the air mass range 1 to 12.  The values of this polynomial increase with air mass values > 17 and reach as high as 6 with air mass 37, which is clearly not realistic.

In the current DISC implementation air mass values up to 37 are allowed, and unrealistic DNI values appear when zenith > 88 / air mass > 19.

I suggest clipping zenith at 85 or 86 instead of clipping cos(zenith) to 0.065.  This produces a more graceful transition to night time.

wholmgren:
- [x] Closes #450 

@cwhanse @adriesse ready for review.


pvlib/irradiance.py
```diff
@@ -1369,6 +1374,7 @@ def disc(ghi, solar_zenith, datetime_or_doy, pressure=101325,
     am = atmosphere.get_relative_airmass(solar_zenith, model='kasten1966')
     am = atmosphere.get_absolute_airmass(am, pressure)
 
+    am = np.minimum(am, max_airmass)  # GH 450
```
cwhanse: what's "GH 450" just curious

docs/sphinx/source/whatsnew/v0.6.1.rst
```diff
@@ -24,6 +24,11 @@ API Changes
 * Changed function name from :py:func:`pvlib.solarposition.get_rise_set_transit` (deprecated)
   to :py:func:`pvlib.solarposition.sun_rise_set_transit_spa. `sun_rise_set_transit_spa`
   requires time input to be localized to the specified latitude/longitude. (:issue:`316')
+* Add `max_airmass` keyword argument to :py:func:`pvlib.irradiance.disc`.
+  Default value (`max_airmass=12`) is consistent with polynomial fit in
+  original paper describing the model. Output of
+  :py:func:`pvlib.irradiance.dirint` and `:py:func:`pvlib.irradiance.dirindex`,
+  are different when `min_cos_zenith` and `max_zenith` kwargs are used.
```
cwhanse: This second sentence is not clear to me.  Are you saying that the two functions produce different output with the same kwarg values, or that each function's output is different if the kwarg is set different from its default?

pvlib/irradiance.py
wholmgren: Github issue 450. I've seen this convention in other python projects. In any case I should add something more useful and then maybe "see GH 450 for more"

docs/sphinx/source/whatsnew/v0.6.1.rst
wholmgren: The outputs of each of these functions will be different than they were before if the user set the kwargs to more permissive values.

wholmgren: to be clear, this is only because of the change in `disc` and not because I've directly changed anything in these other functions.

wholmgren:
Thanks for the review @cwhanse. Would you be ok with *not* exposing the `max_airmass` kwarg in any of the functions? That is @adriesse's preference. If we keep the kwarg in `disc` I will leave it to others to do the work of exposing it in `dirint` and `dirindex`, if desired.

cwhanse:
> Would you be ok with not exposing the max_airmass kwarg in any of the functions? 
Yes. My preference would be to expose it in all three functions, but I don't feel strongly about it.

docs/sphinx/source/whatsnew/v0.6.1.rst
cwhanse: Perhaps "Changing the value(s) of `min_cos_zenith` or `max_zenith` kwargs will change the output of  :py:func:`pvlib.irradiance.dirint` and `:py:func:`pvlib.irradiance.dirindex`.

Issue #552, opened 2018-08-31
ENH: Implement PVSyst Cell Temperature Model


mikofski:
**PVSyst Cell Temperature Model**
For users using PVSyst model chain, there is no equivalent cell temperature model. The only cell temperature model is the Sandia array performance model, [`pvlib.pvsystem.sapm_celltemp`](https://pvlib-python.readthedocs.io/en/latest/generated/pvlib.pvsystem.sapm_celltemp.html). The [PVSyst cell temperature model](https://pvpmc.sandia.gov/modeling-steps/2-dc-module-iv/cell-temperature/pvsyst-cell-temperature-model/) is different.

**Describe the solution you'd like**
Add a corresponding method like `pvlib.pvsystem.pvsyst_celltemp`. I'll bet it's already implemented in PVLIB_MATLAB, right?

**Describe alternatives you've considered**
leave it to the user to implement

```python
def pvsyst_celltemp(ambient_temp, gti, panel_efficiency, windspeed, alpha_absorption=0.9,
                    natural_convenction_coeff=29, forced_convection_coeff=0):
    """
    Calculate PVSyst cell temperature.

    Parameters
    ------------
    ambient_temp : numeric
        Ambient temperature in degrees Celsius
    gti : numeric
        Global tilted irradiance, AKA incident irradiance in the plane of array (POA) in W/m^2
    panel_efficiency : numeric
        PV module efficiency as a fraction
    windspeed : numeric
        Wind speed in m/s
    alpha_absorption : numeric
        Absorption coefficient, default is 0.9
    natural_convenction_coeff : numeric
        Natural convection coefficient, default is 29
    forced_convection_coeff : numeric
        Forced convection coefficient, default is zero

    Returns
    --------
    celltemp : numeric
        Cell temperature in degrees Celsius
    """
    combined_convection_coeff = (
        natural_convenction_coeff + forced_convection_coeff * windspeed
    )
    return (ambient_temp
        + alpha_absorption * gti * (1 - panel_efficiency) / combined_convection_coeff)
```

I'm not married to these names, they are only for demonstration

**Additional context**
* is this related to #487 at all? I tried to look for any corresponding or existing issues
* if this cell temperature model is implmented then should others be as well? what does SAM use?

cwhanse:
[...]

PR #628, opened 2018-12-06
initial implementation of pvsyst_celltemp function (#552)

CameronTStark:
- [x] Closes issue #552 

### DIFF
--- a/docs/sphinx/source/api.rst
+++ b/docs/sphinx/source/api.rst
@@ -271,7 +271,7 @@ PVWatts model
    pvsystem.pvwatts_losses
 
 PVsyst model
--------------
+------------
 
 .. autosummary::
    :toctree: generated/
--- a/docs/sphinx/source/api.rst
+++ b/docs/sphinx/source/api.rst
@@ -268,6 +268,15 @@ PVWatts model
    pvsystem.pvwatts_dc
    pvsystem.pvwatts_ac
    pvsystem.pvwatts_losses
+   pvsystem.pvwatts_losses
+
+PVWatts model
+-------------
+
+.. autosummary::
+   :toctree: generated/
+
+   pvsystem.pvsyst_celltemp
 
 Other
 -----
--- a/docs/sphinx/source/api.rst
+++ b/docs/sphinx/source/api.rst
@@ -270,7 +270,7 @@ PVWatts model
    pvsystem.pvwatts_losses
    pvsystem.pvwatts_losses
 
-PVWatts model
+PVsyst model
 -------------
 
 .. autosummary::
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1886,10 +1886,15 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
 
     return pd.DataFrame({'temp_cell': temp_cell, 'temp_module': temp_module})
 
-def pvsyst_celltemp(poa_global, wind_speed, temp_air,
-                    eta_m=0.1,
-                    alpha_absorption=0.9,
-                    temp_model='freestanding'):
+
+def pvsyst_celltemp(
+    poa_global,
+    wind_speed,
+    temp_air,
+    eta_m=0.1,
+    alpha_absorption=0.9,
+    temp_model="freestanding",
+):
     """
     Calculate PVSyst cell temperature.
 
@@ -1934,25 +1939,29 @@ def pvsyst_celltemp(poa_global, wind_speed, temp_air,
         Cell temperature in degrees Celsius
     """
 
-    temp_models = {'freestanding': (29.0, 0),
-                   'insulated': (15.0, 0),
-                   }
+    temp_models = {"freestanding": (29.0, 0), "insulated": (15.0, 0)}
 
     if isinstance(temp_model, str):
-        natural_convenction_coeff, forced_convection_coeff = temp_models[temp_model.lower()]
+        natural_convenction_coeff, forced_convection_coeff = temp_models[
+            temp_model.lower()
+        ]
     elif isinstance(temp_model, (tuple, list)):
         natural_convenction_coeff, forced_convection_coeff = temp_model
     else:
-        raise TypeError("Please format temp_model as a str, or tuple/list. See function docstring for guidance")
+        raise TypeError(
+            "Please format temp_model as a str, or tuple/list."
+        )
 
-    combined_convection_coeff = (forced_convection_coeff * wind_speed) + \
-        natural_convenction_coeff
+    combined_convection_coeff = (
+        forced_convection_coeff * wind_speed
+    ) + natural_convenction_coeff
 
     absorption_coeff = alpha_absorption * poa_global * (1 - eta_m)
-    temp_difference = absorption_coeff/combined_convection_coeff
+    temp_difference = absorption_coeff / combined_convection_coeff
     temp_cell = temp_air + temp_difference
 
     return temp_cell
+
 
 def sapm_spectral_loss(airmass_absolute, module):
     """
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1887,9 +1887,8 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
     return pd.DataFrame({'temp_cell': temp_cell, 'temp_module': temp_module})
 
 
-def pvsyst_celltemp(poa_global, wind_speed, temp_air,
-    eta_m=0.1, alpha_absorption=0.9, temp_model="freestanding",
-):
+def pvsyst_celltemp(poa_global, wind_speed, temp_air, eta_m=0.1,
+                    alpha_absorption=0.9, temp_model="freestanding"):
     """
     Calculate cell temperature using the PVSyst model.
 
@@ -1916,7 +1915,8 @@ def pvsyst_celltemp(poa_global, wind_speed, temp_air,
         If string, can be:
 
             * 'freestanding' (default)
-                Modules with rear surfaces exposed to open air (e.g. rack mounted).
+                Modules with rear surfaces exposed to open air (e.g. rack
+                mounted).
             * 'insulated'
                 Modules with rear surfaces in close proximity to another
                 surface (e.g. roof mounted).
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1887,56 +1887,61 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
     return pd.DataFrame({'temp_cell': temp_cell, 'temp_module': temp_module})
 
 
-def pvsyst_celltemp(
-    poa_global,
-    wind_speed,
-    temp_air,
-    eta_m=0.1,
-    alpha_absorption=0.9,
-    temp_model="freestanding",
+def pvsyst_celltemp(poa_global, wind_speed, temp_air,
+    eta_m=0.1, alpha_absorption=0.9, temp_model="freestanding",
 ):
     """
-    Calculate PVSyst cell temperature.
+    Calculate cell temperature using the PVSyst model.
 
     Parameters
-    ------------
-    poa_global : float or Series
+    ----------
+    poa_global : numeric
         Total incident irradiance in W/m^2.
 
-    wind_speed : float or Series
+    wind_speed : numeric
         Wind speed in m/s at a height of 10 meters.
 
-    temp_air : float or Series
+    temp_air : numeric
         Ambient dry bulb temperature in degrees C.
 
     eta_m : numeric
-        PV module efficiency as a fraction
+        Module external efficiency as a fraction, i.e., DC power / poa_global.
 
     alpha_absorption : float
-        Absorption coefficient, default is 0.9
+        Absorption coefficient, default is 0.9.
 
-    temp_model : string, list, or dict, default 'freestanding'
+    temp_model : string, tuple, or list, default 'freestanding' (no dict)
         Model to be used.
 
         If string, can be:
 
             * 'freestanding' (default)
+                Modules with rear surfaces exposed to open air (e.g. rack mounted).
             * 'insulated'
+                Modules with rear surfaces in close proximity to another
+                surface (e.g. roof mounted).
 
         If tuple/list, supply parameters in the following order:
 
             * natural_convenction_coeff : float
                 Natural convection coefficient. Freestanding default is 29,
-                fully insulated arrays is 15. Recommended values are between
-                23.5 and 26.5.
+                fully insulated arrays is 15.
 
             * forced_convection_coeff : float
-                Forced convection coefficient, default is 0. Recommended values
-                are between 6.25 and 7.68.
+                Forced convection coefficient, default is 0.
+
     Returns
-    --------
+    -------
     temp_cell : numeric or Series
         Cell temperature in degrees Celsius
+
+    References
+    ----------
+    [1]"PVsyst 6 Help", Files.pvsyst.com, 2018. [Online]. Available:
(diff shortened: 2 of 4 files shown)


pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+def pvsyst_celltemp(
+    poa_global,
+    wind_speed,
+    temp_air,
+    eta_m=0.1,
+    alpha_absorption=0.9,
+    temp_model="freestanding",
+):
+    """
+    Calculate PVSyst cell temperature.
```
cwhanse: Calculate cell temperature using the PVsyst model.

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+        Total incident irradiance in W/m^2.
+
+    wind_speed : float or Series
+        Wind speed in m/s at a height of 10 meters.
+
+    temp_air : float or Series
+        Ambient dry bulb temperature in degrees C.
+
+    eta_m : numeric
+        PV module efficiency as a fraction
```
cwhanse: module external efficiency as a fraction, i.e., DC power / poa_global

cwhanse:
I think the `temp_model` keys you are using are fine.  They trace back to two defaults that PVsyst offers: see the Array Thermal losses page of the PVsyst v6 help files.

Be sure to update api.rst and whatsnew.rst.

Thanks for the contribution!

docs/sphinx/source/api.rst
```diff
@@ -268,6 +268,15 @@ PVWatts model
    pvsystem.pvwatts_dc
    pvsystem.pvwatts_ac
    pvsystem.pvwatts_losses
+   pvsystem.pvwatts_losses
+
+PVWatts model
```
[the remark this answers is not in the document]
CameronTStark: Thanks for noticing that, just pushed the fix.

cwhanse:
@CameronTStark could you `Resolve conversation` for the formatting items you've addressed?  Thanks.

wholmgren:
@cwhanse @CameronTStark why the closing and reopening? can we leave it open until merged?

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+    eta_m=0.1,
+    alpha_absorption=0.9,
+    temp_model="freestanding",
+):
+    """
+    Calculate PVSyst cell temperature.
+
+    Parameters
+    ------------
+    poa_global : float or Series
```
wholmgren: I believe that all of these `float or Series` parameters can be better described as `numeric` (they can also be arrays or even DataFrames).

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+
+            * natural_convenction_coeff : float
+                Natural convection coefficient. Freestanding default is 29,
+                fully insulated arrays is 15. Recommended values are between
+                23.5 and 26.5.
+
+            * forced_convection_coeff : float
+                Forced convection coefficient, default is 0. Recommended values
+                are between 6.25 and 7.68.
+    Returns
```
wholmgren: blank line above Returns

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+    temp_air : float or Series
+        Ambient dry bulb temperature in degrees C.
+
+    eta_m : numeric
+        PV module efficiency as a fraction
+
+    alpha_absorption : float
+        Absorption coefficient, default is 0.9
+
+    temp_model : string, list, or dict, default 'freestanding'
```
wholmgren: string, tuple, or list, default 'freestanding' (no dict)

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
 
 
+def pvsyst_celltemp(
+    poa_global,
+    wind_speed,
+    temp_air,
+    eta_m=0.1,
+    alpha_absorption=0.9,
+    temp_model="freestanding",
+):
```
wholmgren: No other pvlib functions follow this indentation pattern. I guess it's ok according to PEP8, but consistency within the library is nice too.

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+    """
+    Calculate PVSyst cell temperature.
+
+    Parameters
+    ------------
```
wholmgren: I can't remember if sphinx will complain about the two extra `-`. Might as well remove the extra here, below Returns, and in api.rst while you're cleaning up the other items.

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+        If tuple/list, supply parameters in the following order:
+
+            * natural_convenction_coeff : float
+                Natural convection coefficient. Freestanding default is 29,
+                fully insulated arrays is 15. Recommended values are between
+                23.5 and 26.5.
+
+            * forced_convection_coeff : float
+                Forced convection coefficient, default is 0. Recommended values
+                are between 6.25 and 7.68.
```
wholmgren: I don't know much about the PVsyst model, but it seems strange that we'd specify defaults outside of the recommended values.

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+    alpha_absorption : float
+        Absorption coefficient, default is 0.9
+
+    temp_model : string, list, or dict, default 'freestanding'
+        Model to be used.
+
+        If string, can be:
+
+            * 'freestanding' (default)
+            * 'insulated'
```
wholmgren: I think a few words about what these terms imply would help some users.

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,82 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+                23.5 and 26.5.
+
+            * forced_convection_coeff : float
+                Forced convection coefficient, default is 0. Recommended values
+                are between 6.25 and 7.68.
+    Returns
+    --------
+    temp_cell : numeric or Series
+        Cell temperature in degrees Celsius
+    """
```
wholmgren: Do we have a reference to point to?

CameronTStark:
> @cwhanse @CameronTStark why the closing and reopening? can we leave it open until merged?

Sorry, I hit the Comment and close button by accident.

pvlib/pvsystem.py
CameronTStark: Sorry, haven't done documentation in sphinx previously. Just to be sure, the rules are that there are as many dashes as characters in the word above it, correct?

pvlib/pvsystem.py
CameronTStark: I understand and will collapse the variables so they fit in less lines. 

If you're interested, here's a rationalization for this pattern implemented by the formatter I used called [Black](https://github.com/ambv/black#how-black-wraps-lines).

pvlib/pvsystem.py
wholmgren: yes.

pvlib/pvsystem.py
CameronTStark: Same here. My intention was to provide functionally familiar defaults to PVsyst users via the "temp_models" options while trying to give guidance to others manually providing their own coefficients per the insight provided on the [website](https://pvpmc.sandia.gov/modeling-steps/2-dc-module-iv/module-temperature/faiman-module-temperature-model/).

Two options:
- Call them something other than "Recommended values" (e.g. physically valid values)
- Remove recommended ranges entirely and leave it to the user to ensure validity.

wholmgren: @cwhanse, what do you think about these defaults?

pvlib/pvsystem.py
CameronTStark: I will include the reference to Faiman. I quickly looked in the online docs for PVsyst and didn't find a good reference there.

cwhanse: I'd point to the PVsyst user manual, "Array thermal losses" page.  Hard to give a URL as a reference, though.  I'd also reference the Faiman paper (take it off the pvpmc.org site). But the PVsyst model is different from Faiman, including module efficiency.

pvlib/pvsystem.py
cwhanse: 29 / 0 are the PVsyst defaults.  Maybe keep them for that reason.  I'd drop the "Recommended values..." comments because those are from the Faiman paper, not from PVsyst.

pvlib/pvsystem.py
CameronTStark: Awesome, thanks @cwhanse for the suggestions. Both will be included in the next commit.

pvlib/pvsystem.py
CameronTStark: I've got these formatting changes ready for the next commit. When should I use the "resolve conversation" button? After I commit, before committing as I check off each item or let you do it once you're satisfied?

CameronTStark:
> I made a handful of minor comments below (hoping that this still works with the PR closed).

@wholmgren it looks like the PR isn't updating with my latest commit now that its closed again. Maybe it'll update if it's reopened?

cwhanse:
I also must have mis-clicked and closed it.

wholmgren:
@CameronTStark can you try to resolve the merge conflict? Then I think it's ready to merge.

cwhanse:
Thanks for the contribution @CameronTStark

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,87 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+    References
+    ----------
+    [1]"PVsyst 6 Help", Files.pvsyst.com, 2018. [Online]. Available:
+    http://files.pvsyst.com/help/index.html. [Accessed: 10- Dec- 2018].
+
+    [2] Faiman, D. (2008). "Assessing the outdoor operating temperature of
+    photovoltaic modules." Progress in Photovoltaics 16(4): 307-315.
+    """
+
+    temp_models = {"freestanding": (29.0, 0), "insulated": (15.0, 0)}
```
mikofski: My thoughts in case this is ever revisited, sorry for commenting so late.
* Move this constant to the module level. IMO avoid internally scoped constants. The computer needs to find memory for this every time this function is called. If it's defined at the module level then I think this memory will only be allocated once, when the module is imported the first time.

CameronTStark:
> Thanks for the contribution @CameronTStark

Sure thing and thanks for the easy introduction to contributing here. 

If there's another place to contribute next that makes sense from here please let me know. If not I'll just check issues again and dive in.

cwhanse:
Good comment, @Mikofski, Mark<mailto:Mark.Mikofski@dnvgl.com> I hadn’t considered that.

Easy issue and PR, both welcome.

pvlib/pvsystem.py
```diff
@@ -1887,6 +1887,87 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
+
+    if isinstance(temp_model, str):
+        natural_convenction_coeff, forced_convection_coeff = temp_models[
+            temp_model.lower()
+        ]
+    elif isinstance(temp_model, (tuple, list)):
```
mikofski: @camerontstark in the future maybe consider [duck typing](http://www.voidspace.org.uk/python/articles/duck_typing.shtml#problem) instead of type checking.  Even though duck typing is the Pythonic preference, it may not always work best. I think your solution here is fine, but in general I think duck typing usually results in more robust code. Eg:
```python
else:
    natural_convenction_coeff, forced_convection_coeff = temp_model
    # already raises ValueError or TypeError
    # or use try: except to raise a more descriptive exception
```

wholmgren:
@CameronTStark thanks for this good pull request. A related contribution would be to add `PVSystem` and `ModelChain` wrappers for this new function. I suggest making a new issue for that feature if you want to pursue it.

cwhanse:
#633

CameronTStark:
> @CameronTStark thanks for this good pull request. A related contribution would be to add `PVSystem` and `ModelChain` wrappers for this new function. I suggest making a new issue for that feature if you want to pursue it.

Sounds good. I'll get on it. Thanks!

cwhanse:
@CameronTStark thanks!  Just a note, the method tests should verify that the object methods call the underlying function, there's not a need to retest the function's output.

pvlib/pvsystem.py
CameronTStark: @mikofski Yes, thanks for the suggestion! I've been trying to pay attention more to duck typing in my code and thought about it for this instance but since this was my first contribution and  sapm_celltemp() had the Look Before You Leap type check in it already I figured I'd just follow convention.

adriesse:
On a more general note, does this model originate with PVsyst?
  
There are several variations of the same model that differ only in the way they define and/or determine the coefficients.  Would we like to have completely separate functions for these variations?  Or rather one core function and some helper functions for the coefficient calculations?  Or some other organization?

cwhanse:
@adriesse can you point me to an example of the variations?  I'm not as familiar with this family of temperature models as I'd like to be.  If I understand correctly, the underlying model is Faiman's, and PVsyst implements a variation of it.  As best we know, `pvsyst_celltemp` implements the function described in PVsyst user manual.

I can see value in `Faiman_celltemp` and other functions (and now I'm regretting not asking for an API change to name this family `celltemp_pvsyst` `celltemp_sapm` etc.)  I'd prefer to see a different function for each model, where the function for a model variation can call helpers etc. to implement the variation (akin to the way the dirint DNI model is implemented).

cwhanse:
> The variable absorption_coeff has a very misleading name. It is in fact the portion of the incident irradiance that is converted to heat.

I see it as the portion of irradiance that is absorbed by the cell, either as heat or converted to current.  In either case _absorption coefficient_ is the term used in the PVsyst documentation so it seems helpful to retain the name in pvlib.

adriesse:
That one is called `alpha_absorption`.  Check line 1964.

cwhanse:
> That one is called alpha_absorption. Check line 1964.

You are correct, I see the issue now.  We have a parameter named alpha_absorption ( = 0.9) which is referred to in the PVsyst documentation as Alpha

> Alpha  is the absorption coefficient 

We have an internal variable `absorption_coeff` which is a product of three factors, one being Alpha, and which has no name in PVsyst.  It is confusing to call this product `absorption_coeff` because, well, its not a coefficient (it has units of temperature), and PVsyst uses phrase "absorption coefficient" to refer to one factor, alpha.

PR welcome to clarify the naming.

adriesse:
Regarding model variations: The determination of cell/module temperature using an energy balance equation and a simple heat transfer coefficient "U" goes way back.  PVsyst, SAM, TRNSYS, Duffie & Beckman, Faiman all do it, and I expect there are others.  Making part of the U value proportional to wind speed is old too: D&B makes reference to a 1954 textbook, and certain constants from that book persist in SAM today despite the fact they were derived for a 0.5m2 plate.

The variations principally arise from assumptions regarding the fraction of energy absorbed, the relative importance given to wind speed, and various adjustments (direct or indirect) to the U values for mounting.

cwhanse:
> Actually, absorption_coeff has the units of irradiance.

Sigh, then the other variable `combined_convection_coeff` must have units of W/m2 K.  We should add units to the docstring.

PR #635, opened 2018-12-14
pvfactors limited implementation for bifacial calculations

### LINKED ISSUE #421: bifacial PV system modeling
implement NREL's rear-surface irradiance model for regular, fixed tilt arrays of bifacial PV models.

anomam:

 - [x] Closes  #421

Implementation of a bifacial model using open-source [pvfactors](https://github.com/SunPower/pvfactors) package.


pvlib/irradiance.py
```diff
@@ -2819,3 +2820,130 @@ def dni(ghi, dhi, zenith, clearsky_dni=None, clearsky_tolerance=1.1,
...
+    """
+    Calculate front and back surface plane-of-array irradiance on
+    a fixed tilt or single-axis tracker PV array configuration, and using
+    the open-source "pvfactors" package.
+    Please refer to pvfactors' online documentation for more details:
+    https://sunpower.github.io/pvfactors/
+
+    Parameters
+    ----------
+    solar_azimuth: numpy array of numeric
```
cwhanse: The rest of pvlib generally uses pandas.Series or pandas.Dataframes.  If pvfactors must have np.array as input, can you do some type conversion in this function?

pvlib/irradiance.py
```diff
@@ -2819,3 +2820,130 @@ def dni(ghi, dhi, zenith, clearsky_dni=None, clearsky_tolerance=1.1,
...
+        horizon_band_angle=15.,
+        run_parallel_calculations=True, n_workers_for_parallel_calcs=None):
+    """
+    Calculate front and back surface plane-of-array irradiance on
+    a fixed tilt or single-axis tracker PV array configuration, and using
+    the open-source "pvfactors" package.
+    Please refer to pvfactors' online documentation for more details:
+    https://sunpower.github.io/pvfactors/
+
+    Parameters
```
cwhanse: Usually 'Inputs' here I believe

pvlib/irradiance.py
```diff
@@ -2819,3 +2820,130 @@ def dni(ghi, dhi, zenith, clearsky_dni=None, clearsky_tolerance=1.1,
...
+        pvfactors is capable of using multiprocessing. Use this flag to decide
+        to run calculations in parallel (recommended) or not.
+    n_workers_for_parallel_calcs: int, default None
+        Number of workers to use in the case of parallel calculations. The
+        default value of 'None' will lead to using a value equal to the number
+        of CPU's on the machine running the model.
+
+    Returns
+    -------
+    front_poa_irradiance: numpy array of numeric
```
cwhanse: Output convention could be np.array if inputs are np.array, but pandas.Series if inputs are themselves pandas objects.

pvlib/irradiance.py
```diff
@@ -2819,3 +2820,130 @@ def dni(ghi, dhi, zenith, clearsky_dni=None, clearsky_tolerance=1.1,
             (zenith < zenith_threshold_for_zero_dni) &
             (dni > max_dni)] = max_dni
     return dni
+
+
+def pvfactors_bifacial_irradiance(
```
cwhanse: Talking with @wholmgren we would like this function in a new module `pvlib.bifacial`. That creates space for other bifacial contributions, and helps users discover this function.  We might put a wrapper function in `pvlib.irradiance`, `get_bifacial(xxxx, model='pvfactors')` or something like that.

Function name: I think you can drop `bifacial` since the function will be in `pvlib.bifacial`

cwhanse:
@anomam  Thanks for this!

> I just realized that my IDE's autopep8 made a lot of automated changes... 😮 don't hesitate to let me know if that's an issue

Yes it is a bit of an issue, although most the changes are improvements.  Including them makes it hard to review the scope of the PR when unrelated formatting changes are included.  OK to address the formatting in a separate issue/PR with formatting as its scope.

pvlib/irradiance.py
anomam: Thank you @cwhanse ! That makes a lot of sense. I moved everything to new bifacial modules here: 1c976e3. 
I also reverted my ``autopep8`` changes for a couple of modules in a340c20.
I'll be working on the ``pandas`` IO handling shortly! 
I'm currently a bit confused about why the ``Appveyor`` tests fail even though ``Travis`` passes, so I'll keep looking

anomam: Oh I see, Travis actually skipped the bifacial tests

pvlib/irradiance.py
anomam: Thanks @cwhanse , I forgot about that convention. So I created a decorating function that does this (with a test for it) in order to separate the modeling logic from the inputs/outputs conversion logic. Let me know if that approach is okay with you

pvlib/irradiance.py
anomam: Thanks @cwhanse , as mentioned above, the decorating function should handle this request as well. Let me know what you think

pvlib/test/conftest.py
```diff
@@ -125,3 +130,30 @@ def has_numba():
 
 requires_netCDF4 = pytest.mark.skipif(not has_netCDF4,
                                       reason='requires netCDF4')
+
+try:
+    import pvfactors
```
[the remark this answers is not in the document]
wholmgren: ignore

pvlib/irradiance.py
wholmgren: Numpy docstring standard uses "Parameters" https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

pvlib/irradiance.py
wholmgren: Does pvfactors fail with a Series or DataFrame?

In general, I like the decorator approach if we can make it robust enough to be used throughout pvlib. (Applying the decorator throughout pvlib should be done in a separate PR.)

pvlib/tools.py
```diff
@@ -425,3 +426,58 @@ def _golden_sect_DataFrame(params, VL, VH, func):
...
+
+        new_args = []
+        for arg in args:
+            if isinstance(arg, pd.Series):
+                is_return_type_pandas = True
+                new_arg = arg.values
+                new_args.append(new_arg)
+            elif isinstance(arg, pd.DataFrame):
+                is_return_type_pandas = True
+                new_arg = arg.values.ravel()  # make sure that 1D
```
wholmgren: I think it's better to raise an exception than allow a `DataFrame --> 1D array --> Series` pattern.

pvlib/bifacial.py
```diff
@@ -0,0 +1,135 @@
...
+    solar_azimuth: numeric
+        Sun's azimuth angles using pvlib's azimuth convention (deg)
+    solar_zenith: numeric
+        Sun's zenith angles (deg)
+    surface_azimuth: numeric
+        Azimuth angle of the front surface of the PV modules, using pvlib's
+        convention (deg)
+    surface_tilt: numeric
+        Tilt angle of the PV modules, going from 0 to 180 (deg)
+    timestamps: array of :class:datetime.datetime objects
```
wholmgren: is this rst markup valid? thought it needed ticks around datetime.

Can I use a `DatetimeIndex` here? That is much more common in pvlib. Arrays of datetime objects are inefficient.

pvlib/tools.py
```diff
@@ -425,3 +426,58 @@ def _golden_sect_DataFrame(params, VL, VH, func):
...
+        # Run function
+        outputs = f(*new_args, **new_kwargs)
+
+        # If inputs were provided as pandas series or dataframe, convert
+        # outputted numpy arrays to pandas series
+        new_outputs = []
+        if is_return_type_pandas:
+            for output in outputs:
+                if isinstance(output, np.ndarray):
+                    new_output = pd.Series(output)
```
wholmgren: Looks like this function ignores the input index. `Series (with meaningful index) --> 1D array --> Series (without meaningful index)` is not very friendly. Arguably better to just return the 1D array and avoid the complication of a decorator.

pvlib/irradiance.py
wholmgren: You'll need to add the required packages to the ci files in https://github.com/pvlib/pvlib-python/tree/master/ci and https://github.com/pvlib/pvlib-python/blob/master/appveyor.yml

pvlib/test/test_bifacial.py
```diff
@@ -0,0 +1,66 @@
+import numpy as np
+from datetime import datetime
+from pvlib.bifacial import pvfactors_timeseries
+from conftest import (requires_pvfactors, requires_future, requires_shapely,
+                      requires_scipy)
+
+
+@requires_scipy
+@requires_shapely
+@requires_future
```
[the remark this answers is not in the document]
anomam: You're right; ``pvfactors`` does have ``scipy``, ``shapely``, and ``future`` in its requirements. So if the user installs pvfactors using ``pip install pvfactors`` is should just work!

setup.py
```diff
@@ -45,7 +45,7 @@
 TESTS_REQUIRE = ['pytest', 'pytest-cov', 'pytest-mock', 'nose']
 EXTRAS_REQUIRE = {
     'optional': ['scipy', 'tables', 'numba', 'siphon', 'netcdf4',
-                 'ephem', 'cython'],
+                 'ephem', 'cython', 'pvfactors==0.1.5'],
```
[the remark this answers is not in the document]
anomam: Thanks @wholmgren! Understood, I'll remove it. What is the best way to let the users know that they need pvfactors v0.1.5 for this pvlib implementation to work? And that later pvfactors versions may have updated APIs that would break this implementation?

pvlib/irradiance.py
anomam: Thanks! I also just realized that ``pvlib``'s testing relies on ``conda`` and not ``pip``. I'll try to learn it and spend some time on creating conda builds for ``pvfactors``. My first attempt was a failure because the pvlib conda channel has pvlib versions up to ``0.6.0``, and I had ``>=0.6.1b0`` as a requirement in pvfactors, so it seems like it will take more time than I thought

pvlib/irradiance.py
anomam: pvfactors has been tested with numpy arrays so far. But I can try to see if it would work with Series and DataFrames, with some luck duck-typing might work...

Understood; I did implement a test for the decorating function which checks multiple cases. But I can bring the type conversion into the pvfactors function if you prefer. Let me know

pvlib/tools.py
anomam: Thanks @wholmgren . In the specific case of the ``pvfactors_timeseries()`` function, the ``timestamps`` index is required as a separate argument, but your point stays valid in any case.
Would you be ok with the decorating function recording and passing any ``meaningful index`` to potentially outputted Series to answer this concern?

pvlib/irradiance.py
wholmgren: `pvlib`'s CI configurations use `conda` for most packages for convenience, but the tests don't require it. There's no need to make conda builds for `pvfactors` for this purpose. You can specify `pvfactors` in the `pip` section at the bottom of the requirements files. Our appveyor tests only use conda, but we can skip the pvfactors tests on that platform.

pvlib/tools.py
wholmgren: I suggest we make a less-general function that's only designed to handle the inputs for `pvfactors_timeseries`: series/1d array in, numpy out, exception otherwise. Then we can step back and assess how we can make it more general for other functions in another issue/pr.

setup.py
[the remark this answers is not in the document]
wholmgren: > What is the best way to let the users know that they need pvfactors v0.1.5 for this pvlib implementation to work? 

Would a `>=` requirement work? Are you planning on breaking this functionality in an 0.2 release in the near future? 

> And that later pvfactors versions may have updated APIs that would break this implementation?

I'm hoping that you'll keep the pvlib wrappers up to date!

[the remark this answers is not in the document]
anomam: Thanks! The only user request I got on that was to be able to provide timeseries albedo, but I can always make it backwards compatible

pvlib/test/conftest.py
```diff
@@ -125,3 +130,12 @@ def has_numba():
...
```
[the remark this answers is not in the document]
anomam: Thanks @mikofski !

pvlib/tools.py
anomam: Thanks, I removed the decorator function and implemented checks in the pvfactors function

pvlib/bifacial.py
anomam: Thx! ``DatetimeIndex`` does work, I added this in the tests

pvlib/tools.py
anomam: Thanks @wholmgren . I removed the decorator function and am now doing the checks inside the pvfactors function. Let me know if the new implementation is ok

pvlib/irradiance.py
anomam: Ok, thanks a lot for all the help @wholmgren

pvlib/bifacial.py
```diff
@@ -0,0 +1,147 @@
...
+        pvfactors is capable of using multiprocessing. Use this flag to decide
+        to run calculations in parallel (recommended) or not.
+    n_workers_for_parallel_calcs: int, default None
+        Number of workers to use in the case of parallel calculations. The
+        default value of 'None' will lead to using a value equal to the number
+        of CPU's on the machine running the model.
+
+    Returns
+    -------
+    front_poa_irradiance: numeric
```
wholmgren: also for every timestamp of the simulation? Not clear to me why you've included that phrase in `df_registries` docstring but not here.

pvlib/bifacial.py
```diff
@@ -0,0 +1,147 @@
...
+                                      calculate_radiosities_serially_perez,
+                                      get_average_pvrow_outputs)
+    idx_slice = pd.IndexSlice
+
+    # Build up pv array configuration parameters
+    pvarray_parameters = {
+        'n_pvrows': n_pvrows,
+        'pvrow_height': pvrow_height,
+        'pvrow_width': pvrow_width,
+        'surface_azimuth': surface_azimuth[0],  # not necessary
```
wholmgren: why take the first element here and for the other parameters below? why not necessary? seems like they must not actually be used. if they're only placeholders then perhaps `None` is a better value and a comment like `# will be overwritten when used` would be more descriptive. alternatively, perhaps don't include them in the dictionary and add a block comment above if necessary?

pvlib/bifacial.py
```diff
@@ -0,0 +1,147 @@
...
+            inputs)
+
+    # Get the average surface outputs
+    df_outputs = get_average_pvrow_outputs(df_registries,
+                                           values=['qinc'],
+                                           include_shading=True)
+
+    # Select the calculated outputs from the pvrow to observe
+    ipoa_front = df_outputs.loc[:, idx_slice[index_observed_pvrow,
+                                             'front', 'qinc']].values
```
wholmgren: perhaps only a question because I'm not familiar enough with `pvfactors`, but why extract `.values` instead of returning the Series?

pvlib/test/test_bifacial.py
```diff
@@ -0,0 +1,121 @@
...
+    rho_back_pvrow = 0.05
+    horizon_band_angle = 15.
+
+    # Expected values
+    expected_ipoa_front = [1034.96216923, 795.4423259]
+    expected_ipoa_back = [92.11871485, 70.39404124]
+    tolerance = 1e-6
+
+    # Test serial calculations
+    ipoa_front, ipoa_back, _ = pvfactors_timeseries(
```
wholmgren: Why not test the `df_registries` value?

ci/requirements-py27.yml
```diff
@@ -19,3 +19,4 @@ dependencies:
     - pip:
         - coveralls
         - pytest-timeout
+        - pvfactors==0.1.5
```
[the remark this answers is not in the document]
anomam: I'm sorry ``pvfactors`` is complicating things..! Let me know if you would like me to start working on a ``requirements.txt`` file in this PR

pvlib/bifacial.py
anomam: You're right, that's not very clear. I tried to clarify it in the commits below. Let me know if that's better

pvlib/bifacial.py
anomam: Yeah these lines are really useless :sweat_smile: I just got rid of them in the commits below

pvlib/bifacial.py
anomam: I just believed that having a 1D numpy array was simpler for these results, but I can adjust it to ``pvlib``'s needs, just let me know!

pvlib/test/test_bifacial.py
anomam: Very good point: 
- it is partly and indirectly tested as ``ipoa_front`` and ``ipoa_back`` do come from ``df_registries`` (see [here](https://github.com/pvlib/pvlib-python/blob/9829b810ebedfc22172e51d499d1b271f766a537/pvlib/bifacial.py#L134-L136))
- but on a more general note, I guess my intent was to keep things simple. We could see ``df_registries`` as a "master document" containing very detailed results (maybe mostly unused too: see example [here](https://sunpower.github.io/pvfactors/developer/pvfactors_demo.html#Surface-Registry) or also [here](https://sunpower.github.io/pvfactors/developer/pvfactors_demo.html#Calculate-radiosities-for-single-time-point-and-using-isotropic-diffuse-sky-dome-approach)). And for simplicity, I wanted to extract and test what I thought most users would be interested in, which are the aggregated front and back surface POA irradiance values, while still giving users the option of looking at the details.

I can add some tests for other outputs in ``df_registries``, let me know! (I might grab some tests from the pvfactors repo)
Here is what the ``df_registries`` expected values would look like for the given test inputs we have so far:
image
Is it okay to use csv files to store test data in the ``pvlib`` repo?

wholmgren: That makes sense. I don't have a strong opinion on whether or not this should be tested in pvlib.

>  Is it okay to use csv files to store test data in the pvlib repo?

Yes, see [pvlib/data](https://github.com/pvlib/pvlib-python/tree/master/pvlib/data). Your example is a good size for a test file.

ci/requirements-py27.yml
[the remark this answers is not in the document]
wholmgren: Not complicating things, just noting something for future reference. Let's leave the requirements.txt file for another PR.

pvlib/test/test_bifacial.py
anomam: Thanks @wholmgren , I don't think that there is a strong need to implement this here. I tested a lot of different cases in the pvfactors repo, so the probability of getting an issue here should be low. But don't hesitate to let me know if you want it at any point

anomam:
Thank you so much for the time you spent on this @cwhanse @wholmgren 

@cwhanse, from @wholmgren comments above I got the sense that returning 1D arrays was good, but please don't hesitate to let me know if you prefer pandas series (when the inputs are also pandas series)

I'll take a look again tomorrow (Japan time) to check for any necessary clean up or adjustments

pvlib/test/test_bifacial.py
cwhanse: I think we're ok without testing the `df_registries` output.  In pvlib we're confirming that `pvfactors_timeseries` is returning values consistent with separate `pvfactors` calculations and its enough to test `ipoa_front` and `ipoa_back`.

wholmgren:
> @cwhanse, from @wholmgren comments above I got the sense that returning 1D arrays was good, but please don't hesitate to let me know if you prefer pandas series (when the inputs are also pandas series)

Sorry, I didn't fully understand the inputs/outputs and discussions (still might not). After looking more carefully, I suggest that the outputs *always* be pandas Series and DataFrames. My rationale is that 

* the function is specifically meant for timeseries analysis
* the function requires input datetime information (optionally as a DatetimeIndex)
* the function always returns a DataFrame for one of its outputs
* the returned DataFrame always contains a `timeseries` column (at least that's my understanding)

Unlike other functions in pvlib that are more flexible, here we require users to handle at least one pandas object (the DataFrame) and we know the timestamps that each row corresponds to. So, I suggest that we always return the two Series with the appropriate DatetimeIndex. My understanding is that the DataFrame would need MultiIndex with (row, timestamp). I think this works: `df_registries.set_index(['pvrow', 'timestamp'])`.

pvlib/test/test_bifacial.py
```diff
@@ -0,0 +1,121 @@
...
+    np.testing.assert_allclose(ipoa_front, expected_ipoa_front,
+                               atol=0, rtol=tolerance)
+    np.testing.assert_allclose(ipoa_back, expected_ipoa_back,
+                               atol=0, rtol=tolerance)
+
+
+@requires_pvfactors
+def test_pvfactors_timeseries_pandas_inputs():
+    """ Test that pvfactors is functional, using the TLDR section inputs of the
+    package github repo README.md file, but converted to pandas Series"""
```
wholmgren: link to this would be helpful

pvlib/test/test_bifacial.py
```diff
@@ -0,0 +1,121 @@
...
+    # Test serial calculations
+    ipoa_front, ipoa_back, _ = pvfactors_timeseries(
+        solar_azimuth, solar_zenith, surface_azimuth, surface_tilt,
+        timestamps, dni, dhi, gcr, pvrow_height, pvrow_width, albedo,
+        n_pvrows=n_pvrows, index_observed_pvrow=index_observed_pvrow,
+        rho_front_pvrow=rho_front_pvrow, rho_back_pvrow=rho_back_pvrow,
+        horizon_band_angle=horizon_band_angle,
+        run_parallel_calculations=False, n_workers_for_parallel_calcs=None)
+
+    np.testing.assert_allclose(ipoa_front, expected_ipoa_front,
```
wholmgren: This should use the pandas test functions if we adopt the always return a Series approach

pvlib/bifacial.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    Calculate front and back surface plane-of-array irradiance on
+    a fixed tilt or single-axis tracker PV array configuration, and using
+    the open-source "pvfactors" package.
+    Please refer to pvfactors online documentation for more details:
+    https://sunpower.github.io/pvfactors/
+
+    Inputs
+    ------
+    solar_azimuth: numeric (list, numpy array, or pandas Series)
```
wholmgren: I prefer to only document the input type as `numeric` (no `(list...)`). Like it or not, it's consistent with the rest of `pvlib`.

wholmgren:
I'd like see the PR finished early next week or pushed to 0.6.2. I think we're pretty close, but we need to get 0.6.1 out and this (understandably) has not been touched in awhile.

anomam:
Hi @wholmgren , sorry for the long delay, I'm still on my way back from a lot of travel but I will wrap it up early next week. When is the 0.6.1 release?

anomam:
Hi @wholmgren , I updated the function to return pandas Series and I fixed the docstrings as requested. Let me know if it doesn't look like what you expected. Thanks again for your time

cwhanse:
LGTM. @wholmgren ?

docs/sphinx/source/whatsnew/v0.6.1.rst
```diff
@@ -82,6 +84,7 @@ Testing
 ~~~~~~~
 * Add test for :func:`~pvlib.solarposition.hour_angle` (:issue:`597`)
 * Update tests to be compatible with pytest 4.0. (:issue:`623`)
+* Add tests for :py:func:`pvlib.bifacial.pvfactors_timeseries` implementation and :py:func:`pvlib.tools.enforce_numpy_arrays` decorator function (:issue:`421`)
```
wholmgren: The `enforce_numpy_arrays` section is outdated.

anomam: Thanks, done in 55dd3a6

wholmgren:
Great to see bifacial modeling in pvlib - thanks @anomam!

anomam:
Thank you so much for your help and time spent on this @wholmgren @cwhanse !
I hope to see more bifacial models getting ``pvlib``'ed as well :sunny:

pvlib/bifacial.py
```diff
@@ -0,0 +1,148 @@
...
+
+import pandas as pd
+
+
+def pvfactors_timeseries(
+        solar_azimuth, solar_zenith, surface_azimuth, surface_tilt,
+        timestamps, dni, dhi, gcr, pvrow_height, pvrow_width, albedo,
+        n_pvrows=3, index_observed_pvrow=1,
+        rho_front_pvrow=0.03, rho_back_pvrow=0.05,
+        horizon_band_angle=15.,
```
kanderso-nrel, line 14: @anomam do you happen to remember why you chose this particular default value for `horizon_band_angle`?  From what I can tell in the git history, the analogous default on the pvfactors side was 6.5 (both at the time of this PR and today) rather than 15.  I know this PR is ancient history at this point so I'm not expecting much, just curious :)

There are now some small differences between this wrapper and pvfactors itself in the `rho_*` parameters as well, but I think those were only introduced later on after this PR was merged.

anomam: hey @kanderso-nrel ! It was a long time ago so I'm not entirely sure anymore, but what I remember is that these values might have needed a bit of tuning to match the measurements depending on the sites etc. I think for the horizon band it might be safe to check what the Perez paper says about what value to use.
For the reflectivity parameter values, it's something I actually never really had time to study in depth. I think the default values for the PV rows were low enough to not have much impact, but that's probably not accurate in all cases...
Sorry if my answers are vague, I think getting better ones would require more work/studies/experience...

Issue #633, opened 2018-12-12
Add `pvsyst_celltemp` method to `PVSystem`


cwhanse:
#552 #628 

Add `PVSystem` method thats wrap `pvsystem.pvsyst_celltemp`

wholmgren:
closed by #636.

PR #636, opened 2018-12-14
Add PVSystem pvsyst_celltemp methods

CameronTStark:
- [x] Closes issue #633

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.6.1.rst
+++ b/docs/sphinx/source/whatsnew/v0.6.1.rst
@@ -73,6 +73,8 @@ Bug fixes
 * Fix documentation errors when using IPython >= 7.0.
 * Fix error in :func:`pvlib.modelchain.ModelChain.infer_spectral_model` (:issue:`619`)
 * Fix error in ``pvlib.spa`` when using Python 3.7 on some platforms.
+* Fix error in :func:`pvlib.irradiance._delta_kt_prime_dirint` (:issue:`637`). The error affects
+  the first and last values of DNI calculated by the function :func:`pvlib.irradiance.dirint`
 
 
 Testing
--- a/docs/sphinx/source/whatsnew/v0.6.1.rst
+++ b/docs/sphinx/source/whatsnew/v0.6.1.rst
@@ -60,6 +60,7 @@ Enhancements
 * Add option for :py:func:`pvlib.irradiance.disc` to use relative airmass
   by supplying `pressure=None`. (:issue:`449`)
 * Created :py:func:`pvlib.pvsystem.pvsyst_celltemp` to implement PVsyst's cell temperature model. (:issue:`552`)
+* Add `PVSystem` class method :py:func:`~pvlib.pvsystem.PVSystem.pvsyst_celltemp` (:issue:`633`)
 
 
 Bug fixes
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -46,6 +46,16 @@ DC_MODEL_PARAMS = {
 }
 
 
+TEMP_MODELS = {
+    'sapm': {'open_rack_cell_glassback': [-3.47, -.0594, 3],
+             'roof_mount_cell_glassback': [-2.98, -.0471, 1],
+             'open_rack_cell_polymerback': [-3.56, -.0750, 3],
+             'insulated_back_polymerback': [-2.81, -.0455, 0],
+             'open_rack_polymer_thinfilm_steel': [-3.58, -.113, 3],
+             '22x_concentrator_tracker': [-3.23, -.130, 13]},
+    'pvsyst': {'freestanding': (29.0, 0), 'insulated': (15.0, 0)}
+}
+
 # not sure if this belongs in the pvsystem module.
 # maybe something more like core.py? It may eventually grow to
 # import a lot more functionality from other modules.
@@ -1876,13 +1886,7 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
     sapm
     '''
 
-    temp_models = {'open_rack_cell_glassback': [-3.47, -.0594, 3],
-                   'roof_mount_cell_glassback': [-2.98, -.0471, 1],
-                   'open_rack_cell_polymerback': [-3.56, -.0750, 3],
-                   'insulated_back_polymerback': [-2.81, -.0455, 0],
-                   'open_rack_polymer_thinfilm_steel': [-3.58, -.113, 3],
-                   '22x_concentrator_tracker': [-3.23, -.130, 13]
-                   }
+    temp_models = TEMP_MODELS['sapm']
 
     if isinstance(model, str):
         model = temp_models[model.lower()]
@@ -1896,9 +1900,9 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
 
     E0 = 1000.  # Reference irradiance
 
-    temp_module = pd.Series(poa_global*np.exp(a + b*wind_speed) + temp_air)
+    temp_module = pd.Series(poa_global * np.exp(a + b * wind_speed) + temp_air)
 
-    temp_cell = temp_module + (poa_global / E0)*(deltaT)
+    temp_cell = temp_module + (poa_global / E0) * (deltaT)
 
     return pd.DataFrame({'temp_cell': temp_cell, 'temp_module': temp_module})
 
@@ -1960,7 +1964,7 @@ def pvsyst_celltemp(poa_global, wind_speed, temp_air, eta_m=0.1,
     photovoltaic modules." Progress in Photovoltaics 16(4): 307-315.
     """
 
-    temp_models = {"freestanding": (29.0, 0), "insulated": (15.0, 0)}
+    temp_models = TEMP_MODELS['sapm']
 
     if isinstance(temp_model, str):
         natural_convenction_coeff, forced_convection_coeff = temp_models[
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -47,12 +47,12 @@ DC_MODEL_PARAMS = {
 
 
 TEMP_MODELS = {
-    'sapm': {'open_rack_cell_glassback': [-3.47, -.0594, 3],
-             'roof_mount_cell_glassback': [-2.98, -.0471, 1],
-             'open_rack_cell_polymerback': [-3.56, -.0750, 3],
-             'insulated_back_polymerback': [-2.81, -.0455, 0],
-             'open_rack_polymer_thinfilm_steel': [-3.58, -.113, 3],
-             '22x_concentrator_tracker': [-3.23, -.130, 13]},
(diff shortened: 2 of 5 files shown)


cwhanse:
I changed #633 to address only `PVSystem`.  If this PR can close the edited #633, we can take up the model chain work in another issue.

docs/sphinx/source/whatsnew/v0.6.1.rst
```diff
@@ -58,6 +58,8 @@ Enhancements
 * Add option for :py:func:`pvlib.irradiance.disc` to use relative airmass
   by supplying `pressure=None`. (:issue:`449`)
 * Created :py:func:`pvlib.pvsystem.pvsyst_celltemp` to implement PVsyst's cell temperature model. (:issue:`552`)
+* Add `PVSystem` class method :py:func:`~pvlib.pvsystem.PVSystem.pvsyst_celltemp` and
+ `ModelChain` class method :py:func:`~pvlib.modelchain.ModelChain.pvsyst_cell` (:issue:`633`)
```
cwhanse: Edit for scope of PR

pvlib/pvsystem.py
```diff
@@ -511,6 +511,22 @@ def sapm_effective_irradiance(self, poa_direct, poa_diffuse,
             poa_direct, poa_diffuse, airmass_absolute, aoi,
             self.module_parameters, reference_irradiance=reference_irradiance)
 
+    def pvsyst_celltemp(self, poa_global, wind_speed, temp_air, eta_m=0.1,
+                        alpha_absorption=0.9, temp_model="freestanding"):
```
mikofski: I believe the default kwarg `"freestanding"` is redundant because it will be overwritten by the `racking_model='open_rack_cell_glassback'` argument passed for `temp_model` on L528 below

cwhanse: We need to reconcile the two sets of keywords, one for the SAPM temperature model and the other for the PVsyst  model.  A topic for a separate PR, I think.

mikofski:
Have you guys noticed that GitHub now alerts users when a commit is *force* pushed to a PR?

pvlib/pvsystem.py
```diff
@@ -1944,7 +1964,7 @@ def pvsyst_celltemp(poa_global, wind_speed, temp_air, eta_m=0.1,
     photovoltaic modules." Progress in Photovoltaics 16(4): 307-315.
     """
 
-    temp_models = {"freestanding": (29.0, 0), "insulated": (15.0, 0)}
+    temp_models = TEMP_MODELS['sapm']
```
cwhanse: 'pvsyst'

CameronTStark:
The two failing checks seem to be from outside the changed code base from my PR. Is there anything else I need to do to get this accepted? Just making sure you're not waiting for me. Thx.

cwhanse:
I'm ready to merge this.  @wholmgren ?

docs/sphinx/source/whatsnew/v0.6.1.rst
```diff
@@ -73,6 +74,8 @@ Bug fixes
 * Fix documentation errors when using IPython >= 7.0.
 * Fix error in :func:`pvlib.modelchain.ModelChain.infer_spectral_model` (:issue:`619`)
 * Fix error in ``pvlib.spa`` when using Python 3.7 on some platforms.
+* Fix error in :func:`pvlib.irradiance._delta_kt_prime_dirint` (:issue:`637`). The error affects
+  the first and last values of DNI calculated by the function :func:`pvlib.irradiance.dirint`
```
wholmgren: I'm concerned and confused to see this and the associated changes in `irradiance.py` show up in this PR. Can you merge `master` into your branch and see if it cleans up the PR?

pvlib/pvsystem.py
```diff
@@ -46,6 +46,16 @@
...
 
 
+TEMP_MODELS = {
+    'sapm': {'open_rack_cell_glassback': [-3.47, -.0594, 3],
+             'roof_mount_cell_glassback': [-2.98, -.0471, 1],
+             'open_rack_cell_polymerback': [-3.56, -.0750, 3],
+             'insulated_back_polymerback': [-2.81, -.0455, 0],
+             'open_rack_polymer_thinfilm_steel': [-3.58, -.113, 3],
+             '22x_concentrator_tracker': [-3.23, -.130, 13]},
+    'pvsyst': {'freestanding': (29.0, 0), 'insulated': (15.0, 0)}
```
wholmgren: `sapm` models use lists and `pvsyst` models use tuples. Let's be consistent. Perhaps tuples are best for this application but I don't have a strong preference.

pvlib/pvsystem.py
```diff
@@ -511,6 +521,22 @@ def sapm_effective_irradiance(self, poa_direct, poa_diffuse,
             poa_direct, poa_diffuse, airmass_absolute, aoi,
             self.module_parameters, reference_irradiance=reference_irradiance)
 
+    def pvsyst_celltemp(self, poa_global, wind_speed, temp_air, eta_m=0.1,
+                        alpha_absorption=0.9):
```
wholmgren: This does not follow the pattern for `PVSystem` methods. Instead of putting the kwargs in the method signature, the method should pull the values from the `self.module_parameters` dictionary. Please see other PVSystem methods for examples of using the `_build_kwargs` function.

pvlib/test/test_pvsystem.py
```diff
@@ -1118,6 +1118,23 @@ def test_pvsyst_celltemp_with_index():
     assert_series_equal(expected, pvtemps)
 
 
+def test_PVSystem_pvsyst_celltemp(mocker):
+    racking_model = 'insulated'
+    system = pvsystem.PVSystem(racking_model=racking_model)
+    mocker.spy(pvsystem, 'pvsyst_celltemp')
+    irrads = 1000
+    temps = 25
+    alpha_absorption = 0.9
```
wholmgren: Use non-default parameters to ensure that the right values are passed to the function.

pvlib/test/test_pvsystem.py
```diff
@@ -1118,6 +1118,23 @@ def test_pvsyst_celltemp_with_index():
...
+    winds = 1
+    eta_m = 0.1
+    out = system.pvsyst_celltemp(irrads, winds, temps)
+    pvsystem.pvsyst_celltemp.assert_called_once_with(irrads, winds, temps,
+                                                     eta_m, alpha_absorption,
+                                                     temp_model=racking_model)
+    assert isinstance(out, float)
+    assert out == 79.0
```
wholmgren: Strict tests of values belong in the test of the original function. Here we just want to make sure it's reasonable. Maybe `assert out < 90 and out > 70`.

pvlib/pvsystem.py
cwhanse: Good catch, I'd prefer tuples for all

CameronTStark: My choice for PVSyst was made since these parameters are unchanging so tuples made more sense. 

I'll change all to tuples.

docs/sphinx/source/whatsnew/v0.6.1.rst
CameronTStark: I'm confused too especially since I haven't made changes to irradiance.py at all. 

When I do a diff on my branch with upstream/master this isn't highlighted.

pvlib/pvsystem.py
CameronTStark: @wholmgren I looked at the other `PVSystem` implementations but am not sure how to work in your suggestion. The `self.module_parameters` attribute contains module parameters specific to SAPM with no analogues to `eta_m` or `alpha_absorption` that I'm aware of while the other parameters `poa_global`, `wind_speed` and `temp_air` are environmental.  

I modeled this off of `PVSystem.sapm_celltemp` which isn't in this format. Can you provide a little more guidance on how you see this working?

wholmgren: @CameronTStark sorry I was unclear. 

`PVSystem.sapm_celltemp` is not a good template here because `pvsystem.sapm_celltemp`'s only keyword argument is explicitly set in `PVSystem` construction. Check out `PVSystem.physicaliam` for a simple example of how an attribute such as `PVSystem.module_parameters` may be referenced.

`poa_global`, `wind_speed` and `temp_air` should all be method arguments -- these are things that are external to the `PVSystem`.

`eta_m` and `alpha_absorption` describe the `PVSystem`, so they should be stored in an object attribute. In this case they should be stored in `module_parameters`. 

`module_parameters` is a general purpose attribute. It is not specific to SAPM or any other model.

`_build_kwargs` allows us to pull data specified in `module_parameters` if it's there. If it's not there, no kwarg is specified and we use the default values in the function definition.

Hope that helps, but let me know if not.

cwhanse: Suggest using `PVSystem.ashraeiam` as the model. In the `pvsyst_celltemp` method, kwargs = `_build_kwargs(['eta_m', 'alpha_absorption'], self.module_parameters)` 
then
`return pvsyst_celltemp(poa_global, wind_speed, temp_air, temp_model=self.racking_model, **kwargs)`

Note the re-ordering of the arguments in the signature.

What will happen is that if a user has included `eta_m` and `alpha_absorption` in `module_parameters`, they will be extracted and passed in to `pvsystem.pvsyst_celltemp`.  If they aren't present, `pvsystem.pvsyst_celltemp` will use the default values.

cwhanse: `PVSystem.sapm_celltemp` does not have this behavior currently for the `a` and `b` parameters it consumes.

wholmgren: > PVSystem.sapm_celltemp does not have this behavior currently for the a and b parameters it consumes.

I don't follow. `a` and `b` are not explicit parameters in the function signature. `PVSystem.racking_model` is usually a string, but could be a list or dict as specified by the function `pvsystem.sapm_celltemp` function. I think some API changes could help, but that's a separate issue.

cwhanse: I wasn't clear. "this behavior" I meant only that `PVSystem.sapm_celltemp` doesn't use `_build_kwargs` and isn't a model for what we're asking for in `PVSystem.pvsyst_celltemp`.  Agree that changing `sapm_celltemp` is separate from this PR.

CameronTStark: Great, thank you both for your guidance. I wasn't sure about the flexibility of `module_parameters` and now this all makes plenty of sense.

CameronTStark:
Just FYI, the stickler-ci fail deals with code that I didn't contribute but I can change it if necessary.

cwhanse:
@wholmgren Travis failure is a numpy/py27 conflict of some sort.  py27-min build is OK.  AppVeyor failure is `test_read_midc_raw_data_from_nrel`.

I don't know why the merged code from #638 shows up here as a change, it must be because it was merged into @CameronTStark's branch via a pull request from me, rather than from pvlib-python/master.

OK to merge?

PR #651, opened 2019-01-22
Improve clarity of code in pvsyst_celltemp (names and doc)

adriesse:
- [x] Improves on the implementation of issue #552

I have tried to improve clarity of code in `pvsyst_celltemp` through changes to variables names and docstring, and also changed the name of one function parameter.

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -521,7 +521,7 @@ class PVSystem(object):
             poa_direct, poa_diffuse, airmass_absolute, aoi,
             self.module_parameters, reference_irradiance=reference_irradiance)
 
-    def pvsyst_celltemp(self, poa_global, wind_speed, temp_air):
+    def pvsyst_celltemp(self, poa_global, temp_air, wind_speed):
         """Uses :py:func:`pvsyst_celltemp` to calculate module temperatures
         based on ``self.racking_model`` and the input parameters.
 
@@ -535,7 +535,7 @@ class PVSystem(object):
         """
         kwargs = _build_kwargs(['eta_m', 'alpha_absorption'],
                                self.module_parameters)
-        return pvsyst_celltemp(poa_global, wind_speed, temp_air,
+        return pvsyst_celltemp(poa_global, temp_air, wind_speed,
                                model_params=self.racking_model, **kwargs)
 
     def first_solar_spectral_loss(self, pw, airmass_absolute):
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -521,7 +521,7 @@ class PVSystem(object):
             poa_direct, poa_diffuse, airmass_absolute, aoi,
             self.module_parameters, reference_irradiance=reference_irradiance)
 
-    def pvsyst_celltemp(self, poa_global, temp_air, wind_speed):
+    def pvsyst_celltemp(self, poa_global, temp_air, wind_speed=1.0):
         """Uses :py:func:`pvsyst_celltemp` to calculate module temperatures
         based on ``self.racking_model`` and the input parameters.
 
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1915,7 +1915,7 @@ def pvsyst_celltemp(poa_global, temp_air, wind_speed=1.0, eta_m=0.1,
     as implemented in PVsyst.
 
     The heat loss factors provided through the 'model_params' argument
-    represent the combined effect of convection, radiation and conduction, 
+    represent the combined effect of convection, radiation and conduction,
     and their values are experimentally determined.
 
     Parameters
@@ -1926,18 +1926,18 @@ def pvsyst_celltemp(poa_global, temp_air, wind_speed=1.0, eta_m=0.1,
     temp_air : numeric
         Ambient dry bulb temperature in degrees C.
 
-    wind_speed : numeric
-        Wind speed in m/s measured at the same height for which the wind loss 
-        factor was determined.  The default value is 1.0, which is the wind 
-        speed at module height used to determine NOCT.   (Sorry André!).
+    wind_speed : numeric, default 1.0
+        Wind speed in m/s measured at the same height for which the wind loss
+        factor was determined.  The default value is 1.0, which is the wind
+        speed at module height used to determine NOCT.
 
-    eta_m : numeric
+    eta_m : numeric, default 0.1
         Module external efficiency as a fraction, i.e., DC power / poa_global.
 
-    alpha_absorption : float
-        Absorption coefficient, default is 0.9.
+    alpha_absorption : numeric, default 0.9
+        Absorption coefficient
 
-    model_params : string, tuple, or list, default 'freestanding' (no dict)
+    model_params : string, tuple, or list (no dict), default 'freestanding'
         Heat loss factors to be used.
 
         If string, can be:
@@ -1952,7 +1952,7 @@ def pvsyst_celltemp(poa_global, temp_air, wind_speed=1.0, eta_m=0.1,
         If tuple/list, supply parameters in the following order:
 
             * constant_loss_factor : float
-                Combined heat loss factor coefficient. Freestanding 
+                Combined heat loss factor coefficient. Freestanding
                 default is 29, fully insulated arrays is 15.
 
             * wind_loss_factor : float
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -536,7 +536,7 @@ class PVSystem(object):
         kwargs = _build_kwargs(['eta_m', 'alpha_absorption'],
                                self.module_parameters)
         return pvsyst_celltemp(poa_global, wind_speed, temp_air,
-                               temp_model=self.racking_model, **kwargs)
+                               loss_factors=self.racking_model, **kwargs)
 
     def first_solar_spectral_loss(self, pw, airmass_absolute):
 
@@ -1909,9 +1909,13 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
 
 
 def pvsyst_celltemp(poa_global, wind_speed, temp_air, eta_m=0.1,
-                    alpha_absorption=0.9, temp_model="freestanding"):
+                    alpha_absorption=0.9, loss_factors="freestanding"):
     """
-    Calculate cell temperature using the PVSyst model.
+    Calculate cell temperature using an emperical heat loss factor model
+    as implemented in PVsyst.
+
+    The heat loss factors represent the combined effect of convection,
+    radiation and conduction, and their values are experimentally determined.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


pvlib/pvsystem.py
```diff
@@ -536,7 +536,7 @@ def pvsyst_celltemp(self, poa_global, wind_speed, temp_air):
         kwargs = _build_kwargs(['eta_m', 'alpha_absorption'],
                                self.module_parameters)
         return pvsyst_celltemp(poa_global, wind_speed, temp_air,
-                               temp_model=self.racking_model, **kwargs)
+                               loss_factors=self.racking_model, **kwargs)
```
cwhanse: I'm not a fan of this renaming.  Although `loss_factors` may be more in line with PVsyst terms, I think using a term that is common with the SAPM temperature model is beneficial.  The `temp_model` variable name aligns with the `TEMP_MODELS` constant dict in `pvlib.pvsystem`.  I'm open to a term other than `temp_model` (which I don't object to) but only if it applies to all temperature models.

pvlib/pvsystem.py
```diff
@@ -1930,8 +1934,8 @@ def pvsyst_celltemp(poa_global, wind_speed, temp_air, eta_m=0.1,
     alpha_absorption : float
         Absorption coefficient, default is 0.9.
 
-    temp_model : string, tuple, or list, default 'freestanding' (no dict)
-        Model to be used.
+    loss_factors : string, tuple, or list, default 'freestanding' (no dict)
+        Heat loss factors to be used.
```
cwhanse: This is a much more clear description

pvlib/pvsystem.py
```diff
@@ -1944,12 +1948,12 @@ def pvsyst_celltemp(poa_global, wind_speed, temp_air, eta_m=0.1,
 
         If tuple/list, supply parameters in the following order:
 
-            * natural_convenction_coeff : float
-                Natural convection coefficient. Freestanding default is 29,
+            * constant_loss_factor : float
+                Combined heat loss factor coefficient. Freestanding default is 29,
```
cwhanse: OK with this change

pvlib/pvsystem.py
```diff
@@ -1944,12 +1948,12 @@ def pvsyst_celltemp(poa_global, wind_speed, temp_air, eta_m=0.1,
...
                 fully insulated arrays is 15.
 
-            * forced_convection_coeff : float
-                Forced convection coefficient, default is 0.
+            * wind_loss_factor : float
+                Combined heat loss factor influenced by wind. Default is 0.
```
cwhanse: And this change is OK

pvlib/pvsystem.py
```diff
@@ -1965,25 +1969,22 @@ def pvsyst_celltemp(poa_global, wind_speed, temp_air, eta_m=0.1,
     photovoltaic modules." Progress in Photovoltaics 16(4): 307-315.
     """
 
-    temp_models = TEMP_MODELS['pvsyst']
+    pvsyst_defaults = {"freestanding": (29.0, 0), "insulated": (15.0, 0)}
```
cwhanse: Not OK with this change.  We want to group the sets of coefficients into a common dict outside of the function so that it is easier to write an `infer` method for `ModelChain`

pvlib/pvsystem.py
wholmgren: `sapm_celltemp` simply calls this keyword argument `model`. I like `model` or `loss_factors` or `heat_loss_factors` more than `temp_model`. I think `temp_model` could apply to the higher level `sapm_celltemp` vs. `pvsyst_celltemp`, but not here.

pvlib/pvsystem.py
wholmgren: @adriesse what was your reasoning for inlining the defaults here?

@cwhanse I don't think this makes any difference for `ModelChain`, at least at this point in time. I think the idea was to allow people to inspect and modify the dictionary programmatically instead of manually - which could be good or bad.

pvlib/pvsystem.py
cwhanse: I'd prefer the same keyword and behavior in both functions.  Either a string which maps to an entry in `TEMP_MODELS`, a dict, or (cringe) a list or tuple.  I'm OK with `model` but `loss_factors` and variants only describe coefficients for the PVsyst model, the SAPM coefficients have different units.

adriesse: `loss_factors` is more in line with generic heat transfer, at least that's what motivated my suggestion.  I dislike the `*model*` variations because in my view the numbers represent model *parameters*, not models.

pvlib/pvsystem.py
adriesse: I think the version of `pvsystem.py` I started with had this line, and I merely changed the name.  But I see where you're coming from with the grouping.  How easy is it for a use to find out what the available options are from your global list?

adriesse: I see `infer_xxx` more as a work-around generally.  Isn't there some kind of dictum about explicit being better than implicit?  :)

adriesse:
Here are some additional thoughts about things I did not include in the PR:

- In the function argument list, I believe windspeed would be better placed *after* temperature, and given a default value--probably zero.
- If alpha is to be exposed, then it could/should be dealt with in the same way as the loss factors rather than as an independent function argument.  If this were done, then the name `loss_factors` would definitely no longer be appropriate, and I would suggest something like `model_parms` instead.
- For my own work I still like dictionaries for presets, so my functions don't need to know about them.  So I would use:
  - `pvsyst_celltemp (poa, wind, temp, **freestanding)` or 
  - `pvsyst_celltemp (poa, wind, temp, **model_parms['freestanding'])`

I do have some new work on temperature models which will raise some additional interesting code integration questions, but that's still a few months away at best.

pvlib/pvsystem.py
cwhanse: The idea of `infer_` at the `ModelChain` level is to let a user specify a racking model (one of the available keys in `TEMP_MODELS`) for a PVSystem instance and pvlib will figure out which cell temperature model to use.

pvlib/pvsystem.py
cwhanse: I am using the term model to refer to the vector of values, in the same way pvlib uses the keyword `model` for a vector of values for a DC model.  Open to other ideas for naming.

wholmgren:
> In the function argument list, I believe windspeed would be better placed after temperature, and given a default value--probably zero.

> For my own work I still like dictionaries for presets, so my functions don't need to know about them. So I would use: `pvsyst_celltemp (poa, wind, temp, **freestanding)` or `pvsyst_celltemp (poa, wind, temp, **model_parms['freestanding'])`

I agree with these proposals. They make the low level functions more flexible and more explicit. The doc strings can point users to the presets in the module-level dictionary. The `PVSystem` and `ModelChain` abstractions can still make the function a bit easier to use for common use cases.

We'd need a deprecation cycle to make the same changes to `sapm_celltemp`. I don't have a  preference about if that happens for 0.6.1 or not.

pvlib/pvsystem.py
mikofski: IMO there should be a distinction between porcelain and plumbing. Stop me if I'm preaching to the choir.

### plumbing
Low level functions should be explicit, arguments should be all spelled out, more primitive and native as you go lower level, features should simple

### porcelain
High level classes and functions should make things easy for the user. This may mean that they **_infer_** the users desire, they have preset defaults, they do lots of fancy magic without the user having to understand all the details

adriesse: I certainly understand and appreciate the intent. Is there is a difference from the DC_MODEL_PARAMS situation?  It seems that there you're detecting parameter *names*, whereas with the temperature models you're planning to detect parameter *values*.

cwhanse: You are correct, with DC_MODEL_PARAMS we're inferring a DC model from the field names, with TEMP_MODELS we're inferring a cell temperature model from the racking name string.  TEMP_MODELS is a bit clunky, and could be improved with a parallel to DC_MODEL_PARAMS, and a new function that translates from the racking strings to parameter sets.

pvlib/pvsystem.py
adriesse: I'd be fine with that.

cwhanse: I wouldn't.  alpha and eta_m are (should be) module specific, where as U0 and U1 are more tied to the mounting arrangement.

pvlib/pvsystem.py
```diff
@@ -1908,33 +1908,36 @@ def sapm_celltemp(poa_global, wind_speed, temp_air,
...
     poa_global : numeric
         Total incident irradiance in W/m^2.
 
-    wind_speed : numeric
-        Wind speed in m/s at a height of 10 meters.
-
     temp_air : numeric
         Ambient dry bulb temperature in degrees C.
 
+    wind_speed : numeric
```
cwhanse: format is `wind_speed : numeric, default 1.0`  Remove editorial parenthetical.  I'm OK with the default wind_speed value of 1.0.

pvlib/pvsystem.py
cwhanse: Name changes in [commit](https://github.com/pvlib/pvlib-python/pull/651/commits/c37cbf6cccb9d1c076aec5553a612a60e5707afb) are OK with me

adriesse:
Further changes anyone?

pvlib/test/test_pvsystem.py
```diff
@@ -1113,7 +1113,7 @@ def test_pvsyst_celltemp_with_index():
     irrads = pd.Series([0, 500, 0], index=times)
     winds = pd.Series([10, 5, 0], index=times)
 
-    pvtemps = pvsystem.pvsyst_celltemp(irrads, winds, temps)
+    pvtemps = pvsystem.pvsyst_celltemp(irrads, temps, winds)
```
wholmgren: Keyword args should be tested as keyword args rather than positional. So `wind_speed=winds` here and below.

wholmgren:
Ok to merge if CI passes.

pvlib/pvsystem.py
```diff
@@ -536,7 +536,7 @@ def pvsyst_celltemp(self, poa_global, wind_speed, temp_air):
         kwargs = _build_kwargs(['eta_m', 'alpha_absorption'],
                                self.module_parameters)
         return pvsyst_celltemp(poa_global, wind_speed, temp_air,
```
cwhanse: Need to update argument order here.  Change argument order of `PVSystem.pvsyst_celltemp` to match the underlying function's order of irradiance, temperature, windspeed

pvlib/pvsystem.py
```diff
@@ -521,7 +521,7 @@ def sapm_effective_irradiance(self, poa_direct, poa_diffuse,
             poa_direct, poa_diffuse, airmass_absolute, aoi,
             self.module_parameters, reference_irradiance=reference_irradiance)
 
-    def pvsyst_celltemp(self, poa_global, wind_speed, temp_air):
+    def pvsyst_celltemp(self, poa_global, temp_air, wind_speed):
```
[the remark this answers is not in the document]
adriesse: Done.

cwhanse:
@adriesse thanks Anton!

adriesse:
You're welcome!

PR #655, opened 2019-02-04
make whatsnew for v0.6.2

cwhanse:
housekeeping

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.6.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.6.2.rst
@@ -9,8 +9,6 @@ release.
 **Python 2.7 support will end on June 1, 2019**. Releases made after this
 date will require Python 3. (:issue:`501`)
 
-**Minimum pandas requirement bumped 0.14.0=>0.15.0**
-
 
 API Changes
 ~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


docs/sphinx/source/whatsnew/v0.6.2.rst
```diff
@@ -0,0 +1,32 @@
...
+v0.6.2 (exp. 31 March 2019)
+---------------------------
+
+This is a minor release. We recommend all users of v0.6.1 upgrade to this
+release.
+
+**Python 2.7 support will end on June 1, 2019**. Releases made after this
+date will require Python 3. (:issue:`501`)
+
+**Minimum pandas requirement bumped 0.14.0=>0.15.0**
```
wholmgren: this happened in the previous release, so no need to include it here.

cwhanse: removed.  Update Zenodo badge, or is that automatic?  I checked zenodo.org and v0.6.1 appears.

wholmgren: The DOI line README.md file currently links to 0.6.0. I think it's good practice to update it for all releases, but ok with me if we do that for 0.x releases.

PR #666, opened 2019-02-19
add CRN file parser

wholmgren:
- [x] Closes issue #xxxx

Adds a parser for the US Climate Reference Network files available at https://www.ncdc.noaa.gov/crn/qcdatasets.html

The map below shows the locations of these sensors.

screen shot 2019-02-18 at 7 41 22 pm

Question: what should we do about the columns that pvlib does not use (e.g. precip, soil moisture, surface temp)? I'm keeping them at the moment. I could add a kwarg, but I'd prefer to keep it simple unless there's a good reason to complicate it.


wholmgren:
At first I only spec-ed the dtype for irradiance since that was the one that caused problems in the test, then I added the others that I actually cared about, but yes, better to just specify all of them. Latest commit also works around some pandas versions issues.

wholmgren:
I finally have all of the tests passing. I increased the minimum pandas version to 0.16 so that I could use `Series.str.zfill` and convert HMM format to HHMM format. Maybe there's another way, but this version of pandas did not remove much, so I don't think it's a problem.

cwhanse:
About the non-solar fields: I think it's simpler to return the full CRN file content and let the pvlib user extract the data of interest for their application.

wholmgren:
Works for me. 

I've found some issues with the `read_fwf` function default behavior. It seems that only the first line of the file is used to determine the field widths, so we'll need to hard code them. I actually did this in a prototype version of the code but removed them once I had the test data working without them. I'll expand the test data set to cover the problematic cases.

wholmgren:
GHI at Tucson AZ 
crns0101-05-2019-az_tucson_11_w
Boulder CO
crns0101-05-2019-co_boulder_14_w
Oakley KS
crns0101-05-2019-ks_oakley_19_ssw
Socorro NW
crns0101-05-2019-nm_socorro_20_n

All times are UTC

I've looked at ~50 of ~150 sites and they all look reasonable in the first week of 2019.

wholmgren:
If there are no further comments, I think this is ready to merge.

pvlib/iotools/crn.py
```diff
@@ -0,0 +1,101 @@
+"""Functions to read data from the US Climate Reference Network (CRN).
+"""
+
+import pandas as pd
+import numpy as np
+from numpy import dtype
+
+
+HEADERS = 'WBANNO UTC_DATE UTC_TIME LST_DATE LST_TIME CRX_VN LONGITUDE LATITUDE AIR_TEMPERATURE PRECIPITATION SOLAR_RADIATION SR_FLAG SURFACE_TEMPERATURE ST_TYPE ST_FLAG RELATIVE_HUMIDITY RH_FLAG SOIL_MOISTURE_5 SOIL_TEMPERATURE_5 WETNESS WET_FLAG WIND_1_5 WIND_FLAG'  # noqa: E501
```
mikofski: can't this long line be wrapped?

```python
HEADERS = ('WBANNO UTC_DATE UTC_TIME LST_DATE LST_TIME CRX_VN LONGITUDE LATITUDE AIR_TEMPERATURE',
    ' PRECIPITATION SOLAR_RADIATION SR_FLAG SURFACE_TEMPERATURE ST_TYPE ST_FLAG RELATIVE_HUMIDITY',
    ' RH_FLAG SOIL_MOISTURE_5 SOIL_TEMPERATURE_5 WETNESS WET_FLAG WIND_1_5 WIND_FLAG')  # noqa: E501
```

wholmgren: It could be. I copy/pasted that from the CRN HEADERS.txt file. I've gone back and forth on what is most clear in this situation.

pvlib/iotools/crn.py
```diff
@@ -0,0 +1,101 @@
...
+# specify dtypes for potentially problematic values
+DTYPES = [
+    dtype('int64'), dtype('int64'), dtype('int64'), dtype('int64'),
+    dtype('int64'), dtype('int64'), dtype('float64'), dtype('float64'),
+    dtype('float64'), dtype('float64'), dtype('float64'),
+    dtype('int64'), dtype('float64'), dtype('O'), dtype('int64'),
+    dtype('float64'), dtype('int64'), dtype('float64'),
+    dtype('float64'), dtype('int64'), dtype('int64'), dtype('float64'),
+    dtype('int64')
+]
```
mikofski: nit: why not list comprehension?

```python
DTYPES = [np.dtype(t) for t in (
    # not necessary to use addition and multiplication here if undesired for readability
    ['int64']*6 + ['float64']*5
    + ['int64', 'float64', 'O', 'int64', 'float64', 'int64', 'float64', 'float64', 'int64', 'int64',
       'float64', 'int64'])]
```

wholmgren: turns out that you can supply a list of strings and the `dtype` function is not necessary.

wholmgren:
Thanks for the suggestions @mikofski!

PR #667, opened 2019-02-28
add SOLRAD data parser to iotools

wholmgren:
https://www.esrl.noaa.gov/gmd/grad/solrad/index.html

 - [ ] Closes issue #xxxx

~Needs tests.~


pvlib/test/test_solrad.py
```diff
@@ -0,0 +1,106 @@
...
+from numpy import nan
+
+import pytest
+
+from pvlib.iotools import solrad
+
+
+test_dir = os.path.dirname(
+    os.path.abspath(inspect.getfile(inspect.currentframe())))
+testfile = os.path.join(test_dir, '../data/abq19056.dat')
```
cwhanse: should be `testfile = os.path.join(test_dir, 'data', '703165TY.csv')` for cross-platform compatibility

pvlib/test/test_solrad.py
```diff
@@ -0,0 +1,106 @@
...
+testfile_mad = os.path.join(test_dir, '../data/msn19056.dat')
```
cwhanse: see above comment

pvlib/test/test_solrad.py
wholmgren: The path you've specified would be `pvlib/test/data/703165TY.csv`, but the data is in `pvlib/data/abq19056.dat`. I am pretty sure that `os.path` handles the `..` specification as needed for the platform. The appveyor builds do not object to this pattern.

wholmgren:
some examples from the NOAA ftp `realtime` folder:

abq19059
msn19059

pvlib/test/test_solrad.py
cwhanse: Of course.  I substituted a file I have to test.  Doesn't work on Windows.  `C:\python\pvlib-dev\pvlib-python\pvlib\../data/abq19056.dat`

wholmgren: Hm... Appveyor tests run on Windows so I am quite confused.

cwhanse: Hmmm... the string is interpreted correctly when a file operation is performed, so it's OK as coded.  I still think my suggestion is an improvement.
```

import os
import inspect

test_dir = os.path.dirname(
    os.path.abspath(inspect.getfile(inspect.currentframe())))

print(test_dir)

testfile = os.path.join(test_dir, '../data/703165TY.csv')
print(testfile)

with open(testfile) as infile:
    r = infile.readline()
    print(r)
```
produces
```
C:\python\pvlib-dev\pvlib-python\pvlib\test
C:\python\pvlib-dev\pvlib-python\pvlib\test\../data/703165TY.csv
703165,"SAND POINT",AK,-9.0,55.317,-160.517,7
```

wholmgren: Is '..' acceptable? `testfile = os.path.join(test_dir, '..', 'data', '703165TY.csv')`

wholmgren: I see it's called `pardir`: https://docs.python.org/3/library/os.html#os.pardir

cwhanse: that would maintain the OS-specific separator so the string doesn't have a mix of \ and /

cwhanse: or use `os.pardir` to avoid the `..`

wholmgren: Thanks. I thought `os.path.join` was smarter than it was because the tests have never complained about that pattern in the past. Looking forward to the Python 3 only days and using [pathlib](https://docs.python.org/3/library/pathlib.html).

PR #694, opened 2019-04-25
ENH: Create psm3.py

### LINKED ISSUE #592: make new PSM reader for iotools
**Problem**:
This new API is not trivial I think I was able to download data once, but it wasn't immediately intuitive to me.

**Proposal**:
Given a user's NREL API key, they can get PSM data and use it for weather and solar resource

**Alternatives**
just figure it out myself, then maybe an iotool to consume whatever is downloaded

**Additional context**
[nrel psm api v2](https://developer.nrel.gov/docs/solar/nsrdb/psm_data_download/)
[new nsrdbsource](https://nsrdb.nrel.gov/current-version)
[download instructions](https://nsrdb.nrel.gov/download-instructions)
related to #272

mikofski:
initial concept for psm3 reader

 - [x] Closes  #592


cwhanse:
@mikofski how does this pass lat / long ?

mikofski:
@cwhanse the arguments for `get_psm3()` are latitude, longitude, tmy, and interval - tmy defaults to `'tmy'` and interval defaults to 60

cwhanse:
Aha I see it now.

mikofski:
There are a lot of typos in here, but it does work, I'll fix the typos and upload a test ASAP

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Parameters
+    ----------
+    latitude : float or int
+        in decimal degrees, between -90 and 90, north is positive
+    longitude : float or int
+        in decimal degrees, between -180 and 180, east is positive
+    tmy : str
+        PSM3 parameter "name", see notes below for options, default ``'tmy'``
+    interval : integer
```
wholmgren: `int` or `integer`? `int` used above.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+    Get PSM3 data
+
+    Parameters
+    ----------
+    latitude : float or int
+        in decimal degrees, between -90 and 90, north is positive
+    longitude : float or int
+        in decimal degrees, between -180 and 180, east is positive
+    tmy : str
+        PSM3 parameter "name", see notes below for options, default ``'tmy'``
```
wholmgren: numpydoc standard says "Enclose variables in single backticks." 

https://numpydoc.readthedocs.io/en/latest/format.html section 4. Parameters

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    Get PSM3 data
+
+    Parameters
+    ----------
+    latitude : float or int
+        in decimal degrees, between -90 and 90, north is positive
+    longitude : float or int
+        in decimal degrees, between -180 and 180, east is positive
+    tmy : str
```
wholmgren: why is the function kwarg named `tmy` if it maps to the PSM3 parameter `name`?

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+        PSM3 parameter "name", see notes below for options, default ``'tmy'``
+    interval : integer
+        interval size in minutes, can be only either 30 or 60, default 60
+
+    Returns
+    -------
+    headers : dict
+        meta data from NREL PSM-3 about the record, see notes for fields
```
wholmgren: metadata (one word)

PSM-3 or PSM3? both are used in the doc string

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+    -----
+    The ``tmy`` parameter 'names' must be a single value from the following
+    list:
+
+        ['1998', '1999', '2000', '2001', '2002', '2003', '2004', '2005',
+         '2006', '2007', '2008', '2009', '2010', '2011', '2012', '2013',
+         '2014', '2015', '2016', '2017', 'tmy', 'tmy-2016', 'tmy-2017',
+         'tdy-2017', 'tgy-2017']
+
+    The return is a tuple with two items. The first item is a header with meta
```
wholmgren: This phrasing implied to me that you'd describe the second item, too.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+    References
+    ----------
+    [1] `NREL Developer Network - Physical Solar Model (PSM) v3
+        <https://developer.nrel.gov/docs/solar/nsrdb/psm3_data_download/>`_
+    [2] `NREL National Solar Radiation Database (NSRDB)
+        <https://nsrdb.nrel.gov/>`_
+    """
+    longitude = ('%9.4f' % longitude).strip()
+    latitude = ('%8.4f' % latitude).strip()
+    params = {
```
wholmgren: Do we need to provide a way for users to specify at least their own api key? I suspect so. I'd hope that the name, email, affiliation, reason fields are not required and are linked to the api key when a user registers.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+import io
+import csv
+import datetime
+import requests
+import pandas as pd
+import pytz
+
+URL = "http://developer.nrel.gov/api/solar/nsrdb_psm3_download.csv"
+
+# 'relative_humidity', 'total_precipitable_water' are note available
```
wholmgren: note -> not

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+        'attributes':  ','.join(ATTRIBUTES),
+        'leap_day': 'false',
+        'utc': 'false',
+        'interval': interval
+    }
+
+    s = requests.get(URL, params=params)
+    if s.ok:
+        f = io.StringIO(s.content.decode('utf-8'))
+        x = csv.reader(f, delimiter=',', lineterminator='\n')
```
wholmgren: need a comment about metadata parsing and data parsing.

more descriptive variable names would also help

consider using readline to pop off the metadata rows, then pass the buffer to `pd.read_csv`.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+        z = dict(zip(y[0], y[1]))
+        # in USA all timezones are integers
+        z['Local Time Zone'] = int(z['Local Time Zone'])
+        z['Time Zone'] = int(z['Time Zone'])
+        z['Latitude'] = float(z['Latitude'])
+        z['Longitude'] = float(z['Longitude'])
+        z['Elevation'] = int(z['Elevation'])
+        tz = pytz.timezone('Etc/GMT%+d' % -z['Time Zone'])
+        w = pd.DataFrame(y[3:], columns=y[2], dtype=float)
+        w.Year = w.Year.apply(lambda yr: int(yr))
```
wholmgren: I don't understand why you specify dtype=float above and then set the columns to be integers here. Wouldn't it be better to specify the correct dtype when creating the dataframe?

If we keep this, I think `.astype(int)` is preferable to the element by element apply.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+        w.Month = w.Month.apply(lambda mt: int(mt))
+        w.Day = w.Day.apply(lambda dy: int(dy))
+        w.Hour = w.Hour.apply(lambda hr: int(hr))
+        w.Minute = w.Minute.apply(lambda mn: int(mn))
+        w.index = w.apply(
+            lambda dt: tz.localize(datetime.datetime(
```
wholmgren: Use `pd.to_datetime`

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+                int(dt.Year), int(dt.Month), int(dt.Day),
+                int(dt.Hour), int(dt.Minute))), axis=1)
+        return z, w
+    raise requests.HTTPError(s.json()['errors'])
```
wholmgren: all of the above is in an if block, so maybe put this in an else block.

pvlib/iotools/psm3.py
mikofski: good point, `name` just seemed to vague to me, but matching their API is probably more important, should I change it to `name` then?

pvlib/iotools/psm3.py
mikofski: @cdeline do you know what NREL's stance is on this?

pvlib/iotools/psm3.py
wholmgren: I agree that `names` is vague, but I don't think `tmy` right here. I'd make it consistent.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,133 @@
...
+    requests.HTTPError
+
+    Notes
+    -----
+    The ``tmy`` parameter 'names' must be a single value from the following
+    list:
+
+        ['1998', '1999', '2000', '2001', '2002', '2003', '2004', '2005',
+         '2006', '2007', '2008', '2009', '2010', '2011', '2012', '2013',
+         '2014', '2015', '2016', '2017', 'tmy', 'tmy-2016', 'tmy-2017',
```
wholmgren: `tmy` and variants do not appear to be options on the API documentation linked to above. Did I miss it?

mikofski: IMO their documentation is spotty.  `'tmy'` definitely works I haven't tried the variants.  There's an example of downloading one year [here](https://nsrdb.nrel.gov/api-instructions) the docs do say that only one year can be downloaded this way but I haven't tested that.  The example says zenith can be downloaded,  but in tests I couldn't get precipitable water or relative humidity. This api could use more rigorous testing due to the spotty documentation IMO

mikofski:
Bummer, i don't think the `dtype` arg in `read_csv` was available yet in pandas-0.16 so the [python-27-min](https://github.com/pvlib/pvlib-python/blob/master/setup.py#L41) build fails

mikofski:
How's it looking? Ready?

pvlib/iotools/psm3.py
cdeline: I'll double-check with Manajit.  From the API documentation [https://nsrdb.nrel.gov/api-instructions](url) there are several things requested in addition to api_key. Unclear if these are stored at time of API key registration :


`your_name = 'John+Smith'  # Your full name, use '+' instead of spaces.`
`reason_for_use = 'beta+testing' # Your reason for using the NSRDB.`
`your_affiliation = 'my+institution' # Your affiliation`
`your_email = 'john.smith@server.com' # Your email address`
`mailing_list = 'true' # Please join our mailing list so we can keep you up-to-date on new developments.`

cwhanse: @mikofski your timing with this PR could not have been better.  FYI, `get_psm3` works passing my own API key and no other meta data.

cdeline: Here's a response from the NSRDB API developer, Paul Edwards:

>OK, as to the first question about user metadata, technically speaking the email address is the only field that is strictly required because the system uses this to automatically communicate messages back to the user if necessary (such as a download failure etc). However we would strongly encourage your team to include the rest! We use the other fields to collect user stats for internal use and these stats have been vital in our ability to report accurately back to DOE on the usefulness of these services. Also, we don’t have access to the fields associated with the API key. In fact when developer.nrel.gov proxies the request through to our code it doesn’t even include the API key at all. 

>Regarding the “DEMO_KEY” api key, that key is severely rate limited, and is also used in tons of examples all over the place for 100s of endpoints. Consequently it is not uncommon to find that using that key will result in rejected requests due to the rate limit. My recommendation is to go ahead and include it in your examples etc, but with strong disclaimers along the lines of what I just said… nothing is going to work right until one uses their own API key. There is actually a documentation page with more details that your team could use to inform how to write code to report back to users when the rate limit is the cause of failures if that is something you’re interested in.

https://developer.nrel.gov/docs/rate-limits/


>Oh yeah, I missed one of your points. Yes, I think it would be neat if the “reason” attribute was something like “pvlib”. We could even pull stats for the PVLib team after a while if that was useful to them.

>Let me know if this helps, 
Paul

cdeline: So no, it doesn't appear that user information is connected back to the api key for any reason.

Here's what I would recommend based on this feedback:
1. Include 'pvlib' somewhere in the api call, like 'reason':'pvlib'.  So we can track the number of users that take advantage of this function.
2. Have users pass their email address as a separate function input along with the api key.
3. Include in the documentation a suggestion that users optionally hard-code other specific details such as affiliation.  
4. Set mailing_list = false by default if we're requesting email addresses (?)

wholmgren: Thanks @cdeline. 

Maybe arg `api_key` and kwargs for the other parameters (full_name, email, affiliation, reason, mailing_list). Use the demo key in examples/tests. I doubt many people will use their own API key if it's only a kwarg.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,148 @@
...
+        'names': names,
+        'attributes':  ','.join(ATTRIBUTES),
+        'leap_day': 'false',
+        'utc': 'false',
+        'interval': interval
+    }
+    # request CSV download from NREL PSM3
+    response = requests.get(URL, params=params)
+    if not response.ok:
+        raise requests.HTTPError(response.json()['errors'])
```
cwhanse: I think it would be helpful to return the error message, or None.  In my workflow I'm looping over lat, long, and so put the call to `get_psm3` inside a `try:except` block to avoid exiting the loop on the first error.  NREL puts information in the response error when, for example, a lat/long is passed where there is no NSRDB data.  In my case I could decide what to do in the `except` block based on the error message.

To be explicit:
`return header, data, response.json()['errors']`

mikofski: I don't understand. AFAIK, that's exactly what this does. here's my workflow:
```python
try:
    header, data = get_psm3(latitude, longitude)
except requests.HTTPError as exc:
    print(exc)
```

but actually it looks more like this
```python
@app.route('/', methods=['GET', 'POST'])
def myfun(latitude=LATITUDE, longitude=LONGITUDE, tracker=False,
            surface_azimuth=0.0, surface_tilt=20.0, module_type=None,
            project_size=None, dst=False):
    kwargs = {'title': 'PVWatts'}
    if request.method == 'GET':
        try:
            # FYI: calculate() is a method that calls get_psm3()
            mpp, header = calculate(
                latitude, longitude, tracker, surface_tilt, surface_azimuth)
        except requests.HTTPError as exc:
            flash(exc, 'error')
            return render_template('pvwatts.html', title='PVWatts')
        else:
            # do stuff
```
so if I put a location outside the US, like Bristol-UK, then it flashes the errors on the screen, exactly as they were returned by psm3
get_psm3
Does this work for you?

mikofski: @cwhanse also note that if the response, `r`, is successful, it returns the CSV data in `r.content`, and `r.json()` doesn't work. Therefore `r.json()['errors']` only works if the response is **not** okay. And vice-versa, if the response isn't okat, then there's no CSV file, and therefore, no header or data., So the suggestion to do this:
```python
return header, data, response.json()['errors']
```
wouldn't work, it would have to be more complicated, like this:
```python
if r.ok:
    return pvlib.iotools.read_tmy3(StringIO(response.content))
else:
    return response.json()['errors']
```
PyLint would complain about this signature because it returns different types and number of returns depending on the outcome which is not a consistent API. We could do some workarounds to make it consistent, like you say, return `None` when headers, data, or errors are missing, but IMO that's not as clean as just raising an exception and letting the caller decide what they want to do.

mikofski:
@cwhanse and @wholmgren not sure what's up with codecov, it won't even let me log in, everything else is green

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,174 @@
...
+    Get PSM3 data
+
+    Parameters
+    ----------
+    latitude : float or int
+        in decimal degrees, between -90 and 90, north is positive
+    longitude : float or int
+        in decimal degrees, between -180 and 180, east is positive
+    api_key : str
+        required NREL Developer Network API key
```
wholmgren:
```suggestion
        NREL Developer Network API key
```

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,174 @@
...
+    email : str
+        required, system uses this to automatically communicate messages back
```
wholmgren:
```suggestion
        NREL API uses this to automatically communicate messages back
```

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,174 @@
...
+
+# 'relative_humidity', 'total_precipitable_water' are not available
+ATTRIBUTES = [
+    'air_temperature', 'dew_point', 'dhi', 'dni', 'ghi', 'surface_albedo',
+    'surface_pressure', 'wind_direction', 'wind_speed']
+PVLIB_PYTHON = 'pvlib python'
+
+
+def get_psm3(latitude, longitude, api_key, email, names='tmy', interval=60,
+             full_name=PVLIB_PYTHON, affiliation=PVLIB_PYTHON):
```
wholmgren: does sphinx render this as `full_name='pvlib python'` or `full_name=PATH_TO_SOME_OBJECT`? If the former, ok as is. If the latter, better to be repeat the string for the two default values.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,174 @@
...
+    longitude : float or int
+        in decimal degrees, between -180 and 180, east is positive
+    api_key : str
+        required NREL Developer Network API key
+    email : str
+        required, system uses this to automatically communicate messages back
+        to the user only if necessary
+    names : str
+        PSM3 API parameter specifing year or TMY variant to download, see notes
+        below for options, default is ``'tmy'``
```
wholmgren: the "default is" pattern here and below is inconsistent with our current practice of putting "default X" next to the type spec. I don't care too much about consistency in `iotools`, so I'm ok with letting it go. Up to you or others.

pvlib/iotools/psm3.py
wholmgren: Did you try the variants? They don't need to be automatically tested, but I'd like to know that someone checked that they worked.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,174 @@
...
+    References
+    ----------
+
+    * `NREL Developer Network - Physical Solar Model (PSM) v3
+      <https://developer.nrel.gov/docs/solar/nsrdb/psm3_data_download/>`_
+    * `NREL National Solar Radiation Database (NSRDB)
+      <https://nsrdb.nrel.gov/>`_
+
+    """
+    longitude = ('%9.4f' % longitude).strip()
```
wholmgren: If the formatting is important, we need a comment saying so. If not, let's remove it.

pvlib/test/test_psm3.py
```diff
@@ -0,0 +1,57 @@
...
+    assert np.allclose(data['Wind Direction'], expected['Wind Direction'])
+    # check header
+    for hf in HEADER_FIELDS:
+        assert hf in header
+    # check timezone
+    assert (data.index.tzinfo.zone == 'Etc/GMT%+d' % -header['Time Zone'])
+    # check errors
+    with pytest.raises(HTTPError):
+        psm3.get_psm3(LATITUDE, LONGITUDE, api_key='BAD', email=PVLIB_EMAIL)
+    with pytest.raises(HTTPError):
```
wholmgren: why does this one fail?

pvlib/iotools/psm3.py
mikofski: Yes,[it](https://mikofski-pvlib-python.readthedocs.io/en/patch-1/generated/pvlib.iotools.get_psm3.html) renders as "pvlib python" for both. Yoohoo! 🕺

pvlib/iotools/psm3.py
mikofski: I'm hoping that's up for discussion, I don't want to break with convention, but I wish the default was not in the type spec, seems out of place there to me, but I could still be convinced otherwise - tell me to do it and I will :)

pvlib/iotools/psm3.py
mikofski: I disagree, the WKT is unnecessarily strict, I think we can do this nice thing for the user and make it just work. The user inputs two numbers, either integer or float, and we convert it into a WKT which has 4 decimals, longitude before latitude, and exactly one space between them. I think we could possibly add a `format_WKT()` method in tools, instead of embedding it here, so that other tools could use that method, but asking the user to format the WKT is IMHO not a wise choice.

mikofski: maybe these lines need some comments to make it clearer to future developers?

pvlib/test/test_psm3.py
mikofski: * `api_key`="BAD" fails because the response is HTTP/1.1 403 Forbidden and there's a long error message in the content, no JSON, that says something like, "please go to NREL developer network and register for a key"
* Bristol-UK fails because PSM3 only works in the US, maybe South America, and I think some parts of India perhaps around Chennai? I can confirm - perhaps this should be a warning in the docstring?

perhaps I should add some more comments here to explain these tests - like I did here - for future devs?

pvlib/iotools/psm3.py
cwhanse: I think using a helper function that returns the formatted string would help a lot

pvlib/iotools/psm3.py
cwhanse: I prefer (but not strongly) the `names : str, default ``'tmy'`` pattern, for the way it displays on readthedocs.

pvlib/iotools/psm3.py
wholmgren: All I'm asking for is a code comment along the lines of `# format into WKT spec`

pvlib/test/test_psm3.py
wholmgren: BAD was self explanatory. I don't think we need to test that it fails outside the US. Not our problem.

pvlib/iotools/psm3.py
cwhanse: My workflow issues appear to be my misunderstanding of how to use `try/except` in combination with errors raised by a function called in the `try` block.  From my point of view, `get_psm3` is enahnced if it can help with error management without compromising its purpose (get and return PSM3 data for one location). Merely displaying the error and exiting is likely to cause frustration.  But I don't have a concrete suggestion if returning the error isn't desireable.

pvlib/iotools/psm3.py
wholmgren: Yes, up for discussion, no, not in this PR! Please make a new issue and collect some examples of projects that follow different standards.

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,174 @@
...
+    headers : dict
+        metadata from NREL PSM3 about the record, see notes for fields
+    data : pandas.DataFrame
+        timeseries data from NREL PSM3
+
+    Raises
+    ------
+    requests.HTTPError
+        if the request return status is not ok then the ``'errors'`` from the
+        JSON response will be returned as an exception
```
cwhanse: `raised as an exception`

pvlib/iotools/psm3.py
wholmgren: In general in python, if errors cannot be handled within the function, they should be raised rather than returned.

pvlib/test/test_psm3.py
cwhanse: PSM3 has data in a lat/long rectangle around the lower 48 states, so there's coverage in northern Mexico for example.  But it has a coastline filter and excludes data in the Gulf of Mexico.  I think we should test for failure at lat/long where PSM3 does not have data.

wholmgren: Does PSM3 return a sensible error if lat/lon outside of its area?

cwhanse:
```
from get_psm3 import get_psm3

import api_keys

MY_KEY = api_keys.get_api_key('nsrdb')

for lat, lon in zip([25, 25, 25], [-125, -124.5, -124]):
    try:
        meta, data = get_psm3(lat, lon, names='tmy', interval=60,
                              api_key=MY_KEY)
    except Exception as e:
        print("{}, {}: ".format(lat, lon), e)
```

Returns
```
25, -125:  ["The 'wkt' parameter did not match any known data points."]
25, -124.5:  Expecting value: line 1 column 1 (char 0)
25, -124:  Expecting value: line 1 column 1 (char 0)
```

The first line is what I expect, the next 2 are json decoding errors that puzzle me.

cwhanse: Aha, I suspect the 2nd and 3rd messages are empty responses because of the once-per-2second  [throttle](https://developer.nrel.gov/docs/solar/nsrdb/guide/), so `raise requests.HTTPError(response.json()['errors'])` is throwing the exception about Expecting value:...

pvlib/iotools/psm3.py
mikofski: @cwhanse [`requests.HTTPError`](https://2.python-requests.org/en/master/api/#requests.HTTPError) is a subclass of the base `RequestsException` that also has a `response` attribute that has the entire response from the request, including the `json()`, the `content`, and the `status_code` so you can catch the exception and handle it anyway you want. Printing out the message is just one option, you can get the response and do something else:
```python
import requests
URL = "http://developer.nrel.gov/api/solar/nsrdb_psm3_download.csv"
r = requests.get(URL)
x = requests.HTTPError(r.content, response=r)

# print message
try:
    raise x
except requests.HTTPError as exc:
    print(exc)
# b'Error Code,Error Message\n"API_KEY_MISSING","No api_key was supplied. Get one at https://developer.nrel.gov/signup/"'

# print response
try:
    raise x
except requests.HTTPError as exc:
    print(exc.response)
# <Response [403]>

# handle exception based on response status code
try:
    raise x
except requests.HTTPError as exc:
    if exc.response.status_code == 403:
        # do something else

x.response.__dict__
# ouptut
{'_content': b'Error Code,Error Message\n"API_KEY_MISSING","No api_key was supplied. Get one at https://developer.nrel.gov/signup/"',
 '_content_consumed': True,
 '_next': None,
 'status_code': 403,
 'headers': {'Server': 'openresty', 'Date': 'Thu, 02 May 2019 17:06:46 GMT', 'Content-Type': 'text/csv', 'Transfer-Encoding': 'chunked', 'Connection': 'keep-alive', 'Vary': 'Accept-Encoding', 'Access-Control-Allow-Origin': '*', 'X-Cache': 'MISS', 'Strict-Transport-Security': 'max-age=31536000; preload', 'Content-Encoding': 'gzip'},
 'raw': <urllib3.response.HTTPResponse at 0x15e832f92b0>,
 'url': 'http://developer.nrel.gov/api/solar/nsrdb_psm3_download.csv',
 'encoding': 'ISO-8859-1',
 'history': [],
 'reason': 'Forbidden',
 'cookies': <RequestsCookieJar[]>,
 'elapsed': datetime.timedelta(0, 0, 250634),
 'request': <PreparedRequest [GET]>,
 'connection': <requests.adapters.HTTPAdapter at 0x15e832fc2e8>}
```

pvlib/iotools/psm3.py
mikofski: no I didn't try the variants ...

mikofski:
I figured out why codecov is complaining, `psm3.py` has only 95% coverage because there are two lines in the python compatible import of `JSONDecodeError` that I can't test - this is only used in Python-3 and it's not covered by either six or python-future afaik. :(
image

the `JSONDecodeError` is used to get the `"errors"` field out of the response if it exists, other times though the error message is in the content only. We can safely remove the Python-2 compatibility when we officially drop Python-2

pvlib/iotools/psm3.py
wholmgren: or `if 'something I know how to handle' in str(exc): handle_it()`

pvlib/test/test_psm3.py
wholmgren: > ["The 'wkt' parameter did not match any known data points."]

This could be confusing to pvlib users because we're asking the to specify lat/lon instead of API. So I'm ok with catching this in the function and raising a more descriptive error about lat/lon.

pvlib/iotools/psm3.py
cwhanse: Empty or rejected responses due to the throttle limit on requests throw an error when parsing the response, see the comments below.  I suspect this situation will come up.  As far as I can tell, in this situation the status_code is 400 but the errors text is empty.

wholmgren:
@mikofski thanks for looking into the coverage. No need to fix it.

mikofski:
Is it okay? 🚀  I fixed the defaults, but I don't have time to check the variants for "names", has anyone else already done this? Anymore comments?

pvlib/iotools/psm3.py
```diff
@@ -0,0 +1,185 @@
...
+    }
+    # request CSV download from NREL PSM3
+    response = requests.get(URL, params=params)
+    if not response.ok:
+        # if the API key is rejected, then the response status will be 403
+        # Forbidden, and then the error is in the content and there is no JSON
+        try:
+            errors = response.json()['errors']
+        except JSONDecodeError:
+            errors = response.content.decode('utf-8')
```
cwhanse: This helps with the rate limit errors, which now produce
```
Error Code,Error Message
"OVER_RATE_LIMIT","You have exceeded your rate limit. Try again later or contact us at https://developer.nrel.gov/contact/ for assistance"
```

pvlib/iotools/psm3.py
cwhanse: `tmy-2016` works
`2016` works

mikofski:
@cwhanse do you approve of the changes? merge at will 🚀

cwhanse:
> @cwhanse do you approve of the changes? merge at will 🚀

@mikofski OK as is, would be nice to add timeout and retry logic to the request.get, but we can do that later if you want to open an issue.

mikofski:
>timeout and retry logic to the `requests.get`

I don't feel comfortable adding that though. I think current handling is sufficient, and I think anything further could and should be handled by the user because it may differ. Maybe you could add that as an issue?

My preference is for smaller methods that are supplemented later by more features as needed, versus adding a lot of functionality up front that might not be required.

PR #708, opened 2019-05-04
ENH: Port PVsyst and Desoto parameter estimation from MATLAB

mikofski:

 - [x] Closes issue #227

This is a continuation of #229 originally by @mattguttenberg in August, 2016


mikofski:
@cwhanse curious, do you know if the Schumaker quadratic spline parameters could be replaced by [`scipy.interpolate.BSpline`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.BSpline.html#scipy.interpolate.BSpline)? If not would you be interested in contributing your code to Scipy, or if you're not then could I contribute it?

cwhanse:
@mikofski Bspline isn't a replace for the Schumaker spline.  Schumaker's method constrains each spline to be convex down (negative quadratic coefficient) - this is important to mimic the expected shape of the IV curve.

I'd say yes to pushing to scipy, but we should reconsider how I handled the ends of the data. Schumaker's paper wasn't clear on that point.

mikofski:
Hi @cwhanse , @wholmgren , @adriesse , and whoever else, PTAL. I know there are still some documentation issues to clean up, and the code should be profiled to remove some bottlenecks, and I think some code can be swapped out for builtins (_EG_: `lambertw` may be redundant with `scipy.special`) so in other words there's still a lot more that could be done, but it is running and producing reasonable good results.

Also, I believe the codecov/patch fails to hit 100% coverage due to the `ImportError` checks for scipy and statsmodels, hopefully that's okay. I'm really looking forward to using this in the wild soon!

[pvlib-coverage.zip](https://github.com/pvlib/pvlib-python/files/3929884/pvlib-coverage.zip)

PS: Cliff - I only used the first 3000 of the PVsyst demo test points because something, I think `scipy.optimize.least_squares` was really take a long time on one of the last 585 points in that dataset. I will run some profiling to figure it out, maybe later?

mikofski:
Cliff, another note, I removed the graphing from `pvsyst_param_est` but I was planning on putting them back into the testing maybe or perhaps it could be in a jupyter notebook tutorial in the docs? My concerns were as follows:

1. I'm not sure how to test that code, I can make sure a plot was made, but that doesn't mean it plotted the correct thing,
2. it adds Matplotlib to the `extra_requires` which I wasn't sure would be wanted, and
3. there's never been any plotting tools in pvlib before, thought it might require a discussion?

mikofski:
Also Will and Cliff, I moved the old `ivtools.py` to `ivtools/__init__.py` which might be less preferable than say `ivtools/core.py` and then just copying the existing api to `ivtools/__init__.py`:

    # in pvlib/ivtools/__init__.py
    from pvlib.ivtools.core import fit_sde_sandia
    from pvlib.ivtools.core import fit_sdm_cec_sam
    from pvlib.ivtools.core import fit_sdm_desoto
    from pvlib.ivtools.pvsyst_parameter_estimation import (
        pvsyst_parameter_estimation)  # this is actually already there :)

I was lazy and just did the easy thing, but I also wasn't sure what you'd want to call `core.py`, maybe `api.py` is better? Or we can just leave it as is - whatever you prefer.

ci/requirements-py35.yml
```diff
@@ -6,6 +6,7 @@ dependencies:
     - coveralls
     - cython
     - ephem
+    - matplotlib
```
mikofski: not needed now

pvlib/ivtools/PVsyst_parameter_estimation.py
```diff
@@ -0,0 +1,731 @@
...
+
+    if np.isnan(gamma_ref) or np.isnan(mugamma) or not np.isreal(gamma_ref) \
+            or not np.isreal(mugamma):
+        badgamma = True
+    else:
+        badgamma = False
+
+    pvsyst = OrderedDict()
+
+    if ~badgamma:
```
mikofski: * This will **always** evaluate to `True`:
```python
>>> ~True
-2

>>> ~False
-1
```
* The Python tilde operator is the bit-complement, so `~0 -> -` and `~1 -> -2`.
* The **NumPy** tilde operator is `not`
* In Python (and NumPy) implicit booleaness is only false for zero:
```python
>>> if ~True: print('yes')
yes

>>> if ~False: print('yes')
yes

>>> if 0: print('yes')

```

cwhanse:
@mikofski thanks for pushing this PR towards a conclusion. I'm happy to review. Any objections if I push to your branch to address some of the clean up items you mention?

@wholmgren would appreciate your review for the API and changes to `ivtools.py`

wholmgren:
> and changes to ivtools.py

No problem with making `ivtools` a subpackage, but like @mikofski anticipated I don't like seeing all the algorithm code in the `__init__.py`. Seems like `fit.py` might be a better fit than `core.py` or `api.py`. Then use `__init__.py` to bring the functions into a single name space (if that's your preference).

> your review for the API 

sure, but later.

mikofski:
>Seems like `fit.py` might be a better _fit_

Nice pun, I'm okay with that. Cliff, does that sound good to you too?

>Any objections if I push to your branch

No objections, please do, and thanks for your help!

#### one more note

I noticed that I changed the flat slope tolerance in qspline (line 4) from 1e-6 to the cube root of double precision, which might have effects on the number of knots and the shape of the spline, but probably not, right?

cwhanse:
It is not clear to me (yet) how best to organize these functions that fit either the single diode equation (SDE) or one of the single diode models (SDM).  I agree with @wholmgren about not putting function declarations in `__init__.py`.  If we want to put several related  functions in a .py file, I'd suggest dividing by SDE or SDM, e.g. `sde.py` would have `fit_sde_sandia` etc. but the function name now looks redundant.  For now, maybe leave them all as separate .py files, like we are doing in `iotools.py`.

A vision for this part of pvlib is to eventually declare an IVCurve class that contains data, with a method `fit` that accepts kwargs for the model to be fit, and fitting method. Here we're assembling some library functions that could be called from the `fit` method.

For the PVsyst fitting procedure being ported, `rectify_iv_curve` could be useful outside of model fitting, for cleaning up measured IV curve data. `calc_theta_phi_exact` could be moved to `pvlib.singlediode` and renamed so it's purpose is clear. Many of the other functions are helpers that are specific to this Sandia procedure, but they are common to the Matlab code for the PVsyst, CEC and De Soto model fitting. I would make these functions private, and put them all into a .py file named `sandia_fitting_tools` or something similar. The API should expose `fit_sdm_pvsyst_sandia` (if we follow the current naming pattern of "fit" + "sdm" or "sde" + which SDM + method) and not much of the internals.

I think we should import and use the scipy lambertw function rather than this implementation, which is in PVLib Matlab because Matlab puts lambertw in an optional toolbox.

cwhanse:
>> Seems like fit.py might be a better fit
> Nice pun

Yep

> I'm okay with that. Cliff, does that sound good to you too?

I think we could segregate functions by SDE or SDM.

cwhanse:
Thanks @mikofski I'll make another PR to your branch, to improve consistency inside each module and work on the docstrings.

mikofski:
I think you can also just push directly to this PR

cwhanse:
Update:

1. Switching to `scipy.special.lambertw` changed the solution for a few IV curves in `PVsyst_demo.txt` to fall outside the test tolerance. I left the IV curves in the data set, and omit them from testing, since these curves may provide a case to improve the algorithm.
2. The test failure on py35-min is due to a change in `numpy.linalg.lstsq` kwarg `rcond` which happened at numpy v1.14.0. Changing to the new behavior `rcond=None` to silence a bunch of warnings causes the failures since py35-min is pinned to numpy 1.12. I'd say not worth programming around unless we plan for py35-min to be supported for a long time.

I still want to do some significant refactoring in `fit_sdm.fit_pvsyst_sandia`, see if I can silence a lot of the numpy warnings, and improve coverage.

mikofski:
Happy Holidays Cliff!

mikofski:
>test failure on py35-min...
Might be worth creating a `requires_numpy_1_14` decorator in `conftest` just to silence that error for now, since we just bumped to NumPy-1.12, and I don't know how long it will be to the next bump.

wholmgren:
> Might be worth creating a requires_numpy_1_14 decorator in conftest just to silence that error for now, since we just bumped to NumPy-1.12, and I don't know how long it will be to the next bump.

This is only an acceptable pattern for optional dependencies and for version issues with the test code itself (as opposed to issues with the feature code).

pvlib/ivtools/fit_sde.py
```diff
@@ -459,5 +461,13 @@ def _cocontent_regress(v, i, voc, isc, cci):
...
     # translate from coefficients in rotated space (gamma) to coefficients in
     # original coordinates (beta)
-    beta = np.linalg.lstsq(np.dot(mb, ma), gamma[0:5], rcond=None)[0]
+    beta = np.linalg.lstsq(np.dot(mb, ma), gamma[0:5], rcond=rcond)[0]
     return beta
+
+
+def _set_rcond():
+    # handles change in rcond in numpy.linalg.lstsq
+    if np.__version__ < '1.14':
```
kanderso-nrel: This is a risky pattern -- `'1.9' < '1.14'` is False, though since pvlib requires numpy 1.12+, maybe this won't be a problem until numpy version `1.100`...

cwhanse: Rats. Hopefully the intent is clear - any suggestions here?

kanderso-nrel: Ideally there would be something like `sys.version_info >= (3, 5)`, but there's nothing in `np.version` that isn't a string.  Off the top of my head I don't have anything better than a DIY approach:
```
In [53]: v = "1.13.9".split("."); (int(v[0]), int(v[1])) < (1, 14)
Out[53]: True

In [54]: v = "1.14.1".split("."); (int(v[0]), int(v[1])) < (1, 14)
Out[54]: False
```

cwhanse:
> If the issue is only a matter of dealing with warnings, can we wrap the numpy call in with np.errstate?

Possible, but not preferable. The change in numpy makes a numerical routine more reliable, so I'd prefer to use the new behavior. The warning is issued if the old behavior is used. When the new behavior is requested with numpy<1.14, you get a type error.

wholmgren:
We're a few days away from numpy 1.14 being covered by NEP-29. As I stated in #828, I'm ok following NEP-29 scheduling so long as there's a compelling reason (like this one). The major question is if we should do that in 0.7.1. or 0.8. Seems a little aggressive for 0.7.1 but I'm not entirely opposed.

pvlib/ivtools/fit_sde.py
cwhanse: Thanks! Perfectly fine for a workaround that will go away when we pin to numpy 1.14 or later.

wholmgren: we use `pkg_resources.parse_version` in [`conftest.py`](https://github.com/pvlib/pvlib-python/blob/master/pvlib/test/conftest.py#L7) but it appears that [`packaging.version.parse`](https://packaging.pypa.io/en/latest/version/#packaging.version.parse) is now [preferred](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python).

cwhanse: Weird - wonder how my response comment ended up as a review....

cwhanse:
One last item on my list before reviews: naming. Current, `ivtools.fit_sde` module has functions `fit_sandia` etc.; similar for `ivtools.fit_sdm.fit_cec_sam`. I want to drop one "fit" and lean towards `sde.py/fit_sandia` and `sdm.py/fit_cec_sam`.

Any opinions here?

mikofski:
@cwhanse what should users do if they only have the data sheet with STC values, and do not have an IV curve? or would we recommend digitizing IV curve images from datasheets to generate PAN files?

[image](https://www.canadiansolar.com/wp-content/uploads/2019/12/Canadian_Solar-Datasheet-HiDM_CS1H-MS_EN.pdf)
[Canadian_Solar Datasheet (12/2019) HiDM CS1H-MS](https://www.canadiansolar.com/wp-content/uploads/2019/12/Canadian_Solar-Datasheet-HiDM_CS1H-MS_EN.pdf)

adriesse:
I would not trust the images from the marketing department.

mikofski:
>I want to drop one "fit" and lean towards `sde.py/fit_sandia` and `sdm.py/fit_cec_sam`

I'm okay with that :)

cwhanse:
> @cwhanse what should users do if they only have the data sheet with STC values, and do not have an IV curve? or would we recommend digitizing IV curve images from datasheets to generate PAN files?

Personally, I agree with @adriesse - I wouldn't turn the pictures into "data." With only the data from the spec sheet (Isc, Voc, etc.) one could use `sdm/fit_cec_sam` to get a CEC performance model. There's not an equivalent technique to fit PVSyst that I know of, outside of PVSyst.

I know that some will fit one model (e.g. CEC), calculate IV curves, and then fit a second model to the calculated IV curves, as a way to get parameters for the second model when data is lacking. This chaining is questionable, in my opinion. The result is a model of a model, with rather weak validation since the data are lacking from the start.

mikofski:
>There's not an equivalent technique to fit PVSyst that I know of, outside of PVSyst.

I believe this may be a very common use case, maybe we should try to address it?
* I have heard from users that they prefer pvsyst over other models
* I believe having a large set of reliable IV curve is uncommon and difficult to obtain,
* I think many users with only a dataset will try to take the workaround path you mention

cwhanse:
> > There's not an equivalent technique to fit PVSyst that I know of, outside of PVSyst.
> 
> I believe this may be a very common use case, maybe we should try to address it?

I agree that a datasheet-to-PVSyst function would attract users, and a successful method would be a nice journal publication.  PVSyst includes some tools to help users get parameters from their data. Some time ago, Ken Sauer, Tom Roessler and I worked at a way to converge to better parameters if one had reasonable starting values. That method could apply here: datasheet to CEC, then use CEC parameters + PVSyst defaults as a starting point.  FYI - CFV Labs uses the Sauer et al. method in their PAN file service.

K. J. Sauer, T. Roessler and C. W. Hansen, "Modeling the Irradiance and Temperature Dependence of Photovoltaic Modules in PVsyst," in IEEE Journal of Photovoltaics, vol. 5, no. 1, pp. 152-158, Jan. 2015.
doi: 10.1109/JPHOTOV.2014.2364133

adriesse:
It is my impression that the PVsyst company uses a human supervised process to produce model parameters, and the same is probably true for the labs that sell PAN files, so it may be hard to create an equivalent function that will work for anyone.  Nevertheless, I think it's good to pursue this goal--especially if you have funding.

> I know that some will fit one model (e.g. CEC), calculate IV curves, and then fit a second model to the calculated IV curves, as a way to get parameters for the second model when data is lacking. This chaining is questionable, in my opinion. The result is a model of a model, with rather weak validation since the data are lacking from the start.

I wouldn't necessarily presume that this doesn't work well.  The weakest link in the chain might actually be the first link rather than the second.  But someone would have to try it to find out.

adriesse:
> I have heard from users that they prefer pvsyst over other models

Any indications why?

cwhanse:
@mikofski getting closer.

Pvlib for MATLAB has three functions that estimate parameters for the pvsyst, CEC and De Soto single diode models, all using a common core for IV curve fitting which I've factored into `pvlib.ivtools.sde._fit_sandia_cocontent`

Design choice for pvlib-python:

-  one function `fit_sdm_sandia(model=`pvsyst`, `CEC`, `desoto`) with a bunch of if blocks or
-  three functions with common helpers `fit_pvsyst_sandia`, `fit_CEC_sandia`, `fit_desoto_sandia`.

I'm leaning towards `fit_model_sandia`.

I have a request from @toddkarin for the De Soto model fitting code.

mikofski:
>three functions with common helpers fit_pvsyst_sandia, fit_CEC_sandia, fit_desoto_sandia.
I'm leaning towards fit_pvsyst_sandia etc.

me too.

Thank you so much for all your work on this!

cwhanse:
> Thank you so much for all your work on this!

Had an excellent PR to build on, thanks to @mattguttenberg and @mikofski 

@wholmgren keep the LOGGER functionality?  See [here](https://github.com/pvlib/pvlib-python/blob/09127c60838af715e6a5470d3b02c7c8103e1f1b/pvlib/ivtools/sdm.py#L420).

mikofski:
My preference would be to remove logging, unless it adds some useful info about the solver state that the user really wants to know. In which case switch the mode to `INFO` and maybe add a verbosity kwarg.

I only put logging in to help with debugging, so now that everything's working, I'm okay with nixing it.

wholmgren:
> My preference would be to remove logging

Agree.

cwhanse:
This monster is ready for review.

NREL's PySAM package was updated 2 days ago to 2.0 (and to 2.0.1 and 2.0.2 since) and `ivtools.sdm.fit_cec_sam` does not work with the 2.0.X versions. Pinned to 1.2.1 until I can debug the failure.

kanderso-nrel:
Also, many of the convergence failures and parameter checks are untested: https://codecov.io/gh/pvlib/pvlib-python/compare/3183cc9149b7a7d8d42fa707295160a9d98873f2...5bd14c7c41f991ba59694bb464cd273c5a611533/diff

pvlib/ivtools/sde.py
```diff
@@ -0,0 +1,475 @@
...
+    3. Fit the linear portion of the IV curve with a line.
+
+    .. math::
+
+        I &\approx \frac{I_{L}}{1 + G_{p} R_{s}} - \frac{G_{p} V}{1 + G_{p}
+        R_{s}} \\
+        &= \beta_{0} + \beta_{1} V
+
+    4. The exponential portion of the IV curve is defined by
+       :math:`\beta_{0} + \beta_{1} \times V - I > ilim \times i_sc`.
```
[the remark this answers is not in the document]
cwhanse: The sign on beta_1 is incorrect in Eq. 5 and 7 in the paper, thanks for pointing that out. I should amend that paper.

pvlib/ivtools/sdm.py
```diff
@@ -0,0 +1,1255 @@
...
+
+    eps1: float, default 1e-3
+        Tolerance for the IV curve fitting. The parameter updating stops when
+        absolute values of the percent change in mean, max and standard
+        deviation of Imp, Vmp and Pmp between iterations are all less than
+        eps1, or when the number of iterations exceeds maxiter.
+
+    Returns
+    -------
+    OrderedDict
```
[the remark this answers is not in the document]
cwhanse: no reason, picked up somewhere in the port from Matlab and three PR contributors.

pvlib/ivtools/sdm.py
```diff
@@ -0,0 +1,1255 @@
...
+        tc[j] - float
+            cell temperature [C]
+        i_sc[j] - float
+            short circuit current of IV curve [A]
+        v_oc[j] - float
+            open circuit voltage of IV curve [V]
+        i_mp[j] - float
+            current at max power point of IV curve [A]
+        v_mp[j] - float
+            voltage at max power point of IV curve [V]
```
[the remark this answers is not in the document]
cwhanse: That was confusing to me and I wrote it. Changed to e.g.
```
        i_mp : array
            current at max power point for each IV curve [A]
```

pvlib/ivtools/sdm.py
```diff
@@ -0,0 +1,1255 @@
...
+    else:
+        raise RuntimeError('Parameter estimation failed')
+
+
+def fit_desoto(v_mp, i_mp, v_oc, i_sc, alpha_sc, beta_voc, cells_in_series,
+               EgRef=1.121, dEgdT=-0.0002677, temp_ref=25, irrad_ref=1000,
+               root_kwargs={}):
+    """
+    Calculates the parameters for the De Soto single diode model using the
+    procedure described in [1]_.
```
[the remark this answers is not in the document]
cwhanse: Caught a few omissions, thanks.

cwhanse:
> Also, many of the convergence failures and parameter checks are untested: https://codecov.io/gh/pvlib/pvlib-python/compare/3183cc9149b7a7d8d42fa707295160a9d98873f2...5bd14c7c41f991ba59694bb464cd273c5a611533/diff

It's actually hard to compose IV curve data that incur some of those failures. I don't know if it's worth the time to test those lines, if we feel it is needed I'll hunt through some measured data and see if I can find some that fail by each criteria.

adriesse:
Sorry, I started looking through, but it's way too much for me.  I'd really need to try using the code before venturing to make the most minor comment.

cwhanse:
@kanderso-nrel what do you make of the two failed tests? No actual test failures but a mysterious error code from Cmd.exe...

ci/requirements-py35.yml
```diff
@@ -23,6 +23,7 @@ dependencies:
     - scipy
     - shapely  # pvfactors dependency
     - siphon  # conda-forge
+    - statsmodels
```
wholmgren, line 26: You'll also need to add `statsmodels` to the optional dependencies in `setup.py`.

pvlib/ivtools/utility.py
```diff
@@ -0,0 +1,433 @@
...
+import numpy as np
+import pandas as pd
+from collections import OrderedDict
+
+
+# A small number used to decide when a slope is equivalent to zero
+EPS = np.finfo('float').eps**(1/3)
+
+
+constants = OrderedDict()
```
wholmgren, line 16: Why an `OrderedDict` instead of regular `dict`?

pvlib/ivtools/utility.py
```diff
@@ -0,0 +1,433 @@
...
+constants['E0'] = 1000.0
+constants['T0'] = 25.0
+constants['k'] = 1.38066e-23
+constants['q'] = 1.60218e-19
+
+
+def numdiff(x, f):
+    """
+    Compute first and second order derivative using possibly unequally
+    spaced data.
```
wholmgren, line 26: Should compare this implementation to MetPy's

https://github.com/Unidata/MetPy/blob/931c1143379b637a3747b2c3c372568498630155/src/metpy/calc/tools.py#L986

https://github.com/Unidata/MetPy/blob/931c1143379b637a3747b2c3c372568498630155/src/metpy/calc/tools.py#L1076

kanderso-nrel:
> @kanderso-nrel what do you make of the two failed tests? No actual test failures but a mysterious error code from Cmd.exe... I have tried re-running the failed tests on previous commits, sometimes the error clears, but for the py35 windows build it is persistent.

Not 100% sure but it looks like something in the shapely/geos world doesn't like getting garbage collected when the tests are done.  I think it's fine to ignore for this PR and worth digging into further if it continues to be a nuisance.  Could be that we need to change how we install shapely.

pvlib/ivtools/utility.py
```diff
@@ -0,0 +1,433 @@
...
+        * increases in voltage
+        * contains no negative current or voltage values
+        * contains no NaNs
+        * contains no points with duplicate voltage values. Where voltage
+          values are repeated, a single data point is substituted with current
+          equal to the average of current at duplicated voltages.
+    """
+
+    if len(voltage) != len(current):
+        raise ValueError('voltage and current must have the same length')
```
wholmgren, line 170: generally -1 on this kind of check in pvlib. vstack will throw its own informative enough ValueError if they don't match.

pvlib/ivtools/utility.py
```diff
@@ -0,0 +1,433 @@
...
+    .. [2] M. H. Lam, "Monotone and Convex Quadratic Spline Interpolation",
+       Virginia Journal of Science 41(1), Spring 1990
+    .. [3] PVLib MATLAB https://github.com/sandialabs/MATLAB_PV_LIB
+    """
+    # Make sure vectors are 1D arrays
+    x = x.flatten()
+    y = y.flatten()
+
+    n = x.size
+    assert n == y.size
```
wholmgren, line 243: `assert` calls don't belong in code. again -1 on this kind of checking.

pvlib/tests/ivtools/test_sdm.py
```diff
@@ -0,0 +1,377 @@
...
+    modeled['R_s'] = result['R_s']
+    modeled['R_sh_ref'] = result['R_sh_ref']
+    expected = pd.Series(params)
+    assert np.allclose(modeled[params.keys()].values,
+                       expected[params.keys()].values, rtol=5e-2)
+
+
+@requires_scipy
+@requires_statsmodels
+def test_fit_pvsyst_sandia(npts=3000):
```
wholmgren, line 132: can this test be broken up into much smaller chunks?

pvlib/tests/ivtools/test_sdm.py
```diff
@@ -0,0 +1,377 @@
...
+
+        iv_specs.update(
+            cells_in_series=int(Ns), alpha_sc=float(aIsc),
+            beta_voc=float(bVoc), descr=descr)
+
+        strN, strM = f.readline().split(',')
+        N = int(strN)
+        M = int(strM)
+
+        isc = np.empty(N)
```
wholmgren, line 150: generally inefficient to initialize an empty array and assign values element by element but fine to ignore if the test is fast (< 1 s)

wholmgren:
I've seen those errors on a lot of the recent PRs. I'm hoping they magically go away with a future update to something. Python 3.5 is nearing end of life and Python 3.9 is coming soon, so we should probably drop 3.5 soon anyways.

pvlib/ivtools/utility.py
cwhanse: MetPy implements 3-point formulae for n-dimensional data, where the pvlib function uses 5-point formulae only for a function of a single variable. I don't see a way for pvlib to use the MetPy function, because the polynomial expressions for the weight at each point don't extend such that a 5point function can use a 3point function in it's calculation. Way back when I was working out this method there was a reason to use 5-point formulation, probably for some level of smoothing of the input data, but I don't recall exactly.

The MetPy function implement some logic to handle the partial windows at each end of the data, where the pvlib function ducks the issue and returns Nan. That could be an improvement to bring from MetPy to pvlib. I don't think it would substantially improve the downstream function that currently uses the output of this `numdiff` function.

I'm sure the pvlib function coding could be improved - it's very explicit. The MetPy function has sophistication that isn't needed for a function of a single variable.

pvlib/ivtools/utility.py
cwhanse: Hold over from Matlab

pvlib/tests/ivtools/test_sdm.py
cwhanse: it's a long test function but I don't see the advantage of breaking it up. There's only one test with many outputs being checked.

The test produces 1) Pvsyst parameters (9 values), 2) five diode equation values for each IV curve, and 3) a Boolean array which is True if the parameter procedure converged for an IV curve.  Each of these outputs is tested. Also, the estimated Pvsyst parameters are pushed forward through the Pvsyst model and the calculated IV curves are compared with the input IV curves.

Half of the code is extracting variables from the text files containing the input IV curves and the known solution.  We could separate the IO portion into a helper function but it would only be used here.

Long term, if pvlib adopts an explicit convention for IV curve data, there would be code in pvlib.iotools to read/write that data, and the text files used here would conform to that convention.

pvlib/tests/ivtools/test_sdm.py
cwhanse: Agree, the format of the file being read doesn't help (a few header rows, then for each IV curve: one line of 7 floats, a block of numbers for the voltage array, a second block of numbers for the current array, length of voltage and current arrays can be different for different IV curves).

I doubt the test is <1s, but the major time sink is the function being tested `sdm.fit_pvyst_sandia`.

pvlib/ivtools/utility.py
mikofski: Hi @cwhanse, @wholmgren , & others,

Not a blocker to me, but curious if you had considered other numerical differentiation tools like:
* [`scipy.misc.derivative`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.derivative.html#scipy.misc.derivative) - allows you to change the number of points used (from 3 to 5, etc) as long as it's odd - check the [source](https://github.com/scipy/scipy/blob/v1.5.2/scipy/misc/common.py#L75-L145) to see the weights used
* [`statsmodels.tools.numdiff`](https://www.statsmodels.org/stable/tools.html#numerical-differentiation) - also does complex numbers and can return the gradient, Jacobians, and Hessians
* [`numdifftools`](https://numdifftools.readthedocs.io/en/latest/) - latest release was June, 2019, but this is based on the [MATLAB derivest](https://www.mathworks.com/matlabcentral/fileexchange/13490-adaptive-robust-numerical-differentiation) originally authored by legendary Kodak mathematician [John D'errico](https://www.mathworks.com/matlabcentral/profile/authors/869215)
* [algopy](https://pythonhosted.org/algopy/) - I believe there used to be more packages like this, but Algopy seems to be the only one (in Python) that has survived. It requires significant effort, and overloads NumPy, but I believe it calculates derivatives of each operation as it occurs rather than of functional outputs

Anyway, just putting this out there, I don't actually expect you (or MetPy) to change anything, just always curious how or why there are so many tools that seem to all do the same thing, in apparent contradiction to [pep20: the Zen of Python](https://www.python.org/dev/peps/pep-0020/):
>There should be one-- and preferably only one --obvious way to do it.

cwhanse: Thanks @mikofski . I would much prefer to import this capability from numpy or scipy than host it in pvlib. Maybe it's out there in python-land, when we wrote the Matlab code we couldn't find it. Looking at the resources you identified:

- scipy.misc.derivative requires equally spaced x-data, which is not a safe assumption for measured IV curves.
- statsmodels.tools.numdiff uses hardcoded 2 point (forward difference) or 3 point (centered difference) formula, although  it looks like it might accommodate unequal spacing, not clear to me.
- numdifftools looks like the primary use case is automated differentiation, which is related but different than calculating a numerical derivative from data which is the need for the sdm.fit_xxx_sandia functions.

MetPy's numdiff is based on the same reference I used, it could be made more general by adding higher order formulas, then imported here.

pvlib/tests/conftest.py
```diff
@@ -90,6 +90,16 @@ def assert_frame_equal(left, right, **kwargs):
 requires_scipy = pytest.mark.skipif(not has_scipy, reason='requires scipy')
 
 
+try:
+    import statsmodels.api as sm  # noqa: F401
```
wholmgren, line 94:
```suggestion
    import statsmodels  # noqa: F401
```

pvlib/tests/ivtools/test_sdm.py
```diff
@@ -0,0 +1,377 @@
...
+from pvlib import pvsystem
+
+from pvlib.tests.conftest import requires_scipy, requires_pysam
+from pvlib.tests.conftest import requires_statsmodels
+
+BASEDIR = os.path.dirname(__file__)
+TESTDIR = os.path.dirname(BASEDIR)
+PROJDIR = os.path.dirname(TESTDIR)
+DATADIR = os.path.join(PROJDIR, 'data')
+TESTDATA = os.path.join(DATADIR, 'ivtools_numdiff.dat')
```
wholmgren, line 18: prefer pathlib.Path for new code

pvlib/tests/ivtools/test_sdm.py
```diff
@@ -0,0 +1,377 @@
...
+    ivcurves['v'] = v[:npts]
+    ivcurves['i'] = i[:npts]
+
+    pvsyst_specs = dict.fromkeys(spec_list)
+    # order required to match file being read
+    paramlist = [
+        'I_L_ref', 'I_o_ref', 'EgRef', 'R_sh_ref', 'R_sh_0', 'R_sh_exp', 'R_s',
+        'gamma_ref', 'mu_gamma']
+    varlist = ['iph', 'io', 'rs', 'rsh', 'u']
+    pvsyst = OrderedDict(key=(paramlist + varlist))
```
wholmgren, line 190: is ordereddict needed here?

pvlib/tests/ivtools/test_sdm.py
wholmgren: I think a helper function or two would make the test more readable but fine to leave as is.

pvlib/tests/ivtools/test_utility.py
```diff
@@ -0,0 +1,83 @@
...
+import pandas as pd
+import pytest
+from pvlib.ivtools.utility import numdiff, rectify_iv_curve, schumaker_qspline
+
+
+BASEDIR = os.path.dirname(__file__)
+TESTDIR = os.path.dirname(BASEDIR)
+PROJDIR = os.path.dirname(TESTDIR)
+DATADIR = os.path.join(PROJDIR, 'data')
+TESTDATA = os.path.join(DATADIR, 'ivtools_numdiff.dat')
```
wholmgren, line 12: pathlib here too

pvlib/data/ivtools_numdiff.dat
```diff
@@ -0,0 +1,52 @@
+4.50204,0,nan,nan
```
wholmgren, line 1: why `.dat` when the format is csv?

pvlib/tests/ivtools/PVsyst_demo_model.txt
```diff
@@ -0,0 +1,3588 @@
+36,0.0054,-0.0774,Mitsubishi PV-UE125MF5N cSi
```
wholmgren, line 1: .csv here too?

setup.py
```diff
@@ -44,8 +44,9 @@
 TESTS_REQUIRE = ['nose', 'pytest', 'pytest-cov', 'pytest-mock',
                  'pytest-timeout', 'pytest-rerunfailures', 'pytest-remotedata']
 EXTRAS_REQUIRE = {
-    'optional': ['ephem', 'cython', 'netcdf4', 'nrel-pysam', 'numba',
-                 'pvfactors', 'scipy', 'siphon', 'tables', 'cftime >= 1.1.1'],
+    'optional': ['cython', 'ephem', 'netcdf4', 'nrel-pysam', 'numba',
+                 'pvfactors', 'scipy', 'siphon', 'statsmodels', 'tables',
```
wholmgren, line 48: should we specify a minimum version of statsmodels? probably only needed if the feature is not available in a release that occurred within the last few years.

pvlib/ivtools/utility.py
wholmgren: Thanks for checking Cliff. Let's go with this implementation. 

Do we really want to make this part of the public API? Same concern for the qspline function.

setup.py
cwhanse: I don't think so, the only capability being used is the Robust Linear Model (RLM) class which I would guess is stable.

pvlib/tests/ivtools/PVsyst_demo_model.txt
cwhanse: And moved both files to pvlib\data

pvlib/tests/ivtools/test_sdm.py
cwhanse: nope

cwhanse:
Odd that I can't reply to a few of the comments.
- I added helper functions for the test that read the data files, maybe sometime we can replace those custom functions with a pvlib function.
- numdiff and schumaker_qspline are private.

wholmgren:
@cwhanse the 3.8 configuration also needs statsmodels https://github.com/pvlib/pvlib-python/blob/master/ci/requirements-py38.yml

cwhanse:
Feels good to finally merge this. Thanks @mikofski for moving this forward, @Peque for the improvements, and especially @mattguttenberg for the port from Matlab - hope you are still watching.

Issue #501, opened 2018-07-06
Plan for dropping Python 2.7


wholmgren:
The Python Software Foundation is dropping Python 2.7 support in 1.5 years (Jan 1, 2020) ([ref](https://devguide.python.org/#status-of-python-branches)). The major scientific python packages (numpy, pandas, matplotlib, scipy, etc.) are dropping Python 2.7 support in less than a year, though this is a little nuanced and depends on the package ([ref](https://python3statement.org/)). 

The question here is should we commit to a plan for dropping Python 3 support, or should we wait until it's unavoidable?

Here's a list of pros and cons that I see for moving pvlib to python 3 only and a date for dropping python 3 support. This is specific to pvlib and python 3, rather than python 3 overall...

Pros:

- We can advertise pvlib on https://python3statement.org/
- Remove a small amount of compatibility code
- keyword only arguments could be useful for some complicated pvlib function signatures
- Matrix multiplication operator (3.5+)
- Better unpacking (useful for kwargs magic) (3.5+)
- We can (maybe) use f-strings in pvlib (3.6+)
- dict retains ordering, so no need to use OrderedDict in pvlib (3.6+)
- Maybe type annotations (3.7+) but I think the scientific python community is going to have to struggle with this for a bit

Cons:

- It's an unnecessary burden for people that are tied to 2.7

Proposed dates:

- 2019-06-01 (@wholmgren)

I think these pros are small but meaningful, especially once we can commit to 3.6+, while the cons are unknown because I don't have a good understanding of how many pvlib users are on 2.7 and what might be holding them back. My understanding is that download statistics from pypi and anaconda are not very reliable for this question, but I will still try to compile some soon.

Again, this is a discussion about pvlib moving to python 3 only -- it is not a discussion about the merits of python 3 overall. For example, while the new Python 3.7 [``breakpoint``](https://docs.python.org/3/library/functions.html#breakpoint) function looks great, it's not going to be used within pvlib so it's irrelevant to the discussion. 

Anyone watching that is tied to 2.7 should speak now, because I think there are a handful of contributors that would like to push us to python 3 only sooner than later!

So, please post a date you'd like to see pvlib commit to Python 3+ only and list any additional pros and cons. I will update the lists as people post.

thunderfish24:
It might be easiest to follow (and refer to) this plan: https://docs.scipy.org/doc/numpy/neps/dropping-python2.7-proposal.html

AndyMender:
[...]

Issue #727, opened 2019-05-15
Add CI test for bare environment installation


wholmgren:
#725 describes how the CI builds let an installation dependency issue slip through to a new release. We should improve the Azure Pipelines config to prevent that from happening.

Might make sense to do this along with removing the `2.7-min` build next month.

Issue #719, opened 2019-05-13
Add os x tests to azure pipelines config


wholmgren:
Might want to use [templates](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops). See [pandas](https://github.com/pandas-dev/pandas/tree/master/ci/azure) for an example.

PR #728, opened 2019-05-17
Drop Python 2 and 3.4

markcampanelli:

 - [x] Closes issue #501

To ease development for the 0.7.0 release, let's drop support for Python 2 and 3.4 sooner rather than later.


markcampanelli:
@wholmgren Mind if we do this sooner rather than later to simplify development for 0.7.0?

We can always branch off a 0.6.4 release for any hotfixes.

wholmgren:
No objection here. You'll need to drop 2.7 from the azure-pipelines.yml too.

wholmgren:
another mention of python 2.7: https://pvlib-python.readthedocs.io/en/stable/contributing.html#code-style

.travis.yml
```diff
@@ -11,20 +11,14 @@ sudo: false # if false, use TravisCI's container based build
...
-      env: CONDA_ENV=py34
     - python: 3.5
       env: CONDA_ENV=py35
     - python: 3.6
       env:
         - CONDA_ENV=py36
         - TASK="coverage"
         - DEPLOY_ENV="true"
-    - python: 3.6
+    - python: 3.7
```
markcampanelli: @wholmgren Do you know if this is a significant problem in the current builds?

wholmgren: It is not a problem -- this only specifies the base travis python version. The conda requirements file specifies the python version that's actually used in the test environment. I figured it was better to use the same version for consistency. But Travis did not yet have a base 3.7 available when it came time to add it to build matrix, so I used 3.6.

markcampanelli: Can you check my update in 6207754?

It's based on https://github.com/travis-ci/travis-ci/issues/9815.

markcampanelli: Also, I missed the memo on this: Is the plan to retire Travis once the Azure pipelines are deemed stable?

wholmgren: > Can you check my update in 6207754?

I don't know what if any nuance there might be on this, but if it works and it's not noticeably slower then it's ok with me. Is there any benefit to making this change?

> Is the plan to retire Travis once the Azure pipelines are deemed stable?

Main purpose of adding azure was that it let us remove appveyor. I want to see the azure pipelines documentation improve before making a plan to retire Travis.

.travis.yml
```diff
@@ -63,14 +54,6 @@ install:
     - echo "install"
     - conda env create --file ci/requirements-$CONDA_ENV.yml
     - source activate test_env # all envs are named test_env in the yml files
-    # needed to make sure that pandas is compiled against the right
-    # version of numpy
-    - if [[ "$CONDA_ENV" == "py27-min" ]]; then
```
markcampanelli: @wholmgren Do we need to define a new minimum environment in Python 3?

markcampanelli:
@wholmgren One more question: There are four different `environment.yml` files remaining in addition to the `setup.py`. It looks like the dependencies have diverged a bit between them (e.g., versions for numpy and pandas). If I can get some "official" minimums for all the required packages, then I can try to bring these files in line and test out some "clean" installs for compatibility.

.travis.yml
wholmgren: Yes, good catch. I'm pretty sure you'll be able to pin the minimum version numbers in the pip section of the file. I'm less sure about the conda section of the file. Either way is fine with me. I don't think we'll need anything special after that.

wholmgren:
> If I can get some "official" minimums for all the required packages, then I can try to bring these files in line and test out some "clean" installs for compatibility.

Official minimums are in the `setup.py` file. The non-min ci/environment files intentionally don't specify versions so that we always pull the latest dependencies. The py27-min + travis.yml configuration is a workaround for some packaging issue that I don't fully recall. (Maybe conda previously iterated through the pip requirements instead specifying them simultaneously.) I think the workaround is no longer necessary and a py35-min with pinned minimum versions should be adequate. If it's not, let's address the min configuration in another issue/pr.

.travis.yml
wholmgren: Well, I was wrong -- I'm not able to easily create a python 3.5 environment with the current numpy/pandas minimums on my mac. I suspect it's possible but I don't think it's worth the effort to struggle against theses old packages. I can easily build and test the environment with pandas 0.17:

```
# requirements-py35-min.yml
name: test_env
channels:
    - defaults
dependencies:
    - python=3.5
    - pip
    - pip:
        - numpy==1.10.1
        - pandas==0.17.0
        - pytz
        - nose
        - pytest
        - pytest-mock
        - pytest-timeout
        - coveralls
```

We should also specify a newer version of conda in the travis config. Latest is fine.

Only catch is that `test_sun_rise_set_transit_spa` claims it needs pandas 0.17, but it apparently needs 0.18.

markcampanelli:
@wholmgren I think I have all the build changes finished. I did have to update a test, which seemed to fail in strange ways in my Python 3.5 testing with the minimum numpy & pandas installations. The fix seems to make sense, but I'm not sure why it doesn't always fail.

My next (and hopefully final) step will be to remove the `__future__` imports. However, can you provide some guidance on if/how I should update [this](https://github.com/pvlib/pvlib-python/blob/master/pvlib/atmosphere.py#L214)? Also, can you think of any other Python 2 vs. 3 computations changes that might bite us?

docs/environment.yml
```diff
@@ -3,19 +3,29 @@ channels:
...
     - defaults
 dependencies:
-    - python=3.6
-    - mock  # needed for local python 2.7 builds
-    - numpy=1.14.2
-    - scipy
-    - pytables
-    - pandas=0.22.0
-    - pytz
+    - coveralls
```
wholmgren: Why do you want to change the docs environment? This has been a pain to test and debug on RTD in the past. I suggest addressing separately if there's a need.

pvlib/__init__.py
```diff
@@ -1,3 +1,10 @@
+# First ensure proper Python version.
```
wholmgren: No need for this.

setup.py
```diff
@@ -40,13 +40,12 @@
...
-                    'six',
-                    ]
-TESTS_REQUIRE = ['pytest', 'pytest-cov', 'pytest-mock', 'nose']
+                    'requests']
+TESTS_REQUIRE = ['nose', 'pytest', 'pytest-cov', 'pytest-mock',
+                 'pytest-timeout']
 EXTRAS_REQUIRE = {
     'optional': ['scipy', 'tables', 'numba', 'siphon', 'netcdf4',
-                 'ephem', 'cython', 'pvfactors'],
+                 'ephem', 'cython', 'pvfactors == 1.0.1'],
```
wholmgren: I'd rather not pin the dependency our setup.py file.

ci/requirements-py35.yml
```diff
@@ -3,21 +3,25 @@ channels:
...
     - conda-forge
 dependencies:
-    - python=3.5
-    - numpy
-    - scipy
-    - pytables
-    - pandas
-    - pytz
+    - coveralls
+    - cython
```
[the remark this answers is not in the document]
wholmgren: Ah, I forgot it's in the setup.py extras section. I can't remember why it's there. I guess it should be here if it's in the setup.py file.

setup.py
```diff
@@ -40,13 +40,12 @@
 INSTALL_REQUIRES = ['numpy >= 1.10.1',
                     'pandas >= 0.16.0',
```
wholmgren: min version change

docs/sphinx/source/installation.rst
```diff
@@ -216,7 +216,7 @@ environment) when you start a new shell or terminal.
 Compatibility
 -------------
 
-pvlib-python is compatible with Python versions 2.7 and 3.4-3.7.
+pvlib-python is compatible with Python 3.5-7.
```
wholmgren: I suggest "3.5 and above".

wholmgren: Also in readme and contributing.

ci/requirements-py35.yml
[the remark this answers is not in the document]
markcampanelli: Yes, I tried to make all the non-min requirements files match the `optional` setup. You'll also notice that I moved everything I could to the conda install. Anaconda recommends minimizing the mixing of conda and pip.

docs/environment.yml
markcampanelli: I wasn't sure what to do here, but I'm fine reverting.

wholmgren: Ok, please do revert this change.

pvlib/__init__.py
markcampanelli: I think it would help people who need to upgrade, but I'm fine removing.

setup.py
markcampanelli: Is this a minimum? It is pinned in the build environments?

setup.py
markcampanelli: Sorry that got reverted in some of my testing.

markcampanelli:
Travis CI test statistics for 507428f:
- CONDA_ENV=py35-min
  - 402 passed, 201 skipped, 2 xfailed, 1 xpassed, 2 warnings in 174.61 seconds- CONDA_ENV=py35
- CONDA_ENV=py35
  - 604 passed, 3 skipped, 3 xfailed, 1 xpassed, 76 warnings in 244.82 seconds
- CONDA_ENV=py36
  - 604 passed, 3 skipped, 2 xfailed, 2 xpassed, 2 warnings in 262.36 seconds
- CONDA_ENV=py37
  - 605 passed, 2 skipped, 2 xfailed, 2 xpassed, 2 warnings in 132.43 seconds

markcampanelli:
@wholmgren Let me know if you have any other comments. Otherwise, this should be GTG if LGTM analysis passes.

pvlib/test/test_modelchain.py
```diff
@@ -1,9 +1,4 @@
 import sys
-try:
-    from unittest.mock import ANY
-except ImportError:
-    # python 2
-    from mock import ANY
```
markcampanelli: Apparently, this is not used.

setup.py
```diff
@@ -71,7 +67,8 @@
 setuptools_kwargs = {
     'zip_safe': False,
     'scripts': [],
-    'include_package_data': True
+    'include_package_data': True,
+    'python_requires': '>=3.5, <4'
```
markcampanelli: @wholmgren I remembered that this is the more "proper" way to call this out. This way, `pip install -e .` won't work if the Python version is wrong.

markcampanelli: There's an even shorter syntax for this: '~=3.5'. See https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

wholmgren: Good find!

cwhanse:
No objections to merge

wholmgren:
Thanks @markcampanelli!

.travis.yml
markcampanelli: I just set up CI using Azure pipelines for Ubuntu 16.04, Windows, and macOS [here](https://github.com/markcampanelli/pvfit-m/blob/master/azure-pipelines.yml). It might be nice to maintain only a single CI in pvlib.

wholmgren: Thanks for the example! #719 tracks this - well, at least the macOS part. Additional work would be required to transition the deploy section of the travis config to azure. #727 also relevant.

PR #730, opened 2019-05-21
Add a new and faster way to calculate the solarposition

meteoDaniel:

A new way to calculate solar position. Let us benchmark this against `nrel_numba`. A first test against the default configuration shows the code runs nearly 100 times faster: 
1000 timesteps:
Default: 0:00:00.023037
spencer_mc: 0:00:00.003093
10000 timesteps:
Default: 0:00:00.178758
spencer_mc: 0:00:00.007183
100000 timesteps:
Default: 0:00:01.400652
spencer_mc: 0:00:00.049189


meteoDaniel:
Some issue of stickler-ci are not related to my changes. 
Please help me for the other tests.

pvlib/test/test_solarposition.py
```diff
@@ -753,7 +753,78 @@ def test_sun_rise_set_transit_geometric(expected_rise_set_spa, golden_mst):
...
+                                     [101.08062211, -11.08062211, 33.87167089,
+                                      0.9670539,
+                                      0.40898412, -2.10671942],
+                                     [99.63573414, -9.63573414, 37.04545803,
+                                      0.96705293,
+                                      0.40898091, -2.10897939],
+                                     [98.0818867, -8.0818867, 40.15181379,
+                                      0.96705197,
+                                      0.40897769, -2.11123924]],
+                                    index=times,
```
cwhanse: `times[;10]` here, I think

pvlib/solarposition.py
```diff
@@ -1445,3 +1457,118 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
...
+                           'elevation': np.degrees(elevation),
+                           'azimuth': np.degrees(azimuth),
+                           'eccentricity': eccentricity,
+                           'declination': declination,
+                           'equation_of_time': eot},
+                          index=times)
+    return result
+
+
+def datetime2julian(times):
```
cwhanse: Does this function do the same thing as `tools.datetime_to_djd`?

meteoDaniel: I guess it is similar but not the same. djd is limited to days.

meteoDaniel:
@cwhanse I hope i solved all issues for linter and lgtm. Lgtm is still running at the moment. 
Can you tell me the difference between apparent and non apparent solar azimuth/zenith/elevation ? 
Thanks in advance.

mikofski:
Hi @meteoDaniel , FYI: I pushed a PR with new changes to [SolarUtils](https://github.com/SunPower/SolarUtils/pull/14) that allows you to specify either the year or a specific sequence of datetimes, it's very fast

```python
In [1]: from solar_utils import solposAM, get_solposAM
   ...: location = [35.56836, -119.2022, -8.0]
   ...: weather = [1015.62055, 40.0]
   ...: from datetime import datetime, timedelta
   ...: times = [(datetime(2017,1,1,0,0,0)+timedelta(hours=h)).timetuple()[:6] for h in range(1000)]

In [2]: %timeit get_solposAM(location, times, weather)
1.82 ms ± 16.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

On my PC that's about twice as fast as the Spencer method per the 1000-timestep benchmark in this PR. SolarUtils is a very thin Python wrapper around the freely available [NREL SOLPOS algorithm](https://www.nrel.gov/grid/solar-resource/solpos.html) source code from Martin Rymes, 1998-2001.

Although runtime scales linearly, while your runtime stays fairly low, probably due to NumPy? Porting SOLPOS to Python would probably have similar results. The source is freely available, so perhaps if you require even more speed that would be an option.
```python
In [6]: times = [(datetime(2017,1,1,0,0,0)+timedelta(hours=h)).timetuple()[:6] for h in range(10000)]

In [7]: %timeit get_solposAM(location, times, weather)
18.8 ms ± 117 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
One of the stated purpose of pvlib python is to catalog all useful solar algorithms, so the Spencer method is still very useful. Thanks for contributing this. Let me know if there's anyway I can help. But I thought since you were concerned about speed, you might also be interested in SOLPOS from NREL either using SolarUtils or by porting it from the source to Python.

cwhanse:
> Can you tell me the difference between apparent and non apparent solar azimuth/zenith/elevation ?
> Thanks in advance.

Generally, the 'apparent' values account for atmospheric refraction, the non-apparent values are geometric.

docs/sphinx/source/whatsnew/v0.7.1.rst
```diff
@@ -0,0 +1,11 @@
+.. _whatsnew_0701:
+
+v0.7.1 (MONTH DAY, YEAR)
+---------------------
+
+- Add a faster way to calculate the solar position (`spencer_mc`)
```
wholmgren: please move this to the 0.7.0 whats new file https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/whatsnew/v0.7.0.rst

pvlib/solarposition.py
```diff
@@ -1445,3 +1454,96 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
     sunset = _local_times_from_hours_since_midnight(times, sunset_hour)
     transit = _local_times_from_hours_since_midnight(times, transit_hour)
     return sunrise, sunset, transit
+
+
+def spencer_mc(times, latitude, longitude):
```
wholmgren: Is `mc` for meteocontrol? If so, please remove.

pvlib/solarposition.py
```diff
@@ -1445,3 +1454,96 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
...
+    """
+    Calculate the solar position using a python implementation of the
+    Spencer (1972) formulation provided by meteocontrol
+
+    Parameters
+    ----------
+    times : :class:`pandas.DatetimeIndex`
```
wholmgren: no need for the sphinx `:class:`. the documentation build will add it

pvlib/solarposition.py
```diff
@@ -1445,3 +1454,96 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
     sunset = _local_times_from_hours_since_midnight(times, sunset_hour)
     transit = _local_times_from_hours_since_midnight(times, transit_hour)
     return sunrise, sunset, transit
+
+
+def spencer_mc(times, latitude, longitude):
+    """
+    Calculate the solar position using a python implementation of the
+    Spencer (1972) formulation provided by meteocontrol
```
wholmgren: My preference is a code comment saying who added the code and what their affiliation is. I don't think this belongs in rendered documentation.

pvlib/solarposition.py
```diff
@@ -1445,3 +1454,96 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
...
+
+    # Eccentricity: correction factor of the earth's orbit.
+    eccentricity = 1.00011 + 0.034221 * cos_gamma[0]
+    eccentricity += 0.001280 * sin_gamma[0]
+    eccentricity += 0.000719 * cos_gamma[1]
+    eccentricity += 0.000077 * sin_gamma[1]
+
+    declination = np.array(declination_spencer71(times.dayofyear))
+
+    # Equation of time (difference between standard time and solar time).
```
wholmgren: can this be replaced with `equation_of_time_spencer71`?

pvlib/solarposition.py
```diff
@@ -1445,3 +1454,96 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
...
+    lat = np.radians(latitude)
+
+    # Compute fractional year (gamma) in radians
+    gamma = 2 * np.pi * (julians_2000 % JULIAN_YEARS) / JULIAN_YEARS
+    cos_gamma = np.cos(gamma), np.cos(gamma * 2), np.cos(gamma * 3)
+    sin_gamma = np.sin(gamma), np.sin(gamma * 2), np.sin(gamma * 3)
+    day_time = (julians_2000 % 1) * 24
+
+    # Eccentricity: correction factor of the earth's orbit.
+    eccentricity = 1.00011 + 0.034221 * cos_gamma[0]
```
wholmgren: It's usually better to make the assignment in a single expression broken across multiple lines and encapsulated with parenthesis. My apologies if stickler steered you wrongly in this case.

pvlib/solarposition.py
```diff
@@ -1445,3 +1454,96 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
...
+    azimuth = np.arccos(y)
+
+    # Convert azimuth angle from 0-pi to 0-2pi.
+    tlt_filter = 0 <= tlt
+    azimuth[tlt_filter] = 2 * np.pi - azimuth[tlt_filter]
+
+    result = pd.DataFrame({'zenith': np.degrees(zenith),
+                           'elevation': np.degrees(elevation),
+                           'azimuth': np.degrees(azimuth),
+                           'eccentricity': eccentricity,
```
wholmgren: what's the case for providing `eccentricity` to the end user? seems like too much information to me.

pvlib/tools.py
```diff
@@ -425,3 +429,25 @@ def _golden_sect_DataFrame(params, VL, VH, func):
             raise Exception("EXCEPTION:iterations exceeded maximum (50)")
 
     return func(df, 'V1'), df['V1']
+
+
+def datetime_to_julian(times):
```
wholmgren: I don't like the duplication of `datetime_to_djd`. Let's make the old function work with both `datetime` and `DatetimeIndex`. I suspect converting a `datetime` input to a `Timestamp` will do the job.

pvlib/test/test_tools.py
```diff
@@ -1,6 +1,10 @@
+import numpy as np
+import pandas as pd
 import pytest
 
+import pvlib.tools
 from pvlib import tools
+from pvlib.test.test_solarposition import times
```
wholmgren: Please create an appropriate new input within the test function below. Dependencies across test modules are difficult to maintain.

pvlib/test/test_tools.py
```diff
@@ -1,6 +1,10 @@
+import numpy as np
+import pandas as pd
 import pytest
 
+import pvlib.tools
```
wholmgren: `tools` is already imported on the next line.

pvlib/test/test_solarposition.py
```diff
@@ -754,8 +754,125 @@ def test_sun_rise_set_transit_geometric(expected_rise_set_spa, golden_mst):
                        atol=np.abs(expected_transit_error).max())
 
 
-# put numba tests at end of file to minimize reloading
+def test_spencer_mc():
```
wholmgren: Please follow this test pattern so that we can be sure the output is correct to within some tolerance: https://github.com/pvlib/pvlib-python/blob/d68617c03a2bd6cc95100d91d8de5f1069ae13cc/pvlib/test/test_solarposition.py#L420-L455

pvlib/solarposition.py
```diff
@@ -1445,3 +1454,96 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
     sunset = _local_times_from_hours_since_midnight(times, sunset_hour)
     transit = _local_times_from_hours_since_midnight(times, transit_hour)
     return sunrise, sunset, transit
+
+
+def spencer_mc(times, latitude, longitude):
+    """
+    Calculate the solar position using a python implementation of the
+    Spencer (1972) formulation provided by meteocontrol
+
```
wholmgren: I recommend that we add a comment on the relative accuracy and speed of the function. Perhaps it belongs in a `Notes` section below.

pvlib/solarposition.py
```diff
@@ -1445,3 +1454,96 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
...
+    eccentricity += 0.000077 * sin_gamma[1]
+
+    declination = np.array(declination_spencer71(times.dayofyear))
+
+    # Equation of time (difference between standard time and solar time).
+    eot = 0.000075 + 0.001868 * cos_gamma[0]
+    eot -= 0.032077 * sin_gamma[0]
+    eot -= 0.014615 * cos_gamma[1]
+    eot -= 0.040849 * sin_gamma[1]
+    eot *= 229.18
```
mikofski: this entire section is a repeat of [`equation_of_time_spencer71`](https://github.com/pvlib/pvlib-python/blob/b20ca5319fd2ea01b0261401d112b70f49146604/pvlib/solarposition.py#L1072-L1079):
```python
    day_angle = _calculate_simple_day_angle(dayofyear)
    # convert from radians to minutes per day = 24[h/day] * 60[min/h] / 2 / pi
    eot = (1440.0 / 2 / np.pi) * (
        0.0000075 +
        0.001868 * np.cos(day_angle) - 0.032077 * np.sin(day_angle) -
        0.014615 * np.cos(2.0 * day_angle) - 0.040849 * np.sin(2.0 * day_angle)
    )
    return eot
```

I would prefer to be [DRY vs WET](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) -- can you pleases consider reusing the existing code functions?

EG: `eot = equation_of_time_spencer71(dayofyear)`

where I suspect that you can use the julian dates you've calculated in place of day of year, or adjust them slightly

see [`_calculate_simple_day_angle`](https://github.com/pvlib/pvlib-python/blob/b20ca5319fd2ea01b0261401d112b70f49146604/pvlib/solarposition.py#L1000-L1014)

pvlib/solarposition.py
```diff
@@ -1445,3 +1454,96 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
...
+
+    # True local time
+    tlt = (day_time + longitude / 15 + eot / 60) % 24 - 12
```
mikofski: I suspect this is similar as [`hour_angle`](https://github.com/pvlib/pvlib-python/blob/b20ca5319fd2ea01b0261401d112b70f49146604/pvlib/solarposition.py#L1324) - perhaps there is a way to reuse the existing code, or modify it to suit your needs so we don't have duplicate code?

mikofski: looking in more detail they are the nearly the same:
    tlt = hour_angle / 15.

but IMO this version here, looks a little easier to read, with the use of `mod()`, so might be nice to replace the existing `hour_angle` with this, assuming there is better or same performance

pvlib/solarposition.py
meteoDaniel: No not quite the same but it is similar. Actually datetime_to_djd is referencing to the dublin standard and the function is not able to work with arrays. I will have a look If it is possble to adapt my function that it is working with dublin julian days so that we can use my implementation.

pvlib/solarposition.py
meteoDaniel: But I sticler will fail on this lines. I guess you know how to solve it. Thanks fpr helping

pvlib/test/test_tools.py
meteoDaniel: Ähh right! ;D

pvlib/test/test_solarposition.py
meteoDaniel: Good Idea. I will bring it in the next commit

pvlib/solarposition.py
meteoDaniel: Right. New implementatino uses the [equation_of_time_spencer71](https://github.com/pvlib/pvlib-python/blob/b20ca5319fd2ea01b0261401d112b70f49146604/pvlib/solarposition.py#L1072-L1079) function. 
Acually I am not really familiar with the Julian Days and Day Of Year usage. I will check it out if there are any ways to use existend code.

pvlib/solarposition.py
meteoDaniel: I tried `hour_angle` and it was not working in this code.

mikofski:
@meteoDaniel just [add `W504,W504` to .stickler.yml`](https://github.com/mikofski/pvlib-python/blob/7e97fcbc1f8db0582dbb4f74428aa25fa9c79aaf/.stickler.yml#L5). see pep8.org and PyQA. it was a mistake for stickler to raise the warnings

pvlib/solarposition.py
mikofski: But do you agree they are the same equation, that only differ by a factor of 1/15? we have noticed that different formulations have different "offsets" so perhaps adjusting the input may help? Or perhaps leave it,  raise an issue to collapse this later, and maybe add a `#TODO` in the code.  If @cwhanse and @wholmgren are okay then so am I

pvlib/solarposition.py
adriesse: Some subtle form of recognition seems fine with me.  But if the method is identical to spencer's paper, then the suffix '_mc' might be too much.

cwhanse:
@meteoDaniel please merge master into this PR, should take care of the test failures for `get_psm3`

adriesse:
I believe these calculations are identical to the ones in the "analytical" aka "geometric" solar position functions already in pvlib. (#583)

mikofski:
OMG! @adriesse is correct! I guess I didn't review this very well. Might still be worth it, but would be better to use the existing "analytical/geometric" functions or make them better if there's room for improvement.

# azimuth
https://github.com/pvlib/pvlib-python/blob/03bb14384d896b45b485ea88d70521f878d1e23a/pvlib/solarposition.py#L1517-L1524

is the same as 

https://github.com/pvlib/pvlib-python/blob/e3635c2faa4ef990cbbae6cf4432d9773d00f31f/pvlib/solarposition.py#L1237-L1261

# zenith
https://github.com/pvlib/pvlib-python/blob/03bb14384d896b45b485ea88d70521f878d1e23a/pvlib/solarposition.py#L1490-L1511

is the same as

https://github.com/pvlib/pvlib-python/blob/master/pvlib/solarposition.py#L1309-L1312

pvlib/solarposition.py
```diff
@@ -1445,3 +1453,80 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
...
+            np.sin(declination) * np.sin(latitude_radians) +
+            np.cos(declination) * np.cos(latitude_radians) * np.cos(ha)
+    )
+
+    # Compute the sun's elevation and zenith angle.
+    elevation = np.arcsin(sin_sun_elevation)
+    zenith = np.pi / 2 - elevation
+
+    # Compute the sun's azimuth angle.
+    y = -(np.sin(latitude_radians) * np.sin(elevation) - np.sin(declination)) \
```
mikofski: @meteoDaniel can you please replace these lines
https://github.com/pvlib/pvlib-python/blob/03bb14384d896b45b485ea88d70521f878d1e23a/pvlib/solarposition.py#L1517-L1524
with a call to `solar_azimuth_analytical(latitude, hourangle, declination, zenith)` ?

pvlib/solarposition.py
```diff
@@ -1445,3 +1453,80 @@ def sun_rise_set_transit_geometric(times, latitude, longitude, declination,
...
+
+    References
+    ----------
+    [1] Spencer (1972) can be found in
+    Sen, Zekai. Solar energy fundamentals and modeling techniques:
+     atmosphere, environment, climate change and renewable energy.
+     Springer Science & Business Media, 2008.
+    """
+
+    julians, julian_2000 = datetime_to_julian(times)
```
mikofski: @meteoDaniel can you please replace all of these lines
https://github.com/pvlib/pvlib-python/blob/03bb14384d896b45b485ea88d70521f878d1e23a/pvlib/solarposition.py#L1490-L1511
with a call to `solar_zenith_analytical(latitude, hourangle, declination)` ?

adriesse:
Maybe the title could be changed to something like: "Convenience function to calculate the complete solar position using the analytical/geometric methods."

Not sure whether the name Spencer should be used.

adriesse:
By the way, refraction reduces the apparent zenith angle at the horizon by about 0.5 degrees (not 2 degrees as noted above).

wholmgren:
Let's start fresh in a new PR if there's anything we want to continue here.

PR #744, opened 2019-07-02
Add a note to installation docs about dependencies

odow:

 - [x] <s>Closes issue #xxxx</s>
 - [x] <s>Adds description and name entries in the appropriate `docs/sphinx/source/whatsnew` file for all changes.</s>

Add extra documentation to installation as discussed in this Google group post: https://groups.google.com/forum/#!topic/pvlib-python/QtvHELYmrww

Here is a screenshot of the docs as compiled locally:
image

### DIFF
--- a/docs/sphinx/source/installation.rst
+++ b/docs/sphinx/source/installation.rst
@@ -80,12 +80,12 @@ non-editable way, use one of the following commands to install pvlib-python::
 .. note::
     
     By default, pvlib will not install some infrequently used dependencies. 
-    If you run into an error such as 
+    If you run into an error such as  
     `ModuleNotFoundError: No module named 'netCDF4'`
-    you can either add these manually with
+    you can either add these manually with:
     `pip install netCDF4`
-    or you can install pvlib with all optional dependencies
-    `pip install pvlib[optional]`
+    or you can install pvlib and all optional dependencies with:
+    `pip install pvlib[optional]`.
 
 If your system complains that you don't have access privileges or asks
 for a password then you're probably trying to install pvlib into your
--- a/docs/sphinx/source/installation.rst
+++ b/docs/sphinx/source/installation.rst
@@ -77,6 +77,16 @@ non-editable way, use one of the following commands to install pvlib-python::
     # another option if you know what you are doing
     pip install pvlib[optional]
 
+.. note::
+    
+    By default, pvlib will not install some infrequently used dependencies. 
+    If you run into an error such as 
+    `ModuleNotFoundError: No module named 'netCDF4'`
+    you can either add these manually with
+    `pip install netCDF4`
+    or you can install pvlib with all optional dependencies
+    `pip install pvlib[optional]`
+
 If your system complains that you don't have access privileges or asks
 for a password then you're probably trying to install pvlib into your
 system's Python distribution. This is usually a bad idea and you should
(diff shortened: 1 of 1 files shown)


docs/sphinx/source/installation.rst
```diff
@@ -77,6 +77,16 @@ non-editable way, use one of the following commands to install pvlib-python::
     # another option if you know what you are doing
     pip install pvlib[optional]
 
+.. note::
+    
+    By default, pvlib will not install some infrequently used dependencies. 
+    If you run into an error such as  
+    `ModuleNotFoundError: No module named 'netCDF4'`
+    you can either add these manually with:
+    `pip install netCDF4`
```
mikofski: Thanks for this PR. I'm concerned that users may also be using conda instead of or in addition to pip. What are the repercussions for a conda user if they read this and run pip install netCDF4, or put another way, should they take a different action?

- There is an older version of [netcdf4 on anaconda default channel](https://anaconda.org/anaconda/netcdf4), which still meets pvlib minimum requirments
- **WARNING** there is also a [conda-forge version of the current netcdf4](https://anaconda.org/conda-forge/netcdf4) however, the [netcdf4 conda-forge feedstock](../../../conda-forge/netcdf4-feedstock/blob/master/recipe/meta.yaml) lists numpy as a dependent, so mixing with the default anaconda channel may corrupt your conda env, therefore probably better to use an env with conda-forge as default
- the [netCDF4 `setup.py`](../../../Unidata/netcdf4-python/blob/master/setup.py) lists `numpy` and `cftime` as dependencies, so using `pip install netcdf4` will install `cftime` from PyPI not anaconda. I am assuming that the wheels at the Cheeseshop already include the hdf5 and netcdf4 libraries, which are also available from the anaconda main channel.

So my recommendation would be to treat LIKE ALIKE.
- if PyPI then use `pip`
- if anaconda, then use `conda`
- if conda-forge, then use `conda -c conda-forge`

Where this breaks down is for a few packages like siphon which AFAIK doesn't have a good version on the default anaconda channel. [Search anaconda for siphon](https://anaconda.org/search?q=siphon) and you find conda-forge, Unidata, and a few others, each with different versions. Siphon [`setup.py` lists `numpy`, `protobuf`, `beautifulsoup4`, `requests`, and `pandas` as dependencies](../../../Unidata/siphon/blob/master/setup.py#L17), which are all available from the default anaconda channel or conda-forge. If they are installed first using conda, then siphon can be installed from PyPI using pip.

odow: So, I didn't read the comments in the installation instructions carefully enough, particularly the `# best option if using pvlib.forecast module` line added in https://github.com/pvlib/pvlib-python/pull/533. This is my fault, but it would be nice to have a big orange box rather than an (obvious) line buried in a code comment.

Would you accept a change to
```
By default, pvlib will not install some infrequently used dependencies. 
If you run into an error such as  
`ModuleNotFoundError: No module named 'netCDF4'`
you can either install pvlib with all optional dependencies using
`pip install pvlib[optional]`, or you can install pvlib from conda-forge
`conda create -n pvlib -c conda-forge pvlib-python; conda activate pvlib`.
```

wholmgren: I think a big orange box is appropriate for this.

odow: I've updated the wording. Let me know if you want any changes.

mikofski:
Sure, lgtm

cwhanse:
@odow please merge master, with #748 should fix the py35-min test failure.

PR #748, opened 2019-07-09
use pip for py35-min numpy/pandas

wholmgren:
The py35-min builds are failing because the pinned numpy/pandas builds are no longer available on the `defaults` conda channel. I think it's because Anaconda recently cleaned up their package repositories and removed some old package combinations, including the numpy/pandas builds in that test environment. See [here](https://www.anaconda.com/why-we-removed-the-free-channel-in-conda-4-7/) for more info on their cleanup. 

This PR gets the builds working again by switching the numpy/pandas installations from conda to pip.

### DIFF
--- a/ci/requirements-py35-min.yml
+++ b/ci/requirements-py35-min.yml
@@ -13,5 +13,5 @@ dependencies:
     - pytz
     - requests
     - pip:
-        - numpy==1.10.1
-        - pandas==0.18.0
+        - numpy==1.10.4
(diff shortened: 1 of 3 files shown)


wholmgren:
Forgot to add that the minimum numpy and pandas versions are slightly increased. There was a weird installation incompatibility on linux with numpy 1.10.0 and pandas 0.18.0. It went away when switching to the highest bug patch versions of each of those releases series, so that's my recommendation.

PR #753, opened 2019-07-12
Put SAM product renaming code in a separate function, simplify, add warning.

adriesse:
- [ ] Closes issue #xxxx

I needed this function for myself so I thought I might as well submit it.

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1663,11 +1663,11 @@ def _normalize_sam_product_names(names):
 
     n_duplicates = names.duplicated().sum()
     if n_duplicates > 0:
-        warnings.warn ('Original names contain %d duplicate(s).' % n_duplicates)
+        warnings.warn('Original names contain %d duplicate(s).' % n_duplicates)
 
     n_duplicates = norm_names.duplicated().sum()
     if n_duplicates > 0:
-        warnings.warn ('Normalized names contain %d duplicate(s).' % n_duplicates)
+        warnings.warn('Normalized names contain %d duplicate(s).' % n_duplicates)
 
     return norm_names.values
 
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1647,26 +1647,39 @@ def retrieve_sam(name=None, path=None):
     return _parse_raw_sam_df(csvdata)
 
 
+def _normalize_sam_product_names(names):
+    '''
+    Replace special characters within the product names to make them more
+    suitable fur use as Dataframe column names.
+    '''
+    import warnings
+
+    BAD_CHARS  = ' -.()[]:+/",'
+    GOOD_CHARS = '____________'
+
+    mapping = str.maketrans(BAD_CHARS, GOOD_CHARS)
+    names = pd.Series(data=names)
+    norm_names = names.str.translate(mapping)
+
+    n_duplicates = names.duplicated().sum()
+    if n_duplicates > 0:
+        warnings.warn ('Original names contain %d duplicate(s).' % n_duplicates)
+
+    n_duplicates = norm_names.duplicated().sum()
+    if n_duplicates > 0:
+        warnings.warn ('Normalized names contain %d duplicate(s).' % n_duplicates)
+
+    return norm_names.values
+
+
 def _parse_raw_sam_df(csvdata):
+
     df = pd.read_csv(csvdata, index_col=0, skiprows=[1, 2])
-    colnames = df.columns.values.tolist()
-    parsedcolnames = []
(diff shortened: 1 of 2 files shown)


adriesse:
Not too many comments coming in, so maybe I need to send out alerts?  @wholmgren , @markcampanelli , @mikofski , @jsstein This work is not done, but I'm hoping to get early comments to keep me on the straight and narrow path.

pvlib/pvsystem.py
```diff
@@ -1647,26 +1647,39 @@ def retrieve_sam(name=None, path=None):
...
+    mapping = str.maketrans(BAD_CHARS, GOOD_CHARS)
+    names = pd.Series(data=names)
+    norm_names = names.str.translate(mapping)
+
+    n_duplicates = names.duplicated().sum()
+    if n_duplicates > 0:
+        warnings.warn ('Original names contain %d duplicate(s).' % n_duplicates)
+
+    n_duplicates = norm_names.duplicated().sum()
+    if n_duplicates > 0:
```
wholmgren: are either of these warnings triggered in the current file? I would have thought the DataFrame construction would fail if so.

pvlib/pvsystem.py
```diff
@@ -1647,26 +1647,39 @@ def retrieve_sam(name=None, path=None):
...
+    '''
+    Replace special characters within the product names to make them more
+    suitable fur use as Dataframe column names.
+    '''
+    import warnings
+
+    BAD_CHARS  = ' -.()[]:+/",'
+    GOOD_CHARS = '____________'
+
+    mapping = str.maketrans(BAD_CHARS, GOOD_CHARS)
```
wholmgren: nice! this is new to me

pvlib/pvsystem.py
adriesse: I got 6 duplicates after normalization in CECmod, and 12 before normalization in ADRInverter.  It is possible to have duplicate values in DataFrame columns or index.

adriesse: I can add a test if the function gets the thumbs up.

adriesse: Do I see correctly that there is no current test for `retrieve_sam()`?

cwhanse: @janinefreeman FYI. I recognize that both SAM and pvlib are victims of what appears in the CEC table. Currently, pvlib imports the module parameter file from SAM. Perhaps there's a way we can help reduce some of these formatting requirements upstream in SAM, and avoid the duplicates.

pvlib/test/test_pvsystem.py
```diff
@@ -147,6 +147,28 @@ def test_PVSystem_physicaliam(mocker):
     pvsystem.physicaliam.assert_called_once_with(thetas, **module_parameters)
     assert iam < 1.
 
+def test__normalize_sam_product_names():
```
cwhanse: two underscores in name, if that matters.

wholmgren:
We could also relax the restriction on disallowed characters. I think this
was in the earliest versions of pvlib python to support tab completion in
ipython. But that’s only useful to some workflows. And there’s been a lot
of ipython development since then so maybe no longer necessary.

pvlib/test/test_pvsystem.py
adriesse: One from the `test_` prefix and one from the function name.

adriesse:
I've never use the tab completion and find the database more convenient in its non-transposed form.

pvlib/pvsystem.py
janinefreeman: @cwhanse Thanks for the heads up. We get frequent requests to fix info coming from the CEC, but we are trying very hard to avoid error-checking the database. But if this method could be ported into SAM, that could be a possibility. Yet another scenario where we'd all benefit from the oft-pitched crowd-sourced component database...

cwhanse: @janinefreeman I think this is a case of creating code-friendly row and column labels. If you can point me to the part of SAM code that reads the CEC's spreadsheet, or that writes the SAM module database, I'm happy to take a look.

pvlib/pvsystem.py
```diff
@@ -1647,26 +1647,41 @@ def retrieve_sam(name=None, path=None):
     return _parse_raw_sam_df(csvdata)
 
 
+def _normalize_sam_product_names(names):
+    '''
+    Replace special characters within the product names to make them more
+    suitable fur use as Dataframe column names.
```
```suggestion
    suitable for use as Dataframe column names.
```

adriesse:
No problem @wholmgren, it wasn't a high priority for me either.

wholmgren:
thanks @adriesse!

PR #264, opened 2016-11-15
Linke turbidity factor fixes

mikofski:
fixes #263 and fixes #262 

* shift `outputmin` to zero, since (1) zero based indexing, and (2) documentation states that indices of LT data are 0.0833[arcdeg] apart
* shift from edge to center
* raise `IndexError` for out of bound lat/lon
* add `test_linke_turbidity_corners()` to test corners are not out of bounds, PASSES, but older tests fail since increment is different.

PR #755, opened 2019-07-17
fix rounding issue in linke turbidity lookup function

### LINKED ISSUE #754: Rounding issue in clearsky.py _linearly_scale method
**Describe the bug**
In the _linearly_scale method, there are two logic statements (lines 316 and 321) that first round values to one decimal place, then compare if <= .5. Since they're being rounded first, values like .54 would become .5, then the check <= .5 would actually be true. Also, from a performance standpoint, there is no reason to round first, we could instead just do the comparison immediately. If you actually want to check if values are < .55, then write that explicitly and modify the comment on line 314. Or keep <= .5 and the leave the comment as is, but either way, there's no need to round and compare to a float. I can make a pull request for either approach, please just let me know what's preferred.

**To Reproduce**
```
>>> import numpy as np
>>> arr = np.array([.49, .50, .54, .55, .56])
>>> np.around(arr, 1) <= .5  # current behavior
array([ True,  True,  True, False, False])
>>> arr <= .5  # intended behavior and also much faster
array([ True,  True, False, False, False])
>>> arr < .55  # equivalent to current behavior but much faster
array([ True,  True,  True, False, False])

```
**Expected behavior**
I would have expected values like .54 to raise an error in this if block, but they don't.

**Screenshots**

**Versions:**
 - ``pvlib.__version__``: 0.6.3
 - python: 3.5.2

**Additional context**
Removing the unnecessary rounding operation reduces the runtime of these logic checks by about 60%.

alexandermorgan:

 - [X] Closes #754 
 - [X] Adds description and name entries in the appropriate `docs/sphinx/source/whatsnew` file for all changes.


wholmgren:
@alexandermorgan I'm having a hard time understanding this change because the function itself is written using terms that I think are too abstract for the application. I had to look at the code that calls the function to understand what the function itself did. We either need a few comments explaining the comparison or we need to make the function more concrete. Obviously not your fault, but I don't know how to review the change if I don't understand where we're starting from! 

My understanding is the function signature might be more clear like:

```python
def _equirectangular_lookup(coordinate, min_coordinate, max_coordinate, min_index, max_index):
    """Determine array index corresponding to desired latitude or longitude for data on an equirectangular map projection. Min/max coordinates are mapped onto min/max indices."""
```

I'm not totally confident that I've accurately described the situation, so please double check.

alexandermorgan:
@wholmgren I agree with you. It's a little tricky to unpack what this function does at first, but in the end it's not all that complicated. Your function name (_equirectangular_lookup) could work. Another alternative would be something like _degrees_to_index. Also, the inputmatrix and outputmatrix variables should be renamed to show that they're single floats, not iterables or matricies. And the index number obviously has to be an int, so that rounding should happen inside of the _linearly_scale model, not in lines 198-203. So in the end, "outputmatrix" should be called "outputindex" and it should be an int. And a clearer name for "inputmatrix" would be "lat_or_long_degrees" or maybe just "degrees" and explain that it's latitude or longitude in the method documentation.
I'll try to make this as clear as possible and I'll ping you when it's ready for a second review. Thanks.

alexandermorgan:
@wholmgren this turned out to be a bigger change than I intended. I also removed a try/except/finally block because the exception being checked for is already checked for earlier in _degrees_to_index. These changes now include a fair amount of renamed variables, some slight refactoring, and a new test.

docs/sphinx/source/whatsnew/v0.7.0.rst
```diff
@@ -14,6 +14,8 @@ Bug fixes
 ~~~~~~~~~
 * Fix handling of keyword arguments in `forecasts.get_processed_data`.
   (:issue:`745`)
+* Fix rounding issue in `clearsky._linearly_scale`.
```
wholmgren: Can you add a bit more detail here? Most users won't be aware of this private function.

pvlib/clearsky.py
```diff
@@ -301,28 +291,63 @@ def _calendar_month_middles(year):
...
+    """Transform input degrees to an output index integer. The Linke
+    turbidity lookup tables have three dimensions, latitude, longitude, and
+    month. Specify a degree value and either 'latitude' or 'longitude' to get
+    the appropriate index number for the first two of these index numbers.
+
+    Parameters
+    ----------
+    degrees : float or int
+        Degrees of either latitude or longitude.
+    degree_type : string
```
wholmgren: coordinate instead of degree_type?

pvlib/clearsky.py
```diff
@@ -301,28 +291,63 @@ def _calendar_month_middles(year):
...
-                     (inputmatrix, inputmax - inputrange, inputmax))
-    # round down if input is within half an index or else raise index error
-    if outputmatrix > outputmax:
-        if np.around(outputmatrix - outputmax, 1) <= 0.5:
-            outputmatrix = outputmax
+                     (degrees, inputmin, inputmax))
+
+    # If the index is still out of bounds after rounding, raise an error.
+    if index > outputmax:
+        if index - outputmax <= 0.500001:
```
wholmgren: comment on why `0.500001` instead of `0.5`

pvlib/clearsky.py
```diff
@@ -195,22 +195,12 @@ def lookup_linke_turbidity(time, latitude, longitude, filepath=None,
...
     lt_h5_file = tables.open_file(filepath)
-    try:
-        lts = lt_h5_file.root.LinkeTurbidity[latitude_index,
-                                             longitude_index, :]
-    except IndexError:
-        raise IndexError('Latitude should be between 90 and -90, '
-                         'longitude between -180 and 180.')
-    finally:
-        lt_h5_file.close()
+    lts = lt_h5_file.root.LinkeTurbidity[latitude_index, longitude_index, :]
```
wholmgren: while we're at it, let's use a context manager to automatically open/close the file:

```python
with tables.open_file(filepath) as lt_h5_file:
    lts = lt_h5_file.root.LinkeTurbidity[latitude_index, longitude_index, :]
```

alexandermorgan: These are all great ideas, and I've integrated them all with my latest commits.

pvlib/clearsky.py
```diff
@@ -151,9 +151,9 @@ def lookup_linke_turbidity(time, latitude, longitude, filepath=None,
     ----------
     time : pandas.DatetimeIndex
 
-    latitude : float
+    latitude : numeric
```
wholmgren: `numeric` implies non-scalar input is ok, which is not true here. See https://pvlib-python.readthedocs.io/en/stable/contributing.html#documentation

alexandermorgan: Ok, so I'll change it to float or int. I'll also fix the test argument issue.

wholmgren:
@alexandermorgan please address the merge conflict and then I think we're ready to merge.

mikofski:
Just to clear up my name, [`_linearly_scale` was ported from the MATLAB `LinearlyScale` 5 years ago](../blob/c3777211d3019069f41c1886a59702078ee2a82d/pvlib/clearsky.py#L283) perhaps that's why the syntax is weird? Anyway it has nothing to do with me. #264 was about fixing the way Linke Turbidity was interpolated, to confirm to the correct resolution which was off

wholmgren:
@mikofski thanks for the review. Your name is clear - sorry to imply otherwise!

PR #757, opened 2019-07-21
remove python 2 lines

### LINKED ISSUE #729: Remove Python 2 vestiges
With the merging of #728, we can now consider dropping Python2-specific code and workarounds.

Some candidates are:
- https://github.com/pvlib/pvlib-python/blob/master/pvlib/atmosphere.py#L214
- Uses of `from __future__ import ...`
- Other places where Python 2 vs. 3 considerations may have "[complected](https://www.youtube.com/watch?v=rI8tNMsozo0)" the code.

alexandermorgan:
- [X] Closes #729


docs/sphinx/source/whatsnew/v0.7.0.rst
```diff
@@ -8,6 +8,7 @@ recommend all users of v0.6.3 upgrade to this release after checking API
 compatibility notes.
 
 **Python 2.7 support ended on June 1, 2019.** (:issue:`501`)
+**Python 2.7 support ended on June 1, 2019**. (:issue:`501`)
```
markcampanelli: Looks like a duplicate.

alexandermorgan: Good catch, thanks! I put line 10 in to contain the period within the bolding in the rendered version. I guess I didn't get the merge quite right. I'll add this to the PR.

alexandermorgan:
I believe that addresses all the cases. They're hard to find because some not all of the try/except imports are labeled as being needed for py2-py3 compatibility. In one case tkinter was only being imported in the python2 way, so that should all be sorted out now. I think the low diff test coverage is ok, because it's hard to test the case when a module can't be imported, and no new unchecked lines were introduced.
One issue I would particularly appreciate review on is the changes to test_conftest.py. I removed the parts about python2, but now I wonder if we shouldn't just remove test_fail_on_pvlib_version() and test_fail_on_pvlib_version_fail_in_test() altogether.

wholmgren:
The diff looks good. It would not surprise me if there are a few other python 2 things hiding in the code but they can be addressed later.

>  I wonder if we shouldn't just remove test_fail_on_pvlib_version() and test_fail_on_pvlib_version_fail_in_test() altogether.

Why do you want to remove these? They test deprecations within pvlib and are not related to the python version.

alexandermorgan:
I was concerned that with the `strict` argument changed, they would no longer be valid. You're right that we should keep these tests, but I'll remove the `strict` arguments altogether since the specific case (python 2 on Windows) is no longer supported.

wholmgren:
Strict should now be set to True, not removed altogether.

alexandermorgan:
> I'd rather not touch the `versioneer.py`

Ok, I've undone the two changes I previously made to versioneer.py.

> Also, why does your first comment say that the PR partially address the issue instead of closing it? Is there more to do (that we know about)?

I said that because I didn't understand the python 2 comment in atmosphere.py. Later I went back and looked at the history for that file and saw that the cases where the zenith > 90° got filtered out anyway, so the previous changes just did that filtering before all the regular calculations (which are only meant for cases where zenith <= 90°). This change addressed a problem in py2 but it didn't actually change the behavior in py3 and is maybe even slightly preferable in py3 anyway. So all I ended up doing was rewriting the comment (now it reads: "set zenith values greater than 90 to nans") so it would be clearer.

PR #758, opened 2019-07-29
Calculating horizon profiles and associated shading losses

JPalakapillyKWH:

Code to get generate horizon profiles using the googlemaps API. Also has functionality to create shading losses due to the horizon profile. Currently only done with an isotropic sky diffuse model and the direct normal.
Follows much of the methodology in https://doi.org/10.1016/j.solener.2014.09.037

I've been developing this code in recent weeks. Is this something PVlib wants to incorporate?


adriesse:
I think this would make a great addition!

cwhanse:
> Is this something PVlib wants to incorporate?

Yes, I think so. `horizon.py` is OK, I would like to think a bit more about how this module relates to existing modules (`irradiance.py`, `bifacial.py`) and other process models not yet in pvlib, e.g., effects of partial shading on an array.

cwhanse:
@JPalakapillyKWH I'm not familiar with Google maps elevation data. What is the spatial resolution? Do the data represent the near-field or only the far-field horizon?  Near-field accounts for trees, buildings, etc., far-field is terrain such as mountains and hills.

JPalakapillyKWH:
@cwhanse The spatial resolution varies since googlemaps (as I've gathered) uses a patchwork of different signals but is generally at least 90m. The code will capture the near-field horizon to the extent that the googlemaps AP is affected by buildings/trees etc.

I included the googlemaps elevation calls because it was a simple way to get elevation data for sites distributed across the world, but the bulk of the horizon module is agnostic to the source of the elevation data.

cwhanse:
Ok, was just thinking how to describe the output of this module, as near-field or far-field shading.  Horizon is usually interpreted as far field.

JPalakapillyKWH:
It is designed for far-field.

cwhanse:
@JPalakapillyKWH super content for pvlib.  I've read the reference and look through the code. Some high-level comments about modules and connecting with the rest of pvlib. would appreciate  @wholmgren @mikofski input here.

- `horizon.py` main purpose is to provide functions that return the horizon profile (list of (azimuth, dip angle) tuples) and sky diffuse loss factor `dtf`. I'd put `calculate_dtf` and `collection_plane_dip_angle` in `horizon.py`.  I'd add the Boolean "in-front" variable as an output if that is being calculated somewhere.
- Some of the helper functions e.g. `pol2cart`, `latitude_to_geocentric` could go into `pvlib.tools`. I'd leave the grid, sample, filter and gmaps functions in `horizon.py`. 
- There's a mix of "to" and "from" function names, I'd prefer "to".  
- I am wondering if there's a way to use a modified calculation of `dtf` in the current `isotropic` sky diffuse model, rather than making a new model `horizon_adjusted`. Could we change the denominator in the `dtf` calculation to sum over only the cells in front of the array?  Then sky diffuse = dhi * (1 + cos(tilt))/2 * dtf  and `dtf` could be an optional kwarg for the current `isotropic` function.
- it's too bad the reference isn't more explicit about the time series simulation.

wholmgren:
I skimmed the code and it looks like a great start - thanks @JPalakapillyKWH! I've not yet looked at the reference. 

In addition to expanding some of the function documentation, I think we're going to want some documentation that synthesizes the proposed code. 

There are a handful of places that generate tuples or loop over them in calculations. Can the code make more use of numpy arrays to improve performance and readability?

I would encourage marking some of the helper functions as private with an `_` and/or putting some of these functions in `pvlib.tools` (which is not explicitly private but is also not documented in the `api.rst` file). 

Great that the code is agnostic about elevation data source. I am not so sure about including the googlemaps api wrapper code in pvlib python, but I'm definitely ok using it in example documentation. I'd probably want to use the functions with a gridded data source such as [SRTM](https://www2.jpl.nasa.gov/srtm/).

JPalakapillyKWH:
@cwhanse and @wholmgren  I made a lot of the code restructuring changes y'all suggested. Many functions were moved to tools.py.

I am currently working on making the code more numpy-friendly. I'm also working on creating documentation by creating a .rst file in the sphinx source folder and updating api.rst. Is this right way to create documentation?

I did include a few functions for visualizing horizon profiles but they require matplotlib to run. Do these seem worth including in the horizon module itself or is this something that would go in the documentation?

@cwhanse I don't think summing over the front of the panel will work since the horizon could be at a higher elevation angle than the back of the panel (e.g. panel tilt is 20 deg but the horizon is at a 30 deg elevation angle behind it). I just included the horizon adjustment code in the isotropic model with horizon profile as a kwarg. Let me know what you think.

pvlib/irradiance.py
```diff
@@ -661,9 +668,18 @@ def isotropic(surface_tilt, dhi):
...
 
+    surface_azimuth : numeric
+        Surface azimuth angles in decimal degrees. surface_azimuth must
+        be >=0 and <=360. The azimuth convention is defined as degrees
+        east of north (e.g. North = 0, South=180 East = 90, West = 270).
+
     dhi : numeric
         Diffuse horizontal irradiance in W/m^2. DHI must be >=0.
 
+    horizon_profile : list (optional)
```
adriesse: How about making this into two vector arguments?

pvlib/irradiance.py
```diff
@@ -2885,3 +2921,43 @@ def dni(ghi, dhi, zenith, clearsky_dni=None, clearsky_tolerance=1.1,
...
+        The modeled direct normal irradiance in the plane of array.
+    solar_zenith : numeric
+        Solar zenith angle.
+    solar_azimuth : numeric
+        Solar azimuth angle.
+    horizon_profile : list
+        A list of (azimuth, dip_angle) tuples that defines the horizon
+        profile given. Every azimuth from -180 to 180, in 1 degree
+        increments, must be in this list.
+
```
adriesse: If you require 361 values, you do not need the tuples; just a vector of dip angles would do.

pvlib/irradiance.py
```diff
@@ -2885,3 +2921,43 @@ def dni(ghi, dhi, zenith, clearsky_dni=None, clearsky_tolerance=1.1,
...
+    ----------
+    poa_direct : numeric
+        The modeled direct normal irradiance in the plane of array.
+    solar_zenith : numeric
+        Solar zenith angle.
+    solar_azimuth : numeric
+        Solar azimuth angle.
+    horizon_profile : list
+        A list of (azimuth, dip_angle) tuples that defines the horizon
+        profile given. Every azimuth from -180 to 180, in 1 degree
```
adriesse: Is this a different convention than used above?  (0-360 degrees)

pvlib/irradiance.py
```diff
@@ -661,9 +668,18 @@ def isotropic(surface_tilt, dhi):
...
+        be >=0 and <=360. The azimuth convention is defined as degrees
+        east of north (e.g. North = 0, South=180 East = 90, West = 270).
+
     dhi : numeric
         Diffuse horizontal irradiance in W/m^2. DHI must be >=0.
 
+    horizon_profile : list (optional)
+        A list of (azimuth, dip_angle) tuples that defines the horizon
+        profile
+
```
adriesse: Do you  require 361 values here?

pvlib/test/test_irradiance.py
```diff
@@ -940,3 +940,53 @@ def test_clearness_index_zenith_independent(airmass_kt):
...
+                                                        ephem_data["azimuth"],
+                                                        ephem_data["zenith"])
+
+    max_expected = pd.Series([0, 0, 0, 0])
+    assert_allclose(max_adjusted.values, max_expected.values, atol=1e-7)
+
+
+def test_horizon_adjusted_isotropic():
+    zero_horizon = []
+    max_horizon = []
```
adriesse: I suggest adding a couple of intermediate test cases.

mikofski:
Sorry to jump in late. Great addition @JPalakapillyKWH.

One caveat that should probably be mentioned somewhere, is that I believe some irradiance sources, such as NSRDB may already include the far field horizon loss, since they are calibrated by ground stations, please correct me if I'm wrong.

I too would like to see examples from other geographical sources like NASA SRTM and ESRI, but I think those should be separate PRs .

I'll try to take a closer look later today.

Thanks!

cwhanse:
> @cwhanse I don't think summing over the front of the panel will work

I was not clear - I meant sum over the sky cells visible from the plane of array. The revised implementation accomplishes the same end for the isotropic sky and is arguably closer to the published reference.

Not in this PR, but it appears to me that we can extend the use of horizon data to other transposition models. That's what I had in mind when I made the comment.

JPalakapillyKWH:
@mikofski You are correct about ground station data already incorporating far field losses. This is designed for satellite irradiance data (PSM, TWC etc.)
@adriesse I took many of your suggestions and ran with them. 

There's been a lot of restructuring in the code to move towards numpy. I'm still updating docstrings though.

mikofski:
>This is designed for satellite irradiance data (PSM, TWC etc.)

Awesome! Would you consider adding a advisory note in the docstring with this warning on usage?

JPalakapillyKWH:
> Awesome! Would you consider adding a advisory note in the docstring with this warning on usage?

Which docstring do you mean?

mikofski:
I would recommend adding a warning note in any method/function that uses the new `horizon_*` args/kwargs:

- `pvlib.irradiance.isotropic`
- `pvlib.irradiance.get_sky_diffuse`
- `pvlib.irradiance.get_total_irradiance`

pvlib/irradiance.py
```diff
@@ -13,7 +13,7 @@
 import numpy as np
 import pandas as pd
 
-from pvlib import atmosphere, solarposition, tools
+from pvlib import atmosphere, solarposition, tools, horizon
```
cwhanse: alphabetical order please

pvlib/irradiance.py
```diff
@@ -661,9 +670,23 @@ def isotropic(surface_tilt, dhi):
...
+
     dhi : numeric
         Diffuse horizontal irradiance in W/m^2. DHI must be >=0.
 
+    horizon_azimuths: numeric (optional)
+        Azimuth values for points that define the horizon profile. The ith
+        element in this array corresponds to the ith element in horizon_angles.
+
+    horizon_angle: numeric (optional)
+        Dip angle values for points that define the horizon profile. The ith
```
cwhanse: Would help to define 'dip angle' here.

pvlib/irradiance.py
```diff
@@ -661,9 +670,23 @@ def isotropic(surface_tilt, dhi):
...
+        east of north (e.g. North = 0, South=180 East = 90, West = 270).
+
     dhi : numeric
         Diffuse horizontal irradiance in W/m^2. DHI must be >=0.
 
+    horizon_azimuths: numeric (optional)
+        Azimuth values for points that define the horizon profile. The ith
+        element in this array corresponds to the ith element in horizon_angles.
+
+    horizon_angle: numeric (optional)
```
cwhanse: `horizon_angles`

JPalakapillyKWH:
There seem to be a lot of timeout issues in the tests. Anybody have any insight? I pass `pytest pvlib` locally.

pvlib/irradiance.py
adriesse: Perhaps this might be a good place to comment on the term `dip angle`.  If I understand correctly, a typical horizon will consist of negative dip angles.  This seems counter-intuitive to me and contrary to usage I've seen in other places, where the horizon has a height (also an angle of course).  I am curious how the term `dip` came to be used.

wholmgren: me too

adriesse:
Is it really necessary/desirable to pull these options into the existing irradiance functions?  For my purposes I would be happy with two separate functions: 
- one to give me an adjustment factor for diffuse irradiance, and
- one to give me an adjustment factor for DNI (binary in its simplest form).

pvlib/irradiance.py
JPalakapillyKWH: I do agree that dip angle is a confusing term. Even more confusing is that I am not really using "true" dip angles in the module since I treat a typical horizon as having positive dip angles (I use dip angle to mean degrees above the horizontal). The term came from http://sherrytowers.com/2014/04/13/archeoastronomy-calculating-the-horizon-profile-using-online-us-geographic-survey-data/ which I had been following when first starting work on horizon code. 
What would be a better term to use? Horizon elevation angle?

pvlib/irradiance.py
JPalakapillyKWH: fixed

pvlib/irradiance.py
JPalakapillyKWH: fixed

pvlib/test/test_irradiance.py
JPalakapillyKWH: done

pvlib/irradiance.py
JPalakapillyKWH: no. The restructured code might make this easier to see.

pvlib/irradiance.py
JPalakapillyKWH: I've changed all azimuths to be from 0-360 now i believe.

pvlib/irradiance.py
JPalakapillyKWH: ran with this

pvlib/irradiance.py
JPalakapillyKWH: done

pvlib/irradiance.py
cwhanse: I'm ok with "dip angle" but whatever term is used it needs a definition.

JPalakapillyKWH:
@adriesse Fair point. I can include just horizon.py and test_horizon.py in this PR.

pvlib/irradiance.py
JPalakapillyKWH: changed dip angles to elevation angles. I think it makes more sense since its given in degrees above the horizontal

adriesse:
>@adriesse Fair point. I can include just horizon.py and test_horizon.py in this PR.

@JPalakapillyKWH Beware though, that my opinion is only one; there may be others...

JPalakapillyKWH:
> @JPalakapillyKWH Beware though, that my opinion is only one; there may be others...

I tend to agree that this code should be split into multiple PRs. Hopefully this will be easier to review now.

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+                                            sampling_param)
+
+    bearing_deg, elevation_angle_deg = elevation_angle_calc(site, samples)
+
+    return bearing_deg, elevation_angle_deg
+
+
+def sample_using_grid(lat_grid, lon_grid, elev_grid):
+    """
+    Calculates the Elevation angle from the site (center of the grid)
```
cwhanse: Description doesn't seem accurate

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+    grid_shape = lat_grid.shape
+    grid_center_i = (grid_shape[0] - 1) // 2
+    grid_center_j = (grid_shape[1] - 1) // 2
+    site_lat = lat_grid[grid_center_i, grid_center_j]
+    site_lon = lon_grid[grid_center_i, grid_center_j]
+    site_elev = elev_grid[grid_center_i, grid_center_j]
+    site = np.array([site_lat, site_lon, site_elev])
+
+    if sampling_method == "grid":
+        samples = sample_using_grid(lat_grid, lon_grid, elev_grid)
```
cwhanse: Suggest that the three "sample" functions be private

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+
+    elev_grid : ndarray
+        A 2d array containing elevation values that correspond to the other
+        two input grids.
+
+    sampling_method : string
+        A string that specifies the sampling method used to generate the
+        horizon profile.
+
+    sampling_param : variable
```
cwhanse: Suggest calling this input `num_samples`, default `None`, and describe the value that should be passed if `sampling_method` is `triangles` or `interpolator` (suggest "polar" instead of "interpolator")

JPalakapillyKWH:
> Working my way through the docstrings and code. I am a little confused about the purpose of the "sample" functions - if a grid is being supplied as input, why does the grid need to be sampled? Is the input grid not regularly spaced?

The input grid will be regularly spaced. The sampling functions are to produce a smooth horizon profile.
Without sampling, the "smoothness" of the horizon profile is highly dependent on the size of your grid. If you don't have enough points on your grid then you will have a very choppy horizon profile. Some azimuths might not even have elevation angles. So the sampling functions are a way of pseudo-generating more elevation data to make the horizon smoother.

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+
+    Parameters
+    ----------
+    lat : numeric
+        The latitude of the location that is to be the center of the grid.
+
+    lon : numeric
+        The longitude of the location that is to be the center of the grid.
+
+    Returns
```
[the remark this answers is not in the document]
JPalakapillyKWH: will do.

wholmgren:
I stumbled across [earthpy](https://github.com/earthlab/earthpy) and this [hillshade example](https://earthpy.readthedocs.io/en/latest/gallery_vignettes/plot_dem_hillshade.html#). Thought people interested in this PR might want to check it out.

cwhanse:
> Working my way through the docstrings and code. I am a little confused about the purpose of the "sample" functions - if a grid is being supplied as input, why does the grid need to be sampled? Is the input grid not regularly spaced?
> 
> The input grid will be regularly spaced. The sampling functions are to produce a smooth horizon profile.
> Without sampling, the "smoothness" of the horizon profile is highly dependent on the size of your grid. If you don't have enough points on your grid then you will have a very choppy horizon profile. Some azimuths might not even have elevation angles. So the sampling functions are a way of pseudo-generating more elevation data to make the horizon smoother.

Makes sense now, thanks.

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+horizon profile.
+"""
+from __future__ import division
+
+import itertools
+
+import numpy as np
+import warnings
+
+from scipy.interpolate import RegularGridInterpolator
```
cwhanse: Since this is the only use of scipy, I suggest we move the import inside a try/except in `sample_using_interpolation`  See https://github.com/pvlib/pvlib-python/blob/0a14b273517a082603cc157faa88a5ab0ac4cac9/pvlib/solarposition.py#L887 for an example

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+        A 2d array containing elevation values that correspond to the other
+        two input grids.
+
+
+    Returns
+    -------
+    all_samples: Nx3 ndarray
+        Array of lat, lon, elev points that are grid points.
+    """
+    assert(lat_grid.shape == lon_grid.shape == elev_grid.shape)
```
cwhanse: I don't think we need the `assert` here and in the other `sample` functions, since grid size equality is tested in `calculate_horizon_points`

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+
+    lats = lat_grid.flatten()
+    lons = lon_grid.flatten()
+    elevs = elev_grid.flatten()
+    samples = np.stack([lats, lons, elevs], axis=1)
+    # remove site from samples
```
cwhanse: I prefer "center" to "site"

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+
+    all_samples = np.delete(samples, samples.shape[0]//2, axis=0)
+    return all_samples
+
+
+def sample_using_triangles(lat_grid, lon_grid, elev_grid,
+                           samples_per_triangle=10):
+    """
+    Creates triangles using nearest neighbors for every grid point and randomly
+    samples each of these triangles to find elevation angles for the horizon
```
cwhanse: Delete "to find elevation angles...."

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+                                                    samples_per_triangle)
+                all_samples = np.vstack([all_samples, samples])
+    return np.array(all_samples)
+
+
+def sample_using_interpolator(lat_grid, lon_grid, elev_grid, num_samples):
+    """
+    Creates a "grid" using polar coordinates and uses the scipy's grid
+    interpolator to estimate elevation values at each point on the polar grid
+    from the input (rectangular) grid that has true elevation values. Elevation
```
cwhanse: I think the second sentence is not needed.

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+    r2 = np.random.rand(num_samples).reshape((num_samples, 1))
+    sqrt_r1 = np.sqrt(r1)
+
+    random_pts = (1-sqrt_r1)*c1 + sqrt_r1*(1-r2)*c2 + sqrt_r1*r2*c3
+    random_pts = tools.xyz_to_lle(random_pts)
+
+    return random_pts
+
+
+def filter_points(horizon_azimuths, horizon_angles, bin_size=1):
```
cwhanse: "find_horizon_points" maybe?  I don't see that this function is used so I'm unsure what of it's purpose.

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+    """
+    Bins the horizon_points by azimuth values. The azimuth value of each
```
cwhanse: Need to update the description with the change in argument names.

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+
+    horizon_angles: numeric
+        Elevation angle values for points that define the horizon profile. The
+        elevation angle of the horizon is the angle that the horizon makes with
+        the horizontal. It is given in degrees above the horizontal. The ith
+        element in this array corresponds to the ith element in
+        horizon_azimuths.
+
+    bin_size : int
+        The width of the bins for the azimuth values.
```
cwhanse: `bin_size` needs a unit - is it degrees?

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+        filtered_angles.append(wedges[key])
+
+    filtered_angles = np.array(filtered_angles)
+    filtered_azimuths = np.array(filtered_azimuths)
+    return filtered_azimuths, filtered_angles
+
+
+def collection_plane_elev_angle(surface_tilt, surface_azimuth, direction):
+    """
+    Determine the elevation angle created by the surface of a tilted plane
```
cwhanse: The description is unclear to me. I think this function is calculating the elevation angle of the unit normal after rotation around the earth surface normal to a different azimuth `direction`, ?

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+    return dtf
+
+
+def DNI_horizon_adjustment(horizon_angles, solar_zenith, solar_azimuth):
+    '''
+    Calculates an adjustment to DNI based on a horizon profile. The adjustment
+    is a vector of binary values with the same length as the provided
+    solar position values. Where the sun is below the horizon, the adjustment
+    vector is 0 and it is 1 elsewhere. The horizon profile must be given as a
+    vector with 361 values where the ith value corresponds to the ith degree
```
cwhanse: Is the 361st value needed?  I expected 0 - 359, since the last value at 360 should be a duplicate of the first value at 0.

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+        elev = np.maximum(horizon_elev, temp)
+
+        first_term = .5 * (a*np.cos(az) + b*np.sin(az)) * \
+                          (np.pi/2 - elev - np.sin(elev) * np.cos(elev))
+        second_term = .5 * c * np.cos(elev)**2
+        dtf += 2 * (first_term + second_term) / num_points
+    return dtf
+
+
+def DNI_horizon_adjustment(horizon_angles, solar_zenith, solar_azimuth):
```
cwhanse: `dni` in function name.  Spell out the acronym "direct normal irradiance (DNI)" in the description.

pvlib/horizon.py
JPalakapillyKWH: The purpose of this function is to make the horizon profile smooth by consolidating horizon points that have similar azimuth values (depending on bin size). For example, if you calculate horizon points from a grid you will have many points that have an azimuth of 0. But you really only want the horizon point with the greatest elevation angle.

cwhanse: That makes sense.  Maybe it would help if the inputs were `azimuths`, `elevation_angles`, and the term "horizon" is only used to for output quantities, since the inputs are not necessarily points on the visible horizon.  I do think a more specific function name would help.

pvlib/horizon.py
```diff
@@ -0,0 +1,646 @@
...
+
+    filtered_angles: numeric
+        elevation angle values for points that define the horizon profile given
+        in degrees above the horizontal. The ith element in this array
+        corresponds to the ith element in filtered_azimuths.
+
+    """
+    assert(horizon_azimuths.shape[0] == horizon_angles.shape[0])
+
+    wedges = {}
```
cwhanse: I'll guess this loop may be a bit slow. What about instead:
```
rounded_azimuths = tools.round_to_nearest(horizon_azimuths, bin_size)
bins = np.unique(rounded_azimuths)

filtered = np.column_stack((bins, np.nan * bins))

for i in range(filtered.shape[0]):
    idx = np.nonzero(rounded_azimuths == filtered[i, 0])
    filtered[i, 1] = np.max(horizon_angles[idx])

return filtered[:, 0], filtered[:, 1]
```

Now that I've thought that through, there will always be a bin at 0 degrees, so the user ought to be informed of this point. I'd add to the note about `bin_size`.

JPalakapillyKWH: A couple of questions on this.
Is `np.nonzero` necessary?  I think you can just index with the boolean array. 
Why will there always be a bin at 0 degrees?

cwhanse: > A couple of questions on this.
> Is np.nonzero necessary?  I think you can just index with the boolean array.

You're probably right. I started with [argwhere](https://docs.scipy.org/doc/numpy/reference/generated/numpy.argwhere.html#numpy.argwhere), decided it wasn't the right approach, stopped at `nonzero` without thinking further.

> Why will there always be a bin at 0 degrees?

You are right again, there won't always be a bin at 0.  But the bins will be a subset of 0, +/-bin_size, +/-2*bin_size, etc.  That's what I meant to describe and what I think we shoudl tell the user.

cwhanse: Needs tests for the functions added to `tools.py` then I'm OK with the review.

pvlib/horizon.py
```diff
@@ -0,0 +1,649 @@
...
+    x1 = np.atleast_1d(v1.T[0])
+    y1 = np.atleast_1d(v1.T[1])
+    z1 = np.atleast_1d(v1.T[2])
+
+    delta = np.atleast_2d(np.subtract(v1, v2))
+
+    normal = np.atleast_2d(np.stack([2*x1/a**2, 2*y1/a**2, 2*z1/b**2], axis=1))
+
+    # Take the dot product of corresponding vectors
+    dot = np.sum(np.multiply(delta, normal), axis=1)
```
adriesse: Would `np.dot()` do the trick?

pvlib/horizon.py
```diff
@@ -0,0 +1,649 @@
+"""
+The ``horizon`` module contains functions for horizon profile modeling.
+There are various geometric utilities that are useful in horizon calculations
+as well as a method that uses the googlemaps elevation API to create a
```
adriesse: Was the gmaps function removed?

pvlib/horizon.py
```diff
@@ -0,0 +1,649 @@
...
+        The bearings from pt1 to pt2 in degrees East of North.
+
+    elevation_angle_deg: numeric
+        The elevation angles that the points in pt2 make with the horizontal
+        as observed from the points in pt1. Given in degrees above the
+        horizontal.
+
+    '''
+    a = 6378137.0
+    b = 6356752.0
```
adriesse: Comments for magic numbers?

pvlib/horizon.py
JPalakapillyKWH: Okay so I'm taking two Nx3 arrays and trying to do an "element-wise" dot product to get back an Nx1 array. I couldn't get `np.dot()` to do this but let me know if you know how.

pvlib/horizon.py
JPalakapillyKWH: Yep. I'll fix the docstring

adriesse:
I read the paper mentioned at the top of this PR and my impression is that there is not much overlap between the methods discussed there, and those programmed here.  Is this correct? 

I would very much like to see notes in the docstrings regarding the nature of the algorithms used, their origins, and their validation.  If there are new original methods, please also identify and take credit for them!  If there is a forthcoming publication, I would love to see it; but if you're still working on one, I would suggest pausing the PR until it is finished and you can share it.

JPalakapillyKWH:
@adriesse 
You are right in that there is not a whole lot of overlap. Much of the code in this PR is taken from the geometry used in viewshed analysis. The sampling functions and binning of the horizon profile were ideas that I got from http://sherrytowers.com/2014/04/13/archeoastronomy-calculating-the-horizon-profile-using-online-us-geographic-survey-data/. 
I don't have any plans for a publication, just working on this for the summer. 

For my own validation of horizon profile modeling, I used the algorithms in this module with elevation data taken from the googlemaps elevation API to calculate horizon profiles. These were mostly compared by eye to horizon profiles from other sources and the online tool http://www.heywhatsthat.com/. 

I also did validation of shading losses due to the horizon. I found shading loss by comparing the output of a modelchain run with the horizon adjusted irradiance vs without. I then compared this to far field losses outputted by PVSyst for a few modeled projects and found that they agreed.

cwhanse:
> Perhaps a good next step would be to add (or expand) Notes sections for the doc strings, where you explain the algorithms used, and give references to the sources you used, where applicable.

`calculate_dtf` can reference the Goss et al. Solar Energy paper for the idea of `dtf`. We need a reference for the summarization formula (lines 598 - 601).  I thought the code was implementing the sum over equal area sky patches?  I may have missed a decision along the way.

I'd reference the ACM transactions paper instead of the class notes for `uniformly_sample_triangle`.

Add the reference to `elevation_and_azimuth` - it is called out as `[1]` but not listed. The rest of the functions seem straight-forward enough.

> If validation is not provided in some formal publication, then perhaps it can be done in a jupyter notebook or something, but I don't know what the best place would be to store such a thing.  Of course you could also reconsider whether maybe you could make a publication out of this work...

 I vote for a page on readthedocs. Separate PR though.

pvlib/horizon.py
```diff
@@ -0,0 +1,651 @@
...
+
+    lon_grid : ndarray
+        A 2d array of longitude values for the grid.
+
+    elev_grid : ndarray
+        A 2d array of elevation values for the grid.
+
+    sampling_method : string, default "grid"
+        A string that specifies the sampling method used to generate the
+        horizon profile. Acceptable values are: "grid", "triangles", "polar".
```
wholmgren: Need some explanation of what these sampling methods mean. Perhaps an Examples section would work best for this.

pvlib/horizon.py
```diff
@@ -0,0 +1,651 @@
...
+    pts = np.array([tools.polar_to_cart(e[0], e[1]) for e in polar_pts])
+    pts += np.array((center_lat, center_lon))
+    total_num_samples = num_samples[0]*num_samples[1]
+
+    interpolated_elevs = interpolator(pts).reshape(total_num_samples, 1)
+    samples = np.hstack((pts, interpolated_elevs))
+    return samples
+
+
+def uniformly_sample_triangle(p1, p2, p3, num_samples):
```
wholmgren: What's the purpose of this public method?

pvlib/tools.py
```diff
@@ -425,3 +425,112 @@ def _golden_sect_DataFrame(params, VL, VH, func):
...
+def xyz_to_lle(point):
+    """
+    Converts a (x, y, z) tuple into a (lat, lon, elev) tuple.
+    The center of the earth is the origin in the xyz-space.
+    The output latitude is assumed to be a common latitude (geodetic).
+    """
+    # Equatorial Radius of the Earth (ellipsoid model) in meters
+    a = 6378137.0
+    # Polar Radius of the Earth (ellipsoid model) in meters
+    b = 6356752.0
```
wholmgren: Let's define these once at the top of `tools.py` as `earth_radius_equatorial` and `earth_radius_polar`. You can set `a = ` and `b = ` within the functions if you want to keep the expressions short. Or `a_sqrd = `, `b_sqrd = ` in the functions as suggested elsewhere.

setup.py
```diff
@@ -37,10 +37,12 @@
 MAINTAINER_EMAIL = 'holmgren@email.arizona.edu'
 URL = 'https://github.com/pvlib/pvlib-python'
 
-INSTALL_REQUIRES = ['numpy >= 1.10.4',
+INSTALL_REQUIRES = ['numpy >= 1.13.3',
                     'pandas >= 0.18.1',
                     'pytz',
-                    'requests']
+                    'requests',
+                    'scipy',
```
wholmgren: revert this change

setup.py
```diff
@@ -37,10 +37,12 @@
 MAINTAINER_EMAIL = 'holmgren@email.arizona.edu'
 URL = 'https://github.com/pvlib/pvlib-python'
 
-INSTALL_REQUIRES = ['numpy >= 1.10.4',
+INSTALL_REQUIRES = ['numpy >= 1.13.3',
```
wholmgren: what are you using that requires numpy 1.13? Increasing the required version is allowed, but I'd like to understand why.

pvlib/horizon.py
```diff
@@ -0,0 +1,651 @@
+"""
+The ``horizon`` module contains functions for horizon profile modeling.
+There are various geometric utilities that are useful in horizon calculations.
+"""
+from __future__ import division
```
wholmgren: python 3 only now, so this is not necessary

pvlib/horizon.py
wholmgren: I also missed the purpose in my review, so it should be made more explicit in the documentation.

pvlib/horizon.py
```diff
@@ -0,0 +1,651 @@
...
+
+    Parameters
+    ----------
+    lat : numeric
+        The latitude of the location that is to be the center of the grid.
+
+    lon : numeric
+        The longitude of the location that is to be the center of the grid.
+
+    grid_radius : numeric
```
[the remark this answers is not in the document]
JPalakapillyKWH: A lot of elevation data (SRTM) is already gridded rectangularly. I figured a rectangular grid would be easiest to interface with different sources of elevation data.
I agree radius isn't a great term but I'm not sure what's better..."half-width"?

pvlib/horizon.py
```diff
@@ -0,0 +1,651 @@
...
+        as observed from the points in pt1. Given in degrees above the
+        horizontal.
+
+    '''
+    # Equatorial Radius of the Earth (ellipsoid model) in meters
+    a = 6378137.0
+    # Polar Radius of the Earth (ellipsoid model) in meters
+    b = 6356752.0
+
+    lat1 = np.atleast_1d(pt1.T[0])
```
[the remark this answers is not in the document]
JPalakapillyKWH: I had this function accept [3,] shaped arrays earlier.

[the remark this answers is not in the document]
JPalakapillyKWH: I think it's okay to allow [3,] shaped arrays as inputs by including the calls to atleast_1d and atleast_2d

pvlib/horizon.py
JPalakapillyKWH: will make private

pvlib/horizon.py
JPalakapillyKWH: Added a notes section with explanations.

pvlib/horizon.py
```diff
@@ -0,0 +1,651 @@
...
+    sampling_method : string, default "grid"
+        A string that specifies the sampling method used to generate the
+        horizon profile. Acceptable values are: "grid", "triangles", "polar".
+
+    num_samples : variable, default None
+        A parameter that is passed into the function specified by
+        sampling_method.
+        If the sampling method is "triangles" this corresponds
+        to the number of samples taken from each triangle.
+        See _sampling_using_triangles for more info.
```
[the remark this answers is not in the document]
JPalakapillyKWH: haha fair enough

pvlib/horizon.py
```diff
@@ -0,0 +1,651 @@
...
+    tilt_rad = np.radians(surface_tilt)
+    plane_az_rad = np.radians(surface_azimuth)
+    a = np.sin(tilt_rad) * np.cos(plane_az_rad)
+    b = np.sin(tilt_rad) * np.sin(plane_az_rad)
+    c = np.cos(tilt_rad)
+
+    # this gets either a float or an array of zeros
+    dtf = np.multiply(0.0, surface_tilt)
+    num_points = horizon_azimuths.shape[0]
+    for i in range(horizon_azimuths.shape[0]):
```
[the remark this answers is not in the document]
JPalakapillyKWH: The broadcasting issue is in collection_plane_elev_angle. I'll see if I can make it work.

pvlib/horizon.py
```diff
@@ -0,0 +1,685 @@
...
+        the plane.
+
+    Notes
+    _____
+
+    The dtf in this method is calculated by approximating the surface integral
+    over the visible section of the sky dome. The integrand of the surface
+    integral is the cosine of the angle between the incoming radiation and the
+    vector normal to the surface. The method calculates a sum of integrations
+    from the "peak" of the sky dome down to the elevation angle of the horizon.
```
adriesse: Could you mention the relevant section or equation(s) in [1]?

JPalakapillyKWH: Do you mean [2]?

adriesse: I only see one reference...

JPalakapillyKWH: my bad. latest changes have the new reference. Old one was wrong.

pvlib/horizon.py
```diff
@@ -612,11 +614,14 @@ def calculate_dtf(horizon_azimuths, horizon_angles,
     integral is the cosine of the angle between the incoming radiation and the
     vector normal to the surface. The method calculates a sum of integrations
     from the "peak" of the sky dome down to the elevation angle of the horizon.
+    A similar method is used in section II of [1] although it is looking at
+    both ground and sky diffuse irradiation.
 
-    [1] Goss et al. (2014) Solar Energy 110, 410-419
-
+    [2] Wright D. (2019) IEEE Journal of Photovoltaics 9(2), 391-396
```
adriesse: Looks interesting. Could you send me a copy?

JPalakapillyKWH: sent an email

pvlib/horizon.py
adriesse: Ok, I read this paper too.  Which equation(s) are you implementing?

cwhanse:
@JPalakapillyKWH could you contact me privately at cwhanse@sandia.gov? I have a few a questions I'd appreciate your input.

AdamRJensen:
> Maybe probably needs 3 or 4 separate tasks then?
> 
> 1. either get horizon data or elevation data
> 2. if elevation data, convert it to horizon data
> 3. use horizon data to calculate effect on DHI and DNI (I think? I might be wrong)

Tasks 1 and 3 listed by @mikofski have been addressed in #1395 and #1849. I suggest we close this PR as it is unlikely to get merged and already somewhat addressed by other PRs in the mean time. Thoughts @pvlib/pvlib-core ?

PR #764, opened 2019-08-13
Snow coverage model from SAM

### LINKED ISSUE #577: Implement another snow loss model like Bill Marion as an alt to pvwatts
I'll try to track down the papers on this

JPalakapillyKWH:
- [x] Closes #577 

I implemented the snow coverage model first proposed by Marion et al. in [1] and later validated and implemented in SAM [2]. Thought it might be useful to include in pvlib. 

[1] Marion, B.; Schaefer, R.; Caine, H.; Sanchez, G. (2013). “Measured and modeled photovoltaic
system energy losses from snow for Colorado and Wisconsin locations.” Solar Energy 97; pp.
112-121. 
[2] https://www.nrel.gov/docs/fy17osti/68705.pdf


mikofski:
Yes! Thank you! Can you please search the [issues](../issues) for "snow" or "Martin" to find the ticket number and then add a comment that says, "closes #XYZ" thanks!

JPalakapillyKWH:
> Yes! Thank you! Can you please search the [issues](../issues) for "snow" or "Martin" to find the ticket number and then add a comment that says, "closes #XYZ" thanks!

Edited above comment.

cwhanse:
Due diligence question: is it feasible, and desirable, to wrap the appropriate function in SAM SSC using nrel-pysam? I haven't located the snow model function in the SAM SSC so I don't know if 1) it does what we want, or 2) it is exposed by nrel-pysam.

janinefreeman:
@cwhanse The snow model implementation in SAM is exposed in nrel-pysam via cmod_snowmodel (https://github.com/NREL/ssc/blob/develop/ssc/cmod_snowmodel.cpp) which calls a shared library (https://github.com/NREL/ssc/blob/develop/shared/lib_snowmodel.cpp) that allows it to be integrated on the DC side in the detailed PV model.

wholmgren:
I think there's value in a python implementation (particularly for people that want to play with the source code), I think there's value in pysam exposing the c implementation to python users (consistency with SAM, maybe speed), but what's the point of pvlib wrapping the pysam implementation?

JPalakapillyKWH:
> I think there's value in a python implementation (particularly for people that want to play with the source code), I think there's value in pysam exposing the c implementation to python users (consistency with SAM, maybe speed), but what's the point of pvlib wrapping the pysam implementation?

Agree with this. I was able to fit data better by tuning some of the empirically coefficients in Marion's model.

cwhanse:
> I think there's value in a python implementation (particularly for people that want to play with the source code), I think there's value in pysam exposing the c implementation to python users (consistency with SAM, maybe speed), but what's the point of pvlib wrapping the pysam wrapper?

It is not easy (at least for me it isn't) to figure out how to use pysam to access specific SAM functions, so in cases where we are leveraging SAM functions, I think it is worth wrapping the wrapper for pvlib users' convenience.
In this case I think we should implement the snow model native in python for pvlib. It does not appear to me that pysam exposes the model parameters.

janinefreeman:
That is true, the SAM implementation doesn't expose the model parameters because most of our users do not have the data necessary to tweak those. On a related note, @JPalakapillyKWH if any of your analysis tuning the parameters to the model is publish-able, the industry would definitely appreciate that :)

cwhanse:
@wholmgren do we want PR as it's own module, or bundled with other code into a more general "irradiance_loss" module?  I'm inclined to the former, anticipating a "soiling" module, whatever we call the "iam" module, "shading" module, etc.

wholmgren:
I also favor putting this in its own module. Maybe snow instead snowcoverage? Not sure coverage adds anything and Anton recently reminded us that module names should be brief. 

We should probably think harder about if/when/how to interface with pysam elsewhere.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,229 @@
...
+later validated and implemented into NREL's SAM.
+"""
+
+import numpy as np
+import pandas as pd
+
+
+def snow_slide_amount(surface_tilt, sliding_coefficient=1.97):
+    '''
+    Calculates the amount of snow that slides off of the surface of a module
```
cwhanse: The way I read the papers, the equation in this function calculates 10 x the fraction of a row's slant height that is uncovered by snow sliding in an hour. A row is an array row, which can be more than one module.  The description as written implies the function returns a mass or volume of snow sliding off a single module.
The 10x aspect can be handled inside this function so that it returns a fraction rather than a number of tenths.
The time aspect needs to be added to the calculation, maybe a parameter 'timestep'

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,229 @@
...
+    ----------
+    [1] Marion, B.; Schaefer, R.; Caine, H.; Sanchez, G. (2013).
+    “Measured and modeled photovoltaic system energy losses from snow for
+    Colorado and Wisconsin locations.” Solar Energy 97; pp.112-121.
+    [2] Ryberg, D; Freeman, J. "Integration, Validation, and Application
+    of a PV Snow Coverage Model in SAM" (2017) NREL Technical Report
+    '''
+
+    tilt_radians = np.radians(surface_tilt)
+    slide_amount = sliding_coefficient * np.sin(tilt_radians)
```
cwhanse: slide_amount = (sliding_coefficient / 10) * np.sin(tilt_radians) * time_step (in hours).
I favor putting /10 in the calculation to maintain the meaning of slide_coefficient=1.97 in the references.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,229 @@
...
+
+    time_step_hours: float
+        Period of the data in hours. (hours between data points)
+
+    sliding coefficient : numeric
+        Another empirically determined coefficient used in [1]. It determines
+        how much snow slides off of the panel if a sliding event occurs.
+    m : numeric
+        A coefficient used in the model described in [1]. It is an
+        empirically determined value given in W/m^2.
```
cwhanse: W/(m^2 C)

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,229 @@
...
+    time_adjusted = snowfall / time_step_hours
+    fully_covered_mask = time_adjusted >= 1
+    return fully_covered_mask
+
+
+def snow_coverage_model(snow_data, snow_data_type,
+                        poa_irradiance, temperature, surface_tilt,
+                        time_step_hours=1, m=-80, sliding_coefficient=1.97):
+    '''
+    Calculates the fraction of a module covered by snow at every time step
```
cwhanse: fraction of the slant height of a row of modules

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,229 @@
...
+    following the same model first described in [1] and later
+    implemented in SAM [2]. Currently only implemented for fixed tilt systems
```
cwhanse: I'd say "validated for fixed tilt systems." and leave it at that.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,229 @@
...
+
+    temperature : numeric
+        Ambient air temperature at the surface (C)
+
+    surface_tilt : numeric
+        Surface tilt angles in decimal degrees. Tilt must be >=0 and
+        <=180. The tilt angle is defined as degrees from horizontal
+        (e.g. surface facing up = 0, surface facing horizon = 90).
+
+    time_step_hours: float
```
cwhanse: Could the time_step be obtained from the index of snow_data?

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,229 @@
...
+    '''
+    Calculates the DC loss due to snow coverage. Assumes that if a string is
+    even partially covered by snow, it produces 0W.
+
+    Parameters
+    ----------
+    snow_coverage : numeric
+        The fraction of a module covered by snow at each time step.
+
+    num_strings_per_row: int
```
cwhanse: I'd call this num_strings and define it as the number of parallel-connected strings along a row slant height.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,229 @@
...
+
+def DC_loss_factor(snow_coverage, num_strings_per_row):
+    '''
+    Calculates the DC loss due to snow coverage. Assumes that if a string is
+    even partially covered by snow, it produces 0W.
+
+    Parameters
+    ----------
+    snow_coverage : numeric
+        The fraction of a module covered by snow at each time step.
```
cwhanse: fraction of row slant height

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,229 @@
...
+        Time series data on either snowfall or ground snow depth. The type of
+        data should be specified in snow_data_type. The original model was
+        designed for ground snowdepth only. (cm/hr or cm)
+
+    snow_data_type : string
+        Defines what type of data is being passed as snow_data. Acceptable
+        values are "snowfall" and "snow_depth". "snowfall" will be in units of
+        cm/hr. "snow_depth" is in units of cm.
+
+    poa_irradiance : numeric
```
cwhanse: Both poa_irradiance and temperature are expected to be time-averages over the same intervals as snow_data. Would a Dataframe be useful here to ensure concurrent inputs?

wholmgren: Most pvlib functions take separate inputs that are expected to be consistent. There are pros and cons to that vs dataframe input. For now, I suggest consistency with the existing API.

cwhanse: Agreed. @JPalakapillyKWH perhaps use Series here for the input type since time steps are required for the calculation.

pvlib/snowcoverage.py
JPalakapillyKWH: Could be but maybe not always? I thought it might be better to allow it as an input.

pvlib/snowcoverage.py
JPalakapillyKWH: makes sense

pvlib/snowcoverage.py
cwhanse: but by having a `time_step` input, equal time steps are assumed. If `snow_data` is a Series, then possibly unequal time steps could be accommodated.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,146 @@
...
+       “Measured and modeled photovoltaic system energy losses from snow for
+       Colorado and Wisconsin locations.” Solar Energy 97; pp.112-121.
+    .. [2] Ryberg, D; Freeman, J. "Integration, Validation, and Application
+       of a PV Snow Coverage Model in SAM" (2017) NREL Technical Report
+    '''
+    delta = snowfall.index.to_series().diff()  # [0] will be NaT
+    timestep = delta.dt.seconds.div(3600)  # convert to hours
+    time_adjusted = snowfall / timestep
+    time_adjusted.iloc[0] = 0  # replace NaN from NaT / timestep
+    return time_adjusted >= threshold
```
kanderso-nrel: I wonder if testing `time_adjusted > threshold` (`>` instead of `>=`) would be better.  Choosing `threshold=0` might be appropriate in some cases and using `>=` here would say zero snowfall is sufficient to fully cover the array.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,146 @@
+"""
+The ``snow`` module contains functions that model the effect of snow on
```
kanderso-nrel: Filename is still `snowcoverage.py`

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,146 @@
...
+    .. [1] Marion, B.; Schaefer, R.; Caine, H.; Sanchez, G. (2013).
+       “Measured and modeled photovoltaic system energy losses from snow for
+       Colorado and Wisconsin locations.” Solar Energy 97; pp.112-121.
+    .. [2] Ryberg, D; Freeman, J. "Integration, Validation, and Application
+       of a PV Snow Coverage Model in SAM" (2017) NREL Technical Report
+    '''
+    delta = snowfall.index.to_series().diff()  # [0] will be NaT
+    timestep = delta.dt.seconds.div(3600)  # convert to hours
+    time_adjusted = snowfall / timestep
+    time_adjusted.iloc[0] = 0  # replace NaN from NaT / timestep
```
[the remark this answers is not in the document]
cwhanse: Have to assume the time length of the interval leading to the first step. I'm OK with your suggestion for regular time series, and if irregular perhaps we revert to the no-snow behavior.  Or provide an initial condition as an argument as in #860  `initial_soiling : float, default 0`

pvlib/snowcoverage.py
cwhanse: yea. I need to sort out the name, split up one test, move this into `losses.py`.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,146 @@
...
+    Returns
+    -------
+    snow_coverage : numeric
+        The fraction of a the slant height of a row of modules that is covered
+        by snow at each time step.
+
+    Notes
+    -----
+    Initial snow coverage is assumed to be zero. Implements the model described
+    in [1]_ with minor improvements in [2]_. Validated for fixed tilt systems.
```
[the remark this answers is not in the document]
cwhanse: OK with me. The content and placement of Notes isn't consistent through the library.

pvlib/snowcoverage.py
cwhanse: That's a good point. We'd need to rewrite the description of the `threshold_snowfall` parameter from 
```
    threshold_snowfall : float, default 1.0
        Minimum hourly snowfall to cover a row's slant height. [cm/hr]
```

to

```
    threshold_snowfall : float, default 1.0
        Hourly snowfall above which snow coverage is set to the row's slant height. [cm/hr]
```

pvlib/snowcoverage.py
wholmgren: In retrospect I'm not sure `losses.py` is a great idea. There are losses at many steps of the modeling process and `losses.py` is going to look like the junk drawer of loss models. So I'd be ok sticking with `snow.py` or `snowcoverage.py` in this PR.

mikofski: Hope about a losses package then?
```
*-+- pvlib
  |
  *-+- losses  # put API in __init__.py
    |
    *- snow_coverage.py
    |
    *- soiling.py
```

kanderso-nrel: In principle it might make sense to have soiling and snow as neighbors to IAM, whether that means including `iam.py` in Mark's `losses` package or something else.

wholmgren: I don't think a subpackage solves the problem. We have spectral "corrections" in `atmosphere.py`, AOI "modifiers" in `iam.py`, DC "efficiency" in `pvsystem.py`, inverter "efficiency" in `pvsystem.py` (maybe soon `inverter.py`), temperature "corrections" `temperature.py`. I'm comfortable describing all of these as "losses" of some form.

cwhanse: >  junk drawer of loss models.

Nice metaphor.

Editorial: Let me just state that I think the PVsyst-style "loss diagrams" do somewhat of a disservice, by applying the term "loss" to quantities of energy that couldn't have been generated in the first place. An example that's not a loss (IMO) is "temperature loss"; the physical reality is that cells are less efficient at higher temperature and hot air happens. There's only a "temperature loss" if the PV system could operate at lower temperature but doesn't.

If we follow the modeling process diagram, which is a guide not a rule, a module `dc_losses` could contain functions for wiring and mismatch losses, but not soiling, since soiling applies to irradiance not DC current. Debatable whether `snow_coverage` is an effect on DC current or on irradiance. 

It seems that pvlib will have multiple modules `soiling.py`, `snow.py`, `mismatch.py` or `dc_losses.py` etc. at some level, and the question is about grouping these modules into a `losses` folder, or not. @wholmgren's question suggests we should agree on a list of these modules, and maybe it will become more clear where to place them.

Here's an attempt to list categories for functions that could be broadly termed losses using my view of that term:

- irradiance reflection (iam.py)
- soiling
- snow
- shading (could overlap with DC losses)
- mismatch (cell to cell and module-to-module variation, also irradiance variation?)
- DC wiring
- inverter MPPT efficiency (not part of the current inverter models, but could be)
- tracker accuracy
- grid support functions (e.g., use of reactive power as voltage control)

mikofski: As an alternate to "losses", how about the term "effects" which can be either positive or negative deviations from the irradiance-corrected nominal condition?

cwhanse:
The test failure is odd. A pandas function is returning 'T' indicating minutes, and pandas doesn't accept that as a valid offset code. Fails with pandas v0.23.4 (which is what is installed with our py3.5), not with pandas v0.24.0 or later. I didn't see anything in the pandas release notes that provides a clue. 

You can reproduce the failure condition with pandas v0.23.4 by

```
import pandas as pd
pd.Timedelta('1T')
```

kanderso-nrel:
@cwhanse, what do you think about this vectorized approach for the core function?
``` python
# calculate how much to slide, assuming sliding conditions are met
interval_sliding = sliding_coefficient * sind(surface_tilt) * \
    _time_delta_in_hours(poa_irradiance.index)
# if conditions not met, slide amount is zero
can_slide = temp_air > poa_irradiance / m
interval_sliding[~can_slide] = 0
# assign an ID number to each sliding period, partitioned by coverage events
is_covering_event = snowfall > threshold_snowfall
sliding_period_number = is_covering_event.cumsum()
# calculating cumulative sliding for each sliding period
cumulative_sliding = interval_sliding.groupby(sliding_period_number).cumsum().clip(0, 1)
cumulative_sliding.iloc[0] = 0 # deal with nan
coverage = 1 - cumulative_sliding
```
It avoids the repeated indexing into the snow_coverage series in the current approach's for loop, at least.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+        Tilt of module's from horizontal, e.g. surface facing up = 0,
+        surface facing horizon = 90. Must be between 0 and 180. [degrees]
+    initial_coverage : float, default None
+        Fraction of row's slant height that is covered with snow at the
+        beginning of the simulation. If None (default) then the initial
+        coverage is set to the snowfall in the first time period. [unitless]
+    threshold_snowfall : float, default 1.0
+        Hourly snowfall above which snow coverage is set to the row's slant
+        height. [cm/hr]
+    m : float, default 80.
```
wholmgren: Can we use a more informative parameter name?

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+        Coefficient used in [1]_ to determine if snow can slide given
+        irradiance and air temperature. [W/(m^2 C)]
+    sliding coefficient : float, default 0.197
```
wholmgren:
```suggestion
    sliding_coefficient : float, default 0.197
```

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+        surface facing horizon = 90. Must be between 0 and 180. [degrees]
+    initial_coverage : float, default None
+        Fraction of row's slant height that is covered with snow at the
+        beginning of the simulation. If None (default) then the initial
+        coverage is set to the snowfall in the first time period. [unitless]
+    threshold_snowfall : float, default 1.0
+        Hourly snowfall above which snow coverage is set to the row's slant
+        height. [cm/hr]
+    m : float, default 80.
+        Coefficient used in [1]_ to determine if snow can slide given
```
wholmgren:
```suggestion
        Coefficient to determine if snow can slide given
```

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+        irradiance and air temperature. [W/(m^2 C)]
+    sliding coefficient : float, default 0.197
+        Empirical coefficient used in [1]_ to determine how much
```
wholmgren:
```suggestion
        Coefficient to determine how much
```

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+    .. [1] Marion, B.; Schaefer, R.; Caine, H.; Sanchez, G. (2013).
+       "Measured and modeled photovoltaic system energy losses from snow for
+       Colorado and Wisconsin locations." Solar Energy 97; pp.112-121.
+    .. [2] Ryberg, D; Freeman, J. (2017). "Integration, Validation, and
+       Application of a PV Snow Coverage Model in SAM" NREL Technical Report
+       NREL/TP-6A20-68705
+    '''
+
+    # set up output Series
+    snow_coverage = pd.Series(index=poa_irradiance.index, data=np.nan)
```
wholmgren:
```suggestion
    snow_coverage = pd.Series(np.nan, index=poa_irradiance.index)
```

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+    validation focused on fixed tilt systems.
+
+    Parameters
+    ----------
+    snowfall : Series
+        Accumulated snowfall within each time period. [cm]
+    poa_irradiance : Series
+        Total in-plane irradiance [W/m^2]
+    temp_air : Series
+        Ambient air temperature at the surface [C]
```
wholmgren:
```suggestion
        Ambient air temperature [C]
```

wholmgren: I think "at the surface" will cause more confusion that clarity. I don't think we use this phrase anywhere else in the library.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+    surface_tilt : numeric
+        Tilt of module's from horizontal, e.g. surface facing up = 0,
+        surface facing horizon = 90. Must be between 0 and 180. [degrees]
+    initial_coverage : float, default None
```
wholmgren: why None instead of 0?

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+        _time_delta_in_hours(poa_irradiance.index)
+
+    uncovered = pd.Series(0.0, index=poa_irradiance.index)
+    uncovered[can_slide] = slide_amt[can_slide]
+
+    # build list of intervals between snow fall events
+    snow_events = snowfall[snow_nrel_fully_covered(snowfall,
+                                                   threshold_snowfall)].index
+    # define new snow coverage at each event
+    new_snow = pd.Series(index=snow_events, data=1.)
```
wholmgren:
```suggestion
    new_snow = pd.Series(1., index=snow_events)
```

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+    can_slide = temp_air > poa_irradiance / m
+    slide_amt = sliding_coefficient * sind(surface_tilt) * \
+        _time_delta_in_hours(poa_irradiance.index)
+
+    uncovered = pd.Series(0.0, index=poa_irradiance.index)
+    uncovered[can_slide] = slide_amt[can_slide]
+
+    # build list of intervals between snow fall events
+    snow_events = snowfall[snow_nrel_fully_covered(snowfall,
+                                                   threshold_snowfall)].index
```
wholmgren: Hard to follow a function called within an indexer. It will also make the traceback a lot more complicated if there's an exception.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+        new_snow[snowfall.index[-1]] = 0.
+
+    windows = list(zip(new_snow.index[:-1], new_snow.index[1:]))
+    for (ev, ne) in windows:
+        filt = (snow_coverage.index > ev) & (snow_coverage.index <= ne)
+        snow_coverage[ev] = new_snow[ev]
+        snow_coverage[filt] = new_snow[ev] - uncovered[filt].cumsum()
+
+    # clean up periods where row is completely uncovered
+    snow_coverage[snow_coverage < 0] = 0
```
wholmgren:
```suggestion
    snow_coverage = snow_coverage.clip(lower=0)
```

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+    if initial_coverage is None:
+        initial_coverage = 0.
+
+    # determine amount that snow can slide in each timestep
+    can_slide = temp_air > poa_irradiance / m
+    slide_amt = sliding_coefficient * sind(surface_tilt) * \
+        _time_delta_in_hours(poa_irradiance.index)
+
+    uncovered = pd.Series(0.0, index=poa_irradiance.index)
+    uncovered[can_slide] = slide_amt[can_slide]
```
wholmgren: It's not immediately obvious to me what's happening here and why the variable is called `uncovered`.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+import pandas as pd
+from pvlib.tools import sind
+
+
+def _time_delta_in_hours(times):
+    delta = times.to_series().diff()
+    return delta.dt.total_seconds().div(3600)
+
+
+def snow_nrel_fully_covered(snowfall, threshold=1.):
```
wholmgren: if the module name is `snow`, then we can drop `snow_` from the function names (my preference, as discussed above). Obviously they will need to stay if the functions move to `losses.py`.

cwhanse:
```
# assign an ID number to each sliding period, partitioned by coverage events
is_covering_event = snowfall > threshold_snowfall
sliding_period_number = is_covering_event.cumsum()
# calculating cumulative sliding for each sliding period
cumulative_sliding = interval_sliding.groupby(sliding_period_number).cumsum().clip(0, 1)
```
> It avoids the repeated indexing into the snow_coverage series in the current approach's for loop, at least.

That's a good idea. I'll figure out how to combine this with using the first snowfall data point, and an (optional) initial coverage fraction.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+    snow_coverage : numeric
+        The fraction of row slant height covered by snow at each time step.
+
+    num_strings: int
+        The number of parallel-connected strings along a row slant height.
+
+    Returns
+    -------
+    loss : numeric
+        fraction of DC capacity loss due to snow coverage at each time step.
```
wholmgren: I found this docstring hard to understand and thought it would benefit from a short Examples section.

cwhanse:
@kanderso-nrel I'm going to let you work the vectorization. The approach above disagrees with other implementations, in that it allows sliding in the same time step as new snowfall accumulates. As a consequence, snow coverage = 1.0 isn't possible except at the very first time step. The loop is definitely inefficient, but the tests are passing which gives us a standard for working performance improvements.

pvlib/snowcoverage.py
```diff
@@ -0,0 +1,170 @@
...
+    temp_air : Series
+        Ambient air temperature at the surface [C]
+    surface_tilt : numeric
+        Tilt of module's from horizontal, e.g. surface facing up = 0,
+        surface facing horizon = 90. Must be between 0 and 180. [degrees]
+    initial_coverage : float, default None
+        Fraction of row's slant height that is covered with snow at the
+        beginning of the simulation. If None (default) then the initial
+        coverage is set to the snowfall in the first time period. [unitless]
+    threshold_snowfall : float, default 1.0
```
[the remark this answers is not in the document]
cwhanse: Yes, both `threshold_snowfall` now

kanderso-nrel:
Ah interesting.  I'll take a look and try to get it to pass your tests, thanks.  Probably won't get to it before tonight or this weekend.

pvlib/snowcoverage.py
cwhanse: `can_slide_coefficient`

pvlib/snowcoverage.py
cwhanse: Idk now, was in the middle of figuring out how to implement this capability

pvlib/snowcoverage.py
cwhanse: I agree.

Do we want to hold this PR until we bring the losses discussion to a conclusion?  OK with me to wait here.

kanderso-nrel:
I think this passes the current set of tests.  Does it make sense to always prevent sliding for the first timestamp?  I guess I'm not clear on whether `initial_coverage` is reflects coverage at the beginning or the end of the first interval.  And have we abandoned `If None (default) then the initial coverage is set to the snowfall in the first time period`?

<details>

    def snow_nrel(snowfall, poa_irradiance, temp_air, surface_tilt,
                  initial_coverage=0, threshold_snowfall=1.,
                  can_slide_coefficient=-80., slide_amount_coefficient=0.197):
    
    
        # assign an ID number to each sliding period partitioned by coverage events
        is_covering_event = snow_nrel_fully_covered(snowfall, threshold_snowfall)
        sliding_period_ID = is_covering_event.cumsum()
    
        # calculate incremental sliding, assuming sliding conditions are met
        interval_sliding = slide_amount_coefficient * sind(surface_tilt) * \
            _time_delta_in_hours(poa_irradiance.index)
        # if sliding conditions not met, slide amount is zero
        can_slide = temp_air > poa_irradiance / can_slide_coefficient
        interval_sliding[~can_slide] = 0
        # if snowing, slide amount is zero
        interval_sliding[is_covering_event] = 0
        # don't allow sliding on the first interval
        interval_sliding.iloc[0] = 0
    
        # calculate cumulative sliding in each sliding period
        cumulative_sliding = interval_sliding.groupby(sliding_period_ID).cumsum()
    
        # overall coverage = (baseline - slide amount).  baseline is 1.0 for new
        # snow, but the first (ID=0) baseline can be overridden:
        coverage = pd.Series(1, index=poa_irradiance.index)
        coverage[sliding_period_ID == 0] = initial_coverage
        coverage -= cumulative_sliding
        return coverage.clip(0, 1)

</details>

cwhanse:
> I think this passes the current set of tests.  Does it make sense to always prevent sliding for the first timestamp?  I guess I'm not clear on whether initial_coverage is reflects coverage at the beginning or the end of the first interval.

Snowfall is defined in the code and in the reference paper as the depth accumulated at the end of each time interval. Snowfall either covers the full slant height (if the threshold is exceeded), or has no effect. If the snowfall timeseries has a frequency, the first snowfall value is checked if it is a full coverage event. If the first snowfall is not a full coverage event, then the covered slant height is set to `initial_coverage` which is a fraction of module slant height.

>  And have we abandoned If None (default) then the initial coverage is set to the snowfall in the first time period?
>
No

pvlib/snow.py
```diff
@@ -0,0 +1,182 @@
...
+        coverage is set to the snowfall in the first time period. [unitless]
+    threshold_snowfall : float, default 1.0
+        Hourly snowfall above which snow coverage is set to the row's slant
+        height. [cm/hr]
+    can_slide_coefficient : float, default -80.
+        Coefficient to determine if snow can slide given irradiance and air
+        temperature. [W/(m^2 C)]
+    slide_amount_coefficient : float, default 0.197
+        Coefficient to determine fraction of snow that slides off in each
+        time period. [unitless]
```
kanderso-nrel: This parameter is assumed to be hourly in the code, perhaps that should be mentioned here.

pvlib/snow.py
```diff
@@ -0,0 +1,182 @@
...
+       Colorado and Wisconsin locations." Solar Energy 97; pp.112-121.
+    .. [2] Ryberg, D; Freeman, J. "Integration, Validation, and Application
+       of a PV Snow Coverage Model in SAM" (2017) NREL Technical Report
+    '''
+    timestep = _time_delta_in_hours(snowfall.index)
+    hourly_snow_rate = snowfall / timestep
+    # if we can infer a time frequency, use first snowfall value
+    # otherwise the first snowfall value is ignored
+    freq = pd.infer_freq(snowfall.index)
+    freq = freq.upper().replace('T', 'M')  # for pandas <= 0.23.4
```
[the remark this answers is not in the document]
wholmgren: might be able to avoid this hack and the `len(freq) == 1` hack by using `to_offset`. For example:

```python
from pandas.tseries.frequencies import to_offset
index = pd.date_range(start="2019-1-1 12:00:00", end="2019-1-1 18:00:00", freq='1h')
offset = to_offset(pd.infer_freq(index))
hr_norm = offset / pd.Timedelta('1h')
```

I've not thoroughly tested against pandas versions.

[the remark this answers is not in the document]
cwhanse: Works for pandas 0.23.4 and later, thanks!

pvlib/snow.py
```diff
@@ -0,0 +1,182 @@
...
+    ----------
+    snowfall : Series
+        Accumulated snowfall within each time period. [cm]
+    poa_irradiance : Series
+        Total in-plane irradiance [W/m^2]
+    temp_air : Series
+        Ambient air temperature [C]
+    surface_tilt : numeric
+        Tilt of module's from horizontal, e.g. surface facing up = 0,
+        surface facing horizon = 90. Must be between 0 and 180. [degrees]
```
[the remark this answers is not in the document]
cwhanse: Agreed, copy and paste from somewhere, I'd guess

cwhanse:
Test failures are for pvlib.bifacial.pvfactors. I believe this PR is ready.

docs/sphinx/source/api.rst
```diff
@@ -308,25 +308,33 @@ Functions for fitting diode models
...
 
 .. autosummary::
    :toctree: generated/
 
-   pvsystem.retrieve_sam
-   pvsystem.systemdef
-   pvsystem.scale_voltage_current_power
+   snow.coverage_nrel
+   snow.fully_covered_nrel
+   dc_loss_nrel
```
kanderso-nrel:
```suggestion
   snow.dc_loss_nrel
```

docs/sphinx/source/whatsnew/v0.7.2.rst
```diff
@@ -22,6 +22,9 @@ Enhancements
 * Add Kimber soiling model :py:func:`pvlib.losses.soiling_kimber`. (:pull:`860`)
 * Add :func:`~pvlib.iotools.read_pvgis_tmy` for files downloaded using the
   PVGIS tool. (:issue:`880`)
+* Add new module :py:mod:`pvlib.snow` to contain models related to snow coverage and effects on a PV system. (:pull:`764`)
+* Add snow coverage model :py:func:`pvlib.snow.coverage_nrel` and function to identify when modules are fully covered by snow :py:func:`pvlib.snow.fully_covered_nrel`. (:issue:`577`)
+* Add function :py:func:`pvlib.snow.dc_loss_snow` for effect of snow coverage on DC output. (:pull:`764`)
```
kanderso-nrel:
```suggestion
* Add function :py:func:`pvlib.snow.dc_loss_nrel` for effect of snow coverage on DC output. (:pull:`764`)
```

pvlib/snow.py
```diff
@@ -0,0 +1,182 @@
+"""
+The ``snow`` module contains functions that model module snow cover and the
+associated effects on PV module output"
```
kanderso-nrel:
```suggestion
associated effects on PV module output
```

pvlib/snow.py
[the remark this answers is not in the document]
kanderso-nrel: FYI I don't think this suggestion made it into the code.

pvlib/snow.py
```diff
@@ -0,0 +1,182 @@
...
+    if np.isnan(snow_coverage.iloc[0]):
+        snow_coverage.iloc[0] = initial_coverage
+    snow_coverage.ffill(inplace=True)
+    snow_coverage -= cumulative_sliding
+
+    # clean up periods where row is completely uncovered
+    return snow_coverage.clip(lower=0)
+
+
+def dc_loss_nrel(snow_coverage, num_strings):
```
kanderso-nrel: Does this function's docstring need a reference?

kanderso-nrel:
I included a comment in that review but I think it got buried back in the thread history, so I'll repost it -- I think this suggestion didn't make it into the PR: https://github.com/pvlib/pvlib-python/pull/764#discussion_r392282059

pvlib/snow.py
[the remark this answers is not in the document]
cwhanse: It should have, I've got something out of synch between my different computers.

pvlib/snow.py
cwhanse: Yes

CameronTStark:
Looks great! I appreciate your commitment on this @cwhanse. 

I'll leave it open for a little while longer for others to review, if desired, then I'll merge.

Thank you!

PR #773, opened 2019-08-23
ModelChain 0.7 deprecations, remove times kwarg

### LINKED ISSUE #706: Remove times kwarg from ModelChain.prepare_inputs and ModelChain.run_model
The `ModelChain.run_model` and `ModelChain.prepare_inputs` methods `times` kwarg is nearly pointless. I'm guessing/hoping that the vast majority of `ModelChain` users supply a `weather` DataFrame with a meaningful `DatetimeIndex`. So, it would greatly simplify the API if we removed the `times` kwarg and inspected  `weather.index` whenever we need timestamp info (e.g. solar position). `ModelChain` is committed to time series modeling using pandas, so I think it's reasonable to require users to supply a meaningful index. 

Current signatures:

```python
    def prepare_inputs(self, times=None, weather=None):
    def run_model(self, times=None, weather=None):
```

proposed signatures:

```python
    def prepare_inputs(self, weather):
    def run_model(self, weather):
```

wholmgren:
- [x] Closes #706 

Also removes deprecated 'singlediode' option.

Also fixes incorrect documentation about what happens if `weather=None` or if `weather` does not have air temperature or wind speed.

I'll wait to update the whatsnew until after #772


cwhanse:
Probably should delete [this ](https://github.com/pvlib/pvlib-python/blob/a873462eb079f7d57cc8c2e5a9cf447ac68027d1/pvlib/modelchain.py#L483) method

pvlib/modelchain.py
```diff
@@ -746,7 +720,7 @@ def effective_irradiance_model(self):
             fd*self.total_irrad['poa_diffuse'])
         return self
 
-    def complete_irradiance(self, times=None, weather=None):
+    def complete_irradiance(self, weather, times=None):
```
cwhanse: Does the change in argument order need to be mentioned in whatsnew?  Same comment on `prepare_inputs`, `run_model`.

docs/sphinx/source/whatsnew/v0.7.0.rst
```diff
@@ -71,6 +71,11 @@ API Changes
 
 * Calling :py:func:`pvlib.pvsystem.retrieve_sam` with no parameters will raise
   an exception instead of displaying a dialog.
+* The `times` keyword argument has been deprecated in the
+  :py:meth:`pvlib.modelchain.ModelChain.run_model`,
+  :py:meth:`pvlib.modelchain.ModelChain.prepare_inputs`, and
+  :py:meth:`pvlib.modelchain.ModelChain.complete_irradiance` methods.
+  Model times are now determined by the input weather DataFrame index.
```
cwhanse: Add a note that `weather` is now required, and that `weather` must have a DatetimeIndex.

pvlib/modelchain.py
```diff
@@ -782,19 +756,23 @@ def complete_irradiance(self, times=None, weather=None):
...
+        >>> mc.run_model(my_weather)  # doctest: +SKIP
         """
-        if weather is not None:
-            self.weather = weather
+        self.weather = weather
+
         if times is not None:
-            self.times = times
+            warnings.warn('times keyword argument is deprecated and will be '
+                          'removed in 0.8. Set the index of the weather '
```
cwhanse: "The index of the weather DataFrame is used for times."  Same comment on `prepare_inputs`

pvlib/modelchain.py
```diff
@@ -782,19 +756,23 @@ def complete_irradiance(self, times=None, weather=None):
...
+                          'argument instead.', pvlibDeprecationWarning)
+
         self.solar_position = self.location.get_solarposition(
-            self.times, method=self.solar_position_method)
+            self.weather.index, method=self.solar_position_method)
```
cwhanse: We're not using `times` in any fashion so we're not really backward compatible, just providing a warning message and then moving ahead with new behavior.  More of a comment on the docstring, but if we want to maintain backward compatibility, we should use `times`, and warn if `times` conflicts with `weather.index`.   Same comment on `prepare_inputs`

wholmgren: Right, the removal of the deprecated clear sky assumption makes the `times` input superfluous. I'm not opposed to keeping the `times` attribute so long as it's set like `self.times = weather.index`. That's probably the easiest thing to do. We could deprecate the attribute, but I don't think it matters either way. I don't think it makes sense to use `times` to calculate solar position because we'll end up with solar position that's inconsistent with `weather`.

pvlib/modelchain.py
wholmgren: Doesn't hurt, but I think it's considered bad form to rely on the position of keyword arguments in code that you don't control.

wholmgren:
> In modelchain.py the Example described in lines 751-753 needs to be updated.

In `complete_irradiance`? Looks ok to me.

cwhanse:
My only remaining concern, I think we should explicitly tell the user that the `times` argument is no longer used.

PR #780, opened 2019-09-13
Correting times in modelchain

matthiasLevy:
In modelchain, when calculating `complete_irradiance`, the times `self.times` where not set to `weather.index` so the output could output wrong index.
This is my first contribution to the project, please excuse any misunderstanding.


cwhanse:
@matthiasLevy thanks for the contribution.  We're actually in process of removing the `times` kwarg in #773, see #706 for context.  Your review of #773 is welcome, as I think it addresses the problem you describe.  @wholmgren

PR #784, opened 2019-10-02
coefficient estimation method following DeSoto(2006)

tylunel:

Brief description of the problem and proposed solution:
This function computes the 5 sdm parameters coefficients following the procedure proposed by W. De Soto et al., "Improvement and validation of a model for photovoltaic array performance", Solar Energy, vol80, pp. 78-88, 2006.
In my opinion the function has the advantage to be transparent for python users and can be used with standard specifications found in manucfacturer datasheets.

Please just state your interest or not in this function and, if you are interested, I will complete it so as to check all the conditions then.

This is my first pull request so don't hesitate to give me any feedback you think useful, thanks !


cwhanse:
> Please just state your interest or not in this function and, if you are interested, I will complete it so as to check all the conditions then.

thanks @tylunel  this will be a nice addition to the library. I'm OK with the general approach. Please let us know when you are ready for review.

pvlib/ivtools.py
```diff
@@ -262,6 +268,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        The short-circuit current (i_sc) temperature coefficient of the
+        module in units of %/K. It is converted in A/K for the computing
+        process.
+    beta_voc : numeric
+        The open-circuit voltage (v_oc) temperature coefficient of the
+        module in units of %/K. It is converted in V/K for the computing
+        process.
+    cells_in_series : numeric, default 60
+        Number of cell in the module.
+        Optional input, but helps to insure the convergence of the computing.
```
markcampanelli: Since you know `celltype`, you might try making the default value here `None`, in which case you estimate `cells_in_series` by dividing `v_oc` by a "nominal" `v_oc` for a single cell of the given `celltype`.

pvlib/ivtools.py
```diff
@@ -262,6 +268,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    ----------
+    [1] W. De Soto et al., "Improvement and validation of a model for
+    photovoltaic array performance", Solar Energy, vol 80, pp. 78-88,
+    2006.
+
+    [2] John A Dufﬁe ,William A Beckman, "Solar Engineering of Thermal
+    Processes", Wiley, 2013
+    """
+    # Constants
+    k = 1.381e-23  # Boltzmann constant, in J/K, or 8.617e-5 eV/K
```
markcampanelli: Note #483 for establishing a single source of truth for constants.

pvlib/ivtools.py
```diff
@@ -262,6 +268,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        a = params[2]
+        Rsh = params[3]
+        Rs = params[4]
+        # five equation vector
+        y = [0, 0, 0, 0, 0]
+
+        # 1st equation - short-circuit - eq(3) in [1]
+        y[0] = Isc - IL + Io*(np.exp(Isc*Rs/a) - 1.0) + Isc*Rs/Rsh
+        # 2nd equation - open-circuit Tref - eq(4) in [1]
+        y[1] = -IL + Io*(np.exp(Voc/a) - 1.0) + Voc/Rsh
```
markcampanelli: Consider using `numpy`'s [`expm1`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.expm1.html).

pvlib/ivtools.py
```diff
@@ -262,6 +268,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        IL = params[0]
+        Io = params[1]
+        a = params[2]
+        Rsh = params[3]
+        Rs = params[4]
+        # five equation vector
+        y = [0, 0, 0, 0, 0]
+
+        # 1st equation - short-circuit - eq(3) in [1]
+        y[0] = Isc - IL + Io*(np.exp(Isc*Rs/a) - 1.0) + Isc*Rs/Rsh
```
markcampanelli: Consider using [`numpy.expm1`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.expm1.html) here and throughout.

pvlib/ivtools.py
```diff
@@ -262,6 +268,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+
+        return y
+
+    # initial guesses of variables for computing convergence:
+    # Values are taken from [2], p753
+    Rsh_i = 100.0
+    a_i = 1.5*k*Tref*cells_in_series/q
+    IL_i = i_sc
+    Io_i = i_sc * np.exp(-v_oc/a_i)
+    Rs_i = (a_i*np.log((IL_i-i_mp)/Io_i+1) - v_mp)/i_mp
```
markcampanelli: Consider using [`numpy.logp1`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.log1p.html).

pvlib/ivtools.py
```diff
@@ -262,6 +268,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    # params_i : initial values vector
+    params_i = np.array([IL_i, Io_i, a_i, Rsh_i, Rs_i])
+
+    # specs of module
+    specs = np.array([i_sc, v_oc, i_mp, v_mp, beta_voc, alpha_sc])
+
+    # computing
+    sol = root(pv_fct, x0=params_i, args=specs, method='lm').x
```
markcampanelli: Check the [`scipy.optimize.OptimizeResult`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html#scipy.optimize.OptimizeResult) for success. I think the community decided to raise exceptions when fitting fails, e.g., [this example](https://github.com/pvlib/pvlib-python/blob/master/pvlib/ivtools.py#L107) and its corresponding tests.

pvlib/ivtools.py
```diff
@@ -262,6 +268,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    """
+    # Constants
+    k = 1.381e-23  # Boltzmann constant, in J/K, or 8.617e-5 eV/K
+    q = 1.602e-19  # electron charge in J/V, or 1 eV
+
+    Tref = temp_ref + 273.15  # in K
+
+    if celltype.lower() in ['monosi', 'polysi', 'multisi']:
+        C = 0.0002677  # valid for silicon
+        Eg = 1.796e-19  # in J, valid for silicon
```
markcampanelli: Is `C` the same as `dEgdT` in `calcparams_desoto` (up to sign?)? If yes, then use a consistent variable and name. Same with `Eg` and `EgRef` in `calcparams_desoto`.

pvlib/ivtools.py
```diff
@@ -262,6 +268,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    # computing
+    sol = root(pv_fct, x0=params_i, args=specs, method='lm').x
+
+    # results
+    return {'I_L_ref': sol[0],
+            'I_o_ref': sol[1],
+            'a_ref': sol[2],
+            'R_sh_ref': sol[3],
+            'R_s': sol[4],
+            'alpha_sc': alpha_sc}
```
markcampanelli: My personal preference is to also record `EgRef`, `dEgdT`, `irrad_ref`, and `temp_ref` in the output, in order to fully capture the conditions of the fit. Ideally, these can then be passed straight through to `calcparams_desoto`, ensuring consistency.

`irrad_ref` should probably also be added as an input with a default value of `1000`. It's simply a pass through value, but I think it is an "implicit" part of the fit that the user should be aware of, and it corresponds to the default value in `calcparams_desoto`.

markcampanelli:
Thanks for tackling this. I provided some early feedback, which I hope helps.

pvlib/ivtools.py
cwhanse: I think it is safe to assume that `cells_in_series` is known. Should be required input.

markcampanelli: In this case I think it’s better to require explicitness here, and not provide a default value.

pvlib/ivtools.py
tylunel: Thank you for the feedback.
Is numpy.expm1 (and numpy.logp1 in the same way) prefered for computing efficiency reasons ? because I feel it makes the equation less readable for the standard user who is not an expert in numpy

cwhanse: Rationale [here](https://stackoverflow.com/questions/47725508/when-to-use-python-function-expm1-rather-than-exp-1) . Here,  `np.expm1` is preferred when `np.exp(Isc*Rs/a)` is very close to 1 (`Isc*Rs/a` is nearly zero), which is rarely of interest when predicting output using a single diode model, but may be important in this optimization that is attempting to determine the parameters.

pvlib/ivtools.py
tylunel: Good question. And these kind of parameter can easily result. For example if the user gives alpha_sc in A/K instead of %/K it will result in many cases in a negative resistance.

And I wonder if these negative resistances are really impossible ? I mean, as the single diode model is a model simplifying the complexity of the architecture of the semi-conductors layers in a PV cell, maybe a negative resistance could be the translation in the model of a physical interaction (generally minor in standard modules but more important in the module being calculated). Would it be possible that keeping this negative resistance would still improve the prediction made by the single diode model ? And finally, it would rather mean that SDM is not suitable for this particular module. 

But still, as what I say before must be rare, I think that a warning would be good to make the user check the units of the inputs.

pvlib/ivtools.py
tylunel: Yes, that's it. 
You are right, I will do that

tylunel:
I don't fully understand where do the 8 failing checks come from, but I think it is not directly related to my PR. They seem to be the same than those described by cwhanse in here : https://github.com/pvlib/pvlib-python/pull/783#issuecomment-538537240

So I think my pull request is ready for review now,  I will welcome any feedback you can do.

cwhanse:
@tylunel thanks very much. I will get to the review tomorrow or the day after. You can ignore the failure for `test_psm3`, and the failure for py35_min indicates that some tests need a `requires_scipy` mark. I can show you how to add that in the review.

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    temp_ref: numeric, default 25
+        Reference temperature condition [C]
+    irrad_ref: numeric, default 1000
+        Reference irradiance condition [W/m2]
+
+    Returns
+    -------
+    Dictionnary with the following elements:
+
+    * 'I_L_ref': numeric
```
cwhanse: Remove the 's from the parameters.  Have you checked how the docstring renders for readthedocs?  You have to go to readthedocs.org, sign in, and link to your pvlib-python project. You should then see a list of versions that has this branch. Click the version (branch) and it should take you to a readthedocs with your PR included.

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        Light-generated current in amperes at std conditions.
```
cwhanse: "at reference conditions. [A]."  Same comment to lines below

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        Number of cell in the module.
+        Optional input, but helps to insure the convergence of the computing.
+    temp_ref: numeric, default 25
+        Reference temperature condition [C]
+    irrad_ref: numeric, default 1000
+        Reference irradiance condition [W/m2]
+
+    Returns
+    -------
+    Dictionnary with the following elements:
```
cwhanse: spelling

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        dEgdT = -0.0002677  # valid for silicon
+        EgRef = 1.796e-19  # in J, valid for silicon
+    elif celltype.lower() in ['cis', 'cigs', 'cdte', 'amorphous', 'thin film']:
+        raise NotImplementedError
+    else:
+        raise ValueError('Unknown cell type.')
+
+    # Conversion from %/K to A/K & V/K
+    alpha_sc = alpha_sc*i_sc/100
+    beta_voc = beta_voc*v_oc/100
```
cwhanse: These lines are removed if the comments above are accepted.

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+
+    def pv_fct(params, specs):
+        """Returns the system of equations used for computing the
+        single-diode 5 parameters.
```
cwhanse: "Evaluates the systems of equations used to solve for the single diode equation parameters."

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    [1] W. De Soto et al., "Improvement and validation of a model for
+    photovoltaic array performance", Solar Energy, vol 80, pp. 78-88,
+    2006.
+
+    [2] John A Dufﬁe ,William A Beckman, "Solar Engineering of Thermal
+    Processes", Wiley, 2013
+    """
+    # Constants
+    k = constants.Boltzmann  # in J/K, or 8.617e-5 eV/K
+    q = constants.elementary_charge  # in J/V, or 1 eV
```
cwhanse: `q = 1  # in eV` if units of eV are adopted.

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    ----------
+    [1] W. De Soto et al., "Improvement and validation of a model for
+    photovoltaic array performance", Solar Energy, vol 80, pp. 78-88,
+    2006.
+
+    [2] John A Dufﬁe ,William A Beckman, "Solar Engineering of Thermal
+    Processes", Wiley, 2013
+    """
+    # Constants
+    k = constants.Boltzmann  # in J/K, or 8.617e-5 eV/K
```
cwhanse: `k = constants.value('Boltzmann constant in eV/K')` for the Boltzmann constant in eV/K.

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    The six values returned by this function can be used by
+    pvsystem.calcparams_desoto to calculate the values at different
+    irradiance and cell temperature.
+
+    Parameters
+    ----------
+    celltype: str, case insensitive
+        Value is one of 'monosi', 'multisi', 'polysi' or'gaas'.
+        Others like 'cis', 'cigs', 'cdte', 'amorphous' are not implemented yet
+    v_mp: numeric
```
cwhanse: type must be `float`. `numeric` implies that I could input a vector of `v_mp` values, and I'm pretty certain that `scipy.optimize.root` will not solve if `v_mp` etc. are vectors.

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    Io_i = i_sc * np.exp(-v_oc/a_i)
+    Rs_i = (a_i*np.log1p((IL_i-i_mp)/Io_i) - v_mp)/i_mp
+    # params_i : initial values vector
+    params_i = np.array([IL_i, Io_i, a_i, Rsh_i, Rs_i])
+
+    # specs of module
+    specs = np.array([i_sc, v_oc, i_mp, v_mp, beta_voc, alpha_sc])
+
+    # computing
+    result = root(pv_fct, x0=params_i, args=specs, method='lm')
```
cwhanse: I would elevate `method` to an kwarg for `fit_sdm_desoto`, default is OK.

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
                                      v_oc)
 
 
+def fit_sdm_desoto(celltype, v_mp, i_mp, v_oc, i_sc, alpha_sc, beta_voc,
+                   cells_in_series, temp_ref=25, irrad_ref=1000):
+    """
+    Calculates the five parameters for the single diode equation using
```
cwhanse: "Calculates the parameters for the De Soto single diode model using the procedure described in [1]".

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    the De Soto et al. procedure described in [1]. This procedure has the
+    advantage of using common specifications given by manufacturers in the
+    datasheets of PV modules.
+    The six values returned by this function can be used by
```
cwhanse: The parameters returned by this function

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+
+        # five equation vector
+        y = [0, 0, 0, 0, 0]
+
+        # 1st equation - short-circuit - eq(3) in [1]
+        y[0] = Isc - IL + Io*np.expm1(Isc*Rs/a) + Isc*Rs/Rsh
+        # 2nd equation - open-circuit Tref - eq(4) in [1]
+        y[1] = -IL + Io*np.expm1(Voc/a) + Voc/Rsh
+        # 3rd equation - Imp & Vmp - eq(5) in [1]
+        y[2] = Imp - IL + Io*np.exp((Vmp+Imp*Rs)/a) + \
```
cwhanse: `np.exp1m` here

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+            (Vmp+Imp*Rs)/Rsh
+        # 4th equation - Pmp derivated=0 -
+        # caution: eq(6) in [1] seems to be incorrect, take eq23.2.6 in [2]
```
cwhanse: Concur, Eq. 6a in [1] has a sign error.

pvlib/ivtools.py
```diff
@@ -262,6 +264,177 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    advantage of using common specifications given by manufacturers in the
+    datasheets of PV modules.
+    The six values returned by this function can be used by
+    pvsystem.calcparams_desoto to calculate the values at different
+    irradiance and cell temperature.
+
+    Parameters
+    ----------
+    celltype: str, case insensitive
+        Value is one of 'monosi', 'multisi', 'polysi' or'gaas'.
```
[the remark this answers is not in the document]
tylunel: Ok, then using celltype allowed to be consistent with fit_sdm_cec_sam(). So we should maybe change it in fit_sdm_cec_sam() as well.
Using eV more than J make it inconsistent with scipy.constants (#483), is it still ok ?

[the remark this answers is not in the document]
cwhanse: `scipy.constants` has the value in eV/K, but have to get it using a key rather than as `scipy.constants.Boltzmann`. The statement is in a comment below. So I think it's OK for #483 .

The SAM function requires celltype as an input, and SAM has an internal lookup table for Egref etc.  I'd prefer to drop celltype and pass Egref but that requires a change to SAM.  Your point about consistent signatures is well-taken; feel free to adjust the argument order for fit_sdm_cec_sam() to match, as best we can.

pvlib/ivtools.py
tylunel: hmmm not sure I can. Do you mean by factorizing with IL in something like 
IL * ((Io/IL)*np.exp(Vmp+Imp*Rs)/a) - 1) ?
In this case the factor Io/IL prevents from using np.expm1 in my opinion

Or do you mean that I misspelled 'exp1m' in 'expm1'. I 've only found 'expm1' in numpy doc

cwhanse: The misspelling is mine. Unless I'm wrong, the equation on line 387 is missing a " -1" term, should be `Imp - IL + Io * (np.exp((Vmp + Imp * Rs) / a) - 1) + \`

tylunel: Oh you are right !! Thanks for that

pvlib/ivtools.py
```diff
@@ -262,6 +264,163 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    EgRef: float, default 1.121 eV - value for silicon
+        Energy of bandgap of semi-conductor used [eV]
+    dEgdT: float, default -0.0002677 - value for silicon
+        Variation of bandgap according to temperature [eV/K]
+    temp_ref: float, default 25
+        Reference temperature condition [C]
+    irrad_ref: float, default 1000
+        Reference irradiance condition [W/m2]
+    solver_method: str, default 'lm'
+        See help of scipy.optimize.root() for more details.
```
markcampanelli: This gives me some pause. None of the tests cover the numerous other choices of solver. Are they all compatible?

What happens if you loop through each of the 9 other `root` solvers in the tests?

pvlib/test/test_ivtools.py
```diff
@@ -102,6 +109,29 @@ def test_fit_sdm_cec_sam(get_cec_params_cansol_cs5p_220p):
                 cells_in_series=1, temp_ref=25)
 
 
+@requires_scipy
+def test_fit_sdm_desoto(get_test_specs_params):
+    result = ivtools.fit_sdm_desoto(cells_in_series=60,
```
markcampanelli: This is where you might test "all" the `root` solvers:
```python
for solver_method in (‘hybr’, ‘lm’, ‘broyden1’, ‘broyden2’,‘anderson’,‘linearmixing’, ‘diagbroyden’, ‘excitingmixing’, ‘krylov’, ‘df-sane’):
```
However, this is somewhat fragile w.r.t. scipy updating the various solvers.

cwhanse: We could restrict the `method` values, if we decide we're not testing/supporting some of the options.

pvlib/ivtools.py
```diff
@@ -6,6 +6,8 @@
 """
 
 import numpy as np
+from scipy.optimize import root
```
cwhanse: The `py35_min` environment purposefully omits scipy. To get the tests to complete, move the `from scipy import …` statements into `fit_sdm_desoto`.  You could include a `try/except` clause like [here ](https://github.com/pvlib/pvlib-python/blob/a873462eb079f7d57cc8c2e5a9cf447ac68027d1/pvlib/solarposition.py#L881) to more gently handle ImportErrors.

pvlib/test/test_ivtools.py
tylunel: Good point. 
I did not explore that before, but I realize that only two of the methods proposed in scipy.optimize.root() actually work for fit_sdm_desoto, namely 'hybr' and 'lm'.

If we don't mention the solver_method of scipy.optimize.root(), it may be less fragile maybe ? I propose to remove the kwarg solver_method and other later reference of it. This way, scipy.optimize.root() will just choose the default value (currently 'hybr'), and if scipy is updated, we can expect scipy to still use a robust method as default value.

And moreover it would simplify a bit the code of fit_sdm_desoto. I don't think that being able to choose the rooting method is a very important feature for most users of pvlib. What do you think ?

markcampanelli: I would be ok with using the default solver.

pvlib/ivtools.py
```diff
@@ -262,6 +264,163 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    # specs of module
+    specs = np.array([i_sc, v_oc, i_mp, v_mp, beta_voc, alpha_sc])
+
+    # computing
+    result = root(pv_fct, x0=params_i, args=specs, method=solver_method)
+
+    if result.success:
+        sdm_params = result.x
+    else:
+        raise RuntimeError('Parameter estimation failed')
```
markcampanelli: Maybe include the full `result` here in the exception for better debugging?

pvlib/ivtools.py
```diff
@@ -262,6 +264,163 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    return {'I_L_ref': sdm_params[0],
+            'I_o_ref': sdm_params[1],
+            'a_ref': sdm_params[2],
+            'R_sh_ref': sdm_params[3],
+            'R_s': sdm_params[4],
+            'alpha_sc': alpha_sc,
+            'EgRef': EgRef,
+            'dEgdT': dEgdT,
+            'irrad_ref': irrad_ref,
+            'temp_ref': temp_ref}
```
markcampanelli: I personally like to be able to inspect the full `result` object from the solver (possibly as a second output in a returned tuple), but I also understand if the majority would rather omit it from the return interface.

pvlib/ivtools.py
```diff
@@ -262,6 +264,163 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    Io_i = i_sc * np.exp(-v_oc/a_i)
+    Rs_i = (a_i*np.log1p((IL_i-i_mp)/Io_i) - v_mp)/i_mp
+    # params_i : initial values vector
+    params_i = np.array([IL_i, Io_i, a_i, Rsh_i, Rs_i])
+
+    # specs of module
+    specs = np.array([i_sc, v_oc, i_mp, v_mp, beta_voc, alpha_sc])
+
+    # computing
+    result = root(pv_fct, x0=params_i, args=specs, method=solver_method)
```
markcampanelli: Have you attempted to compute the Jacobian of `pv_fct` that could be passed as the `jac` parameter to `root`?

This _might_ be both faster and more robust, but it also can easily be left as a followup improvement.

mikofski: Not necessarily true. I believe the secant method is more stable (*) than the Newton-Raphson, and even though it converges slower, it might still be faster, if the Jacobian is computationally expensive to calculate.

(*) I believe secant or Broyden's is less sensitive to poor initial guess, discontinuity, and relatively flat or stiff systems, which can cause N-R to get stuck in an infinite loop, but I can't find a good reference for this, so maybe I'm wrong, sorry

mikofski: BTW, if you do need the derivatives, check `singlediode.py` they're there in [`bishop88()`](../blob/a72aca77a071937ccd740d5f0c6b570f030b6538/pvlib/singlediode.py#L70) by seeing `gradients=True`

pvlib/test/test_ivtools.py
cwhanse: > I would be ok with using the default solver.

I am also ok with this now; thanks for this discussion.

> Probably should mention this in the docstring.

yes.

pvlib/ivtools.py
cwhanse: Since we're not exposing most of the scipy.optimize.root input options, I'm not in favor of returning the `OptimizeResult`.  We could `OptimizeResult.message` to the error message.

pvlib/ivtools.py
cwhanse: > but it also can easily be left as a followup improvement.

Agree, let's tackle that with a follow-on PR.

pvlib/ivtools.py
tylunel: Ok. 
Then do you think it is worth testing the result object in the test ? I feel like some attribute of the complete OptimizeResult object can easily change (like nit the number of iteration, or nfev) if scipy is improved. That's why I would remove the optimizeresult from the dict output in test_fit_sdm_desoto() before testing the equivalence with expected result.

pvlib/test/test_ivtools.py
```diff
@@ -102,6 +109,29 @@ def test_fit_sdm_cec_sam(get_cec_params_cansol_cs5p_220p):
...
+                       'R_sh_ref': 125.79869968976132,
+                       'R_s': 0.2978148476600744,
+                       'alpha_sc': 0.005658,
+                       'EgRef': 1.121,
+                       'dEgdT': -0.0002677,
+                       'irrad_ref': 1000,
+                       'temp_ref': 25}
+    assert np.allclose(pd.Series(result), pd.Series(result_expected),
+                       rtol=1e-4)
+    with pytest.raises(RuntimeError):
```
cwhanse: Should check that the exception raised is the exception we're after.  Could do it like this (learned from @wholmgren quite recently):
```
with pytest.raises(RuntimeError) as exc:
    ivtools.fit….
    assert expected_text in str(exc.value)
```

pvlib/ivtools.py
cwhanse: I don't think it is worth testing the content of `result` in the CI tests. 

However, your question led me to look at the default settings for `solver_method in 'lm', 'hybr '`. The default settings terminate based on relative change of the solution, so it is possible that the root finder stalls in some flat portion of the objective surface rather than finding the minimum. We could build our own function to test the quality of the solution that `root` finds, but let's do that in a follow-on PR.

tylunel:
I think I have incorporated all your feedbacks now :)
Note that the coverage of diff cannot be improved because the requires_scipy mark prevents codecov from testing the raising of ImportError.

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+            'I_o_ref': sdm_params[1],
+            'a_ref': sdm_params[2],
+            'R_sh_ref': sdm_params[3],
+            'R_s': sdm_params[4],
+            'alpha_sc': alpha_sc,
+            'EgRef': EgRef,
+            'dEgdT': dEgdT,
+            'irrad_ref': irrad_ref,
+            'temp_ref': temp_ref,
+            'optimize_result': result}
```
markcampanelli: Unfortunately, including the `optimize_result` right in this dictionary makes chaining computations more difficult:
```python
result = ivtools.fit_sdm_desoto(cells_in_series=60, v_mp=31.0, i_mp=8.71, v_oc=38.3, i_sc=9.43, alpha_sc=0.005658, beta_voc=-0.13788)
pvlib.pvsystem.calcparams_desoto(1000, 25, **result)
```
```console
TypeError                                 Traceback (most recent call last)
<ipython-input-17-57980407c8d1> in <module>
----> 1 pvlib.pvsystem.calcparams_desoto(1000, 25, **result)

TypeError: calcparams_desoto() got an unexpected keyword argument 'optimize_result'
```
There are at least a few options here:
1. Ignore my suggestion to include the `optimize_result`.
2. Take my original suggestion of making `optimize_result` the second element in a returned tuple.
  ```python
      return {'I_L_ref': sdm_params[0], 'I_o_ref': sdm_params[1], 'a_ref': sdm_params[2], 'R_sh_ref': sdm_params[3],
                  'R_s': sdm_params[4], 'alpha_sc': alpha_sc, 'EgRef': EgRef, 'dEgdT': dEgdT, 'irrad_ref': irrad_ref,
                  'temp_ref': temp_ref}, result
  ```
3. Make the returned dictionary have a nested key for the parameters, such as:
  ```python
      return {
                  'params': {
                                  'I_L_ref': sdm_params[0],
                                  'I_o_ref': sdm_params[1],
                                  'a_ref': sdm_params[2],
                                  'R_sh_ref': sdm_params[3],
                                  'R_s': sdm_params[4],
                                  'alpha_sc': alpha_sc,
                                  'EgRef': EgRef,
                                  'dEgdT': dEgdT,
                                  'irrad_ref': irrad_ref,
                                  'temp_ref': temp_ref
                                 },
                  'optimize_result': result
  }
  ```

pvlib/test/test_ivtools.py
```diff
@@ -102,6 +109,31 @@ def test_fit_sdm_cec_sam(get_cec_params_cansol_cs5p_220p):
                 cells_in_series=1, temp_ref=25)
 
 
+@requires_scipy
+def test_fit_sdm_desoto(get_test_specs_params):
+    result = ivtools.fit_sdm_desoto(cells_in_series=60,
```
markcampanelli: move `cell_in_series` to the `get_test_specs_params` fixture.

pvlib/test/test_ivtools.py
```diff
@@ -102,6 +109,31 @@ def test_fit_sdm_cec_sam(get_cec_params_cansol_cs5p_220p):
...
+                       'I_o_ref': 3.2246097466679494e-10,
+                       'a_ref': 1.5912875522463978,
+                       'R_sh_ref': 125.79869968976132,
+                       'R_s': 0.2978148476600744,
+                       'alpha_sc': 0.005658,
+                       'EgRef': 1.121,
+                       'dEgdT': -0.0002677,
+                       'irrad_ref': 1000,
+                       'temp_ref': 25}
+    del result['optimize_result']
```
markcampanelli: Having to delete this entry here suggests that it shouldn't be lumped in with the rest of the parameters.

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        module [A/K].
+    * EgRef: float
+        Energy of bandgap of semi-conductor used [eV]
+    * dEgdT: float
+        Variation of bandgap according to temperature [eV/K]
+    * irrad_ref: float
+        Reference irradiance condition [W/m2]
+    * temp_ref: float
+        Reference temperature condition [C]
+    * result: scipy.optimize.OptimizeResult
```
markcampanelli: Key is actually named `optimize_result`. Don't lump this in with the rest of parameters, see comment below.

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+def fit_sdm_desoto(v_mp, i_mp, v_oc, i_sc, alpha_sc, beta_voc,
+                   cells_in_series, EgRef=1.121, dEgdT=-0.0002677,
+                   temp_ref=25, irrad_ref=1000):
+    """
+    Calculates the parameters for the De Soto single diode model using the
+    procedure described in [1]. This procedure has the advantage of
+    using common specifications given by manufacturers in the
+    datasheets of PV modules. The solution is found using the
+    scipy.optimize.root() function, with the correpsonding default solver
+    method 'hybr'.
```
markcampanelli: It's probably worth mentioning that there are no restrictions on the fit variables, i.e., series resistance _could_ go negative.

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    Dictionary with the following elements:
+
+    * I_L_ref: float
+        Light-generated current at reference conditions [A]
+    * I_o_ref: float
+        Diode saturation current at reference conditions [A]
+    * R_s: float
+        Series resistance [ohms]
+        Note that '_ref' is not mentionned in the name because
+        this resistance is not sensible to the conditions of test.
```
markcampanelli: I don't readily understand what this note means. I always thought the missing `_ref` was merely the legacy of an original oversight. @cwhanse Do you have an insight here?

pvlib/ivtools.py
cwhanse: if you are touching this text, "correpsonding " is misspelled.

pvlib/ivtools.py
cwhanse: `R_s` is used to remain consistent with SAM variable names. Let's remove the "Note that.." text.

pvlib/ivtools.py
cwhanse: I apologize, I overlooked this change.  _If_ we are returning the `result` object, return as 2nd element in a tuple.  `optimize_result` would be a better variable name.

docs/sphinx/source/whatsnew/v0.7.0.rst
```diff
@@ -96,6 +96,8 @@ Enhancements
   the single diode equation to an IV curve.
 * Add :py:func:`~pvlib.ivtools.fit_sdm_cec_sam`, a wrapper for the CEC single
   diode model fitting function '6parsolve' from NREL's System Advisor Model.
+* Add :py:func:`~pvlib.ivtools.fit_sdm_desoto`, a method to fit the single
```
wholmgren: Also add to `api.rst`.

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    datasheets of PV modules. The solution is found using the
+    scipy.optimize.root() function, with the correpsonding default solver
+    method 'hybr'.
+    The parameters returned by this function can be used by
+    pvsystem.calcparams_desoto to calculate the values at different
+    irradiance and cell temperature.
+
+    Parameters
+    ----------
+     v_mp: float
```
wholmgren:
```suggestion
    v_mp: float
```

wholmgren: removes extra space in indentation

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    -------
+    Dictionary with the following elements:
+
+    * I_L_ref: float
+        Light-generated current at reference conditions [A]
+    * I_o_ref: float
+        Diode saturation current at reference conditions [A]
+    * R_s: float
+        Series resistance [ohms]
+        Note that '_ref' is not mentionned in the name because
```
wholmgren:
```suggestion
        Note that '_ref' is not mentioned in the name because
```

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        Optimization result of scipy.optimize.root().
+        See scipy.optimize.OptimizeResult for more details.
+
+    References
+    ----------
+    [1] W. De Soto et al., "Improvement and validation of a model for
+    photovoltaic array performance", Solar Energy, vol 80, pp. 78-88,
+    2006.
+
+    [2] John A Dufﬁe ,William A Beckman, "Solar Engineering of Thermal
```
wholmgren:
```suggestion
    [2] John A Dufﬁe, William A Beckman, "Solar Engineering of Thermal
```

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        from scipy.optimize import root
+        from scipy import constants
+    except ImportError:
+        raise ImportError("The fit_sdm_desoto function requires scipy.")
+
+    # Constants
+    k = constants.value('Boltzmann constant in eV/K')
+    Tref = temp_ref + 273.15  # [K]
+
+    def pv_fct(params, specs):
```
wholmgren: Why is this function inside `fit_sdm_desoto` instead of being its own (perhaps private) function at the module level?

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+        """Evaluates the systems of equations used to solve for the single
+        diode equation parameters.
+        To avoid the confusion in names with variables of container
```
wholmgren: This sentence is confusing. I am guessing it can just be deleted.

pvlib/ivtools.py
tylunel: Sorry, I made this change too quickly.
@markcampanelli I think your view on the package is more complete than mine, so I will include the optimize_result as you suggested. (after all, the goal of this pull request is also to learn from experienced users/contributors of pvlib-python)
@cwhanse Why do you prefer to return it in a tuple? Intuitively I would have use the dict (3rd option of markcampanelli) for improving the readability in the later use of the output.

cwhanse: so that it can be omitted if not of interest, e.g., `params, _ = fit_sdm_desoto(`

pvlib/ivtools.py
```diff
@@ -262,6 +262,173 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    Rsh_i = 100.0
+    a_i = 1.5*k*Tref*cells_in_series
+    IL_i = i_sc
+    Io_i = i_sc * np.exp(-v_oc/a_i)
+    Rs_i = (a_i*np.log1p((IL_i-i_mp)/Io_i) - v_mp)/i_mp
+    # params_i : initial values vector
+    params_i = np.array([IL_i, Io_i, a_i, Rsh_i, Rs_i])
+
+    # specs of module
+    specs = np.array([i_sc, v_oc, i_mp, v_mp, beta_voc, alpha_sc])
```
wholmgren: If I understand correctly, this would be better written as `((i_sc, v_oc, i_mp, v_mp, beta_voc, alpha_sc), )`. Or make `pv_fct` take arguments `params, i_sc, v_oc, i_mp, v_mp, beta_voc, alpha_sc` and then you can avoid the nested tuple and the tuple/array unpacking within the function.

pvlib/test/test_ivtools.py
tylunel: cells_in_series is not in the get_test_specs_params fixture because it changes at line 129 of the test in order to cause the RuntimeError. I feel like it would be more complicated to add it the fixture and to change it later to have the RuntimeError. What do you think ?

pvlib/ivtools.py
wholmgren: Like option 2. I don't like option 3.

pvlib/test/test_ivtools.py
```diff
@@ -102,6 +109,31 @@ def test_fit_sdm_cec_sam(get_cec_params_cansol_cs5p_220p):
...
+    assert np.allclose(pd.Series(result), pd.Series(result_expected),
+                       rtol=1e-4)
+    with pytest.raises(RuntimeError) as exc:
```
wholmgren: this belongs in its own test such as `test_fit_sdm_desoto_failure`

pvlib/ivtools.py
markcampanelli: Thanks for chiming in @cwhanse and @wholmgren. I'm good with option 2. Interestingly, @tylunel, I use option 3 in my own APIs (I always return a single dictionary, nested if necessary), but for some reason I guessed that option 2 would be more preferable than option 3 in pvlib ;).

pvlib/test/test_ivtools.py
wholmgren: I'd prefer to see it added to the fixture and then modified in the function that causes the `RuntimeError`. That way the fixture returns a complete, consistent specification.

pvlib/test/test_ivtools.py
```diff
@@ -28,6 +28,13 @@ def get_cec_params_cansol_cs5p_220p():
                        'R_sh_ref': 837.51, 'R_s': 1.004, 'Adjust': 2.3}}
 
 
+@pytest.fixture
+def get_test_specs_params():
```
wholmgren: I'd also change the fixture name to something more descriptive. For example, `fit_sdm_desoto_params`. While you'll sometimes see it, `get_` is superfluous in a fixture.

pvlib/test/test_ivtools.py
markcampanelli: Does this test warrant its own fixture that includes `cells_in_series`?

TBH, I don't see the fixture `test_fit_sdm_desoto` doing much here, because there is only one test. It might be more extensible for the addition of future tests, but in that case I would also put the expected values into the fixture. See [this example](https://github.com/markcampanelli/pvfit/blob/master/pvfit/modeling/single_diode/equation_test.py#L9).

markcampanelli:
@tylunel Welcome to the world of open-source code reviews! Don't worry, you'll get this code in. Thanks for your patience ;).

pvlib/test/test_ivtools.py
wholmgren: I agree that a fixture is not necessary for these tests. The failure test hard codes the values anyways, so it wouldn't even use it.

pvlib/ivtools.py
tylunel: Mainly because I thought until now that as long as this function was used by only one function, it was better to put it inside. 
I will put it outside in my next commit as suggested

docs/sphinx/source/whatsnew/v0.7.0.rst
```diff
@@ -1,4 +1,4 @@
-.. _whatsnew_0700:
+﻿.. _whatsnew_0700:
```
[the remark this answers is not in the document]
tylunel: I don't know either... This line did not showed in the previous commit moreover. Let's see if it stays with my next commit

pvlib/ivtools.py
```diff
@@ -515,3 +483,62 @@ def _calculate_sde_parameters(beta0, beta1, beta3, beta4, v_mp, i_mp, v_oc):
...
+    y[3] = Imp - Vmp * ((Io/a)*np.exp((Vmp+Imp*Rs)/a) + 1.0/Rsh) / \
+        (1.0 + (Io*Rs/a)*np.exp((Vmp+Imp*Rs)/a) + Rs/Rsh)
+    # 5th equation - open-circuit T2 - eq (4) at temperature T2 in [1]
+    T2 = Tref + 2
+    Voc2 = (T2 - Tref)*betaoc + Voc  # eq (7) in [1]
+    a2 = a*T2/Tref  # eq (8) in [1]
+    IL2 = IL + alphasc*(T2-Tref)  # eq (11) in [1]
+    Eg2 = EgRef*(1 + dEgdT*(T2-Tref))  # eq (10) in [1]
+    Io2 = Io * (T2/Tref)**3 * np.exp(1/k * (EgRef/Tref-Eg2/T2))  # eq (9)
+    y[4] = -IL2 + Io2*np.expm1(Voc2/a2) + Voc2/Rsh  # eq (4) at T2
```
markcampanelli: Note that these references are no longer in this function’s docstring.

pvlib/ivtools.py
```diff
@@ -262,6 +262,142 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
 
+def fit_sdm_desoto(v_mp, i_mp, v_oc, i_sc, alpha_sc, beta_voc,
+                   cells_in_series, EgRef=1.121, dEgdT=-0.0002677,
+                   temp_ref=25, irrad_ref=1000):
+    """
+    Calculates the parameters for the De Soto single diode model using the
+    procedure described in [1]. This procedure has the advantage of
+    using common specifications given by manufacturers in the
+    datasheets of PV modules.
+    The solution is found using the scipy.optimize.root() function,
```
wholmgren: If you want this to be a separate paragraph then you'll need to add a blank line between them. Same for potential paragraph below. Either is fine with me. Now that you've added the line to the `api.rst` file, you can view the rendered documentation by clicking on the readthedocs link in the CI box at the bottom of the PR.

pvlib/ivtools.py
```diff
@@ -262,6 +262,142 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+              Tref, k),)
+
+    # computing with system of equations described in [1]
+    optimize_result = root(_system_of_equations, x0=params_i, args=specs)
+
+    if optimize_result.success:
+        sdm_params = optimize_result.x
+    else:
+        raise RuntimeError('Parameter estimation failed:\n'
+                           + optimize_result.message)
```
wholmgren: I agree with stickler in this case - better to move the plus sign to the end of the line above

tylunel: Actually the plus sign was above in the previous commit I made, but stickler found a lint error as well. I choose to keep it below according to this:
https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

markcampanelli: I humbly suggest:
```python
        raise RuntimeError(
            'Parameter estimation failed:\n' + optimize_result.message)
```
which keeps the string on a single line and which I think is a short enough line.

pvlib/ivtools.py
```diff
@@ -348,3 +486,64 @@ def _calculate_sde_parameters(beta0, beta1, beta3, beta4, v_mp, i_mp, v_oc):
     else:  # I0_voc > 0
         I0 = I0_voc
     return (IL, I0, Rsh, Rs, nNsVth)
+
+
+def _system_of_equations(params, specs):
```
adriesse: Do we need Desoto in the function name?

pvlib/ivtools.py
```diff
@@ -262,6 +262,144 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    temp_ref: float, default 25
+        Reference temperature condition [C]
+    irrad_ref: float, default 1000
+        Reference irradiance condition [W/m2]
+
+    Returns
+    -------
+    Tuple of the following elements:
+
+        * Dictionary with the following elements:
```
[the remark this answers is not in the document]
markcampanelli: I agree that it’s good to be consistent. However, I very much prefer interfaces where order does _not_ matter, because this ordering is a form of complexity, which is harder to maintain/extend without breakages. Also, the dictionary output is perhaps _best_ paired with keyword-only inputs to subsequent functions/methods, but it still works with pvlib’s current mixture of named positional and keyword arguments. However, it would not be compatible with Python 3.8’s new positional-only arguments.

pvlib/ivtools.py
```diff
@@ -262,6 +262,144 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+    a_i = 1.5*k*Tref*cells_in_series
+    IL_i = i_sc
+    Io_i = i_sc * np.exp(-v_oc/a_i)
+    Rs_i = (a_i*np.log1p((IL_i-i_mp)/Io_i) - v_mp)/i_mp
+    # params_i : initial values vector
+    params_i = np.array([IL_i, Io_i, a_i, Rsh_i, Rs_i])
+
+    # specs of module
+    specs = ((i_sc, v_oc, i_mp, v_mp, beta_voc, alpha_sc, EgRef, dEgdT,
+              Tref, k),)
```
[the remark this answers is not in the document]
wholmgren: The outer parentheses could be removed, but the inner parentheses and the trailing comma must stay.

```python
In [1]: ((1, 2),)
Out[1]: ((1, 2),)

In [2]: (1, 2),
Out[2]: ((1, 2),)

In [3]: (1,2,)
Out[3]: (1, 2)
```

I think the outer parentheses add a little bit of clarity, but I'm also ok with removing them.

pvlib/ivtools.py
```diff
@@ -262,6 +262,144 @@ def fit_sde_sandia(voltage, current, v_oc=None, i_sc=None, v_mp_i_mp=None,
...
+
+    # computing with system of equations described in [1]
+    optimize_result = root(_system_of_equations, x0=params_i, args=specs)
```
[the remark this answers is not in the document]
wholmgren: @adriesse  suggestion above could work if this were `args=(specs, )`. No preference here.

pvlib/ivtools.py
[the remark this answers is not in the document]
wholmgren: I don't think there's a right answer. It would be most helpful if people could provide some good references for pros/cons of modern scientific python design patterns. I am not aware of any references that tackle non-trivial cases.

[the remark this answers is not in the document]
markcampanelli: 24 minutes into this keynote talk on simplicity, Rich Hickey (creator of Clojure) discusses the complexity of ordering things in lists:

https://m.youtube.com/watch?v=rI8tNMsozo0&t=3s

This is not gospel of course, and indeed ordering can provide performance gains, etc. (which is why dictionaries are now ordered in CPython, which I personally treat as an implementation detail to not to be relied upon).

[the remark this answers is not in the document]
cwhanse: Let's address consistency among functions returns in a follow-on PR. I can't say I had a pattern in mind when I started `ivtools.py`

pvlib/test/test_ivtools.py
```diff
@@ -102,6 +102,34 @@ def test_fit_sdm_cec_sam(get_cec_params_cansol_cs5p_220p):
...
+                       'a_ref': 1.5912875522463978,
+                       'R_sh_ref': 125.79869968976132,
+                       'R_s': 0.2978148476600744,
+                       'alpha_sc': 0.005658,
+                       'EgRef': 1.121,
+                       'dEgdT': -0.0002677,
+                       'irrad_ref': 1000,
+                       'temp_ref': 25}
+    assert np.allclose(pd.Series(result), pd.Series(result_expected),
+                       rtol=1e-4)
```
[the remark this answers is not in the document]
tylunel: Actually I have no source for the expected result, it is no more than the result of the function fit_sdm_desoto(). I have just checked that these parameters result in an IV curve very close to the one described in the manufacturer datasheet. The relative tolerance doesn't even match with the degree at which I can assess how good the IV curve generated matches with the datasheet curve. 
Then I consider the test more like a way to warn developers if their changes impact significantly the output of this function. 
If someone find a source where the same system of equations is solved, and the input specs and ouput parameters are given, I guess the test could be changed in a follow-on PR.

cwhanse:
@tylunel I fixed the merge conflict in `v0.7.0.rst` it was nothing you did, that file is a bother sometimes.  Please add yourself to the Contributors list at the bottom of this file and we'll merge.  Thanks again!

pvlib/test/test_ivtools.py
```diff
@@ -102,6 +102,34 @@ def test_fit_sdm_cec_sam(get_cec_params_cansol_cs5p_220p):
...
+                       'R_s': 0.297814,
+                       'alpha_sc': 0.005658,
+                       'EgRef': 1.121,
+                       'dEgdT': -0.0002677,
+                       'irrad_ref': 1000,
+                       'temp_ref': 25}
+    assert np.allclose(pd.Series(result), pd.Series(result_expected),
+                       rtol=1e-4)
+
+
```
cwhanse: @requires_scipy

cwhanse:
@tylunel one last edit, add requires_scipy mark to the new test

cwhanse:
test failures are `get_psm3` and `get_midc`, no unexpected warnings. I'll merge tonight and create new issues for follow-on work:

1. investigate improving convergence of `root`
2. pass select kwargs through `fit_sdm_desoto` to `scipy.optimize.root`
3. quality check on returned parameters
4. improve signature consistency in ivtools.py

PR #787, opened 2019-10-08
Get rid of `re` deprecation warnings

Peque:
It seems it was the only regular expression missing an update (all the others are already written as `r'expression'`).

`DeprecationWarning` is given since Python 3.6. In some future version of Python it will be a `SyntaxError`.


cwhanse:
Thanks, @Peque . Test failures are the `get_psm3` pressure [issue](https://github.com/pvlib/pvlib-python/pull/783#issuecomment-538537240)  @wholmgren I'm okay to merge this.

wholmgren:
Thanks!

Issue #751, opened 2019-07-11
Implement IEC61853 IAM calculations


adriesse:
The standard gives the option of using interpolation of measured data, or the Martin & Ruiz model which has the single parameter `a_r`.  Both should be implemented.

cwhanse:
The [Martin and Ruiz model](https://github.com/sandialabs/MATLAB_PV_LIB/blob/master/pvl_iam_martinruiz.m) is in PVLib for MATLAB. In a later paper Martin and Ruiz extended the model to estimate reflection of diffuse irradiance, also in PVLib for MATLAB [here](https://github.com/sandialabs/MATLAB_PV_LIB/blob/master/pvl_iam_martinruiz_components.m)  I think both versions are useful, the component version shows up in various bifacial irradiance models.

cwhanse:
#679 and #680 are relevant

PR #793, opened 2019-10-19
Implement IEC 61853 IAM calculations for diffuse irradiance

adriesse:
- [x] Follows up on issue #751

IAM function for diffuse irradiance as proposed by Martin & Ruiz and adopted in the IEC 61853 energy rating procedure.


adriesse:
@wholmgren @cwhanse Are you ok with this function returning a tuple?

wholmgren:
@adriesse yes, I would like to get away from returning DataFrames.

adriesse:
Ready for review.

pvlib/iam.py
```diff
@@ -274,6 +274,111 @@ def martin_ruiz(aoi, a_r=0.16):
...
+
+        # avoid undefined results for horizontal or upside-down surfaces
+        small_angle = 1e-06
+        slope = np.clip(slope, small_angle, 180 - small_angle)
+
+    if np.any(np.less_equal(a_r, 0)):
+        raise RuntimeError("The parameter 'a_r' cannot be zero or negative.")
+
+    if c2 is None:
+        c2 = 0.5 * a_r - 0.154
```
cwhanse: Add an inline comment that ties this default value to [3] Sect. 7.2

pvlib/iam.py
```diff
@@ -274,6 +274,111 @@ def martin_ruiz(aoi, a_r=0.16):
...
+
+    Notes
+    -----
+    Sky and ground modifiers are complementary: iam_sky for slope = 30 is
+    equal to iam_ground for slope = 180 - 30.  For vertical surfaces,
+    slope = 90, the two factors are equal.
+
+    References
+    ----------
+    [1] N. Martin and J. M. Ruiz, "Calculation of the PV modules angular
```
cwhanse: I think the primary reference should be

[1] Martín, N., Ruiz, J. M. 2005. Annual angular reflection losses in PV modules. Progress in Photovoltaics: Research and Applications, 13(1), 75–84.  The corrigendum [2] doesn't apply.

pvlib/iam.py
```diff
@@ -274,6 +274,111 @@ def martin_ruiz(aoi, a_r=0.16):
...
+    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Oct. 2019
+
+    if isinstance(slope, pd.Series):
+        out_index = slope.index
+    else:
+        out_index = None
+
+    slope = np.asanyarray(slope)
+
+    with np.errstate(invalid='ignore'):
```
cwhanse: I'm not following the purpose of the slope manipulations - what values of slope are anticipated here?

pvlib/iam.py
```diff
@@ -274,6 +274,111 @@ def martin_ruiz(aoi, a_r=0.16):
...
+    a_r : numeric
+        The angular losses coefficient described in equation 3 of [1].
+        This is an empirical dimensionless parameter. Values of a_r are
+        generally on the order of 0.08 to 0.25 for flat-plate PV modules.
+        a_r must be a positive numeric scalar.
+
+    c1, c2 : numeric scalar
+        Two fitting parameters for the expressions that approximate the
+        integral of diffuse irradiance coming from different directions.
+        c1 is contant at 4 / 3 / pi (0.4244) and c2 varies with a_r.  The
```
cwhanse: spelling: contant

pvlib/iam.py
```diff
@@ -274,6 +274,111 @@ def martin_ruiz(aoi, a_r=0.16):
...
+
+        # flip backward tilts forward
+        slope = np.abs(slope)
+
+        # avoid undefined results for horizontal or upside-down surfaces
+        small_angle = 1e-06
+        slope = np.clip(slope, small_angle, 180 - small_angle)
+
+    if np.any(np.less_equal(a_r, 0)):
+        raise RuntimeError("The parameter 'a_r' cannot be zero or negative.")
```
cwhanse: I think this should be `ValueError`

pvlib/iam.py
```diff
@@ -274,6 +274,111 @@ def martin_ruiz(aoi, a_r=0.16):
...
 
 
+def martin_ruiz_diffuse(slope, a_r=0.16, c1=0.4244, c2=None):
+    '''
+    Determine the incidence angle modifiers (iam) for diffuse sky and
+    ground-reflected irradiance using the Martin and Ruiz incident angle model.
+
+    Parameters
+    ----------
+    slope : numeric, degrees
```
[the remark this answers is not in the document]
adriesse: "slope" is just the word I've always used.  "tilt" is fine.  I'll lift the description from another function.

pvlib/iam.py
```diff
@@ -274,6 +274,111 @@ def martin_ruiz(aoi, a_r=0.16):
...
+        The slope or tilt angle of the module, where 0 is horizontal.
+
+    a_r : numeric
+        The angular losses coefficient described in equation 3 of [1].
+        This is an empirical dimensionless parameter. Values of a_r are
+        generally on the order of 0.08 to 0.25 for flat-plate PV modules.
+        a_r must be a positive numeric scalar.
```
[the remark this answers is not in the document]
adriesse: Copied from the other function.  Arrays work for the other parameters by default.  If I advertise it, then I should make lists work too perhaps.

pvlib/iam.py
adriesse: I disagree.  The typos may not be there, but neither are all the explanations.

pvlib/iam.py
adriesse: Any value is allowed.  Maybe I got carried away.

pvlib/iam.py
adriesse: Do you want the other function changed too?

adriesse: I'll change the other function too.

pvlib/iam.py
```diff
@@ -274,6 +274,111 @@ def martin_ruiz(aoi, a_r=0.16):
...
+
+    c1, c2 : numeric scalar
```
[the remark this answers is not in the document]
adriesse: I'll add the line and see what happens.

pvlib/iam.py
cwhanse: You are correct, the FA and FD equations are in the 2001 paper.  Disregard this comment.

pvlib/iam.py
cwhanse: Yes please, and add the `Raises` section to the docstring.

pvlib/iam.py
```diff
@@ -274,6 +274,114 @@ def martin_ruiz(aoi, a_r=0.16):
...
+    from numpy import pi, radians, sin, cos, exp
+
+    beta = radians(surface_tilt)
+
+    with np.errstate(invalid='ignore'):
+        # because sin(pi) isn't exactly zero
+        sin_beta = np.where(surface_tilt < 90, sin(beta), sin(pi - beta))
+
+        trig_term_sky = sin_beta + (pi - beta - sin_beta) / (1 + cos(beta))
+        trig_term_gnd = sin_beta +      (beta - sin_beta) / (1 - cos(beta))
```
wholmgren:
```suggestion
        trig_term_gnd = sin_beta +      (beta - sin_beta) / (1 - cos(beta))  # noqa: E222
```

pvlib/iam.py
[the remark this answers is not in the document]
wholmgren: rst issues aside, I think c2 should have its own line and description because it's allowed to be None.

pvlib/iam.py
```diff
@@ -274,6 +274,114 @@ def martin_ruiz(aoi, a_r=0.16):
...
+    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Oct. 2019
+
+    if isinstance(surface_tilt, pd.Series):
+        out_index = surface_tilt.index
+    else:
+        out_index = None
+
+    surface_tilt = np.asanyarray(surface_tilt)
+
+    with np.errstate(invalid='ignore'):
```
wholmgren: Have you confirmed that all of these numpy operations here and below need to be wrapped with `np.errstate(invalid='ignore')`? Generally better to only put the offending functions inside the context block.

pvlib/test/test_iam.py
```diff
@@ -100,10 +100,52 @@ def test_martin_ruiz():
...
+    # check various inputs as Series
+    surface_tilt = pd.Series(surface_tilt)
+    expected_sky = pd.Series(expected_sky, name='iam_sky')
+    expected_gnd = pd.Series(expected_gnd, name='iam_ground')
+    iam = _iam.martin_ruiz_diffuse(surface_tilt, a_r)
+    assert_series_equal(iam[0], expected_sky)
+    assert_series_equal(iam[1], expected_gnd)
+
+    # check exception clause
+    with pytest.raises(ValueError):
```
wholmgren: separate test please (I see I missed this complaint on the test above from which this test is styled)

pvlib/iam.py
adriesse: Let's see if it complains about the line length now...

pvlib/iam.py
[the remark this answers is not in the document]
adriesse: Just let me know what the preferred scalar/float/numeric/default/units presentation should be.

pvlib/iam.py
adriesse: Ok, I can silence that one too.

adriesse:
Ready for another review.

pvlib/iam.py
cwhanse: According to the description in the docstring, 0 <= `surface_tilt` <= 180. I don't think this function should adjust `surface_tilt` for negative values or values greater than 360. I'd prefer to raise `ValueError` and let the user figure out where the incorrect tilt values came from.

pvlib/iam.py
```diff
@@ -274,6 +274,119 @@ def martin_ruiz(aoi, a_r=0.16):
...
+        # This equation is from [3] Sect. 7.2
+        c2 = 0.5 * a_r - 0.154
+
+    from numpy import pi, radians, sin, cos, exp
+
+    beta = radians(surface_tilt)
+
+    with np.errstate(invalid='ignore'):
+        # because sin(pi) isn't exactly zero
+        sin_beta = np.where(surface_tilt < 90, sin(beta), sin(pi - beta))
```
cwhanse: Since line 360 limits `surface_tilt` between 0 and 180, this line is unnecessary.

pvlib/test/test_iam.py
```diff
@@ -99,11 +100,59 @@ def test_martin_ruiz():
...
+    iam = _iam.martin_ruiz_diffuse(surface_tilt)
+    assert_allclose(iam, expected)
+
+    # will fail if parameter names change
+    iam = _iam.martin_ruiz_diffuse(surface_tilt=surface_tilt, a_r=a_r)
+    assert_allclose(iam, expected)
+
+    a_r = 0.18
+    surface_tilt = [0, 30, 90, 120, -30, np.nan, np.inf]
+    expected_sky = [0.9407678, 0.9452250, 0.9407678, 0.9055541, 0.9452250,
```
cwhanse: `surface_tilt` = -30 should not produce a value that appears valid.

pvlib/iam.py
```diff
@@ -274,6 +274,119 @@ def martin_ruiz(aoi, a_r=0.16):
...
+    # undo possible surface rotations
+    with np.errstate(invalid='ignore'):
+        surface_tilt = (surface_tilt + 180) % 360 - 180
+
+    # flip backward tilts forward
+    surface_tilt = np.abs(surface_tilt)
+
+    # avoid undefined results for horizontal or upside-down surfaces
+    small_angle = 1e-06
+    surface_tilt = np.clip(surface_tilt, small_angle, 180 - small_angle)
```
cwhanse: This line is limiting `surface_tilt` between (nearly) 0 and 180, correct?

pvlib/iam.py
```diff
@@ -274,6 +274,119 @@ def martin_ruiz(aoi, a_r=0.16):
...
+
+    if np.any(np.less_equal(a_r, 0)):
+        raise ValueError("The parameter 'a_r' cannot be zero or negative.")
+
+    if c2 is None:
+        # This equation is from [3] Sect. 7.2
+        c2 = 0.5 * a_r - 0.154
+
+    from numpy import pi, radians, sin, cos, exp
```
cwhanse: minor item - `numpy` is already imported, so these functions can be referenced rather than imported again.

pvlib/iam.py
```diff
@@ -274,6 +274,119 @@ def martin_ruiz(aoi, a_r=0.16):
...
+        generally on the order of 0.08 to 0.25 for flat-plate PV modules.
+        a_r must be greater than zero.
+
+    c1 : numeric scalar
+        First fitting parameter for the expressions that approximates the
+        integral of diffuse irradiance coming from different directions.
+        c1 is given as the constant 4 / 3 / pi (0.4244) in [1].
+
+    c2 : numeric scalar
+        Second fitting parameter for the expressions that approximates the
```
```suggestion
        Second fitting parameter for the expressions that approximate the
```

pvlib/iam.py
```diff
@@ -274,6 +274,119 @@ def martin_ruiz(aoi, a_r=0.16):
...
+        (e.g. surface facing up = 0, surface facing horizon = 90)
+
+    a_r : numeric
+        The angular losses coefficient described in equation 3 of [1].
+        This is an empirical dimensionless parameter. Values of a_r are
+        generally on the order of 0.08 to 0.25 for flat-plate PV modules.
+        a_r must be greater than zero.
+
+    c1 : numeric scalar
+        First fitting parameter for the expressions that approximates the
```
```suggestion
        First fitting parameter for the expressions that approximate the
```

pvlib/iam.py
```diff
@@ -274,6 +274,119 @@ def martin_ruiz(aoi, a_r=0.16):
...
+        The tilt angle is defined as degrees from horizontal
+        (e.g. surface facing up = 0, surface facing horizon = 90)
+
+    a_r : numeric
+        The angular losses coefficient described in equation 3 of [1].
+        This is an empirical dimensionless parameter. Values of a_r are
+        generally on the order of 0.08 to 0.25 for flat-plate PV modules.
+        a_r must be greater than zero.
+
+    c1 : numeric scalar
```
wholmgren:
```suggestion
    c1 : float
```

pvlib/iam.py
```diff
@@ -274,6 +274,119 @@ def martin_ruiz(aoi, a_r=0.16):
...
+        First fitting parameter for the expressions that approximates the
+        integral of diffuse irradiance coming from different directions.
+        c1 is given as the constant 4 / 3 / pi (0.4244) in [1].
+
+    c2 : numeric scalar
```
wholmgren:
```suggestion
    c2 : float
```

pvlib/iam.py
wholmgren: I'm generally opposed to input checking on the grounds that it's a huge amount of work to maintain and it gives users a false sense of security (they prevented me from doing something stupid in function a, so I'm not going to worry about doing something stupid in function b). Garbage in, garbage out. I'd relax my position for higher-level wrapper functions/methods.

pvlib/iam.py
wholmgren: @cwhanse are you asking for `sin = np.sin` etc or are you asking for the equations below to be modified to use `np.sin` etc?

cwhanse:
> In any case, I don't think we should be giving "approve" reviews until we are actually ok with never reviewing it again.)

+1

mikofski:
Regarding Will's comment on input checking, I believe this is also a Python style preference, EG: see [Easier to ask for forgiveness than permission](https://docs.python.org/3/glossary.html#term-eafp) vs. [Look before you leap](https://docs.python.org/3/glossary.html#term-lbyl). For a discussions on the virtues of EAFP vs. LBYL, see [this post by Brett Cannon from Microsoft](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/)

pvlib/iam.py
adriesse: Yes, but it only works properly when the angles are reflected/rotated into the range 0-180 beforehand.  Otherwise it would force -30 to nearly 0 and 210 to nearly 180, etc.

pvlib/iam.py
adriesse: This is a numeric precision issue.

pvlib/test/test_iam.py
adriesse: Is there consensus on this?

pvlib/iam.py
adriesse: I think the equations generally look tidier and are more compact without the `np.`, although for `radians` it was not really necessary.

pvlib/iam.py
adriesse: To docstring doesn't say " 0 <= surface_tilt <= 180 ", does it? In my own code I interpret azimuth 0 with slope -30 to the the same as azimuth 180, slope 30.  Sometimes that's convenient.
Anyway, I'll make it do whatever the majority/authority would like, but it's not clear what that is at the moment.

cwhanse: It doesn't and perhaps it should. It says "degrees from horizontal" and we don't define a positive rotation direction , so in my view it is simply an angle without a rotation direction. Your example has an implicit direction (positive is toward the azimuth). 

The [styel guide](https://pvlib-python.readthedocs.io/en/latest/variables_style_rules.html) is less clear on this point.

pvlib/iam.py
cwhanse: Since sin(theta) = sin( pi - theta) I don't see what this line accomplishes.

pvlib/iam.py
wholmgren: How many functions would need to gain these incantations to keep things consistent through the library?

cwhanse: @wholmgren the manipulations to ensure `surface_tilt` is between 0 and 180?  

To clarify my point of view, 0 < surface_tilt < 180 is how I've seen the quantity used, and I think we can state that in the docstring and omit the manipulations and value checking.

wholmgren: > the manipulations to ensure surface_tilt is between 0 and 180?

Yes. 

> To clarify my point of view, 0 < surface_tilt < 180 is how I've seen the quantity used, and I think we can state that in the docstring and omit the manipulations and value checking.

This is my preference too.

pvlib/iam.py
adriesse: It ensures that when surface_tilt is 180, sin_beta is exactly 0.

pvlib/iam.py
adriesse: Ok, I will remove all that stuff.  Shall I remove the test on `a_r` also?

pvlib/iam.py
cwhanse:
```
In [37]: np.sin(np.pi)
Out[37]: 1.2246467991473532e-16
```

Aha.

pvlib/iam.py
cwhanse: > Shall I remove the test on a_r also?

I don't have a preference. The docstring states the restriction on the value.

wholmgren:
> I would just like to bring the input-checking topic to the general discussion.

I don't have much to add to the general discussion beyond what's already been said in #433. Finding the time and resources to consistently implement that across the library will be a challenge. For the specific discussions above, my concerns were related to setting a new standard and the need to implement across the API, not the merits of the standard itself. A fair criticism of my position is "pvlib doesn't have a clear enough standard for this input, and my proposal is reasonable and useful to me." I don't know the right answer, so I am taking the position that seems to require the least commitment (I am a millennial after all).

wholmgren:
@cwhanse any blockers on this or ok to merge?

cwhanse:
> @cwhanse any blockers on this or ok to merge?

OK to merge.

wholmgren:
thanks @adriesse!

adriesse:
Enjoy!

PR #796, opened 2019-10-22
handle warnings from temperature model tests

### LINKED ISSUE #795: test__infer_temperature_model_params, test_deprecated_08 emit warnings
`test_pvsystem.py::test__infer_temperature_model_params` and `test_modelchain.py::test_deprecated_08` emit `pvlibDeprecationWarning`. They should instead assert that the warnings are raised and silence them so they don't confuse the test output.

cwhanse:

 - [x] Closes  #795 

Move one warning to avoid raising two warnings for the same cause

### DIFF
--- a/pvlib/test/test_modelchain.py
+++ b/pvlib/test/test_modelchain.py
@@ -596,31 +596,34 @@ def test_deprecated_08():
     # without PVSystem.racking_model and PVSystem.module_type
     system = PVSystem(module_parameters=module_parameters)
     # deprecated temp_model kwarg
-    with pytest.warns(pvlibDeprecationWarning) as exc:
+    with pytest.warns(pvlibDeprecationWarning) as recw:
         ModelChain(system, location,
                    dc_model='desoto',
                    aoi_model='no_loss', spectral_model='no_loss',
                    temp_model='sapm',
                    ac_model='snlinverter')
-        assert 'temp_model keyword argument is deprecated' in str(exc.info)
+        w = recw.pop(pvlibDeprecationWarning)
+        assert 'temp_model keyword argument is deprecated' in str(w.message)
     # provide both temp_model and temperature_model kwargs
-    with pytest.warns(pvlibDeprecationWarning) as exc:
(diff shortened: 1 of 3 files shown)


pvlib/test/test_modelchain.py
```diff
@@ -593,29 +593,28 @@ def test_deprecated_08():
...
-                   temperature_model='pvsyst',
-                   temp_model='sapm',
-                   ac_model='snlinverter')
+    # deprecated temp_model kwarg
+    with pytest.warns(pvlibDeprecationWarning) as recw:
+        ModelChain(system, location, dc_model='desoto', aoi_model='no_loss',
+                   spectral_model='no_loss', ac_model='snlinverter',
+                   temp_model='sapm')
+        w = recw.pop(pvlibDeprecationWarning)
+        assert 'temp_model keyword argument is deprecated' in str(w.message)
```
wholmgren: all of these `asserts` need to be brought out of the context blocks.

cwhanse: Working from [this](https://docs.pytest.org/en/latest/warnings.html) documentation

wholmgren: I've not tried this before, but maybe it would be easier to use the `match` kwarg like

```python
    match = 'temp_model keyword argument is deprecated'
    with pytest.warns(pvlibDeprecationWarning, match=match):
        ModelChain(system, location, dc_model='desoto', aoi_model='no_loss',
                   spectral_model='no_loss', ac_model='snlinverter',
                   temp_model='sapm')
```

cwhanse: `match` = regexp would probably work. The example I saw used an exact match and I didn't want to copy the exception message in full, so I stopped looking at it.  Probably `match` is better than the `in str()` pattern

cwhanse: appears to work now. I think it is better to switch to using `match` rather than rely on the desired warning or exception to be the last in the stack.  Another commit coming.

cwhanse:
It all passed on py35_min, so I'll merge.

PR #797, opened 2019-10-23
replace Pandas item() implementation with numpy's using .values

### LINKED ISSUE #794: fix tools._scalar_out to account for Series.item deprecation
`tools._scalar_out` calls `input.item()` to reduce an array or Series to a scalar. But the `Series.item` method is deprecated as of pandas 0.25, so we need to do something different. Maybe first call `np.asarray` so we know we have a numpy array and then we can call `.item()`.

Also should replace the reserved argument name `input` with something else.

CameronTStark:

 - [X] Closes  #794 
 - [X] Fully tested. Added and/or modified tests to ensure correct behavior for all reasonable inputs. Tests (usually) must pass on the TravisCI and Appveyor testing services.
* More tests passed, less warnings (probably from removing the deprecated) but one more test skipped.

 - [X] Updates entries to `docs/sphinx/source/api.rst` for API changes.
* I'll do this once this method is review/approved.

Pandas is deprecating Series.item() and warns against it for 0.25+. I've replaced the item() call with the values.item() which is a numpy method that's not going anywhere. This is exactly how Series.item() works and shouldn't change how _scalar_out() works.

Edit: [Here's a link to the pandas source code for .item() for reference](https://github.com/pandas-dev/pandas/blob/v0.25.2/pandas/core/base.py#L724-L740)

### DIFF
--- a/pvlib/tools.py
+++ b/pvlib/tools.py
@@ -212,7 +212,7 @@ def _scalar_out(input):
     else:  #
         # works if it's a 1 length array and
         # will throw a ValueError otherwise
-        output = input.values.item()
+        output = np.asarray(input).item()
 
     return output
 
--- a/pvlib/tools.py
+++ b/pvlib/tools.py
@@ -212,7 +212,7 @@ def _scalar_out(input):
     else:  #
         # works if it's a 1 length array and
         # will throw a ValueError otherwise
-        output = input.item()
+        output = input.values.item()
 
     return output
 


cwhanse:
I did a quick scan over other use of ".item" I didn't see another instances of the pandas method. LGTM

pvlib/tools.py
```diff
@@ -212,7 +212,7 @@ def _scalar_out(input):
     else:  #
         # works if it's a 1 length array and
         # will throw a ValueError otherwise
-        output = input.item()
+        output = input.values.item()
```
wholmgren: this won't work if `input` is a numpy array because it doesn't have a `values` attribute. 

```python
In [3]: np.array([0]).values
AttributeError                            Traceback (most recent call last)
<ipython-input-3-6306ba001445> in <module>
----> 1 np.array([0]).values

AttributeError: 'numpy.ndarray' object has no attribute 'values'
```

`np.asarray` seems to work:

```python
In [9]: np.asarray(pd.Series([0])).item()
Out[9]: 0

In [10]: np.asarray(np.array([0])).item()
Out[10]: 0
```

Interesting that `np.asscalar` also emits a warning:

```python
In [7]: np.asscalar(pd.Series([0]))
/Users/holmgren/miniconda3/envs/sfacore/lib/python3.7/site-packages/numpy/lib/type_check.py:547: FutureWarning: `item` has been deprecated and will be removed in a future version
  return a.item()
Out[7]: 0
```

CameronTStark: Good suggestion @wholmgren. I've implemented that solution. 

Also, the Numpy docs suggest similarly under [`np.asscalar`](https://docs.scipy.org/doc/numpy-1.17.0/reference/generated/numpy.asscalar.html#numpy.asscalar).

CameronTStark:
Ah, I glossed passed the last part of #794 about the function parameter colliding with the input reserved word but thought about it because my IDE yelled at me too. 

I'll create a new issue about fixing the parameter names for scalar_out and _array_out. I don't mind opening it and addressing it since I'm trying to get some as much GitHub exercise as I can manage. 

The Azure pipelines and travis-ci are failing for this build. Is that expected currently?

PR #805, opened 2019-10-30
Don't expose model parameter dictionaries to users

### LINKED ISSUE #789: rethink module-level model parameters dictionaries
#783 is adding a `iam.IAM_MODEL_PARAMS` dict that's simply sets of required parameters for IAM models. 

#682 added `temperature.TEMPERATURE_MODEL_PARAMETERS` that includes default values.

#548 added `pvsystem.DC_MODEL_PARAMS` that's also only sets of required parameters.

We should reconsider if we want to expose these dictionaries to users.

These dictionaries are used by methods in `PVSystem` and `ModelChain` to automagically determine what functions to call. In principle, user code might do something similar with them, but I doubt that happens.

The `TEMPERATURE_MODEL_PARAMETERS` dictionary with default values is probably the most useful to an end user because the default parameters are useful when calling `temperature` module functions directly (i.e. not through ModelChain/PVSystem).

My short term recommendation is to:

* deprecate `pvsystem.DC_MODEL_PARAMS` into `pvsystem._DC_MODEL_PARAMS`. I'd also be ok with just changing it without deprecation because 1. 0.7 is a big release and 2. this change won't effect many people and 3. My time is limited.
* change `iam.IAM_MODEL_PARAMS` to `iam._IAM_MODEL_PARAMS` to prevent problems before they start
* leave `TEMPERATURE_MODEL_PARAMETERS` alone

veronicaguo:
- [x] Closes  #789 

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.7.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.0.rst
@@ -77,7 +77,6 @@ API Changes
          - `pvsystem.physicaliam` is `iam.physical`
          - `pvsystem.ashraeiam` is `iam.ashrae`
          - `pvsystem.sapm_aoi_loss` is `iam.sapm`
-      - Created dict `iam._IAM_MODEL_PARAMS` to aid in identifying IAM models
    * Changes to `PVSystem` class
       - IAM models are provided by `PVSystem.get_iam` with kwarg `iam_model`.
       - Methods `PVSystem.ashraeiam`, `PVSystem.physicaliam` and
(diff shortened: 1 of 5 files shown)


docs/sphinx/source/whatsnew/v0.7.0.rst
```diff
@@ -77,7 +77,7 @@ API Changes
          - `pvsystem.physicaliam` is `iam.physical`
          - `pvsystem.ashraeiam` is `iam.ashrae`
          - `pvsystem.sapm_aoi_loss` is `iam.sapm`
-      - Created dict `iam.IAM_MODEL_PARAMS` to aid in identifying IAM models
+      - Created dict `iam._IAM_MODEL_PARAMS` to aid in identifying IAM models
```
wholmgren: I think we can delete this line entirely. No need to advertise the private dictionary's existence.

docs/sphinx/source/whatsnew/v0.7.0.rst
```diff
@@ -94,6 +94,7 @@ API Changes
   The `weather` argument of the above methods is now the first, required
   positional argument and the `times` argument is kept as the second keyword
   argument for capability during the deprecation period.
+* Parameter `pvsystem.DC_MODEL_PARAMS` is renamed to `pvsystem._DC_MODEL_PARAMS`
```
wholmgren: We do need to keep this one because it's changing from a public to a private attribute. Let's add another sentence like "Users should not rely on this dictionary's existence or structure."

docs/sphinx/source/whatsnew/v0.7.0.rst
veronicaguo: Yeah makes sense, will drop it.

PR #808, opened 2019-11-03
Fix clearsky Ineichen rounding

cedricleroy:

 - [ ] Closes issue #xxxx
 - [X] Fully tested. Added and/or modified tests to ensure correct behavior for all reasonable inputs. Tests (usually) must pass on the TravisCI and Appveyor testing services.

**Brief description of the problem and proposed solution (if not already fully described in the issue linked to above):**

Maybe this gets too deep into the weeds, but opening a PR in case we want to do something about it.

I originally followed the SE 73 pp. 307-317 [2] paper when I noticed a small discrepancy in the `b` term due to what seemed like a rounding. After writing the fix, I realized that the other paper SE 73 pp. 151-157 [1] match perfectly with PVLib current implementation. 0.163 seems to be a rounding of 0.16268, so opening the PR anyway even if the fix is probably up to debate.

We should perhaps include a comment explaining where it is coming from (similar to `bnci_2`) as `b` can be slightly different depending on which paper we are looking at.


cwhanse:
Agree we should explain the connection between code and sources. I'm having trouble finding the value 0.16268 in the Perez paper [2] - can you point me to where you found it?

cedricleroy:
[2] page 312:
image

0.83 * 0.196 = 0.16268

[1]:

image

cedricleroy:
Looks like a bunch of unrelated commits popped up in this PR, might not be ready to merge as is in master.

cwhanse:
> Looks like a bunch of unrelated commits popped up in this PR, might not be ready to merge as is in master.

That was my mistake. It is fixed in master but we may need to ask you to close and resubmit this PR.  

Thanks for the explanation. I should visit with my colleague who coded the original MATLAB version to confirm his reasons for rounding.

cwhanse:
@cedricleroy I talked with my colleague who originally wrote the MATLAB function and sorted through various discrepancies and typos in the two papers. We'll stick with [1] as the primary reference. For that reason I'd prefer to leave the factor as 0.163 rather than change to 0.16268. You are correct that the discrepancy between [1] and [2] is simply rounding to 3 digits. I don't know why that term in [1] is factored as it appears in [2], I'm guessing the authors of [2] have a physical interpretation of the leading constant 0.83 as the fraction extraterrestrial normal irradiance that arrives at the surface under clear sky conditions.

We can certainly add a comment explaining the coefficients in the code and their source. Rather than carry this PR forward (since it has unrelated commits due to my error), if you'd like to suggest a comment to add, close this PR and open a new one. Or post your suggestion and I'll make a PR for it.

cedricleroy:
Thanks @cwhanse. I followed up here: https://github.com/pvlib/pvlib-python/pull/814.

cwhanse:
Closing, superceded by #814

PR #815, opened 2019-11-07
Change units on SAPM effective irradiance from suns to W/m2

### LINKED ISSUE #472: Effective irradiance units
`pvsystem.sapm` expected `effective_irradiance` in units of suns (1 sun = 1000 W/m2).  `pvsystem.calcparams_desoto` expects `effective_irradiance` in W/m2.  I want to change to consistent units for effective irradiance.  

Preferences for suns or W/m2?

cwhanse:
- [x] Closes  #472

Changes units on `pvsystem.sapm` input `effective_irradiance` from suns to W/m2, for consistency with other DC models.  Deprecation is handled by raising a `pvlibDeprecationWarning` if all `effective_irradiance` values are less than 2.
Adds kwargs `irrad_ref` and `temp_ref` to `pvsystem.sapm`; these constants were hardwired.
Changes units on output of `pvsystem.sapm_effective_irradiance` from suns to W/m2.
Docstring enhancement for `pvsystem.sapm_effective_irradiance.


pvlib/pvsystem.py
```diff
@@ -1580,15 +1580,21 @@ def sapm(effective_irradiance, temp_cell, module):
...
+        Cell temperature [C].
 
     module : dict-like
         A dict or Series defining the SAPM parameters. See the notes section
         for more details.
 
+    irrad_ref : float (optional, default=1000)
+        Reference irradiance [W/m^2].
+
+    temp_ref : float (optional, default=25)
```
wholmgren: Will people find this kwarg useful for this model? I'd rather not expose it unless there's a demand.

pvlib/pvsystem.py
```diff
@@ -1659,12 +1665,19 @@ def sapm(effective_irradiance, temp_cell, module):
     temperature.sapm_module
     '''
 
-    T0 = 25
+    # TODO: remove this check after deprecation period for change in effective
+    # irradiance units, made in v0.7
```
wholmgren: we'll need to make a follow up issue for this and label it with the appropriate milestone

pvlib/pvsystem.py
```diff
@@ -1840,44 +1853,69 @@ def sapm_spectral_loss(airmass_absolute, module):
...
     aoi : numeric
-        Angle of incidence in degrees.
+        Angle of incidence. [degrees]
 
     module : dict-like
         A dict, Series, or DataFrame defining the SAPM performance
         parameters. See the :py:func:`sapm` notes section for more
         details.
 
-    reference_irradiance : numeric, default 1000
```
wholmgren: removed doc string but did not remove from function signature.

pvlib/pvsystem.py
```diff
@@ -1580,15 +1580,21 @@ def sapm(effective_irradiance, temp_cell, module):
...
 
     temp_cell : numeric
-        The cell temperature (degrees C).
+        Cell temperature [C].
 
     module : dict-like
         A dict or Series defining the SAPM parameters. See the notes section
         for more details.
 
+    irrad_ref : float (optional, default=1000)
```
wholmgren: I prefer `reference_irradiance` as used below. I don't see an entry in the variables style guide.

pvlib/pvsystem.py
wholmgren: Similar to above, I'd prefer something like `reference_temperature`.

pvlib/pvsystem.py
wholmgren: what's the logic behind removing the kwarg from this function and adding it to the other? I suggest we should consistently provide flexibility or consistently enforce W/m^2.

pvlib/pvsystem.py
```diff
@@ -1571,7 +1571,7 @@ def _parse_raw_sam_df(csvdata):
     return df
 
 
-def sapm(effective_irradiance, temp_cell, module):
+def sapm(effective_irradiance, temp_cell, module, irrad_ref=1000, temp_ref=25):
```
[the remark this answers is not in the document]
wholmgren: Are these new kwargs thoroughly tested? I only skimmed the tests but it appeared maybe not. The penance for new kwargs is more test permutations.

pvlib/pvsystem.py
cwhanse: `irrad_ref` and `temp_ref` are the current kwarg names for the same quantities in the Desoto, CEC and pvsyst functions. I'm OK changing them but let's do all together and not in this PR.

I do think that exposing the `temp_ref` quantity is useful, in a minority of cases. Some of the parameter sets in the old SAPM database are referenced to 50C rather than 25C.

pvlib/pvsystem.py
cwhanse: Removed from signature (was my oversight) to be consistent with the input expected by other irradiance-to-DC power functions, i.e., effective irradiance in W/m2. SAPM was the oddball.

pvlib/pvsystem.py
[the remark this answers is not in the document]
cwhanse: > It appears that irrad_ref sets the units, so the original behavior could be retained by supplying irrad_ref=1, correct? 

Yes, if irrad_ref=1 a user could supply `effective_irradiance` in units of suns and get a correct result. I don't follow the idea for the deprecation cycle. Are you suggesting that `pvsystem.sapm()` should continue to expect `effective_irradiance` in units of suns, or that it should allow `effective_irradiance` in either W/m2 or suns?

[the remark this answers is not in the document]
cwhanse: > Are these new kwargs thoroughly tested? I only skimmed the tests but it appeared maybe not. The penance for new kwargs is more test permutations.

They are not tested, and should be.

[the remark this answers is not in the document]
wholmgren: Maybe code will make my suggestion more clear...

```python
def sapm(effective_irradiance, temp_cell, module, irrad_ref=1.0, temp_ref=25):
    if irrad_ref == 1.0:
        warnings.warn('default reference irradiance is changing to 1000 in pvlib python 0.9. this has the effect of changing the expected units from suns to W/m^2. Set irrad_ref=1000 to silence this message and maintain future compatibility.'
    # the rest of the function
```

pvlib/pvsystem.py
cwhanse: #825

pvlib/pvsystem.py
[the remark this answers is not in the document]
cwhanse: @wholmgren see if the last 2 commits provide a softer deprecation. Accepting input in either suns or W/m2 seems awkward. I'll fix the test failures after we have concurrence on the approach.

pvlib/pvsystem.py
```diff
@@ -1839,45 +1856,87 @@ def sapm_spectral_loss(airmass_absolute, module):
 
 
 def sapm_effective_irradiance(poa_direct, poa_diffuse, airmass_absolute, aoi,
-                              module, reference_irradiance=1000):
-    """
+                              module, reference_irradiance=1.):
```
wholmgren: If I understand correctly, the function returns the same result as it used to if the function is called without `reference_irradiance`. But if the user did set `reference_irradiance`, then this code will always return a different result. I don't know if that's better or worse than a clean API break.

pvlib/pvsystem.py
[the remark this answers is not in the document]
wholmgren: no objection to the approach here in `sapm`, only concerned about `sapm_effective_irradiance`.

pvlib/pvsystem.py
cwhanse: You understand correctly, if I understand you correctly. 

Old behavior: inputs in W/m2, output in suns, `reference_irradiance=1000` W/m2. 

Interim behavior: input in W/m2, output in suns or W/m2, `reference_irradiance=1.` is used to signal the suns units on the output (which is a change I find awkward). An alternative is to change the default output unit to W/m2, `reference_irradiance=1000.` by default, and tell users to set `reference_irradiance=1.0` to get output in suns. That would change the default behavior.

Target Behavior after deprecation period: input and output in W/m2, and no reference irradiance needed, so that the output is ready as input for any of the DC power functions.

wholmgren: I can't say I like any of the options, so I'm ok with just removing the kwarg and changing the API. In that case, you could make an argument that we shouldn't bother with the deprecation cycle for `sapm`.

cwhanse: Neither did I. I changed back to W/m2 in/out in both functions, removed `reference_irradiance` from the effective irradiance function, and also removed `irrad_ref` and `temp_ref` as kwargs for `sapm`. We can expose those kwargs as `reference_irradiance` and `reference_temperature` when #825 renames throughout.

pvlib/pvsystem.py
```diff
@@ -1839,45 +1851,70 @@ def sapm_spectral_loss(airmass_absolute, module):
...
 
-    reference_irradiance : numeric, default 1000
-        Reference irradiance by which to divide the input irradiance.
-
     Returns
     -------
     effective_irradiance : numeric
-        The SAPM effective irradiance.
+        Effective irradiance accounting for reflections and spectral content.
+        Returned in units of reference_irradiance.
```
```suggestion
        Returned in units of W/m2.
```

pvlib/pvsystem.py
```diff
@@ -1659,12 +1660,23 @@ def sapm(effective_irradiance, temp_cell, module):
...
+    # reference_irradiance and expose
+    temp_ref = 25
+    irrad_ref = 1000
+    # TODO: remove this warning in v0.8 after deprecation period for change in
+    # effective irradiance units, made in v0.7
+    if np.all(effective_irradiance) < 2.0:
+        import warnings
+        warnings.warn('All effective_irradiance inputs to SAPM are low.'
+                      ' Units changed in v0.7 from suns to W/m2',
+                      pvlibDeprecationWarning)
```
wholmgren: I'm not sure if a deprecation warning is the right class of warning if the behavior is suddenly changing. Maybe `RuntimeWarning`?

pvlib/pvsystem.py
```diff
@@ -1659,12 +1660,23 @@ def sapm(effective_irradiance, temp_cell, module):
...
-    T0 = 25
+    # TODO: someday, change temp_ref and irrad_ref to reference_temperature and
+    # reference_irradiance and expose
+    temp_ref = 25
+    irrad_ref = 1000
+    # TODO: remove this warning in v0.8 after deprecation period for change in
+    # effective irradiance units, made in v0.7
+    if np.all(effective_irradiance) < 2.0:
+        import warnings
+        warnings.warn('All effective_irradiance inputs to SAPM are low.'
```
```suggestion
        warnings.warn('effective_irradiance inputs appear to be in suns.'
```

docs/sphinx/source/whatsnew/v0.7.0.rst
```diff
@@ -99,6 +99,11 @@ API Changes
   The `weather` argument of the above methods is now the first, required
   positional argument and the `times` argument is kept as the second keyword
   argument for capability during the deprecation period.
+* The `effective_irradiance` argument for :py:func:`pvsystem.sapm` now requires
+  units of W/m^2. Previously, units for this input were suns. A deprecation
+  warning is raised if all `effective_irradiance` values are extremely low.
```
wholmgren: revise warning type depending on resolution of comment below. revise 'extremely low' to something like "appear to be in suns (value < 2.0)"

docs/sphinx/source/whatsnew/v0.7.0.rst
```diff
@@ -99,6 +99,11 @@ API Changes
...
+* The output of :py:func:`pvsystem.sapm_effective_irradiance` is now in units
+  of W/m2 rather than suns.
```
wholmgren: I wonder if we need to separate the API changes section into deprecations and breaking changes.

PR #816, opened 2019-11-11
fix for *_TMY3.epw reading issue #782

### LINKED ISSUE #782: reading bug on '*_TMY3.epw' file
**Describe the bug**
When a tmy3 file formatted in epw is read, the precipitable_water is not converted from mm to cm, and it results in diverging results for spectral mismatch factor (like negative factor for instance), then resulting in negative 'effective_irradiance' results

**To Reproduce**
1. Create a ChainModel object with a .epw weather file, coming from TMY3 data and imported with 'pvlib.iotools.epw.read_epw()'
2. Run the model
3. Check for inconsistencies in effective_irradiance with 'chain1.effective_irradiance.loc[chain1.effective_irradiance<0]'

or 
1. Run the code attached
    
**Expected behavior**
All the effective_irradiance should be positive or null.

[bug_report_read_epw.txt](https://github.com/pvlib/pvlib-python/files/3672185/bug_report_read_epw.txt)

**Versions:**
 - ``pvlib.__version__``: 0.6.3+24.gbbb4580.dirty
 - ``pandas.__version__``: 0.23.4
 - python: 3.7.1

tylunel:
This PR is for solving epw file reading problem.
A note is added in docstring of epw.py and a validity check is added in atmosphere.first_solar_spectral_correction. This validity check is done according to kwargs min_pw and max_pw. If the value is higher, np.nan is returned.

 - [x] Closes  #782

### DIFF
--- a/pvlib/atmosphere.py
+++ b/pvlib/atmosphere.py
@@ -320,8 +320,9 @@ def gueymard94_pw(temp_air, relative_humidity):
     return pw
 
 
-def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
-                                    coefficients=None):
+def first_solar_spectral_correction(pw, airmass_absolute,
+                                    module_type=None, coefficients=None,
+                                    min_pw=0.1, max_pw=8):
     r"""
     Spectral mismatch modifier based on precipitable water and absolute
     (pressure corrected) airmass.
@@ -363,6 +364,14 @@ def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
 
     airmass_absolute : array-like
         absolute (pressure corrected) airmass.
+
+    min_pw : float, default 0.1
+        minimum atmospheric precipitable water (cm). A lower pw value will be
+        automatically set to this minimum value to avoid model divergence.
+
+    max_pw : float, default 8
+        maximum atmospheric precipitable water (cm). If a higher value is
+        encountered it will be set to np.nan to avoid model divergence.
 
     module_type : None or string, default None
         a string specifying a cell type. Can be lower or upper case
@@ -422,16 +431,18 @@ def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
     # *** Pwat ***
     # Replace Pwat Values below 0.1 cm with 0.1 cm to prevent model from
     # diverging"
-
-    if np.min(pw) < 0.1:
-        pw = np.maximum(pw, 0.1)
-        warn('Exceptionally low Pwat values replaced with 0.1 cm to prevent' +
-             ' model divergence')
+    pw = np.atleast_1d(pw)
+    pw = pw.astype('float64')
+    if np.min(pw) < min_pw:
+        pw = np.maximum(pw, min_pw)
+        warn('Exceptionally low pw values replaced with {0} cm to prevent '
+             'model divergence'.format(min_pw))
 
     # Warn user about Pwat data that is exceptionally high
-    if np.max(pw) > 8:
-        warn('Exceptionally high Pwat values. Check input data:' +
-             ' model may diverge in this range')
+    if np.max(pw) > max_pw:
+        pw[pw > max_pw] = np.nan
+        warn('Exceptionally high pw values replaced by np.nan: '
+             'check input data.')
 
     # *** AMa ***
     # Replace Extremely High AM with AM 10 to prevent model divergence
--- a/pvlib/iotools/epw.py
+++ b/pvlib/iotools/epw.py
@@ -67,9 +67,9 @@ def read_epw(filename, coerce_year=None):
     ===============   ======  =========================================
 
 
-    =============================       ==============================================================================================================================================================
+    =============================       ==================================================================================================================================================================================================================  # noqa: E501
     EPWData field                       description
-    =============================       ==============================================================================================================================================================
+    =============================       ==================================================================================================================================================================================================================
     index                               A pandas datetime index. NOTE, times are set to local standard time (daylight savings is not included). Days run from 0-23h to comply with PVLIB's convention
     year                                Year, from original EPW file. Can be overwritten using coerce function.
     month                               Month, from original EPW file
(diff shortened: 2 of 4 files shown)


tylunel:
Hi, 
I need some help to understand why stickler-ci finds lints only for the lines I have edited in epw.read_epw docstring and not for the others previously existing lines (but which have the same lint however: E501 line too long). 
And for the rest, any other feedback is welcomed as well.
Thanks in advance!

CameronTStark:
> I need some help to understand why stickler-ci finds lints only for the lines I have edited in epw.read_epw docstring and not for the others previously existing lines (but which have the same lint however: E501 line too long).

I believe this is to assign lint responsibility to the most recent editor. It would  interfere with usage of git-blame to troubleshoot code since it would insert a format change in that line that might obscure a bug. The alternative would be the CI making you responsible for more lines than you edited. You can imagine this being dangerous if done on running code as opposed to the documentation in your case. 

I'm sure it'll stop barking at you if you shorten those lines.

pvlib/test/test_atmosphere.py
```diff
@@ -128,6 +128,21 @@ def test_first_solar_spectral_correction_ambiguous():
         atmosphere.first_solar_spectral_correction(1, 1)
 
 
+def test_first_solar_spectral_correction_range():
+    with pytest.warns(UserWarning):
+        out = atmosphere.first_solar_spectral_correction(np.array([.1, 3, 10]),
+                                                         np.array([1, 3, 5]),
+                                                         module_type='monosi')
+        expected = np.array([0.96080878, 1.03055092,        nan])
```
cwhanse: I believe that this line and the next need to shift left. If `pytest.warns` works like `pytest.raises` the code exits the `with` block as soon as the `UserWarning` is raised, so the `assert_allclose` is never executed.  I could be mistaken.

pvlib/test/test_atmosphere.py
```diff
@@ -128,6 +128,21 @@ def test_first_solar_spectral_correction_ambiguous():
         atmosphere.first_solar_spectral_correction(1, 1)
 
 
+def test_first_solar_spectral_correction_range():
+    with pytest.warns(UserWarning):
```
cwhanse: We've started testing the content of the warning messages. Example [here](https://github.com/pvlib/pvlib-python/blob/a349bea38d928696671026dc1e126d44d8d2e529/pvlib/test/test_modelchain.py#L600). OK if you don't want to add, but I thought you'd like to see how it's being done.

pvlib/atmosphere.py
```diff
@@ -364,6 +365,14 @@ def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
     airmass_absolute : array-like
         absolute (pressure corrected) airmass.
 
+    min_pw : float, default 0.1
+        minimum atmospheric precipitable water (cm). A lower value will be
```
cwhanse: A lower `pw` value will be

pvlib/iotools/epw.py
```diff
@@ -67,9 +67,9 @@ def read_epw(filename, coerce_year=None):
     ===============   ======  =========================================
 
 
-    =============================       ==============================================================================================================================================================
+    =============================       ==================================================================================================================================================================================================================
```
cwhanse: If you have the time and interest to edit this table (which was cut/pasted from some document) for line length, it would be much appreciated. If not, that's OK, mark the complaining lines with `# noqa: E501`

pvlib/test/test_atmosphere.py
tylunel: Seems like it is different in this case. I have simply tried by changing the value of 'expected', and the test actually failed. So lines 136 and 137 are executed.

cwhanse: Good to know, `warns` behaves different from `raises`

tylunel:
Thanks for your feedbacks!
New version is pushed. I have made the suggested change, except for editing the table in docstring of read_epw() (sorry for that, I took the easy way to get around the problem ^^')

pvlib/atmosphere.py
```diff
@@ -423,15 +432,16 @@ def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
...
+        pw = np.maximum(pw, min_pw)
+        warn('Exceptionally low pw values replaced with {0} cm to prevent '
+             'model divergence'.format(min_pw))
 
     # Warn user about Pwat data that is exceptionally high
-    if np.max(pw) > 8:
-        warn('Exceptionally high Pwat values. Check input data:' +
-             ' model may diverge in this range')
+    if np.max(pw) > max_pw:
+        pw[pw > max_pw] *= np.nan
```
wholmgren: The slice syntax requires array input. I personally find it extremely useful to be able to use scalar input to understand functions. So, I suggest adding something like `pw = np.atleast1d(pw)`. Why `*=` instead of `=`?

pvlib/test/test_atmosphere.py
wholmgren: While not strictly required, the lines should shift left for style purposes. A `with` block should contain as little code as possible, in this case only the `first_solar_spectral_correction` call.

The difference between the `pytest.warns` and `pytest.raises` requirement is due to what happens within the actual library code (functions don't stop when they emit warnings, they do stop when the raise exceptions) - not anything to do with pytest.

pvlib/atmosphere.py
tylunel: Good point.

> Why *= instead of =?
because of a stupid copy and paste

pvlib/atmosphere.py
```diff
@@ -320,8 +320,9 @@ def gueymard94_pw(temp_air, relative_humidity):
     return pw
 
 
-def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
-                                    coefficients=None):
+def first_solar_spectral_correction(pw, airmass_absolute,
+                                    module_type=None, coefficients=None,
+                                    min_pw=0.1, max_pw=8):
```
wholmgren: Need to add a test or two that calls the function using these kwargs, ideally different from the default.

pvlib/atmosphere.py
```diff
@@ -422,16 +431,18 @@ def first_solar_spectral_correction(pw, airmass_absolute, module_type=None,
     # *** Pwat ***
     # Replace Pwat Values below 0.1 cm with 0.1 cm to prevent model from
     # diverging"
-
-    if np.min(pw) < 0.1:
-        pw = np.maximum(pw, 0.1)
-        warn('Exceptionally low Pwat values replaced with 0.1 cm to prevent' +
-             ' model divergence')
+    pw = np.atleast_1d(pw)
+    pw = pw.astype('float64')
```
wholmgren: why is the type casting necessary?

tylunel: It is necessary for the case when the input is an integer (or array of integers). In this case, the data type of the array is int32 and is not compatible with np.nan (which is a float) for the executing of `pw[pw > max_pw] = np.nan`. It raises an error if I don't cast the type here.

Issue #802, opened 2019-10-29
use `xfail` decorator on `test_get_psm3`


wholmgren:
The ongoing failure of `test_get_psm3` is conditioning us (or at least me*) to accept PRs with failing tests. We've been told NREL is going to fix their API, but it keeps not happening, so I really think we should go ahead and mark this test with `pytest.xfail`. Perhaps `pytest.xfail(strict=True)` so that the test suite alerts us to when NREL does fix it. 

@CameronTStark can you fix this if you agree it makes sense?

* I wrongly approved #800 and I've not been checking #793 because I've been thinking that the test failures were only because of `test_get_psm3`.

mikofski:
So sorry all for any inconveniences or issues as a result of the continued `test_get_psm3` failure!
Thanks @CameronTStark and @wholmgren for cleaning up after my PR puppy!

![image](https://user-images.githubusercontent.com/1385621/67796604-8b634a80-fa3d-11e9-87b2-ce287924aab4.png)

PR #842, opened 2019-12-23
refactor get_psm3 code into parse_psm3, read_psm3

### LINKED ISSUE #841: Add ability to read local NSRDB PSM3 CSVs
Unless I've missed something in the docs, the only pvlib function that parses PSM3 CSVs is `iotools.get_psm3`, which requires making a network request to the NSRDB PSM3 API.  It would be nice to be able to read local PSM3 files (e.g. those downloaded manually from the [NSRDB Viewer](https://maps.nrel.gov/nsrdb-viewer/)) in addition to requesting them from the web API.  Here are a couple proposals:

- Break out the parsing code in `iotools.get_psm3` into its own function `iotools.read_psm3`, leaving `get_psm3` as a networking wrapper around the PSM3 API that then calls `read_psm3` on the API response.  
- Modify `get_psm3` to check if the input is a filename instead of API parameters and retrieve data as appropriate. 

I'd favor the former since it separates networking code from parsing code.  Some other `iotools.read_*` functions can make web requests, but those are just simple urls and more or less equivalent to reading a local file whereas `get_psm3` has to construct a full API request.

kanderso-nrel:
- [x] Closes #841 
 - [x] Tests added

Basically just cut/pasting the PSM3 parsing code into a new function.  I also added some metadata fields that were missing in the docstring (maybe they were added to PSM3 exports after those docs were written).  Similar to `iotools.read_midc` and `pd.read_csv`, `read_psm3` can accept a filename (for users) or a file-like buffer (for `get_psm3`).  Let me know if you think it would be better to have separate functions for files and buffers.


kanderso-nrel:
Why is the get_psm3 test currently expected to fail?  I see #742 but #802/#803 seems unrelated.  Is it just that the API TMYs were updated when 2018 data was added to the NSRDB and the test CSV is out of date because it was created before the update?  I notice that the TMY data currently returned from the API does include a month from 2018.

cwhanse:
> Why is the get_psm3 test currently expected to fail?  I see #742 but #802/#803 seems unrelated. 

I can't say why, I think we just got weary of the test failures for unrelated PRs. After the 30minute shift #742 (summer of 2018) there was a period of frequent server outages, then new failures which I (for one) didn't investigate. Here's one [example](https://travis-ci.org/pvlib/pvlib-python/jobs/604608639?utm_medium=notification&utm_source=github_status); it appears that PSM3 is returning 14 columns now where the test data is 15 columns.

kanderso-nrel:
Thanks.  I did some poking around:
- > that PSM3 is returning 14 columns now where the test data is 15 columns

    I believe this is not the API's fault; the expected dataframe includes a spurious 'Unnamed: 0' column for the index in the CSV.  The actual data columns are the same between the API response and the test data.

- Running `get_psm3` with `names='tmy-2016'`, `'tmy-2017'`, and `'tmy-2018'` return fewer rows than expected (~4380, depending on the year).  The timestamps are 2hrs apart, but they sometimes jump around.  I think this must be an error in the PSM3 API, though to be fair it is undocumented that this endpoint accepts `tmy*` at all. 
- `'tmy'` returns the correct number of rows and contains mostly the same data as the pvlib test file with the exception of a couple of months.  I'm guessing that the TMY datasets were recalculated including 2018 in the pool of years and those two months changed. 
- However, changing URL from `/api/solar/nsrdb_psm3_download.csv` to the TMY endpoint `/api/nsrdb_api/solar/nsrdb_psm3_tmy_download.csv` and fetching `'tmy-2017'` does return the same data as the test file -- the entire `get_psm3` test passes except for the test for bad input interval since the TMY endpoint doesn't pay attention to the interval parameter.  

From this, I think the conclusion is that we're using the wrong endpoint for TMY requests and we're comparing an old TMY against a new TMY.  I think there are a couple options to get the code+tests back to passing, though both require pinning the test against tmy-2017 so that it doesn't have a moving target:
- Wait for the current API endpoint to start returning reasonable data for `tmy-2017`.  Not necessarily clear that this will ever happen. 
- Change `get_psm3` to make the request to the TMY endpoint for `tmy*` requests and the current endpoint for single-year requests.  I think this is what the NSRDB documentation would support.

pvlib/iotools/psm3.py
```diff
@@ -9,7 +9,8 @@
 import pandas as pd
 from json import JSONDecodeError
 
-URL = "http://developer.nrel.gov/api/solar/nsrdb_psm3_download.csv"
+PSM_URL = "http://developer.nrel.gov/api/solar/nsrdb_psm3_download.csv"
+TMY_URL = "http://developer.nrel.gov/api/nsrdb_api/solar/nsrdb_psm3_tmy_download.csv"  # noqa
```
mikofski: Suggestion, not blocker. There's some repeat text in the URLs, perhaps add `NREL_API`:
```python
NREL_API = 'http://developer.nrel.gov/api/'
PSM_URL = NREL_API + "solar/nsrdb_psm3_download.csv"
TMY_URL = NREL_API + "nsrdb_api/solar/nsrdb_psm3_tmy_download.csv"
```
This also results in shorter lines without using `#noqa`.

Also can you double check the URL scheme? I'm a little surprised they are not using HTTPS for endpoints that transmit credentials

kanderso-nrel:
> pinning the test against tmy-2017 ... Change get_psm3 to make the request to the TMY endpoint for tmy* requests and the current endpoint for single-year requests. 

I've added the above to this PR.  `get_psm3` now chooses which API endpoint to use based on whether the `names` parameter is a single year or a TMY dataset.  The TMY and single-year flows are tested separately now as well.  As far as I can tell, the API functionality and tests behave as expected now that we're comparing the test data file against `tmy-2017` with the TMY endpoint.  

I also added a whatsnew for v0.7.1.  I think this is ready for a detailed review whenever folks get a chance!

pvlib/iotools/psm3.py
```diff
@@ -154,30 +141,132 @@ def get_psm3(latitude, longitude, api_key, email, names='tmy', interval=60,
...
+    .. [2] `Standard Time Series Data File Format
+       <https://rredc.nrel.gov/solar/old_data/nsrdb/2005-2012/wfcsv.pdf>`_
+    """
+    if hasattr(filename, 'readline'):
+        # if passed a file-like object, not our job to close it
+        close = False
+        fbuf = filename
+    else:
+        close = True
+        fbuf = open(filename, 'r')
```
mikofski: Is [duck typing](https://en.m.wikipedia.org/wiki/Duck_typing) possible here instead of `if-then`? I believe it would be more Pythonic:
```python
try:
    # always use binary, in case of Windows?
    fbuf = open(filename, 'rb')
except (OSError, IOError):
    pass
```

Also you can use `with` if you delegate passing the contents to a private function, _EG_:

```python
def _parse_psm3(fbuf):
    # The first 2 lines of the response are headers with metadata
    header_fields = fbuf.readline().split(',')

    # etc. all the stuff after try
    return contents

def read_psm3(fbuf):
    """..."""
    try:
        with = open(filename, 'rb') as fbuf:
            content = _parse_psm3(fbuf)
    except (OSError, IOError):
        content = _parse_psm3(fbuf)
    return content
```

pvlib/iotools/psm3.py
kanderso-nrel: Good suggestions.  I've split out the domain from the endpoint paths as you suggest.

The example python snippets in the docs use HTTP but it looks like HTTP and HTTPS both work. I've switched to HTTPS.

pvlib/iotools/psm3.py
kanderso-nrel: Agreed that the private `_parse_psm3` function is better.

I'll nitpick that using `hasattr(obj, 'method')` is an example of duck typing since it's asking "does `filename` `readline()` like a duck?", but I agree that trying to open the file would be more pythonic -- I'm just not sure that combining it with `with open...` is a good idea.  The try/except is wrapping the `_parse_psm3` call as well as the `open()` call when really we only want to catch errors from `open()` and I'm not sure it's necessarily possible to only catch errors from the `open` call.  Catching errors from both would cause the function to try to interpret a filename as a file-like object instead of raising a useful error.  For instance calling `open` on a `StringIO` raises a TypeError so we would have to trap `TypeError`, causing any TypeError raised in `_parse_psm3` to make the code try going down the other path and raise `AttributeError: 'str' object has no attribute 'readline'` as the primary error -- not great.  

I'm not opposed to using try/except for duck typing, but I'm not sure there's a way to safely combine it with a `with open(..) as fbuf:` call.  I don't like storing file handles outside of `with` and keeping track of whether or not to close the object, but I can't think of a better way to separate errors from `open()` and `_parse_psm3`:

```
def read_psm3(fbuf):
    """..."""
    try:
        # using rb causes error later on with fbuf.readline().split(',')
        fbuf = open(filename, 'r')
        close = True
    except TypeError:
        fbuf = filename
        close = False
    finally:
    
    try:
        content = _parse_psm3(fbuf)
    finally:
        if close:
            fbuf.close()
    return content
```

I know using try/except is generally looked on more favorably than attribute testing (EAFP vs LBYL), but to me it seems better to just ask the object "can I `readline` you?" than try to figure out what exceptions `open` might raise.  I'm not even sure that the above code is correct -- if `open` can raise TypeError in other situations besides when we give it a StringIO, we'd want to propagate that to the user.

In case you're interested, I got the `hasattr` pattern from [pandas's CSV writing code](https://github.com/pandas-dev/pandas/blob/master/pandas/io/formats/csvs.py#L177), though I think the additional complexity of that situation warrants the more flexible solution. 

Also RE: the file mode -- using 'rb' would require some other changes since you have to split a bytes object with a bytes object, not a str.  The tests pass on windows for me, so do we need to worry about it?

mikofski: Ok, I agree, "simple is better than complex, but complex is better than complicated." Looks good as is. I'll try to get through the rest later today. Thanks again!

cwhanse:
> I believe this is not the API's fault; the expected dataframe includes a spurious 'Unnamed: 0' column for the index in the CSV.  The actual data columns are the same between the API response and the test data.

Since that [static file](https://github.com/pvlib/pvlib-python/blob/master/pvlib/data/test_psm3.csv) was downloaded from the NSRDB and used without modification, I'm guessing there was a bug in the NSRDB that has been corrected.
 
> 'tmy' returns the correct number of rows and contains mostly the same data as the pvlib test file with the exception of a couple of months.  I'm guessing that the TMY datasets were recalculated including 2018 in the pool of years and those two months changed.

What? Are TMY calculated on a rolling basis now?

> However, changing URL from /api/solar/nsrdb_psm3_download.csv to the TMY endpoint /api/nsrdb_api/solar/nsrdb_psm3_tmy_download.csv and fetching 'tmy-2017' does return the same data as the test file -- the entire get_psm3 test passes except for the test for bad input interval since the TMY endpoint doesn't pay attention to the interval parameter.
> 
> From this, I think the conclusion is that we're using the wrong endpoint for TMY requests and we're comparing an old TMY against a new TMY.  I think there are a couple options to get the code+tests back to passing, though both require pinning the test against tmy-2017 so that it doesn't have a moving target:

I'm confused again - what is a "tmy-2017" file? is this the NSRSDB data for the year 2017, or a TMY file calculated with the year 2017 as a boundary? If its the former, that's an unfortunate re-use of "tmy".

> Wait for the current API endpoint to start returning reasonable data for tmy-2017.  Not necessarily clear that this will ever happen.

The current API endpoint is what is named in the [example script](https://nsrdb.nrel.gov/data-sets/api-instructions.html).

> Change get_psm3 to make the request to the TMY endpoint for tmy* requests and the current endpoint for single-year requests.  I think this is what the NSRDB documentation would support.

I can't tell what we are supposed to get since it appears the webpages are changing again [NSRDB download instructions](https://developer.nrel.gov.docs/solar/nsrdb/) and the earlier comment about TMY being different after 2018 conflicts with the TMY description (1998 - 2017).

Thanks for taking on the work to sort this out. If you have better access to the NSRDB team maybe you can encourage them to adopt some more transparent practices; versioning and release notes would go a long way.

kanderso-nrel:
@cwhanse Sorry for not being more clear, I hope this clarifies what I've dug up:

> Are TMY calculated on a rolling basis now?

I believe this is indeed the case -- as individual years get added to the PSM3 dataset, the TMYs are regenerated to consider the new data.  I can't find explicit documentation of this, but it is implied in the [TMY endpoint docs](https://developer.nrel.gov/docs/solar/nsrdb/psm3_tmy_data_download/):

- Allowed values for the `names` parameter:
    - `tmy, tmy-2016, tmy-2017, tdy-2017, tgy-2017. tmy-2018, tdy-2018, tgy-2018`
- Description for the `names` parameter::
    - The TMY dataset for which data should be extracted. **'tmy' will return the most recently generated TMY file.** (emphasis mine)

And in Table 2 here: https://nsrdb.nrel.gov/about/tmy.html

- Description of the Year header in a TMY file:
    - `TMY/TGY/TDY2016 1998-2016; TMY/TGY/TDY2017 1998-2017` 

So when speaking of TMY data, it's important to distinguish between the "current TMY" and older TMYs.

> what is a "tmy-2017" file? is this the NSRSDB data for the year 2017, or a TMY file calculated with the year 2017 as a boundary?

The latter.  "tmy-2017" is the TMY calculated from the 1998-2017 dataset.  The single-year NSRDB data would just be called "2017".  Because Mark wrote the pvlib test code in May 2019 before 2018 data was added to the NSRDB, requesting the 'tmy' dataset returned the TMY calculated from 1998-2017 data.  Now that 2018 is included, 'tmy' refers to the 1998-2018 TMY, so we need to test the static file against the original 'tmy-2017' dataset.  

> The current API endpoint is what is named in the example script.

Yes, which is consistent with this PR because the NREL example script is requesting a single year (2010), not a TMY.  To be clear, I do think pvlib's current endpoint is correct for single-year requests.  I'm proposing to leave that endpoint as-is for single years but change the endpoint for TMY requests because only the TMY endpoint returns correct data for older TMY datasets.

pvlib/iotools/psm3.py
```diff
@@ -74,50 +83,25 @@ def get_psm3(latitude, longitude, api_key, email, names='tmy', interval=60,
...
-    * Temperature Units
-    * Pressure Units
-    * Wind Direction Units
-    * Wind Speed
-    * Surface Albedo Units
-    * Version
-
-    The second item is a dataframe with the timeseries data downloaded.
+         '2014', '2015', '2016', '2017', '2018', 'tmy', 'tmy-2016', 'tmy-2017',
+         'tdy-2017', 'tgy-2017', 'tmy-2018', 'tdy-2018', 'tgy-2018']
```
cwhanse: This list is going to be outdated in ~6 months, but I don't have a better idea for describing the values. The information on the NSRDB websites are also out of date.

wholmgren:
I'm -1 on putting all of the useful parsing code inside a giant try block. I'd put that block in a public function `parse_psm3` that requires a buffer as an input. Maybe add a usage example for reading a saved file.

That being said, `iotools` is meant to be easy to contribute to, so if you all disagree and/or would rather move on with life then I won't get in the way.

mikofski:
Similar [comment earlier](https://github.com/pvlib/pvlib-python/pull/842#discussion_r361526849) @kanderso-nrel seemed in partial agreement.

Similar [comment earlier](#discussion_r361526849)

kanderso-nrel:
I agree that it would be good to separate parsing logic from file io logic, but I combined reading files and buffers into one function to preempt possible objections to:
1) the API bloat of having so many functions for just PSM3 files
2) inconsistency with `iotools.read_midc`, which presents a single public function for both filenames and file-like objects

If those objections don't exist, I'm happy to "Split out PSM3 parsing code into its own function ~read_psm3~ parse_psm3" again 😄

So to confirm, the overall structure would be three public functions:
- `get_psm3`: make API call, pass buffer to `parse_psm3`
- `read_psm3`: expect filename, pass file object to `parse_psm3`
- `parse_psm3`: expect buffer, return parsed header+df

Out of scope for this PR, but an idea for the future: apply a decorator to all of the `iotools` parsing functions that translates filenames to buffers as needed.

wholmgren:
I personally don't see the need for `read_psm3`. This is python so I think it's reasonable to expect users to supply their own `with` block if they have a file.

pvlib/iotools/psm3.py
kanderso-nrel: I think these values are up to date in the developer docs at least: [TMY endpoint](https://developer.nrel.gov/docs/solar/nsrdb/psm3_tmy_data_download/#request-parameters) and [Single-Year endpoint](https://developer.nrel.gov/docs/solar/nsrdb/psm3_data_download/#request-parameters).  Side note, these pages are developed on GH: https://github.com/NREL/developer.nrel.gov/commit/5c03e14815f4a58dab36dc7584585672b2060115

I just put in a request for the [TMY Description](https://nsrdb.nrel.gov/about/tmy.html) to be updated to include 2018.

cwhanse:
> So to confirm, the overall structure would be three public functions:
> 
> get_psm3: make API call, pass buffer to parse_psm3
> read_psm3: expect filename, pass file object to parse_psm3
> parse_psm3: expect buffer, return parsed header+df

This seems like a good pattern to migrate `iotools` towards. If you are looking for a tie-breaker, I'd vote for the `parse_psm3` function. Public preferred, but not a strong preference.

cwhanse:
The latest push does away with `read_psm3` - I thought we were discussing only moving the parsing code to a new function `parse_psm3`?

kanderso-nrel:
@cwhanse I guess I misunderstood -- I took Will's "I personally don't see the need for read_psm3" and your "If you are looking for a tie-breaker, I'd vote for the parse_psm3 function" to be in favor of removing read_psm3 for this PR, not just adding parse_psm3.  I don't care one way or the other, though I am generally in favor of passing file-like objects around instead of filenames.  Do you want both `read_psm3` and `parse_psm3` to be included as it was listed above?  I'll add it back if so.  

> maybe I should have held back!

@wholmgren Generally I appreciate that not voicing minor objections can be helpful to reduce the barrier to contributing, but don't hold back on my account -- I'm always happy for feedback!

cwhanse:
I could have been more clear. Yes, I think both `read_psm3` and `parse_psm3` have value, the former as a convenience for those less experienced with python.  All it would do is open the file and pass the object to `parse_psm3` as @mikofski suggests.

mikofski:
>inconsistency with `iotools.read_midc`, which presents a single public function for both filenames and file-like objects

I'm not sure what the best paradigm is, it would be nice to have consistency. I think some users may struggle with files and string/byte buffers. _EG_ if using `get_psm3()` it returns a string not a buffer.

I already approved as is (before last commit), I don't have any blockers, I would err toward pushing something now, testing with users, and making changes later as needed, I think this PR may be starting to take on two different issues:
1. the tmy issue
2. how to deal with filenames, strings, and buffers

Sorry

kanderso-nrel:
I'm fine with leaving a broader conversation about iotools inputs for another day.  The current state of this PR is as listed in https://github.com/pvlib/pvlib-python/pull/842#issuecomment-570309454: three PSM3 functions total -- I hope everyone is happy with this.  Thanks for the input and reviewing!

cwhanse:
> Thanks @kanderso-nrel! @cwhanse ok to merge?

Yes

PR #844, opened 2020-01-02
Fix error in the irradiance unit checker.

dzimmanck:
This is a fix to a very simple error, so I am bypassing the checklist.  It is a one line correction to what is clearly a simple error.

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1678,7 +1678,7 @@ def sapm(effective_irradiance, temp_cell, module):
     irrad_ref = 1000
     # TODO: remove this warning in v0.8 after deprecation period for change in
     # effective irradiance units, made in v0.7
-    if np.all(effective_irradiance) < 2.0:
+    if np.all(effective_irradiance < 2.0):
         import warnings
         warnings.warn('effective_irradiance inputs appear to be in suns.'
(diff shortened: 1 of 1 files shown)


wholmgren:
Thanks! It is a simple error, but it means the deprecation warning does not work as intended. We should make a 0.7.1 release very soon. I suggest we go ahead and merge this as is and add a note when #842 is merged.

adriesse:
Thanks for fixing this.  I had wondered about the deprecation messages I was getting but hadn't had time to investigate.  I do think that all zeros should not trigger the message (ideally).

wholmgren:
>  I do think that all zeros should not trigger the message (ideally).

That seems reasonable in principle. Do you actually run into this problem?

cwhanse:
>> I do think that all zeros should not trigger the message (ideally).
> 
> That seems reasonable in principle.

Want me to refine the test condition to exclude 0.0 and NaN?

wholmgren:
> Want me to refine the test condition to exclude 0.0 and NaN?

sure, let's hold 0.7.1 for that. I don't think it needs extensive testing - let's just make sure it looks right this time.

adriesse:
> Do you actually run into this problem?

Not in typical time series, but when comparing models/graphing performance matrices, single or multiple zeroes do come up.

PR #854, opened 2020-01-10
Improves #844 

cwhanse:
Checks for irradiance < 2.0 only when irradiance > 0.0 and not NaN


pvlib/pvsystem.py
```diff
@@ -1679,12 +1679,12 @@ def sapm(effective_irradiance, temp_cell, module):
...
     # effective irradiance units, made in v0.7
     with np.errstate(invalid='ignore'):  # turn off warning for NaN
-        if np.all(
-            effective_irradiance[np.where(effective_irradiance > 0.0)] < 2.0):
-                import warnings
-                warnings.warn('effective_irradiance inputs appear to be in',
-                              ' suns. Units changed in v0.7 from suns to W/m2',
-                              RuntimeWarning)
+        ee = effective_irradiance
+        if np.all(ee[ee > 0.0] < 2.0):
```
```suggestion
        if np.all(np.logical_and(ee > 0.0, ee < 2.0)):
```

wholmgren: this should preserve compatibility with scalar input

kanderso-nrel: Not to continue complicating this simple warning, but `np.all(np.logical_and(ee > 0.0, ee < 2.0))` will return False if ANY zeros are present, which doesn't seem desirable.  Would using `ee = np.asarray(effective_irradiance)` with Cliff's code work?

adriesse: How about:
```(np.nanmin(ee) > 0.0) and (np.nanmax(ee) < 2.0)```

Comment (cwhanse) [reply pvlib/pvsystem.py]: Thanks for the helpful suggestions.

**Comment by kanderso-nrel** [comment]
Maybe I'm misunderstanding the purpose of this PR, but just in case -- I don't think using `np.logical_and` with `ee > 0.0` is quite right, or at least not the whole picture.  For example I think we would want to raise a warning for these two cases but with the current state of the PR, no warning is emitted:

```
effective_irradiance = np.array([0, 0.5, 1.0])
effective_irradiance = np.array([np.nan, 0.5, 1.0])
```
Here's a version that I think behaves correctly:

```
import numpy as np

cases = [
    # warn for these
    np.array([np.nan, 0.5, 1.0]),
    np.array([0, 0.5, 1.0]),
    0.5,
    # not for these
    np.array([0, 0, 0]),
    np.array([np.nan, np.nan, np.nan]),
    np.array([0, 100, 1000]),
    np.array([np.nan, 100, 1000]),
    0,
    np.nan,
    100,
]

for effective_irradiance in cases:
    with np.errstate(invalid='ignore'):  # turn off warning for NaN
        ee = np.asarray(effective_irradiance)
        ee_gt0 = ee[ee > 0.0]
        if ee_gt0.size > 0 and np.all(ee_gt0 < 2.0):
            # raise warning
            print(ee)
```

cwhanse:
You understand the intent correctly. Thanks for the fix!

PR #856, opened 2020-01-12
Interactive example gallery pages

kanderso-nrel:

Experimenting with embedding interactive bokeh-based pages in the example gallery.  I'm using the bokeh-plot sphinx extension; I originally tried to use the bokeh-gallery sphinx extension but linking doesn't work correctly between the gallery index and example pages when combining it with sphinx-gallery... something to do with whether the index page expects to be in the same directory as the pages or not. 

https://external-builds.readthedocs.io/html/pvlib-python/856/gallery/plot_single_axis_tracking.html#interactive-demo

### DIFF
--- a/docs/examples/interactive_examples/interactive_singlediode.py
+++ b/docs/examples/interactive_examples/interactive_singlediode.py
@@ -1,0 +1,127 @@
+from bokeh.io import curdoc
+from bokeh.layouts import column, row
+from bokeh.models import (
+    ColumnDataSource, Select, CustomJS, ColorBar, LinearColorMapper
+)
+from bokeh.transform import transform
+from bokeh.plotting import figure, show
+
+import pandas as pd
+import numpy as np
+from pvlib import pvsystem
+
+irradiance_range = np.arange(0, 1300, 50)
+temperature_range = np.arange(-10, 70, 10)
+
+parameters = {
+    'alpha_sc': 0.004539,
+    'a_ref': 2.6373,
+    'I_L_ref': 5.114,
+    'I_o_ref': 8.196e-10,
+    'R_s': 1.065,
+    'R_sh_ref': 381.68,
+}
+
+dataset = []
+for irrad in irradiance_range:
+    for temp in temperature_range:
+
+        IL, I0, Rs, Rsh, nNsVth = pvsystem.calcparams_desoto(
+            irrad,
+            temp,
+            alpha_sc=parameters['alpha_sc'],
+            a_ref=parameters['a_ref'],
+            I_L_ref=parameters['I_L_ref'],
+            I_o_ref=parameters['I_o_ref'],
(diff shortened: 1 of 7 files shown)


kanderso-nrel:
I know there's still some stickler issues but before I go too far, I wonder what folks think about this kind of interactive example?  There are two in the PR docs now; one on SAT backtracking and another on generating IV curves.

I'm finding them less useful than I'd like because they are restricted to running only JS at runtime, meaning all pvlib calls have to be done up front.  So for instance if we wanted to make an interactive IV curve generator like [this one](https://www.pveducation.org/pvcdrom/solar-cell-operation/shunt-resistance), we'd either have to code up the all the SDM math in JS (a non-starter) or generate every possible IV curve ahead of time and ship them all with the page so that they're available for display.

mikofski:
Sounds like too much work. Can it be static vs. interactive? Save it for another day? Sorry if I sent you on a wild goose chase

cwhanse:
I'm a fan of PVEducation.org and I think that interactive content is valuable. In my view, the gallery is intended for scripts, notebooks etc. that illustrate use of pvlib-python. I'm not convinced that interactive content provided enough additional value to warrant the effort.

kanderso-nrel:
Well, I wasn't trying to assert that there's no good application for interactivity in the pvlib docs.  My question is more "given these limitations, does anyone have the imagination to see a good use case?"

I also wonder if publishing interactive examples is straying outside of pvlib's scope.  For instance, is it pvlib's job to demonstrate what backtracking is and does, or just provide an implementation and link to a reference? 

> Can it be static vs. interactive? 

Can you clarify this question?  The bokeh plots could be generated without any interactivity, but at that point I don't see any benefit over just doing it with matplotlib. 

> a wild goose chase

No worries -- if nothing else, it's good to have an example of this sort of thing in case a good use case pops up in the future.

mikofski:
One option might be to leave it where it is, and just link to it from pvlib?

wholmgren:
Cool idea, but I'm worried about maintenance. I generally find bokeh to be a pain. But if you all are up to committing to maintaining it for the long run then we can consider.

kanderso-nrel:
I think if nobody has an idea for it that really shines, I agree with @cwhanse and it's a better use of time to make more (static) examples.  So I'd say let's close this PR and just keep the possibility in mind for the future.  

Any interest in a PR for just the static portion of the [IV curves example](https://external-builds.readthedocs.io/html/pvlib-python/856/gallery/plot_singlediode.html#sphx-glr-gallery-plot-singlediode-py)?  I'll admit to having almost no experience with those functions, so no worries if it's not worth salvaging.

cwhanse:
> Any interest in a PR for just the static portion of the IV curves example? 

Yes! @bmeyers

bmeyers:
Hi, yes, this certainly looks like it would be very helpful!

PR #859, opened 2020-01-15
reorganize tests into subfolders and use pathlib for conftest DATA_DIR

### LINKED ISSUE #848: ENH: MAINT: STY: group iotool tests into folder
there are so many tests now, would it be okay to group some of them, eg iotools into a folder?

move all of the iotools tests into a folder called `iotools_tests` inside the `pvlib/test` folder

just keep letting the number of files in test grow until someone else thinks it might be useful to group them

**Additional context**
[pandas/tests](https://github.com/pandas-dev/pandas/tree/master/pandas/tests) are grouped in folders, but forgo the `_tests` suffix, it does seem redundant, but will pytest still work without it? I'm not set on any particular convention, as long as it's consistent

so what do folks think? OK to post here? Should I have posted in pvlib-python google group instead? thx!

mikofski:
* closes #848
* create pvlib/tests/iotools and move all iotools tests into that
subfolder
* use fixtures in test_ecmwf_macc.py for expected_test_data
* use conftest.data_dir instead of redundant boilerplate process of
retrieving test folder for each tests, mostly for iotools
* fix test_ivtools.py was using full path to conftest but tests is not a
python package, so no
* change "pvlib/test" -> "pvlib/tests" to conform to popular convention

Signed-off-by: Mark Mikofski <bwana.marko@yahoo.com>

 - [x] Tests added


mikofski:
FYI: no build error, all tests pass, but coveralls said something weird, I think we can ignore. Is "what's new" required?

cwhanse:
> Is "what's new" required?

I think a note about the reorganization is warranted, if only to describe this new test structure for future contributors.

Thanks for the clean up with those directory constants.

mikofski:
@wholmgren and @cwhanse ready for review ✔️

pvlib/tests/iotools/test_pvgis.py
```diff
@@ -8,10 +8,9 @@
 import pytest
 import requests
 from pvlib.iotools import get_pvgis_tmy
+from conftest import data_dir
 
-TESTS = Path(__file__).parent
-PROJECT = TESTS.parent
-DATA = PROJECT / 'data'
+DATA = Path(data_dir)
```
cwhanse: Would it be better (or not) to use the `os.path.join` pattern here instead of importing and using `Path`? I don't have a strong preference except for consistency.

mikofski: Good point. I was just playing around with Python-3, because always wanted to start using `Pathlib` and Python-3 only features more. I think `os` is an old library, so maybe it's slightly more Pythonic to use `Pathlib`, but I'm not sure. I'm open with doing whatever for consistency, but here are a few posts in favor of `Pathlib`:
* [**Why you should be using pathlib**](https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/)
* [**No really, pathlib is great**](https://treyhunner.com/2019/01/no-really-pathlib-is-great/)
* [**Python 3's pathlib Module: Taming the File System**](https://realpython.com/python-pathlib/)
* [**Is pathlib a viable replacement for os.path?**](https://python-forum.io/Thread-Is-pathlib-a-viable-replacement-for-os-path)
* [**Easier Python paths with pathlib**](https://www.linuxjournal.com/content/easier-python-paths-pathlib)
* [**Why use the Pathlib Module?**](https://stackabuse.com/introduction-to-the-python-pathlib-module/#whyusethepathlibmodule)
* etc. I just [Googled `Pathlib` vs. `os`](https://www.google.com/search?q=os.path.join+vs+pathlib)


There is one solid benefit to using `Path` objects over os, which is that it has the `open` method built in, so the old pattern:

    with open(os.path.join('path', 'to', 'file')) as f:
        # do stuff with f

becomes

    with (Path('path') / 'to' / 'file').open() as f
        # do stuff with f

which seems a little clearer perhaps? You can see this in action right here:
https://github.com/pvlib/pvlib-python/blob/d635c09123319c68182898001712377089199744/pvlib/test/test_pvgis.py#L61-L64

Pretty cool, no?

wholmgren:
+1 for `pathlib`. We're python 3.5+ now, so no worries about compatibility.

docs/sphinx/source/whatsnew/v0.7.1.rst
```diff
@@ -26,6 +26,8 @@ Testing
 ~~~~~~~
 * Added single-year PSM3 API test for `iotools.get_psm3`.
 * Added tests for `iotools.parse_psm3` and `iotools.read_psm3`.
+* Change test folder to `tests` and reorganize tests into subfolders, _eg_
```
wholmgren:
```suggestion
* Change `pvlib/test` folder to `pvlib/tests` and reorganize tests into subfolders, _eg_
```

pvlib/tests/iotools/test_pvgis.py
wholmgren: what about defining `data_dir` and `test_dir` as `Path` objects in `conftest`? Seems like that would avoid a lot of `os` and `pathlib` imports in individual test modules.

docs/sphinx/source/whatsnew/v0.7.1.rst
mikofski: PS I had to check the "allow edits from maintainers" box before the "Commit suggestion" button was accessible

mikofski:
Note, the built-in `open()` was updated in Python-3.6 to accept path objects, see [this SO answer](https://stackoverflow.com/a/42694113/1020470)

pvlib/iotools/midc.py
```diff
@@ -249,12 +249,12 @@ def read_midc_raw_data_from_nrel(site, start, end, variable_map={},
     args = {'site': site,
             'begin': start.strftime('%Y%m%d'),
             'end': end.strftime('%Y%m%d')}
-    endpoint = 'https://midcdmz.nrel.gov/apps/data_api.pl?'
-    url = endpoint + '&'.join(['{}={}'.format(k, v) for k, v in args.items()])
+    url = 'https://midcdmz.nrel.gov/apps/data_api.pl'
+    # NOTE: just use requests.get(url, parmams=args) to build querystring
```
cwhanse:
```suggestion
    # NOTE: just use requests.get(url, params=args) to build querystring
```

cwhanse:
> net reduction in lines, so it must be good!
> why the capitalization of DATA_DIR and similar? I guess I'm ok with either lower case or caps in this case, I'm just wondering why you're changing the style.

I have a slight preference to only capitalize module-level constants, e.g., DATA_DIR, and not capitalize variables with fixed values, e.g., TEST_DATA, in an individual test file.

mikofski:
>>why the capitalization of DATA_DIR and similar? I guess I'm ok with either lower case or caps in this case, I'm just wondering why you're changing the style.

>I have a slight preference to capitalize module-level constants, e.g., DATA_DIR, and not capitalize variables with fixed values, e.g., TEST_DATA, in an individual test file. Can't say I've given that a lot of thought.

Just habit. I follow [pep8 guidelines for constants](https://pep8.org/#constants):

>Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include `MAX_OVERFLOW` and `TOTAL`.

I appreciate that what works for Python in general, may not be applicable everywhere, especially _e.g_ in science or math (_a foolish consistency is the hobgoblin of simple minds_, R.W.E.), so I'm happy to conform to pvlib style preferences, but PyCharm and some other linters may complain, although evidently stickler is okay? Note: pycodestyle specifically denies responsibility for enforcing pep8 naming conventions.

PR #850, opened 2020-01-09
add Boyle/Coello (Humboldt State Univ) soiling model

nappaillav:
- [X] Closes #739
 - [X] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [X] Updates entries to [`docs/sphinx/source/api.rst`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes.
 - [X] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [x] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels and Milestone are assigned to the Pull Request and linked Issue.

This is my first contribution, Hence I'm not with the procedure of Pull request.
Please do let me know how to proceed further to complete this pull request.

Issue #828, opened 2019-12-04
adopt NEP-29


mikofski:
**Is your feature request related to a problem? Please describe.**
the travis-ci test fails for #823 because the minimum NumPy version for py35-min is version 1.10.4 - the `np.roll()` function used was updated to allow multiple axes in NumPy-1.12

**Describe the solution you'd like**
[NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) is a proposal by maintainers of IPython, Matplotlib, SciPy, and NumPy that proposes the support window for Python and Numpy.

TL;DR: in the immediate future, drop support for Python-3.5 and Numpy<1.14

**Describe alternatives you've considered**
- statically create the 3 rotation matrices, removing the use of `np.roll` completely
- use the pre-NumPy-1.12 version of `np.roll`

**Additional context**
Since many other major packages that are used by pvlib are conforming to NEP-29, it makes a lot of sense for pvlib to also conform.

wholmgren:
I think NEP-29 is too aggressive. I'm fine with adopting NEP-29 rules for *expected* compatibility. But NEP-29 seems to go a step farther and suggest that every downstream project release should only support the prescribed versions regardless of if it's actually necessary. In the specific case above, why should we restrict pvlib to numpy 1.14 when numpy 1.12 and 1.13 will work just as well? 

I suggest that we instead aggressively drop numpy/python versions when 1. there is a reason to do so and 2. the versions are outside the NEP-29 windows.

mikofski:
an alternate approach would also be to create a `requires_numpy_1_12` decorator in `conftest.py`

wholmgren:
better to just bump the requirement.

PR #860, opened 2020-01-15
add Kimber soiling model

### LINKED ISSUE #669: Kimber Dynamic Soiling model
related to #61 and #668.

[Kimber dynamic soiling model](https://doi.org/10.1109/WCPEC.2006.279690) has 3 parameters:

* soiling rate: decrease in Ee per day as percent, typically 0.15%?
* threshold: amount of daily rainfall in millimeters required to completely restore panels, ie: reset cumulative soiling to zero, typically what? 13mm? I forgot!
* grace period: amount of time after rainfall that panels stay clean, before they begin to accumulate at the soiling rate, typically 2 weeks

mikofski:
- [x] Closes #669
 - [x] Tests added

kimber

generated with the following snippet:
```python
from pvlib.losses import soiling_kimber
import pvlib
x = pvlib.iotools.read_tmy3('path/to/722287TYA.CSV', coerce_year=1990)
z = soiling_kimber(x[0].Lprecipdepth)
z.plot()
import datetime
manwash = [datetime.date(1990,11,1), ]
z = soiling_kimber(x[0].Lprecipdepth, manual_wash_dates=manwash)
z.plot()
plt.grid()
plt.title('Kimber Soiling Model')
plt.ylabel('soiling build-up fraction')
x[1]
plt.legend([x[1]['Name'], 'manual wash Nov. 1st'])
```


cwhanse:
with this and #850 do we want to: 

1. put all loss functions into `losses.py` (for now),
2. put all soiling loss functions into `soiling.py` and anticipate N x `other_loss_type.py`, or
3. create a `losses` folder to contain `soiling.py`, etc.?

mikofski:
Oops! Sorry, I can move this to `losses.py`. I say we kick this question down the road until we are forced to decide. I don't think it will make a big difference now, and hopefully, we should be able to handle the API changes later without too much stress. Is that okay?

Also I was thinking of waiting on this until #850 and #859 are merged, just to make it easier to sync. Thanks!

pvlib/soiling.py
```diff
@@ -0,0 +1,120 @@
...
+    # loop over days
+    for today in rainfall.index:
+
+        # did rain exceed threshold?
+        rain_exceed_thresh = rainfall[today] > threshold
+
+        # if yes, then set soiling to zero
+        if rain_exceed_thresh:
+            soiling[today] = 0
+            initial_soiling = 0
```
mikofski: anyone know if it's okay to combine these into one line:

    soiling[today] = initial_soiling = 0

or will this cause some type of referencing issue or is a pep8 violation?

mikofski:
@kanderso-nrel  I get this error from my example:
```
  File "/home/docs/checkouts/readthedocs.org/user_builds/pvlib-python/checkouts/860/docs/examples/plot_greensboro_kimber_soiling.py", line 48, in <module>
    soiling_no_wash.index, soiling_no_wash.values*100.0)
TypeError: float() argument must be a string or a number, not 'Timestamp'
```
But it works on my machine, do I have the wrong matplotlib version? How do you plot timestamps? I guess I can plot from pandas instead of using `pyplot`

kanderso-nrel:
I installed the same matplotlib and numpy versions as RTD is using and can recreate the error with pandas 1.0.1, but when using pandas 0.25.1 (instead of 1.0.1 like RTD) I get this warning:

```
/home/kevin/anaconda3/lib/python3.7/site-packages/pandas/plotting/_matplotlib/converter.py:103: 
FutureWarning: Using an implicitly registered datetime converter for a matplotlib plotting method. 
The converter was registered by pandas on import.
Future versions of pandas will require you to explicitly register matplotlib converters.

To register the converters:
	>>> from pandas.plotting import register_matplotlib_converters
	>>> register_matplotlib_converters()
  warnings.warn(msg, FutureWarning)
```

Switching to 1.0.1 and implementing that matplotlib registration line appears to fix it.  I wonder why they stopped registering it by default, that's a little annoying.

mikofski:
All tests are passing, the gallery example tenders, I don't know what that patch problem is. Any reviewers?

pvlib/losses.py
```diff
@@ -85,3 +86,116 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+        # soiling rate to soiling from previous day
+        total_soil = initial_soiling + soiling_rate
+
+        # check if soiling has reached the maximum
+        soiling[today] = (
+            max_soiling if (total_soil >= max_soiling) else total_soil)
+
+        initial_soiling = soiling[today]  # reset initial soiling
+
+    return soiling.interpolate()
```
kanderso-nrel: Since this interpolates between points spaced 24h apart, the soiling signal does not immediately drop straight to zero.  Is this intentional?  For example this shows a smooth drop to zero across jan 3rd:

```
rainfall = pd.Series(0, index=pd.date_range('2019-01-01', '2019-01-05', freq='h', closed='left'))
rainfall.loc['2019-01-03 10:00'] = 10
soiling_kimber(rainfall).plot()
```

pvlib/losses.py
```diff
@@ -85,3 +86,116 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
     soiling_ratio = 1 - 0.3437 * erf(0.17 * accum_mass**0.8473)
 
     return soiling_ratio
+
+
+def soiling_kimber(rainfall_timeseries, threshold=6, soiling_rate=0.0015,
+                   grace_period=14, max_soiling=0.3, manual_wash_dates=None,
+                   initial_soiling=0):
```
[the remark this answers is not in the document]
mikofski: Figure 3 in the 2006 IEEE 4th WCPEC paper shows soiling rates
image
The soiling build up rates are classified into four geographical regions:
* central valley
* Norcal
* Socal
* desert

and in 3 "environment" types:
* rural/ag
* suburban
* urban/highway/airport

So 0.15% is kind of the median, but really it depends on those two factors: region & enviro-type.

As far as the rain threshold or grace period, these defaults come from SunPower. We are also using the Kimber model here, so I can ask what our defaults are. And there may be a white paper somewhere that has them, or perhaps @zdefreitas or @mdeceglie would know. Maybe we could even just ask Adie.

You can kind of see how 6-7[mm] and 2-weeks work in Figure 5:
image
but again it really depends on the region, deserts/Socal need slightly less rain, and Norcal/Central valley need slightly more.

Sorry, I can't do better, but I will try to hunt down better references. Any other recommendations?

pvlib/losses.py
mikofski: I'm okay with it like this, I figure it takes a day of rain to clean the panels, so in that example, it rained on Jan 3rd so by the end of the day, the panels should be clean
Figure_1
Are you okay with it like this? Any other suggestions?

mikofski:
okay, any more comments? @CameronTStark maybe add to v0.7.2?

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+    """
+    Kimber soiling model [1]_ assumes soiling builds-up at a daily rate unless
```
cwhanse:
```suggestion
    Kimber soiling model [1]_ assumes soiling builds up at a daily rate unless
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+    the daily rainfall is greater than a threshold. The model also assumes that
+    if daily rainfall has exceeded the threshold within a grace period, then
+    the ground is too damp to cause soiling build-up. The model also assumes
+    there is a maximum soiling build-up. Scheduled manual washes and rain
+    events are assumed to reset soiling to zero.
+
+    Parameters
+    ----------
+    rainfall_timeseries : pandas.Series
```
cwhanse:
```suggestion
    rainfall: pandas.Series
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+        a timeseries of rainfall in millimeters
```
cwhanse:
```suggestion
        a timeseries of rainfall in millimeters
```
```suggestion
        Accumulated rainfall at the end of each time period [mm]
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+    threshold : float, default 6
```
cwhanse:
```suggestion
    cleaning_threshold: float, default 6
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+        the amount of rain in millimeters [mm] required to clean the panels
+    soiling_rate: float, default 0.0015
+        daily soiling rate, enter as fraction, not percent, default is 0.15%
```
cwhanse:
```suggestion
        fraction of energy lost to one day of soiling [unitless]
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+    there is a maximum soiling build-up. Scheduled manual washes and rain
+    events are assumed to reset soiling to zero.
+
+    Parameters
+    ----------
+    rainfall_timeseries : pandas.Series
+        a timeseries of rainfall in millimeters
+    threshold : float, default 6
+        the amount of rain in millimeters [mm] required to clean the panels
+    soiling_rate: float, default 0.0015
```
cwhanse:
```suggestion
    soiling_loss_rate: float, default 0.0015
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+        daily soiling rate, enter as fraction, not percent, default is 0.15%
+    grace_period : int, default 14
+        The number of days after a rainfall event when it's assumed the ground
+        is damp, and so it's assumed there is no soiling. Change to smaller
+        value for dry climate, default is 14-days
```
cwhanse:
```suggestion
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+    ----------
+    rainfall_timeseries : pandas.Series
+        a timeseries of rainfall in millimeters
+    threshold : float, default 6
+        the amount of rain in millimeters [mm] required to clean the panels
+    soiling_rate: float, default 0.0015
+        daily soiling rate, enter as fraction, not percent, default is 0.15%
+    grace_period : int, default 14
+        The number of days after a rainfall event when it's assumed the ground
+        is damp, and so it's assumed there is no soiling. Change to smaller
```
cwhanse:
```suggestion
        is damp, and so it's assumed there is no soiling.
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+    ===================  =======  =========  ======================
+    Region/Environment   Rural    Suburban   Urban/Highway/Airport
+    ===================  =======  =========  ======================
+    Central Valley       0.0011   0.0019     0.0020
+    Northern CA          0.0011   0.0010     0.0016
+    Southern CA          0        0.0016     0.0019
+    Desert               0.0030   0.0030     0.0030
+    ===================  =======  =========  ======================
+
+    Rainfall thresholds may also vary slightly by region. Please consult [1]_
```
cwhanse:
```suggestion
    Rainfall thresholds and grace periods may also vary by region. Please consult [1]_
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+        the initial fraction of soiling on the panels at time zero in the input
+
+    Returns
+    -------
+    soiling : pandas.Series
+        soiling build-up fraction
+
+    Notes
+    -----
+    The soiling rate depends on both the geographical region and the soiling
```
cwhanse:
```suggestion
    The soiling loss rate depends on both the geographical region and the soiling
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+        A list or tuple of Python ``datetime.date`` when the panels were
+        manually cleaned. Note there is no grace period after a manual
+        cleaning, so soiling begins to build-up immediately after a manual
+        cleaning
+    initial_soiling : float, default 0
+        the initial fraction of soiling on the panels at time zero in the input
+
+    Returns
+    -------
+    soiling : pandas.Series
```
cwhanse:
```suggestion
    pandas.Series
```

pvlib/losses.py
```diff
@@ -85,3 +86,128 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+        soiling build-up fraction
```
cwhanse:
```suggestion
        fraction of daily energy lost to soiling. Series has intervals of one day.
```

kanderso-nrel:
> A minor quibble: I'd prefer that the file names for the sample data and the example start with soiling_kimber to make it easier to associate these files with the function

FYI sphinx-gallery defaults to only executing examples if their filename starts with `plot_`.  If you want to name gallery examples some other way, you'll need to configure that behavior: https://sphinx-gallery.github.io/stable/configuration.html#parsing-and-executing-examples-via-matching-patterns

I suggest changing that regex to just match everything -- if for some reason in the future we want to include examples in the gallery without executing them, we can make it more specific at that point.

cwhanse:
> FYI sphinx-gallery defaults to only executing examples if their filename starts with plot_.  If you want to name gallery examples some other way, you'll need to configure that behavior: https://sphinx-gallery.github.io/stable/configuration.html#parsing-and-executing-examples-via-matching-patterns

Did not know that, thanks. `plot_soiling_kimber` works for me as well. Or if we object to the `plot_` prefix, change regex as you suggest. My concern is managing lots of these examples with no pattern to the file names.

It would be good to agree on a naming convention for these files, and document it in the contributing guide. Separate issue :)

pvlib/losses.py
mikofski: the output has the same intervals as the input, so if the input is in hours, the output is also hours

pvlib/losses.py
mikofski: Would `rain_threshold` be okay here? I'm a tiny bit concerned that users will somehow confuse this with manual washes, so I'm looking for some way to clarify that this is the threshold of rain required to clean the panels. Any suggestions?

pvlib/losses.py
cwhanse: Right, my mistake.

pvlib/losses.py
mikofski: leaving it as `cleaning_threshold` for consistency with `soiling_hsu` that already defines this term, also updated definition to be consistent and define accumulation period as one day

pvlib/losses.py
```diff
@@ -85,3 +86,123 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+    Region/Environment   Rural    Suburban   Urban/Highway/Airport
+    ===================  =======  =========  ======================
+    Central Valley       0.0011   0.0019     0.0020
+    Northern CA          0.0011   0.0010     0.0016
+    Southern CA          0        0.0016     0.0019
+    Desert               0.0030   0.0030     0.0030
+    ===================  =======  =========  ======================
+
+    Rainfall thresholds and grace periods may also vary by region. Please
+    consult [1]_ more information.
```
kanderso-nrel:
```suggestion
    consult [1]_ for more information.
```

pvlib/losses.py
```diff
@@ -85,3 +86,123 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
     soiling_ratio = 1 - 0.3437 * erf(0.17 * accum_mass**0.8473)
 
     return soiling_ratio
+
+
+def soiling_kimber(rainfall, cleaning_threshold=6, soiling_loss_rate=0.0015,
+                   grace_period=14, max_soiling=0.3, manual_wash_dates=None,
+                   initial_soiling=0):
+    """
+    Kimber soiling model [1]_ assumes soiling builds up at a daily rate unless
```
kanderso-nrel:
```suggestion
    Calculate soiling ratio with rainfall data and a daily soiling rate using 
    the Kimber soiling model [1]_.
    
    Kimber soiling model assumes soiling builds up at a daily rate unless
```
Add a 1-line explanation for the sphinx docs:
https://external-builds.readthedocs.io/html/pvlib-python/860/api.html#losses

pvlib/tests/test_losses.py
```diff
@@ -90,3 +96,53 @@ def test_soiling_hsu(rainfall_input, expected_output_2):
...
+    soiling_norain = soiling_kimber(norain)
+    # expected soiling reaches maximum
+    # NOTE: TMY3 data starts at 1AM, not midnite!
+    norain = np.ones(8760) * 0.0015/24
+    norain[0] += 0.0015/24
+    norain = np.cumsum(norain)
+    norain[:22] = np.arange(0, 0.0015, 0.0015/22)
+    norain = np.where(norain > 0.3, 0.3, norain)
+    # test no rain, soiling reaches maximum
+    assert np.allclose(soiling_norain.values, norain)
```
kanderso-nrel: FYI I think Will has said elsewhere that smaller individual test functions are preferred.

wholmgren: yes please!

mikofski:
@kanderso-nrel you are right, vectorizing this code makes it run about 20 to 30 times faster on my machine (from 130-140ms to 4-6ms, depending on washes), but it requires the input rainfall time-series to be monotonic - which means that it needs some special code to fix TMY3, just like #889 , but this doesn't add much overhead, about 0.5-1 ms, and I don't think it would effect other non-TMY3 time-series.

There are some pros/cons to the vectorized version.
* it can handle rainfall at any time of the day or the week by using rolling windows, so it picks up some soiling that the discreet-daily (unvectorized) version doesn't
* also it can start building up soiling at any time of the day, or be cleaned anytime
* and cleaning is instantaneous, not a day-long process

Here are some plots to compare against, orange is the vectorized version, blue is not. There are minor differences due to the differences above, but I think they might be improvements? I'd have to save new test data.

No Wash
kimber_vec

With Wash
kimber_vec_wash

The code is copied from the HSU model, which was already vectorized:
<details>

```python
def soiling_kimber_vec(rainfall, cleaning_threshold=6, soiling_loss_rate=0.0015,
                       grace_period=14, max_soiling=0.3, manual_wash_dates=None,
                       initial_soiling=0):

    # convert grace_period to timedelta
    grace_period = datetime.timedelta(days=grace_period)

    # protect against TMY3 by rolling back and then adding 1
    rain_index = np.roll(rainfall.index.values, 1) + np.timedelta64(1, 'h')
    periods_per_day = (rain_index[1] - rain_index[0]) / np.timedelta64(24, 'h')
    rain_tz = rainfall.index.tz
    rain_index = pd.DatetimeIndex(
        rain_index).tz_localize('UTC').tz_convert(rain_tz)

    rainfall = pd.Series(rainfall.values, index=rain_index, name='rainfall')

    daily_rainfall = rainfall.rolling('D').sum()  # closed right

    norain = np.ones_like(rainfall.values) * soiling_loss_rate * periods_per_day
    norain[0] = initial_soiling
    norain = np.cumsum(norain)

    rain_events = daily_rainfall > cleaning_threshold

    grace_windows = rain_events.rolling(grace_period).sum() > 0

    soiling = pd.Series(float('NaN'), index=rainfall.index)
    soiling.iloc[0] = initial_soiling
    soiling[grace_windows] = norain[grace_windows]

    # manual wash dates
    if manual_wash_dates is not None:
        manual_wash_dates = pd.DatetimeIndex(manual_wash_dates).tz_localize(rain_tz)
        norain = pd.Series(norain, index=rain_index, name='soiling')
        soiling[manual_wash_dates] = norain[manual_wash_dates]

    return norain - soiling.ffill()
```
</details>

mikofski:
I guess it goes without saying that the faster, more flexible version is everyone's preference?

mikofski:
#Todo

- [ ] fix `soiling[0]` is zero, not `initial_condition`
- [ ] test for initial condition
- [ ] where greater than max, use max, else soiling less than max
- [ ] fix time roll interval use `dt` not 1H, or check for TMY3 input arg?

cwhanse:
> I guess it goes without saying that the faster, more flexible version is everyone's preference?

As long as it's consistent with the publication, yes for me.

mikofski:
>As long as it's consistent with the publication, yes for me.

Yes, they are consistent, the only difference is that when Adie did it, I think she was mainly looking at _daily_ rainfall, so that was the default accumulation period and time-step interval, which meant that everything begins and ends at midnnight.

With the vectorized version, we can interpret the _accumulation period_ more broadly, as the total rainfall over the last 24-hours. So it could be 2020-02-02 08:00 to 2020-02-03 08:00 instead of always midnight to midnight.

The difference occurs if there's rain overnight. For example, if the threshold is 6mm, what if there's 3mm from 8pm to midnight, and then another 3mm from midnight to 6am? In the daily discreet version (unvectorized) this would not clean the panels because the accumulated rainfall 3mm on day-1 and 3mm on day-2 was never greater than the threshold. But in the rolling (vectorized) version, this _would_ clean the panels.

Personally I think this is better, and now that hourly rainfall is more widely available, I don't see why we wouldn't prefer it.

Here's a copy of the paper: [Kimber_2006_Full_Paper_r3.pdf](https://github.com/pvlib/pvlib-python/files/4206868/Kimber_2006_Full_Paper_r3.pdf)

cwhanse:
> With the vectorized version, we can interpret the accumulation period more broadly, as the total rainfall over the last 24-hours. So it could be 8AM to 8AM instead of always midnight to midnight.

The rolling window approach is definitely better than the midnight to midnight approach. Because rainfall threshold is a depth rather than a rate, what about an argument `accumulation_period` that defines the time window over which the threshold applies? If you agree, maybe default to 24 hours to be faithful to the reference.

mikofski:
>what about an argument `accumulation_period` that defines the time window over which the threshold applies? If you agree, maybe default to 24 hours to be faithful to the reference.

an improvement on the model for sure, I like it, thanks!

mikofski:
Ok @CameronTStark, it's ready, any last comments?

pvlib/losses.py
```diff
@@ -85,3 +86,130 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+
+    # get rainfall timezone, timestep as timedelta64, and timestep in int days
+    rain_tz = rainfall.index.tz
+    rain_index = rainfall.index.values
+    timestep_interval = (rain_index[1] - rain_index[0])
+    day_fraction = timestep_interval / np.timedelta64(24, 'h')
+
+    # if TMY fix to be monotonically increasing by rolling index by 1 interval
+    # and then adding 1 interval, while the values stay the same
+    if istmy:
```
CameronTStark: NIT: change `istmy` to `is_tmy`.

pvlib/tests/test_losses.py
```diff
@@ -90,3 +97,101 @@ def test_soiling_hsu(rainfall_input, expected_output_2):
...
+
+
+@pytest.fixture
+def expected_kimber_soiling_nowash():
+    return pd.read_csv(
+        DATA_DIR / 'greensboro_kimber_soil_nowash.dat',
+        parse_dates=True, index_col='timestamp')
+
+
+@needs_pandas_0_22
```
CameronTStark: I didn't get to try this but I assume these `@needs_pandas_0_22` are necessary. If so I'm even more inclined to bump pandas version to unify the output between versions. No change necessary, just thinking out loud mostly.

mikofski: Pandas `DataFrame.rolling()` kwarg `closed` isn't available for [py35-min on Travis](https://travis-ci.org/pvlib/pvlib-python/jobs/650693240#L857) but everything else passes, py35-min uses pandas-0.18.1

CameronTStark:
Thanks for your great contribution @mikofski and thanks to @kanderso-nrel and @cwhanse for your feedback!

mikofski:
Thanks! Sorry I didn't have a chance to fix the nit. Maybe we can address it later.

pvlib/tests/test_losses.py
wholmgren: My understanding is that the function does not work at all with pandas 0.18. Correct? If so, that's a different kind of breakage than we've traditionally allowed across dependency versions. The function documentation doesn't give any hint as to why it's going to fail in that scenario either.

I thought we were in agreement that we'd bump the pandas spec for 0.8, not 0.7.2. This suggests that we need to do it for 0.7.2. Not the end of the world, but not ideal.

docs/examples/plot_greensboro_kimber_soiling.py
```diff
@@ -0,0 +1,63 @@
...
+# ---------
+# The example shown here demonstrates how the threshold affects soiling.
+# Because soiling depends on rainfall, loading weather data is always the first
+# step.
+
+from datetime import datetime
+from matplotlib import pyplot as plt
+from pvlib.iotools import read_tmy3
+from pvlib.losses import soiling_kimber
+from pvlib.tests.conftest import DATA_DIR
```
wholmgren: Is there a better way to access the data directory from the examples? I don't think it's a good idea to suggest to people that they should import anything from modules in `pvlib.tests`.

pvlib/tests/test_losses.py
```diff
@@ -90,3 +97,101 @@ def test_soiling_hsu(rainfall_input, expected_output_2):
...
+    # get TMY3 data with rain
+    greensboro = read_tmy3(DATA_DIR / '723170TYA.CSV', coerce_year=1990)
+    # NOTE: can't use Sand Point, AK b/c Lprecipdepth is -9900, ie: missing
+    return greensboro[0].Lprecipdepth
+
+
+@pytest.fixture
+def expected_kimber_soiling_nowash():
+    return pd.read_csv(
+        DATA_DIR / 'greensboro_kimber_soil_nowash.dat',
```
wholmgren: why `.dat` instead of `.csv`?

docs/examples/plot_greensboro_kimber_soiling.py
mikofski: Sure, sorry me just being lazy:

1. I can create the path to `DATA_DIR`, something like `import os;DATA_DIR=os.path.join(is.path.dirname(__file__),'..','..','pvlib','data')` except maybe using `pathlib` and pep8 formatting. This is my preference so no network calls
2. I could retrieve the actual Greensboro TMY3 file from the NSRDB by URL, but this would require a network call for every CI run, unless it's decorated with the remote network pytest thing, see #858

pvlib/tests/test_losses.py
mikofski: `*.csv` is ignored by Git

https://github.com/pvlib/pvlib-python/blob/64a6498b689d27d77bb5967e13eab9bc0018d68e/.gitignore#L82

So I have to force commit csv files, but `.dat` isn't. Sorry, me lazy again. Also I kinda like dat since it kinda sounds/looks like data, so it feels like it belongs in the data folder. Silly

pvlib/tests/test_losses.py
mikofski: Oops, I didn't realize that. The existing `soiling_hsu` also used this `closed` kwarg in `DataFrame.rolling()`

https://github.com/pvlib/pvlib-python/blob/64a6498b689d27d77bb5967e13eab9bc0018d68e/pvlib/losses.py#L69

but it's pandas dependency was masked in the test by `required_scipy`

https://github.com/pvlib/pvlib-python/blob/64a6498b689d27d77bb5967e13eab9bc0018d68e/pvlib/tests/test_losses.py#L83

One idea is just dropping this `closed` kwarg from `DataFrame.rolling()` which works fine in pandas-0.18.1, the default behavior is 'both' which isn't the end of the world, but probably not strictly faithful to the references, which imply exclusive of starting timestamp.

Any other ideas @cwhanse? @kanderso_nrel?

docs/examples/plot_greensboro_kimber_soiling.py
kanderso-nrel: This seems like it might come up with future gallery examples too.  Some packages have a function that provides access to the built-in datasets.  That way they are accessible to users without them having to monkey around with filepaths.  Examples:
- https://scikit-learn.org/stable/datasets/index.html#toy-datasets
- https://seaborn.pydata.org/generated/seaborn.load_dataset.html

What about a new function (maybe in `pvlib.tools`?) that looks like `load_dataset('greensboro-tmy')` or something?

pvlib/tests/test_losses.py
CameronTStark: >These @needs* decorators have traditionally been used for only optional dependencies or quirks between versions (like pandas sum of nans = nan or = 0).

Thanks for clarifying this further for us @wholmgren. 

> I thought we were in agreement that we'd bump the pandas spec for 0.8, not 0.7.2. This suggests that we need to do it for 0.7.2. Not the end of the world, but not ideal.

We can revert this merge and chamber it for v0.8. It'll give time to address some of the issues raised brought up and stay consistent with our plan to bump pandas at v0.8.  What do you think @wholmgren?

pvlib/tests/test_losses.py
wholmgren: I have a few reasons to prefer csv to dat... I have to guess at what I need to open a `.dat` file. If I run `head file.dat` is it going to print binary gibberish and maybe nuke my terminal session? And what is in a csv file if not data? 

We could change the ignore to be more like the notebook rule

https://github.com/pvlib/pvlib-python/blob/64a6498b689d27d77bb5967e13eab9bc0018d68e/.gitignore#L87-L89

so 

```
 # Ignore some csv 
 *.csv 
 !pvlib/data/*.csv 
```

That being said, I've previously added `.dat` files for SURFRAD and SOLRAD tests because it follows the naming convention for the raw data from those networks. Maybe those should be changed - I don't know if consistency is preferable there or not.

pvlib/tests/test_losses.py
wholmgren: I don't think we need to revert. @mikofski pointed out that the same issue exists in the already released `soiling_hsu`. We could add a note to the doc string, or a try/except with a better error message around the rolling call, or do nothing. I'm ok with any of those.

We should just be more careful with the requires and needs decorators. I know I've been lazy about them in the past.

This is all also an argument for 1) adding scipy to the base requirements so the test suite doesn't have gotchas like with the hsu function (issue #483) and 2) being more aggressive with bumping dependency specs so that contributors don't have to spend as much time supporting older libraries (issue #828).

docs/examples/plot_greensboro_kimber_soiling.py
mikofski: Hi Kevin, fo you know if there is already an open issue for this? Thanks

kanderso-nrel: See #924 :)

PR #873, opened 2020-01-29
make test_psm3.py robust to API overuse errors

### LINKED ISSUE #870: obtain a dedicated API key for testing NREL psm3
came up in #866 see this [comment](../pull/866#issuecomment-578345464) but related to most PR and testing in general - due to  [NREL PSM3 throttle limits for the DEMO key](https://developer.nrel.gov/docs/rate-limits/) the CI's frequently fail

obtain a dedicated API key from [NREL developer network](https://developer.nrel.gov/) and upload it to [Azure pipelines as a secret variable](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables), ditto for Travis, then for testing, get this from `os.environ[<PVLIB TEST NREL API KEY>]` and use it instead of the DEMO key which has severe [throttle limits](https://developer.nrel.gov/docs/rate-limits/) and is shared by everyone

just live with the test failures, or skip theses tests from the CI's

**Additional context**

CameronTStark:
- [x] Closes #870

test_psm3.py has been dealing with an random failures due to a dependency on a general API key from NREL's Development Network. This PR implements two features to deal with this.

1. Implement a pvlib-python specific NREL API key for testing.
1. Calling get_psm3() inside a try/except statement that repeats the call if an "OVER_RATE_LIMIT" error occurs. 

For 1, Implementing the NREL API key requires an admin of our Azure Pipelines to create the appropriate variable for this to work. The azure-pipelines.yml exports this value into the environment before pytest is run. There is a fallback to the general API code of "DEMO_KEY" that allows for tests to still be run locally. This only alleviates part of the problem and doesn't fix NREL's protections that deny multiple API calls from the same IP address.

Feature 2 works to repeat requests that return an HTTPError caused by an "OVER_RATE_LIMIT" and increases the sleep time between calls in the case of multiple errors. This should allow even local testing without a registered NREL API key to avoid these errors although they may be more susceptible to these tests taking longer.

The only thing I can think of that might cause an error from the API after implementing these strategies would be if there were more than 1000 calls by it within a day. I don't believe we need to accommodate for that possibility now but I just want to highlight it for future reference.


pvlib/tests/iotools/test_psm3.py
```diff
@@ -21,7 +23,12 @@
     'Temperature Units', 'Pressure Units', 'Wind Direction Units',
     'Wind Speed', 'Surface Albedo Units', 'Version']
 PVLIB_EMAIL = 'pvlib-admin@googlegroups.com'
-DEMO_KEY = 'DEMO_KEY'
+try:
+    DEMO_KEY = os.environ["NREL_API_KEY"]
+except KeyError:
+    print("WARNING: NREL API KEY environment variable not set!")
+    print("Using DEMO_KEY instead. This may cause unexpected failures.")
```
mikofski: Please either using [`logging`](https://docs.python.org/3/library/logging.html) framework or [`warnings`](https://docs.python.org/3/library/warnings.html). _EG_:

```python
import warnings

# in try
warnings.warn("WARNING: NREL API KEY ...")
```

Or
```python
import logging

logging.basicConfig()
LOGGER = logging.grtLogger(__name__)
LOGGER.setLevel(logging.WARNING)

# in try
LOGGER.warning("WARNING: NREL API KEY ...")
```

I prefer warnings over logger, but I think either would work better than `print`, which is unpythonic because it blocks the main thread and is not customizable -- or: you can't silence it from the command line. EG you can silence warnings and logging from `python`

wholmgren: I think `warnings.warn` is the better approach here. 

It would also be better if this lookup were part of a fixture that returns the api key rather than existing at the module level. That way you only get the warning if you're running a test that needs to access the live api.

pvlib/tests/iotools/test_psm3.py
```diff
@@ -49,48 +61,72 @@ def assert_psm3_equal(header, data, expected):
...
                                  names='2017', interval=30)
     expected = pd.read_csv(YEAR_TEST_DATA)
     assert_psm3_equal(header, data, expected)
     # check leap day
-    _, data_2012 = psm3.get_psm3(LATITUDE, LONGITUDE, DEMO_KEY, PVLIB_EMAIL,
+    _, data_2012 = delay_get_psm3(LATITUDE, LONGITUDE, DEMO_KEY, PVLIB_EMAIL,
                                  names='2012', interval=60, leap_day=True)
     assert len(data_2012) == (8760+24)
     # check errors
     with pytest.raises(HTTPError):
```
wholmgren: why are these error checks essentially repeated from `test_get_psm3_tmy`? can we safely delete them and save some of our api calls?

CameronTStark:
@wholmgren can you work with me to add the "nrelApiKey" secret variable to Azure Pipelines? 

Either you can follow [the Azure Pipeline's documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables) or give me admin rights to my work email then I'll be happy to do it. 

The tests are failing sporatically still but I anticipate them passing once the key is in place. Thanks.

pvlib/tests/iotools/test_psm3.py
```diff
@@ -49,48 +61,72 @@ def assert_psm3_equal(header, data, expected):
     assert (data.index.tzinfo.zone == 'Etc/GMT%+d' % -header['Time Zone'])
 
 
+def delay_get_psm3(*args, **kwargs):
```
wholmgren: how about [`pytest-rerunfailures`](https://github.com/pytest-dev/pytest-rerunfailures) instead of this custom approach?

pvlib/tests/iotools/test_psm3.py
```diff
@@ -49,48 +61,72 @@ def assert_psm3_equal(header, data, expected):
...
+    """Repeat get_psm3() call if OVER_RATE_LIMIT error occurs otherwise
+    allow HTTPError to raise. Sleep times increase by one second for
+    every OVER_RATE_LIMIT HTTPError in a row. Return (header, data) from
+    get_psm3() as normal.
+    """
+    sleep_time = 0
+    while True:
```
wholmgren: I'm always leery of while loops... what about a for loop with a limit e.g. `for _ in range(max_retries)`

wholmgren:
@CameronTStark I think I've added the variable to the pipeline.

CameronTStark:
> @CameronTStark I think I've added the variable to the pipeline.

Thanks!
 
> Do we also need to add to travis?

Yes. Let me see if I can set that up too.

pvlib/tests/iotools/test_psm3.py
CameronTStark: I had the same thought in discussing with @kanderso-nrel in #870. The error catching seems to test NREL's API more than get_psm3() but I do appreciate the thoroughness of the tests for injecting "bad" inputs and demonstrating that they fail.  I'm not attached to it either way.

pvlib/tests/iotools/test_psm3.py
CameronTStark: I'll look into it.

pvlib/tests/iotools/test_psm3.py
CameronTStark: I get that. I also considered using `for sleep_time in range(max_retries):` to consolidate the iteration of sleep_time and the looping. I'd say `max_retries = 5` would suffice, would you agree?

pvlib/tests/iotools/test_psm3.py
wholmgren: It seems to me that these failures can safely be expected for both the tmy and non-tmy requests. Would be even better if they were in their own test function.

CameronTStark: I was just going to suggest that the tests be split up so we can use the `@pytest.mark.flaky(reruns=5, reruns_delay=2)` decorator over each API call. We still have to check if the error contains "OVER_RATE_LIMIT" and fail the test since that too is an HTTPError.

mikofski: I'm okay with dumping these api tests, sorry for adding them initially

CameronTStark: Not at all @mikofski, I appreciate the thoroughness. 

There could be great use in these as potential early warnings of potentially breaking changes to the API that could warn us by giving a clearer picture of what is failing. This could backfire if there's a lot of API changes that create a nuisance but that by itself could be a warning as well. 

I'm sure this type of issue has been raised and resolved before so I'm ok either way. The newest PR consolidates all error tests into a single parametrized test so that way help keep things tidy enough to move forward.

pvlib/tests/iotools/test_psm3.py
CameronTStark: Thanks for the tip. I've implemented pytest-rerunfailures and it works nicely.

azure-pipelines.yml
```diff
@@ -74,6 +75,7 @@ jobs:
     displayName: 'List installed dependencies'
   - script: |
       source activate test_env
+      export NREL_API_KEY=$(nrelApiKey)
```
CameronTStark: @wholmgren was the secret variable set as "nrelApiKey"? The linux/MacOS tests don't seem to be picking up the API key correctly and failing. I need to fix the Windows version "export" equivalent command as a separate issue.

wholmgren: Screen Shot 2020-01-30 at 9 35 31 AM

azure-pipelines.yml
```diff
@@ -28,6 +28,7 @@ jobs:
       pip install pytest pytest-cov pytest-mock pytest-timeout pytest-azurepipelines
       pip install -e .
       export NREL_API_KEY=$(nrelApiKey)
+      echo $NREL_API_KEY
```
wholmgren: not a big deal in this instance but it defeats the purpose of a secret key. the CI logs are available for all to see.

CameronTStark: I played with this on my fork and noticed that Azure blocked it out as "**". Are you able to see it?

mikofski: I think Travis and other CI will not show any secrets that have the word "SECRET" or "PASSWORD" in it, **but I wouldn't count on it**

wholmgren: @CameronTStark no I see [`/home/vsts/work/_temp/0cb00519-40af-46ec-994a-cab3c0d8db0c.sh: line 2: nrelApiKey: command not found
`](https://dev.azure.com/solararbiter/pvlib%20python/_build/results?buildId=2533&view=logs&j=ea977d73-b8d6-59cf-b2df-c65a31e93ffc&t=2690bf14-3208-5a31-bf74-aa7a625c04e8&l=18) instead of `*`. Nice to know that there is some built in protection but like @mikofski says I wouldn't count on it.

CameronTStark: [Here's a link to the pipeline run for my fork on the same commit where it outputs "***" for the key.](https://dev.azure.com/ctstark/pvlib-test/_build/results?buildId=55&view=logs&j=83ed75f2-3f6e-51e0-dc7b-dfceecefbbcd&t=0b9cc177-13eb-5a9e-870b-984e27ba97f8&l=81) It is printed like that in the raw output too so I think it's safe. I'm aware of the security implications and wouldn't do this for a more valuable key. I tested this out on my own fork initially with a dummy variable and witnessed the "***" for that before putting in the actual key. 

A quirk of this secret variable is that it is sheltered from the Python code even though its in the environment. That's why it has be renamed because I calling `os.environ['nrelApiKey']` didn't work.

wholmgren: Interesting. I get a 401 not authorized error when I try your link.

CameronTStark: > Interesting. I get a 401 not authorized error when I try your link.

Yes, I've been getting that for pvlib-python's Azure links as well for an hour or so. Did permissions change for it?

wholmgren: I didn't change anything.

azure-pipelines.yml
CameronTStark: I have a PR to my pvlib-python fork from the same branch that this PR is from [and Azure sees the variable.](https://dev.azure.com/ctstark/pvlib-test/_build/results?buildId=55&view=logs&j=06aebe0f-636a-5f93-19d5-0c37b58d55c8&t=0878738c-c430-5911-4cfd-99d3ab677436&l=20) You can see from [the latest Azure run that it's not able to see this variable.](https://dev.azure.com/solararbiter/pvlib%20python/_build/results?buildId=2533&view=logs&j=ee50eb0a-7467-5494-0f35-e3b863355bb0&t=2efe8f58-8c60-5d45-8189-b00aa6aac1e4&l=79) 

Can you double check that the variable is set in the pvlib-python pipeline? I'm guessing that you have access to multiple pipelines. Maybe an trailing whitespace?

Sorry to keep bugging you about this. If you'd rather wait until I have Azure Pipelines access I'm happy jumping into that.

azure-pipelines.yml
CameronTStark: If I logout then I can see it. Strange...

setup.py
```diff
@@ -42,7 +42,7 @@
                     'pytz',
                     'requests']
 TESTS_REQUIRE = ['nose', 'pytest', 'pytest-cov', 'pytest-mock',
-                 'pytest-timeout']
+                 'pytest-timeout', 'pytest-rerunfailures']
```
kanderso-nrel:
```suggestion
                 'pytest-timeout', 'pytest-rerunfailures', 'pytest-remotedata']
```

pvlib/tests/iotools/test_psm3.py
```diff
@@ -50,47 +65,47 @@ def assert_psm3_equal(header, data, expected):
...
-        psm3.get_psm3(LATITUDE, LONGITUDE, api_key='BAD', email=PVLIB_EMAIL)
-    with pytest.raises(HTTPError):
-        # coordinates were not found in the NSRDB
-        psm3.get_psm3(51, -5, DEMO_KEY, PVLIB_EMAIL)
-    with pytest.raises(HTTPError):
-        # names is not one of the available options
-        psm3.get_psm3(LATITUDE, LONGITUDE, DEMO_KEY, PVLIB_EMAIL, names='bad')
 
 
 @needs_pandas_0_22
```
kanderso-nrel:
```suggestion
@needs_pandas_0_22
@pytest.mark.remote_data
```

pvlib/tests/iotools/test_psm3.py
```diff
@@ -50,47 +65,47 @@ def assert_psm3_equal(header, data, expected):
...
-                      interval=15)
+
+
+@pytest.mark.parametrize('latitude, longitude, api_key, names, interval',
+                         [(LATITUDE, LONGITUDE, 'BAD', 'tmy-2017', 60),
+                          (51, -5, DEMO_KEY, 'tmy-2017', 60),
+                             (LATITUDE, LONGITUDE, DEMO_KEY, 'bad', 60),
+                             (LATITUDE, LONGITUDE, DEMO_KEY, '2017', 15),
+                          ])
+@needs_pandas_0_22
```
kanderso-nrel:
```suggestion
@needs_pandas_0_22
@pytest.mark.remote_data
```

pvlib/tests/iotools/test_psm3.py
```diff
@@ -50,47 +65,47 @@ def assert_psm3_equal(header, data, expected):
 
 
 @needs_pandas_0_22
```
kanderso-nrel:
```suggestion
@needs_pandas_0_22
@pytest.mark.remote_data
```

CameronTStark:
> Left some suggestions for marking tests with `@pytest.mark.remote_data` from [pytest-remotedata](https://github.com/astropy/pytest-remotedata) in reference to #878. This disables the tests unless including `--remote-data` from the command line when starting the tests.

Thanks for pointing this out to me. This addresses #878 more than #870 to be clear but I'll try to include it.

CameronTStark:
There's good news is that I have the nrelApiKey secret working for PRs. This was due to a setting buried in Azure Pipelines that allows secret variables to be accessed by PRs. I can see how this would be good to have off for other keys but should be fine for our case. 

After a small hiccup dealing with conda's version of pytest-rerunfailures being >Python3.6 I believe this is ready for review. 

I'm opting for putting `pytest-remotedata` in another PR since this one has so many moving parts between pytest and Azure Pipelines. It could be done now but this is currently working and would create another debug iteration that could open up another can of worms.

ci/requirements-py35-min.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
     - pytest-cov
     - pytest-mock
     - pytest-timeout
+    - pytest-rerunfailures
```
wholmgren: does this also need to move to the pip section?

azure-pipelines.yml
```diff
@@ -153,8 +160,10 @@ jobs:
       versionSpec: '$(python.version)'
 
   - script: |
-      pip install pytest pytest-cov pytest-mock pytest-timeout pytest-azurepipelines
+      pip install pytest pytest-cov pytest-mock pytest-timeout pytest-azurepipelines pytest-rerunfailures
       pip install -e .
+      export NREL_API_KEY=$(nrelApiKey)
+      echo $NREL_API_KEY
```
wholmgren: ok to delete all of the `echo` commands now that it's working?

ci/requirements-py36.yml
```diff
@@ -16,6 +16,7 @@ dependencies:
     - pytest
     - pytest-cov
     - pytest-mock
+    - pytest-rerunfailures
```
wholmgren: perhaps put all of the `pytest-rerunfailures` lines in the pip sections for simplicity/consistency. either way is fine though

pvlib/tests/iotools/test_psm3.py
```diff
@@ -21,7 +23,20 @@
     'Temperature Units', 'Pressure Units', 'Wind Direction Units',
     'Wind Speed', 'Surface Albedo Units', 'Version']
 PVLIB_EMAIL = 'pvlib-admin@googlegroups.com'
-DEMO_KEY = 'DEMO_KEY'
+
+
+@pytest.fixture(scope="module")
+def DEMO_KEY():
```
wholmgren: let's name this function `api_key` and also rename the variables below for clarity.

pvlib/tests/iotools/test_psm3.py
```diff
@@ -50,47 +65,47 @@ def assert_psm3_equal(header, data, expected):
...
-        # intervals can only be 30 or 60 minutes
-        psm3.get_psm3(LATITUDE, LONGITUDE, DEMO_KEY, PVLIB_EMAIL, names='2017',
-                      interval=15)
+
+
+@pytest.mark.parametrize('latitude, longitude, api_key, names, interval',
+                         [(LATITUDE, LONGITUDE, 'BAD', 'tmy-2017', 60),
+                          (51, -5, DEMO_KEY, 'tmy-2017', 60),
+                             (LATITUDE, LONGITUDE, DEMO_KEY, 'bad', 60),
+                             (LATITUDE, LONGITUDE, DEMO_KEY, '2017', 15),
```
wholmgren: I'm not sure why stickler doesn't complain, but looks to me like there is extra white space on these two lines.

pvlib/tests/iotools/test_psm3.py
wholmgren: can you remind me why all of these tests need pandas 0.22?

CameronTStark: I was hoping you could tell me =)

ci/requirements-py35-min.yml
CameronTStark: Do you remember where this is referenced? I didn't notice it being called by any of the conda versions.

ci/requirements-py36.yml
CameronTStark: I was thinking of just putting py35's version in pip with the note to highlight why its there instead of conda. This stems from the idea that in the conda environment it would be ideal if everything came from conda.

pvlib/tests/iotools/test_psm3.py
CameronTStark: I named it nrel_api_key since there was a conflict with the get_psm3() keyword of api_key.

pvlib/tests/iotools/test_psm3.py
CameronTStark: AutoPEP8 didn't do anything about it either so I just aligned it manually.

pvlib/tests/iotools/test_psm3.py
CameronTStark: I'll try taking them off locally and see what errors pop out.

ci/requirements-py35-min.yml
wholmgren: Sorry, I don't understand your comment. My understanding is that `pytest-rerunfailures` is not available for python 3.5 in conda, so I'm surprised to see it here in the conda section of the `py35-min` configuration.

CameronTStark: You're right, it should be moved to the pip section but could the entire file be deleted?

My question is basically what the use is of the file this requirements-py35-min.yml file? I've only run across the use of the .yml files in the "ci" directory in the "azure-pipelines.yml" file where they are used to create the environment for the test via the line:
`  - script: conda env create --quiet --file ci/requirements-py$(python.version).yml`

There's no listed py35-min version so this file probably isn't used. This is corroborated by there being no errors  caused by it. 

Not necessarily pertinent to this PR but tangentially relevant.

wholmgren: I'm not sure if it's redundant at this point, but it is used on Travis https://travis-ci.org/pvlib/pvlib-python/jobs/646624536?utm_medium=notification&utm_source=github_status

let's move the line to the pip section for this PR and consider whether the file should be removed in a separate issue

CameronTStark: Thanks for pointing that out. I haven't really touched the Travis system so it's good to know this wrinkle. 

I'll move that over and confirm Travis is happy.

pvlib/tests/iotools/test_psm3.py
```diff
@@ -21,7 +23,20 @@
     'Temperature Units', 'Pressure Units', 'Wind Direction Units',
     'Wind Speed', 'Surface Albedo Units', 'Version']
 PVLIB_EMAIL = 'pvlib-admin@googlegroups.com'
-DEMO_KEY = 'DEMO_KEY'
+
+
+@pytest.fixture(scope="module")
+def nrel_api_key():
+    """"""
```
mikofski: NIT: remove empty docstring, or add meaningful message

pvlib/tests/iotools/test_psm3.py
```diff
@@ -50,47 +65,47 @@ def assert_psm3_equal(header, data, expected):
...
+                          (51, -5, nrel_api_key, 'tmy-2017', 60),
+                          (LATITUDE, LONGITUDE, nrel_api_key, 'bad', 60),
+                          (LATITUDE, LONGITUDE, nrel_api_key, '2017', 15),
+                          ])
+@needs_pandas_0_22
+@pytest.mark.flaky(reruns=5, reruns_delay=2)
+def test_get_psm3_tmy_errors(
+    latitude, longitude, api_key, names, interval
+):
+    """Test get_psm3() for multiple error scenarios:
```
mikofski: NIT: I'm not a fan of this docstring style, and it seems inconsistent with pvlib, [pep8](https://pep8.org/#documentation-strings), and other packages (pandas, scipy, numpy, mpl, etc.)

Can you please dedent (undent? outdent?) the docstring so that it's the same indentation as the `"""` either like this?
```python
    """Test get_psm3() for multiple error scenarios:
    * Bad api key -> HTTP 403 forbidden because api_key is rejected
    * Bad latitude/longitude -> Coordinates were not found in the NSRDB
    * Bad name -> names is not one of the available options
    * Bad interval, single year -> intervals can only be 30 or 60 minutes
    """
```
or imho even better would be this:
```python
    """
    Test get_psm3() for multiple error scenarios:
    * Bad api key -> HTTP 403 forbidden because api_key is rejected
    * Bad latitude/longitude -> Coordinates were not found in the NSRDB
    * Bad name -> names is not one of the available options
    * Bad interval, single year -> intervals can only be 30 or 60 minutes
    """
```
thanks!

pvlib/tests/iotools/test_psm3.py
```diff
@@ -50,47 +65,47 @@ def assert_psm3_equal(header, data, expected):
...
+        * Bad api key -> HTTP 403 forbidden because api_key is rejected
+        * Bad latitude/longitude -> Coordinates were not found in the NSRDB
+        * Bad name -> names is not one of the available options
+        * Bad interval, single year -> intervals can only be 30 or 60 minutes
+    """
+    with pytest.raises(HTTPError) as e:
```
mikofski: NIT: please use verbose names, instead of single letters for variable names. I'm okay with either `exc` or `err` or anything longer, just not `e` -- thanks!

pvlib/tests/iotools/test_psm3.py
```diff
@@ -50,47 +65,47 @@ def assert_psm3_equal(header, data, expected):
...
+        psm3.get_psm3(latitude, longitude, api_key, PVLIB_EMAIL,
+                      names=names, interval=interval)
+    # ensure the HTTPError caught isn't due to overuse of the API key
+    assert "OVER_RATE_LIMIT" not in str(e.value)
```
mikofski: nice catch! 😎

pvlib/tests/iotools/test_psm3.py
mikofski: @CameronTStark see [this comment](../694#issuecomment-487272722) in #694
>Bummer, I think the `pandas.to_datetime()` API changed after pandas-0.16 so the python-27-min build fails. I'll try to decorate test_psm3 with `@needs_pandas_0_22`

**So** since we've dropped Python-2.7 support, we just need to check if it's still valid for Python-3.5, otherwise we can remove these.

**Also** since we bumped pandas to v0.18.1 in [pvlib-0.7.0](https://pvlib-python.readthedocs.io/en/stable/whatsnew.html#v0-7-0-december-18-2019) we can probably definitely drop these now.

mikofski: Yep, the ability to assemble a datetimes from a sequence `(year, month, ...)` was new in [pandas-0.18](https://pandas.pydata.org/pandas-docs/version/0.18/whatsnew.html#whatsnew-0181-enhancements-assembling)

CameronTStark: Thanks for researching this. I took off the decorators locally to run the tests and didn't have an issue with my pandas==1.0.1 environment. I tried setting up an environment with 0.18.1 and couldn't get it to install so I can't confirm for that scenario. 

Like the `pytest-remotedata` I think this should be addressed in another issue so we can get this across the finishline.

pvlib/tests/iotools/test_psm3.py
CameronTStark: Missed this, thanks for highlighting. 

I've included some verbage about what is going on for our future selfs.

pvlib/tests/iotools/test_psm3.py
CameronTStark: I dedented the bullets to be in line and conformed it more closely to [PEP 257](https://www.python.org/dev/peps/pep-0257/). The second suggestion doesn't conform to this PEP with the extra \n at the beginning. 

Please let me know if there's a better way to list the scenarios.

pvlib/tests/iotools/test_psm3.py
CameronTStark: I did a `grep "as e:" pvlib/**/*.py"` and found that there were two instances of that not including this one. I repeated for exc which had none and err which also had two. I decided the tie breaker would be [the pytest docs that describe](https://docs.pytest.org/en/latest/assert.html)  `pytest.raises()` that used excinfo.

pvlib/tests/iotools/test_psm3.py
mikofski: [PEP 8 says consistency within a project is more important, and PEP 20 says readability counts.](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds). Take a look at docstrings in pvlib and the other major projects, and you'll see both of the styles I listed above are in common usage. I personally prefer the leading newline, unless the docstring is a one-liner, because I like the first line to be just as readable as the following ones, and I think the quotation marks detract from the readability. Anyway my point is, I don't think we have to follow PEP 257 verbatim.

wholmgren: my 2 cents: I'm really happy to see the explanation for the test. But let's not worry too much about doc string formatting in the test suite. Or use comments and avoid any debate!

pvlib/tests/iotools/test_psm3.py
```diff
@@ -50,47 +71,49 @@ def assert_psm3_equal(header, data, expected):
...
-    header, data = psm3.get_psm3(LATITUDE, LONGITUDE, DEMO_KEY, PVLIB_EMAIL,
-                                 names='2017', interval=30)
+    header, data = psm3.get_psm3(LATITUDE, LONGITUDE, nrel_api_key,
+                                 PVLIB_EMAIL, names='2017', interval=30)
     expected = pd.read_csv(YEAR_TEST_DATA)
     assert_psm3_equal(header, data, expected)
-    # check leap day
-    _, data_2012 = psm3.get_psm3(LATITUDE, LONGITUDE, DEMO_KEY, PVLIB_EMAIL,
-                                 names='2012', interval=60, leap_day=True)
-    assert len(data_2012) == (8760+24)
```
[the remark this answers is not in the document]
mikofski: @kanderso-nrel ?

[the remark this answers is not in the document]
kanderso-nrel: A colleague of mine reached out the other day asking if I knew why his homemade PSM3 API wrapper didn't work anymore and it turned out that the API rejects the default urllib user-agent string now.  Between that and the 30-minute timestamp shift from several months back, I'd have a preference for "over-testing" here, especially if #882 gets implemented.  That said, it's not pvlib's job to act as the API's test suite so I don't mind getting rid of it if other folks feel it's unnecessary.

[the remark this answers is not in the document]
wholmgren: what's the verdict here? add this two lines back or not? your call @CameronTStark. I think this is the only thing preventing a merge.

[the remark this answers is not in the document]
kanderso-nrel: Sorry for not responding -- I did send a note over to the NREL API team asking about this but haven't heard back yet.  I'll share anything that comes out of that conversation.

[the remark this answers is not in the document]
CameronTStark: @wholmgren I added it back in. I didn't mean to take it out but it got lost in the shuffle. It's already been committed so please let me know if it's good to merge. Thanks!

Thanks @kanderso-nrel for sending a note over to the NREL API team.

wholmgren:
thanks @CameronTStark @kanderso-nrel @mikofski !

PR #885, opened 2020-02-11
remove needs_pandas decorator

### LINKED ISSUE #883: Clean up @needs_pandas_0_22 and @needs_pandas_0_17 pytest decorators
It's apparent that the @needs_pandas decorators used in test_psm3.py are irrelevant now and should be removed. I'll create a PR to remove them and make sure there's no errors in the CI. 

For reference, @mikofski made good observations about this issue [here](https://github.com/pvlib/pvlib-python/pull/873#discussion_r375571469).

CameronTStark:
- [x] Closes #883

Decorators meant to deal with advances in pandas in the test suite are now antiquated and can be removed. Tests passed locally but it's TBD for everything on the CI.

### DIFF
--- a/pvlib/tests/test_irradiance.py
+++ b/pvlib/tests/test_irradiance.py
@@ -13,7 +13,8 @@ from pandas.util.testing import assert_frame_equal, assert_series_equal
 
 from pvlib import irradiance
 
-from conftest import needs_numpy_1_10, requires_ephem, requires_numba
+from conftest import (needs_numpy_1_10, pandas_0_22,
+                      requires_ephem, requires_numba)
 
 
 # fixtures create realistic test input data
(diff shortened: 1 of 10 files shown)


pvlib/tests/test_irradiance.py
```diff
@@ -237,11 +236,8 @@ def test_perez_components(irrad_data, ephem_data, dni_et, relative_airmass):
         columns=['sky_diffuse', 'isotropic', 'circumsolar', 'horizon'],
         index=irrad_data.index
     )
-    if pandas_0_22():
-        expected_for_sum = expected['sky_diffuse'].copy()
-        expected_for_sum.iloc[2] = 0
-    else:
-        expected_for_sum = expected['sky_diffuse']
+    expected_for_sum = expected['sky_diffuse'].copy()
+    expected_for_sum.iloc[2] = 0
```
wholmgren: this test fails on the travis 3.5 min configuration.

CameronTStark: I put needs_pandas_0_22 back for that specific test and it seems all is well. [The py35-min config](https://github.com/pvlib/pvlib-python/blob/master/ci/requirements-py35-min.yml) forces pandas=0.18.1 where the others do not. This should be good to merge as is unless there's a discussion on advancing the minimum pandas version.

CameronTStark: I keep wondering if we should discuss updating our minimum pandas version to 0.22+ and so I'll just ask here, should we advance pandas to 0.22 or higher? It would be in the user's best interest if functions would provide consistent output across supported versions.

wholmgren: 0.22 was [released](https://pandas.pydata.org/pandas-docs/version/0.22/) on Dec 30, 2017, so it's old enough to qualify under NEP-29. I think we should wait for 0.8, though.

CameronTStark: I agree that that's the right way to go. I'll open an issue and mark it with an 0.8.0 milestone to remind us to do this ahead of that release.

PR #886, opened 2020-02-11
Create inverter.py

cwhanse:
 - [x] Tests added

Creates `inverter.py`
- `pvsystem.snlinverter` to `inverter.sandia`
- `pvsystem.adrinverter` to `inverter.adr`
- `pvsystem.pvwatts_ac` to `inverter.pvwatts`


adriesse:
I fully support the new module, but I wonder about the naming.  Perhaps 'efficiency' or 'power' should be part of it, since that's what the current functions model?  I can imagine inverter-related functions that model other aspects.

pvlib/tests/test_pvsystem.py
```diff
@@ -1436,13 +1436,17 @@ def test__sapm_celltemp_translator():
 
 
 @fail_on_pvlib_version('0.9')
-def test_deprecated_09(cec_inverter_parameters, adr_inverter_parameters):
+def test_deprecated_09():
+    # get parameters from fixtures
+    from conftest import (cec_inverter_parameters, adr_inverter_parameters)
```
cwhanse, line 1441: @wholmgren can you suggest a way to use the `cec_inverter_parameters` fixture (from `conftest.py`) here?  The decorator `fail_on_pvlib_version` doesn't allow any arguments for the function to be tested, and pytest doesn't like fixtures to be imported and treated as functions.

wholmgren: You can probably use a magic mock object.

cwhanse: Is it possible to pass arguments through `fail_on_pvlib_version`?  [This](https://realpython.com/primer-on-python-decorators/) article suggests it could be done.

wholmgren: I don’t remember if that limitation is due to my bad understanding of decorators or something about pytest interaction with a custom decorator. 

There’s no reason this function needs the working parameters, so I suspected using magic mock would be a lot faster to implement. 

I’m speculating but another option might be to turn the fail decorator into a pytest fixture.

wholmgren: After further thought you'll probably need patch too:

```python
    # here we only care that warnings are emitted, 
    # so we use mock instead of real parameters
    from unittest.mock import MagicMock, patch
    cec_params = MagicMock()
    adr_params = MagicMock()
    with pytest.warns(pvlibDeprecationWarning):
        with patch('pvlib.pvsystem.snlinverter'):
            pvsystem.snlinverter(250, 40, cec_params)
```

Peque: @wholmgren This worked for me:

https://github.com/Peque/pvlib-python/commit/44e86c20ab80fb19246058beb78a9c361ab95300

Feel free to fetch the commit from my `inverter` branch.

Peque: Okay, *now* I had a look at https://github.com/pvlib/pvlib-python/pull/973, it seems I was late to the party... :sweat_smile:

pvlib/modelchain.py
```diff
@@ -56,7 +56,7 @@ def basic_chain(times, latitude, longitude,
         See temperature.sapm_cell for details.
 
     inverter_parameters : None, dict or Series
-        Inverter parameters as defined by the CEC. See pvsystem.snlinverter for
+        Inverter parameters as defined by the CEC. See inverter.sandia for
```
kanderso-nrel, line 59: consider using `py:func:` here

pvlib/pvsystem.py
```diff
@@ -758,21 +758,32 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
 
     # inverter now specified by self.inverter_parameters
     def snlinverter(self, v_dc, p_dc):
-        """Uses :func:`snlinverter` to calculate AC power based on
+        """Uses :pyfunc:`pvlib.inverter.sandia` to calculate AC power based on
```
kanderso-nrel, line 761:
```suggestion
        """Uses :py:func:`pvlib.inverter.sandia` to calculate AC power based on
```

pvlib/pvsystem.py
```diff
@@ -758,21 +758,32 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
...
 
         Parameters
         ----------
-        See pvsystem.snlinverter for details
+        See pvlib.inverter.sandia for details
 
         Returns
         -------
-        See pvsystem.snlinverter for details
+        See pvlib.inverter.sandia for details
```
kanderso-nrel, line 770: Here and above, consider `:py:func:`

pvlib/pvsystem.py
```diff
@@ -758,21 +758,32 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
...
         """
-        return snlinverter(v_dc, p_dc, self.inverter_parameters)
+        return inverter.sandia(v_dc, p_dc, self.inverter_parameters)
 
     def adrinverter(self, v_dc, p_dc):
-        return adrinverter(v_dc, p_dc, self.inverter_parameters)
+        """Uses :pyfunc:`pvlib.inverter.adr` to calculate AC power based on
```
kanderso-nrel, line 775:
```suggestion
        """Uses :py:func:`pvlib.inverter.adr` to calculate AC power based on
```

pvlib/pvsystem.py
```diff
@@ -758,21 +758,32 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
...
+        ``self.inverter_parameters`` and the input parameters.
+
+        Parameters
+        ----------
+        See pvlib.inverter.adr for details
+
+        Returns
+        -------
+        See pvlib.inverter.adr for details
```
kanderso-nrel, line 784: `:py:func:` here and above

pvlib/pvsystem.py
```diff
@@ -835,7 +846,8 @@ def pvwatts_ac(self, pdc):
         kwargs = _build_kwargs(['eta_inv_nom', 'eta_inv_ref'],
                                self.inverter_parameters)
 
-        return pvwatts_ac(pdc, self.inverter_parameters['pdc0'], **kwargs)
+        return inverter.pvwatts(pdc, self.inverter_parameters['pdc0'],
+                                **kwargs)
```
kanderso-nrel, line 850: Docstring still references the old function

pvlib/tests/test_modelchain.py
```diff
@@ -447,22 +447,29 @@ def acdc(mc):
     mc.ac = mc.dc
 
 
+# TODO in v0.9: remove 'snlinverter', 'adrinverter', 'pvwatts'
```
kanderso-nrel, line 450: Same here about 'pvwatts' being removed

pvlib/pvsystem.py
cwhanse: @kanderso-nrel is there any reason we don't use the full path when a function is within the module?  E.g., for a docstring within `pvsystem.py`, :py:func:`pvlib.pvsystem.pvwatts_ac` instead of :py:func:`pvwatts_ac`

pvlib/tests/test_modelchain.py
cwhanse: I've edited the TODO comment to drop the 'snlinverter' and 'adrinverter' values for `ModelChain.ac_model`. 

There's a loose thread outside of this PR, in that `PVSystem.snlinverter`, `PVSystem.adrinverter` and `PVSystem.pvwatts_ac` methods aren't renamed. I couldn't come up with a naming scheme for the `PVSystem` methods that I liked, when the methods wrap functions in a different module. The method name needs to package both the method purpose (inverter model) and the model (snl, adr, pvwatts).

pvlib/pvsystem.py
kanderso-nrel: As far as I know, it's just a style preference for whether RTD will display `pvlib.pvsystem.pvwatts_ac` or just `pvwatts_ac` in the link text.  I don't think pvlib consistently follows one style or the other currently.  The only reason I can think of to not include the full path is just to reduce clutter -- seems silly to have a bunch of `pvlib.` prefixes everywhere if the bare function names are sufficient.  But sometimes just the bare function name alone isn't very informative.  

It is possible to display custom text in these links:  ```:py:func:`title <target>` ```  ([docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#xref-syntax)).  Not saying we should be using it, it's just another option.

cwhanse: My (not strong) preference would be to display full paths, although its more text there's no question where the function is found in the library. I'm sure that part of the reason that there's no consistent use is my lack of understanding of these markups.

cwhanse: In `pvsystem.py` I've put full paths in where `:py:func:` is used.

cwhanse:
Test failures are unrelated (NREL server not responding), otherwise I think I've addressed the review comments (thanks @kanderso-nrel )

kanderso-nrel:
@cwhanse Did you see that GitHub collapsed several of the review comments in this main thread (`11 hidden conversations`)?  Not all of them were important, I just want to make sure they didn't get overlooked.  In particular, one of the docstring tables isn't rendering on RTD.

cwhanse:
> @cwhanse Did you see that GitHub collapsed several of the review comments in this main thread (11 hidden conversations)?  

No, I overlooked that. Will take care of it shortly.

pvlib/inverter.py
```diff
@@ -0,0 +1,305 @@
...
+    inverter : dict-like
+        Defines parameters for the inverter model in [1]_.  See Notes for
+        required model parameters. A copy of the parameter database from the
+        System Advisor Model (SAM) [2]_ is provided with pvlib and may be read
+        using :py:func:`pvlib.pvsystem.retrieve_sam.
+
+    Returns
+    -------
+    ac_power : numeric
+        AC power output. [W]
```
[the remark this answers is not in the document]
cwhanse: I'd prefer `power_dc` but then I'd feel like we should change the argument names for `PVSystem.snlinverter`.

how strongly do you feel about this naming?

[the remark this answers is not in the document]
kanderso-nrel: > how strongly do you feel about this naming?

Not strongly at all, just saw a small inconsistency and pointed it out.  If the parameters do need to be renamed, might make the most sense to do it in a follow-on PR rather than this one.

pvlib/inverter.py
```diff
@@ -0,0 +1,305 @@
...
+    v_dc : numeric
+        DC voltage input to the inverter, should be >= 0. [V]
+
+    p_dc : numeric
+        DC power input to the inverter, should be >= 0. [W]
+
+    inverter : dict-like
+        Defines parameters for the inverter model in [1]_.  See Notes for
+        required model parameters. A parameter database is provided with pvlib
+        and may be read using :py:func:`pvlib.pvsystem.retrieve_sam.
```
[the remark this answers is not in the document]
cwhanse: There's a file of coefficients for the ADR inverter model, which is also read by `pvlib.pvystem.retrieve_sam`. I really do think that this generic function could be improved and moved to `pvlib.iotools`, but not in this PR.

pvlib/inverter.py
```diff
@@ -0,0 +1,305 @@
...
+              maximum AC power output. [W]
+
+    Vnom      Nominal DC input voltage. Typically the level at which the
+              highest efficiency is achieved. [V]
+
+    Vmax      Maximum DC input voltage. [V]
+
+    Vmin      Minimum DC input voltage. [V]
+
+    Vdcmax    . [V]
```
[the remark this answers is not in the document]
cwhanse: @adriesse for the ADR inverter model, what is the definition of the parameter Vdcmax?  The definition is missing from the function docstring. I'm having trouble distinguishing Vdcmax from Vmax, which I understand to be the maximum DC input voltage. It look like Vdcmax is an alternate upper bound on DC voltage.

pvlib/inverter.py
```diff
@@ -0,0 +1,305 @@
...
+    MPPTLow   Minimum DC voltage for MPPT range. [V]
+
+    Pacmax    Maximum AC output power, used to clip the output power
+              if needed. [W]
+
+    ADRCoefficients  A list of 9 coefficients that capture the influence
+              of input voltage and power on inverter losses, and thereby
+              efficiency. Corresponds to terms from [1]_ (in order): :math:
+              `b_{0,0}, b_{1,0}, b_{2,0}, b_{0,1}, b_{1,1}, b_{2,1}, b_{0,2},
+               b_{1,2},  b_{1,2}`.
```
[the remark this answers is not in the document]
cwhanse: Difficult to describe. From Eq. 9 in the reference, b0,x is W, b1,x is 1/W, b2,x is 1/W^2 after one finds the note that "v_in" is normalized voltage so unitless.  I think it is better to point to the reference. Do you agree?

adriesse:
Would you anticipate putting DC/DC converters, optimizers and/or charge controllers in a different module?  They have some things in common.

cwhanse:
> Would you anticipate putting DC/DC converters, optimizers and/or charge controllers in a different module?  They have some things in common.

I haven't considered that. I suppose these functions could go in `inverter.py`, there isn't a better place at the moment. Do you think the module name is an issue?

We decided not to create `losses.py`. But that doesn't preclude e.g. `dc_losses.py` where wiring loss modules could be deployed, among other functions. 

How would you deploy inverter, DC/DC converter or optimizer and DC-side loss functions?

adriesse:
Names are always an issue, but we can't live without them. :)  You could call the module `powerconversion`.  Then it would seem like home for inverters, converters and transformers.  Or simply `electrical` and you could put wiring in there too.  If you want them bundled, that is.

cwhanse:
> Names are always an issue, but we can't live without them. :)  You could call the module powerconversion.  Then it would seem like home for inverters, converters and transformers.  Or simply electrical and you could put wiring in there too.  If you want them bundled, that is.

Do you know of performance models for devices other than inverters that are more involved than, say, a factor applied to DC or AC power to account for efficiency or internal consumption?

In the case of DC-DC converters or optimizers, I could envision a function sitting between the PV module function output (IV curves) and the input to the inverter model, but I'm not aware of published models (and I haven't diligently looked).

Maybe we move forward with `inverter.py` with its stated scope, which should cover most of the common uses. Although not acknowledged by the name `inverter` I think functions related to MPPT could be put in this module also.

adriesse:
All the active electronics have a control aspect to them, which hasn't been considered much in system simulation so far.   I just wanted to open the field a bit, but go ahead as you deem best.

cwhanse:
@kanderso-nrel can you see if the NSRDB PSM3 data service is offline?  Looks that way to me.

kanderso-nrel:
I'm getting tests passing except for `get_psm3_tmy` too.  On first glance, it seems like a few of the data columns have changed for that TMY dataset, possibly the date labeling too.  Not sure what to make of that -- I had the impression that `tmy-2017` was static and should never change.  I'll open an issue on the NREL github to see if this was an unintentional change.

pvlib/inverter.py
```diff
@@ -0,0 +1,305 @@
...
+
+    Pnt       AC power consumed by inverter at night (night tare) to
+              maintain circuitry required to sense PV array voltage. [W]
+
+    =======   ============================================================
```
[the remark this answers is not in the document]
kanderso-nrel: Sphinx still doesn't like this.  I'm not if all of these steps are necessary, but I got it working locally for me by:

1) Adding spacing between terms and definitions so that there is no overlap between columns (the ADRCoefficients term sticks out pretty far)
2) Extending the left `====` strings so that they completely cover the column width
3) Removing the empty line between Pnt and the trailing line.

Here's the text that works for me (with some text reflows to stay under 80 chars):

```
    ===============   ========================================================
    Column            Description
    ===============   ========================================================
    Pnom              Nominal DC power, typically the DC power needed to
                      produce maximum AC power output. [W]
    Vnom              Nominal DC input voltage. Typically the level at which
                      the highest efficiency is achieved. [V]
    Vmax              Maximum DC input voltage. [V]
    Vmin              Minimum DC input voltage. [V]
    Vdcmax            . [V]
    MPPTHi            Maximum DC voltage for MPPT range. [V]
    MPPTLow           Minimum DC voltage for MPPT range. [V]
    Pacmax            Maximum AC output power, used to clip the output power
                      if needed. [W]
    ADRCoefficients   A list of 9 coefficients that capture the influence
                      of input voltage and power on inverter losses, and
                      thereby efficiency. Corresponds to terms from [1]_
                      (in order): :math:`b_{0,0}, b_{1,0}, b_{2,0}, b_{0,1},
                      b_{1,1}, b_{2,1}, b_{0,2}, b_{1,2}, b_{1,2}`. See [1]_
                      for the use of each coefficient and the associated unit.
    Pnt               AC power consumed by inverter at night (night tare) to
                      maintain circuitry required to sense PV array voltage.
                      [W]
    ===============   ========================================================
```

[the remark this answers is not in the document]
kanderso-nrel: Two more notes:  Vdcmax is still undefined and `b_{1,2}` is repeated in `ADRCoefficients`, presumably should be `b_{2,2}`?

docs/sphinx/source/whatsnew/v0.8.0.rst
```diff
@@ -20,6 +29,8 @@ Testing
 ~~~~~~~
 * Decorator :py:func:`pvlib.conftest.fail_on_pvlib_version` can now be
   applied to functions that require args or kwargs. (:pull:`973`)
+* Test added for :py:meth:`ModelChain.ac_model` to confirm ValueError when
+  ac_model is an invalid string. (:pull:`886`)
```
kanderso-nrel, line 33: Nitpick: this renders as `ModelChain.ac_model()`, which might imply the check happens when the AC model is run but it actually happens when the `ac_model` string is stored by the ModelChain constructor.  Might be clearer to just link ```:py:class:`ModelChain` ``` instead.

docs/sphinx/source/whatsnew/v0.8.0.rst
```diff
@@ -3,6 +3,15 @@
...
+API Changes with Deprecations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* Moved functions related to inverters from `pvsystem.py` to `inverter.py`.
+  Functions are renamed to follow a more consistent pattern, as follows (:pull:`886`):
+   - :py:func:`pvlib.pvsystem.snlinverter` is now :py:func:`pvlib.inverter.sandia`
+   - :py:func:`pvlib.pvsystem.pvwatts_ac` is now :py:func:`pvlib.inverter.pvwatts`
+   - :py:func:`pvlib.pvsystem.adrinverter` is now :py:func:`pvlib.inverter.adr`
+* ModelChain.ac_model now accepts `'sandia'`, `'pvwatts'` and `'adr'` for the
+  inverter models.  (:pull:`886`)
 API Changes
```
kanderso-nrel, line 15: Sphinx complains no blank line between bullet list and following title

cwhanse:
> git grep -n snlinverter shows a handful of docs files still referencing pvsystem.snlinverter.  I checked the other two and didn't see any offenders.

Many thanks for checking this. I've made these corrections where the function is referenced; other instances are references to `PVSystem` and `ModelChain` methods that aren't being renamed in this PR.

pvlib/inverter.py
```diff
@@ -235,24 +238,23 @@ def adr(v_dc, p_dc, inverter, vtol=0.10):
 
 def pvwatts(pdc, pdc0, eta_inv_nom=0.96, eta_inv_ref=0.9637):
     r"""
-    Implements NREL's PVWatts inverter model [1]_.
+    Implements NREL's PVWatts inverter model.
+
+    The PVWatts inverter model [1]_ calculates inverter efficiency :math:'\eta'
```
kanderso-nrel, line 243: Backticks instead of apostrophes

pvlib/inverter.py
```diff
@@ -145,30 +144,34 @@ def adr(v_dc, p_dc, inverter, vtol=0.10):
...
-    =======   ============================================================
+                     of input voltage and power on inverter losses, and thereby
+                     efficiency. Corresponds to terms from [1]_ (in order):
+                     :math: `b_{0,0}, b_{1,0}, b_{2,0}, b_{0,1}, b_{1,1},
+                     b_{2,1}, b_{0,2}, b_{1,2},  b_{2,2}`. See [1]_ for the
+                     use of each coefficient and its associated unit.
+    Pnt              AC power consumed by inverter at night (night tare) to
+                     maintain circuitry required to sense the PV array
+                     voltage. [W]
+    =======          ==========================================================
```
kanderso-nrel, line 170: Still not rendering, I think it's because the three `====` strings don't span the width of the first column. 

Sorry this is such a pain, I wish sphinx was smarter about tables.

cwhanse:
> docs/tutorials/pvsystem.ipynb:576:    "### snlinverter"

What does the ### formatting do? I assumed this was just an internal label or link.

kanderso-nrel:
> > docs/tutorials/pvsystem.ipynb:576:    "### snlinverter"
> 
> What does the ### formatting do? I assumed this was just an internal label or link.

That's inside a markdown cell, just making a small header:

image

It's not code so nbd to leave as-is.

cwhanse:
@wholmgren ok to merge?

pvlib/pvsystem.py
```diff
@@ -827,15 +838,16 @@ def pvwatts_losses(self):
     def pvwatts_ac(self, pdc):
         """
         Calculates AC power according to the PVWatts model using
-        :py:func:`pvwatts_ac`, `self.module_parameters['pdc0']`, and
-        `eta_inv_nom=self.inverter_parameters['eta_inv_nom']`.
+        :py:func:`pvlib.inverter.pvwatts`, `self.module_parameters['pdc0']`,
+        and `eta_inv_nom=self.inverter_parameters['eta_inv_nom']`.
```
wholmgren, line 842: While we're at it, probably change the single ticks around the module parameters items to double ticks so they render using code formatting.

pvlib/tests/test_modelchain.py
```diff
@@ -447,22 +447,30 @@ def acdc(mc):
...
+                  'sandia': sapm_dc_snl_ac_system,
                   'adrinverter': cec_dc_adr_ac_system,
+                  'adr': cec_dc_adr_ac_system,
                   'pvwatts': pvwatts_dc_pvwatts_ac_system}
+    ac_method_name = {'snlinverter': 'snlinverter',
+                      'sandia': 'snlinverter',
+                      'adrinverter': 'adrinverter',
+                      'adr': 'adrinverter',
+                      'pvwatts': 'pvwatts_ac'}
     system = ac_systems[ac_model]
```
wholmgren, line 470: Ideally we would test that the deprecation warning is raised if the ac method name is snlinverter or adrinverter. Ok with me to skip it though since the code looks correct and is not too complicated.

pvlib/tests/test_modelchain.py
```diff
@@ -481,6 +489,14 @@ def test_ac_model_user_func(pvwatts_dc_pvwatts_ac_system, location, weather,
...
 
 
+def test_ac_model_not_a_model(pvwatts_dc_pvwatts_ac_system, location, weather):
+    exc_text = 'not a valid AC power model'
+    with pytest.raises(ValueError, match=exc_text):
+        ModelChain(pvwatts_dc_pvwatts_ac_system, location,
+                   ac_model='not_a_model', aoi_model='no_loss',
+                   spectral_model='no_loss')
+
+
```
wholmgren, line 499: Add a test that will fail on pvlib 0.9 if the old ac method names are provided.

docs/sphinx/source/whatsnew/v0.8.0.rst
```diff
@@ -3,9 +3,19 @@
...
 -------------------------
 
+API Changes with Deprecations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* Moved functions related to inverters from `pvsystem.py` to `inverter.py`.
+  Functions are renamed to follow a more consistent pattern, as follows (:pull:`886`):
+   - :py:func:`pvlib.pvsystem.snlinverter` is now :py:func:`pvlib.inverter.sandia`
+   - :py:func:`pvlib.pvsystem.pvwatts_ac` is now :py:func:`pvlib.inverter.pvwatts`
+   - :py:func:`pvlib.pvsystem.adrinverter` is now :py:func:`pvlib.inverter.adr`
+* ModelChain.ac_model now accepts `'sandia'`, `'pvwatts'` and `'adr'` for the
```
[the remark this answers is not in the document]
cwhanse: `'sandia'` is intended to display as a string value. Should it be marked differently?

pvlib/tests/test_modelchain.py
cwhanse: I'll add a test for the deprecation warning if ModelChain.ac_model is assigned 'snlinverter' or 'adrinverter'

The ModelChain and PVSystem methods are still snlinverter and adrinverter.  Do we want to rename those as well? I favor doing that, but in a separate PR.

wholmgren: I agree we should rename the methods in a separate PR.

docs/sphinx/source/whatsnew/v0.8.0.rst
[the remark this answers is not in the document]
wholmgren: The current rendering is fine with me (shown below), but the meaning of single backticks is context dependent in rst. If you want to keep these in italics, I prefer `*'sandia'*`. Also fine to format as code (e.g. double backtick 'sandia' double backtick)

Screen Shot 2020-07-06 at 8 07 48 AM

pvlib/tests/test_modelchain.py
cwhanse: @wholmgren PTAL not sure I got the test right

https://github.com/pvlib/pvlib-python/blob/7663321a3936820265b89c6c88a7c0ee60e263fa/pvlib/tests/test_modelchain.py#L716

pvlib/tests/test_modelchain.py
```diff
@@ -447,29 +447,52 @@ def acdc(mc):
     mc.ac = mc.dc
 
 
+# TODO in v0.9: remove 'snlinverter', 'adrinverter' since these values for
+# ModelChain.ac_model are superseded by 'sandia' and 'adr'
 @pytest.mark.parametrize('ac_model', [
-    'snlinverter', pytest.param('adrinverter', marks=requires_scipy),
+    'sandia', 'snlinverter', pytest.param('adrinverter', marks=requires_scipy),
+    pytest.param('adr', marks=requires_scipy),
```
wholmgren, line 454: remove the old keywords from this test to avoid deprecation warnings in the logs

https://dev.azure.com/solararbiter/pvlib%20python/_build/results?buildId=3731&view=logs&j=fc432b8b-e2e3-594e-d8b1-15597b6c1d62&t=309866e1-2cf4-5f00-3d0a-999fc3a0f279&l=235

pvlib/tests/test_modelchain.py
```diff
@@ -682,6 +713,22 @@ def test_deprecated_08():
                    temperature_model='pvsyst', temp_model='sapm')
```
wholmgren, line 714: commenting here since it's the closest I can get to the relevant line... change the `ac_model` specs in `test_deprecated_08` so that new warnings are not emitted.

docs/sphinx/source/api.rst
```diff
@@ -265,10 +265,20 @@ Low-level functions for solving the single diode equation.
    singlediode.bishop88_v_from_i
    singlediode.bishop88_mpp
 
-SAPM model
-----------
```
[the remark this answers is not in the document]
cwhanse: Section headings are back

docs/sphinx/source/whatsnew/v0.8.0.rst
```diff
@@ -3,15 +3,25 @@
 v0.8.0 (Month day, year)
 -------------------------
 
+API Changes with Deprecations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* Moved functions related to inverters from `pvsystem.py` to `inverter.py`.
```
wholmgren, line 8:
```suggestion
* Moved functions related to inverters from ``pvsystem.py`` to ``inverter.py``.
```

docs/sphinx/source/whatsnew/v0.8.0.rst
```diff
@@ -3,15 +3,25 @@
...
+  Functions are renamed to follow a more consistent pattern, as follows (:pull:`886`):
+   - :py:func:`pvlib.pvsystem.snlinverter` is now :py:func:`pvlib.inverter.sandia`
+   - :py:func:`pvlib.pvsystem.pvwatts_ac` is now :py:func:`pvlib.inverter.pvwatts`
+   - :py:func:`pvlib.pvsystem.adrinverter` is now :py:func:`pvlib.inverter.adr`
+* Attribute :py:attribute:`modelchain.ModelChain.ac_model` now accepts ``sandia``,
+  ``pvwatts`` and ``adr`` for the inverter models.  (:pull:`886`)
```
wholmgren, line 14: I'm guessing the vast majority of people set this through the `ac_model` kwarg rather than the attribute. So I would link to ``:py:class:`pvlib.modelchain.ModelChain` `` instead. Also suggest adding `'` within the code blocks.

docs/sphinx/source/whatsnew/v0.8.0.rst
```diff
@@ -23,6 +33,8 @@ Testing
 ~~~~~~~
 * Decorator :py:func:`pvlib.conftest.fail_on_pvlib_version` can now be
   applied to functions that require args or kwargs. (:pull:`973`)
+* Test added for :py:meth:`ModelChain` to confirm ValueError when
```
wholmgren, line 36: use same link code as commented above

pvlib/inverter.py
```diff
@@ -0,0 +1,308 @@
...
+    if isinstance(p_dc, pd.Series):
+        power_ac = pd.Series(power_ac, index=p_dc.index)
+
+    return power_ac
+
+
+def adr(v_dc, p_dc, inverter, vtol=0.10):
+    r'''
+    Converts DC power and voltage to AC power using Anton Driesse's
+    Grid-Connected PV Inverter efficiency model
```
wholmgren, line 112:
```suggestion
    Grid-Connected PV Inverter efficiency model.
```

wholmgren: grid-connected inverter?

PR #897, opened 2020-02-21
Expose temperature.faiman in PVSystem and ModelChain 

### LINKED ISSUE #836: Expose temperature.faiman at PVSystem and ModelChain
`ModelChain.temperature_model` accepts `pvsyst` and `sapm` but not `faiman`.

Add `faiman` as an option for `ModelChain.temperature_model`.
Create `PVSystem.faiman_celltemp`

kanderso-nrel:
- [x] Closes #836 
 - [x] Tests added

I assume that ModelChain needs tests as well, but I wanted to ask for advice on how best to do it.  Seems like adding 'faiman' to [`test_infer_temp_model`](https://github.com/pvlib/pvlib-python/blob/master/pvlib/tests/test_modelchain.py#L345) will require a new PVSystem fixture, but beyond that I'm not sure how best to integrate with the existing tests. 

Not directly related to this PR: maybe I'm missing it because it's too late at night, but these two tests seem redundant: https://github.com/pvlib/pvlib-python/blob/master/pvlib/tests/test_pvsystem.py#L398-L434


pvlib/pvsystem.py
```diff
@@ -618,6 +611,32 @@ def pvsyst_celltemp(self, poa_global, temp_air, wind_speed=1.0):
...
+        poa_global : numeric
+            Total incident irradiance [W/m^2].
+
+        temp_air : numeric
+            Ambient dry bulb temperature [C].
+
+        wind_speed : numeric, default 1.0
+            Wind speed in m/s measured at the same height for which the wind
+            loss factor was determined.  The default value 1.0 m/s is the wind
+            speed at module height used to determine NOCT. [m/s]
```
wholmgren: I see N0CT is used in other docstrings, but I don't see a definition anywhere. Could be a follow up issue.

kanderso-nrel:
Thanks for pointing me in the right direction.  After taking a closer look, I don't think the temp model inference test was doing what it was supposed to do -- both the SAPM and PVSyst tests were being inferred as SAPM, but the inferred model wasn't checked so the test passed anyway.  I've changed it to check the inferred model and created a new PVSystem fixture that uses the PVSyst temp model since I didn't see one defined elsewhere.  And added a Faiman test case of course. 

Test failures seem unrelated?

cwhanse:
> I agree. I don't know/remember the context for this but it looks safe to delete one of them. @cwhanse?

I agree, delete one.

pvlib/pvsystem.py
cwhanse: NOCT = Nominal Cell Operating Temperature. Maybe spell out the acronym?

We could add a Note with the definition: NOCT is cell temperature in conditions of irradiance 800 W/m2, ambient air temperature 20C, wind speed 1m/s and cell at Voc, reference to IEC 61215-2 but I'm guessing on the part, I can check when in the office tomorrow.

cwhanse:
> Thanks for pointing me in the right direction.  After taking a closer look, I don't think the temp model inference test was doing what it was supposed to do -- both the SAPM and PVSyst tests were being inferred as SAPM, 

Yep, both fixtures had parameters for the SAPM temperature model. Nice catch and test correction.

pvlib/pvsystem.py
kanderso-nrel: @cwhanse I'm happy to copy/paste a Note snippet from you into this PR, but I'll point out that NOCT is also used in the docstrings of `PVSystem.pvsyst_celltemp`, `temperature.pvsyst_cell`, and `temperature.faiman`.  I wonder if instead of including definitions in docstrings, it might be better to have a central list of definitions somewhere in the sphinx docs?  Or just rely on the PVPMC website or PVCDROM (which as [a page](https://www.pveducation.org/pvcdrom/modules-and-arrays/nominal-operating-cell-temperature) on NOCT already)?

wholmgren: Ok. Let’s address it separately if at all.

wholmgren:
@CameronTStark @kanderso-nrel looks like the failing test is due to issues with remote data. Ok with me to overrule it, but thought it was worth pointing out given the recent activity on this subject.

pvlib/pvsystem.py
cwhanse: @CameronTStark I re-approve this PR for merge :)

wholmgren:
thanks @kanderso-nrel!

PR #872, opened 2020-01-29
Add calcparams_desoto+singlediode example to gallery

kanderso-nrel:
- [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels and Milestone are assigned to the Pull Request and linked Issue.

Bringing over the static part of #856.  This is my first time touching the IV curve part of pvlib, so suggested improvements are welcome.

link to RTD PR build: https://external-builds.readthedocs.io/html/pvlib-python/872/auto_examples/plot_singlediode.html#sphx-glr-auto-examples-plot-singlediode-py

PR #909, opened 2020-02-23
BLD: build docs on Azure Pipelines

### LINKED ISSUE #771: add sphinx documentation build to CI
A handful of recent PRs have broken the documentation build. We should add the documentation build to the azure pipelines CI configuration and fail it if the documentation errors.

CameronTStark:
- [x] Closes #771

Building docs in the CI will help raise awareness of documentation problems and the aid in troubleshooting. This PR aims to add an additional step in the azure-pipelines.yml to accomplish this task.


CameronTStark:
I think the CI is setup right but I could use some help dealing with the warnings that result in the the exit code '1' that causes the test to "fail".

kanderso-nrel:
FYI the `-W` option to `sphinx-build` turns warnings into errors.  We could just disable that flag, but it seems like if we can find a way to fix the warnings currently being raised it would be good to treat warnings as errors in the CI. 

As far as fixing the warnings -- one of them is from #872, oops.  I think all the other ones that look like `docs/sphinx/source/api.rst:568:<autosummary>:1: WARNING: Unknown target name: "1".` are because the docstring-level references get copied into api.rst and don't point to anything there.  One solution would be to not include references in the first line of a docstring, but ideally there would be a way to configure sphinx to handle it better, for instance by propagating the link to the citation entry all the way to the API page.  I'm not having luck finding anything helpful on google in that regard.  Maybe this should be treated as a sphinx bug?

mikofski:
>One solution would be to not include references in the first line of a docstring

+1 for no references in first line. How would they be handled in `api.rst` there's no way other than ignoring them that makes sense to be. I think that other global links like [`:ref:some-arb-location`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations) or [`:doc:another-page`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-documents) would probably work fine.

CameronTStark:
That did the trick @kanderso-nrel, thanks for highlighting that! 

I agree that it'll be ideal to address those warnings but I don't believe that's necessary to address the issue with this PR. Now that this works I'll reapply the rest of the tests and ask for review/merge. 

Would you or @mikofski like to create a separate issue for addressing the warnings? Your familiarity with doc creation would do it more justice than I.

cwhanse:
> One solution would be to not include references in the first line of a docstring
> 
> +1 for no references in first line. 

+2

azure-pipelines.yml
```diff
@@ -194,3 +194,24 @@ jobs:
...
+  steps:
+  - bash: echo "##vso[task.prependpath]/usr/share/miniconda/bin"
+    displayName: Add conda to PATH
+  - script: conda env create -n test_env --quiet --file ci/doc.yml
+    displayName: Create Anaconda environment
+  - script: |
+      source activate test_env
+      # Replicate the exact environment created by the readthedocs CI
+      # conda install --yes --quiet -c pkgs/main mock pillow sphinx sphinx_rtd_theme
+      pip install -e .[all]
```
wholmgren: maybe just `pip install .[docs]` on top of the python3.8 environment?

wholmgren: rtd uses this file to manage its environment https://github.com/pvlib/pvlib-python/blob/master/readthedocs.yml

ci/doc.yml
```diff
@@ -0,0 +1,28 @@
+name: xarray-docs
```
wholmgren: It's not clear to me that we actually need a new doc env specification. We also have the old https://github.com/pvlib/pvlib-python/blob/master/docs/environment.yml from when we did use conda on rtd. Could probably be deleted.

CameronTStark: Yes, I tried this initially but it got left in after I transitioned to a pip install only. I'll delete it.

azure-pipelines.yml
CameronTStark: I tried using just .[docs] in a previous commit but that was unsuccessful. The RTD file is helpful to confirm that using .[all] is necessary. Do you think it's worth it's own issue to address that the .[docs] is insufficient to build?

CameronTStark:
I think this is good to go now and can use a final review. 

Just FYI, my plan is to next try to simplify the azure-pipelines.yml through use of templates.

azure-pipelines.yml
wholmgren: some of the doc pages, e.g. [Forecasting](https://pvlib-python.readthedocs.io/en/stable/forecasts.html), need the optional dependencies. I don't know what the right approach is long term, but I agree it's better to use all for now.

azure-pipelines.yml
```diff
@@ -194,3 +194,24 @@ jobs:
...
+    vmImage: 'ubuntu-16.04'
+  steps:
+  - bash: echo "##vso[task.prependpath]/usr/share/miniconda/bin"
+    displayName: Add conda to PATH
+  - script: conda env create -n test_env --quiet --file ci/doc.yml
+    displayName: Create Anaconda environment
+  - script: |
+      source activate test_env
+      # Replicate the exact environment created by the readthedocs CI
+      # conda install --yes --quiet -c pkgs/main mock pillow sphinx sphinx_rtd_theme
```
wholmgren: can we delete this line?

azure-pipelines.yml
```diff
@@ -194,3 +194,24 @@ jobs:
...
+  pool:
+    vmImage: 'ubuntu-16.04'
+  steps:
+  - bash: echo "##vso[task.prependpath]/usr/share/miniconda/bin"
+    displayName: Add conda to PATH
+  - script: conda env create -n test_env --quiet --file ci/doc.yml
+    displayName: Create Anaconda environment
+  - script: |
+      source activate test_env
+      # Replicate the exact environment created by the readthedocs CI
```
wholmgren: do we agree now that this is not the goal?

azure-pipelines.yml
```diff
@@ -194,3 +194,24 @@ jobs:
...
     displayName: 'Build sdist'
+
+
+- job: DocsTest
+  pool:
+    vmImage: 'ubuntu-16.04'
+  steps:
+  - bash: echo "##vso[task.prependpath]/usr/share/miniconda/bin"
+    displayName: Add conda to PATH
+  - script: conda env create -n test_env --quiet --file ci/doc.yml
```
wholmgren: `requirements-py38.yml`?

azure-pipelines.yml
```diff
@@ -194,3 +194,24 @@ jobs:
...
+    displayName: Create Anaconda environment
+  - script: |
+      source activate test_env
+      # Replicate the exact environment created by the readthedocs CI
+      # conda install --yes --quiet -c pkgs/main mock pillow sphinx sphinx_rtd_theme
+      pip install -e .[all]
+    displayName: Replicate readthedocs CI environment
```
wholmgren: install pvlib for doc build?

azure-pipelines.yml
CameronTStark: This has been removed, now only dealing standard Python/pip.

azure-pipelines.yml
CameronTStark: I think it is the goal for the CI isn't it?

azure-pipelines.yml
CameronTStark: I believe this is necessary to run the sphinx-gallery examples at least. I was inspired by xarray's Azure config and this is also [a step that xarray uses as well](https://github.com/pydata/xarray/blob/18e34cc8746ea2ce7be785edcaeee0bd7d4a0200/ci/azure/install.yml#L39). 

Based on the readthedocs.yml file, isn't this at least part of how ReadTheDocs builds our documentation for us? In our readthedocs.yml file it states that the `extra_requirements` are listed as `-all` and based on [The ReadTheDocs documentation](https://docs.readthedocs.io/en/latest/config-file/v1.html#python-extra-requirements) that is what is installed during their build of our docs. I'm inexperienced with ReadTheDocs so please let me know if I am missing something.

CameronTStark:
Sorry again for jumping the gun. I think this is in a state that could use another review and discussion. Thanks!

azure-pipelines.yml
wholmgren: I think the goal is to test the docs, not debug the service that builds/serves the docs. We test pvlib on many different environments, so it's reasonable to test the docs in different environments too.

azure-pipelines.yml
wholmgren: I'm not sure we're on the same page here. My comment should have been a suggestion like

```
    displayName: Install pvlib[all] for doc build
```

wholmgren:
Safe to ignore the coverage failure. codecov has been really bad lately...

azure-pipelines.yml
CameronTStark: I like that better, thanks for the recommendation.

CameronTStark:
Alright, I think this is good for a final review/approval/merge. 

Thanks for your patience and feedback @wholmgren!

PR #910, opened 2020-02-24
fix read_tmy3 with year coerced not monotonic, breaks soiling

### LINKED ISSUE #889: soiling_hsu doesn't work with read_tmy3 - index must be monotonic
**Describe the bug**
calling `pvlib.losses.soiling_hsu(read_tmy3(<tmy file>, coerce_year=<year>), ...)` raises `ValueError: index must be monotonic` because the last hour of a TMY3 file has been coerced to the same year as the first index, instead of the first hour of the next year

**To Reproduce**
```python
from pvlib.losses import soiling_hsu
from pvlib.tests.conftest import DATA_DIR

greensboro = read_tmy3(DATA_DIR / '723170TYA.CSV', coerce_year=1990)
greensboro_rain = greensboro[0].Lprecipdepth
soiling_hsu(greensboro_rain, cleaning_threshold=6, tilt=10, pm2_5=1e-5, pm10=1e-5)
```
raises `ValueError: index must be monotonic`
<details>

```
ValueError                                Traceback (most recent call last)
<ipython-input-95-ff36a88e4e3a> in <module>
----> 1 soiling_hsu(greensboro_rain, cleaning_threshold=6, tilt=10, pm2_5=1e-5, pm10=1e-5)

~\Projects\pvlib-python\pvlib\losses.py in soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10, depo_veloc, rain_accum_period)
     67
     68     # accumulate rainfall into periods for comparison with threshold
---> 69     accum_rain = rainfall.rolling(rain_accum_period, closed='right').sum()
     70     # cleaning is True for intervals with rainfall greater than threshold
     71     cleaning_times = accum_rain.index[accum_rain >= cleaning_threshold]

~\projects\pvlib-python\venv\lib\site-packages\pandas\core\generic.py in rolling(self, window, min_periods, center, win_type, on, axis, closed)
  10145                                    min_periods=min_periods,
  10146                                    center=center, win_type=win_type,
> 10147                                    on=on, axis=axis, closed=closed)
  10148
  10149         cls.rolling = rolling

~\projects\pvlib-python\venv\lib\site-packages\pandas\core\window.py in rolling(obj, win_type, **kwds)
   2624         return Window(obj, win_type=win_type, **kwds)
   2625
-> 2626     return Rolling(obj, **kwds)
   2627
   2628

~\projects\pvlib-python\venv\lib\site-packages\pandas\core\window.py in __init__(self, obj, window, min_periods, center, win_type, axis, on, closed, **kwargs)
     63         self.win_freq = None
     64         self.axis = obj._get_axis_number(axis) if axis is not None else None
---> 65         self.validate()
     66
     67     @property

~\projects\pvlib-python\venv\lib\site-packages\pandas\core\window.py in validate(self)
   1576                                          timedelta))):
   1577
-> 1578             self._validate_monotonic()
   1579             freq = self._validate_freq()
   1580

~\projects\pvlib-python\venv\lib\site-packages\pandas\core\window.py in _validate_monotonic(self)
   1610             formatted = self.on or 'index'
   1611             raise ValueError("{0} must be "
-> 1612                              "monotonic".format(formatted))
   1613
   1614     def _validate_freq(self):
```
</details

**Expected behavior**
I'm not sure, I guess I would expect it to be able to handle TMY3 data, but maybe it's the TMY3 with the year coerced to the same number that's the problem? This would also fail TMY3 if you didn't coerce the year, but it might fail silently in some weird ugly way that you didn't notice at first.

**Versions:**
 - pvlib: 0.7.1+25.gc7b1c0e
 - pandas: 0.24.2
 - python: 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 24 2018, 00:16:47) [MSC v.1916 64 bit (AMD64)]

**Additional context**
related to #865 and #866

mikofski:
* if tmy is coerced to year, then the last hour is in the same year,
 before the other timestamps, so timestamps are no monotonic
* closes #889
* adds new function to `iotools` called `fix_tmy3_coerce_year_monotonicity`
* also fix reference footnote in first line of docstring,
 since used in api.rst, causes doc warning
* don't use mutable input args
* always create soiling pd.Series for consistency and robust to future edits
* fix rain_index is not a tz-aware pd.DatetimeIndex series,
 so rename as rain_index_vals and use rainfall.index in Soiling
* add new test for `fix_tmy3_coerce_year_monotonicity` and use it for Kimber soiling example and tests

 - [x] Tests added

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.7.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.2.rst
@@ -14,6 +14,8 @@ Enhancements
   the original ``Date (MM/DD/YYYY)`` and ``Time (HH:MM)`` columns that the
   indices were parsed from (:pull:`866`)
 * Add Kimber soiling model :py:func:`pvlib.losses.soiling_kimber` (:pull:`860`)
+* Add :py:func:`~pvlib.iotools.fix_tmy3_coerce_year_monotonicity` to fix TMY3
+  data coerced to a single year to be monotonically increasing (:pull:`910`)
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.7.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.2.rst
@@ -15,35 +15,35 @@ Enhancements
 ~~~~~~~~~~~~
 * TMY3 dataframe returned by :py:func:`~pvlib.iotools.read_tmy3` now contains
   the original ``Date (MM/DD/YYYY)`` and ``Time (HH:MM)`` columns that the
-  indices were parsed from (:pull:`866`)
-* Add Kimber soiling model :py:func:`pvlib.losses.soiling_kimber` (:pull:`860`)
+  indices were parsed from. (:pull:`866`)
+* Add Kimber soiling model :py:func:`pvlib.losses.soiling_kimber`. (:pull:`860`)
 
 Bug fixes
 ~~~~~~~~~
 * Fix :py:func:`~pvlib.iotools.read_tmy3` parsing when February contains
-  a leap year (:pull:`866`)
+  a leap year. (:pull:`866`)
 * Implement NREL Developer Network API key for consistent success with API
-  calls in :py:mod:`pvlib.tests.iotools.test_psm3` (:pull:`873`)
+  calls in :py:mod:`pvlib.tests.iotools.test_psm3`. (:pull:`873`)
 * Fix issue with :py:class:`pvlib.location.Location` creation when
-  passing ``tz=datetime.timezone.utc`` (:pull:`879`)
+  passing ``tz=datetime.timezone.utc``. (:pull:`879`)
 * Fix documentation homepage title to "pvlib python" based on first heading on
   the page. (:pull:`890`) (:issue:`888`)
 * Implement `pytest-remotedata <https://github.com/astropy/pytest-remotedata>`_
   to increase test suite speed. Requires ``--remote-data`` pytest flag to
-  execute data retrieval tests over a network.(:issue:`882`)(:pull:`896`)
+  execute data retrieval tests over a network. (:issue:`882`)(:pull:`896`)
 * Fix missing
   `0.7.0 what's new <https://pvlib-python.readthedocs.io/en/stable/whatsnew.html#v0-7-0-december-18-2019>`_
   entries about changes to ``PVSystem.pvwatts_ac``. Delete unreleased
   0.6.4 what's new file. (:issue:`898`)
 * Compatibility with cftime 1.1. (:issue:`895`)
-* Add Python3.8 to Azure Pipelines CI (:issue:`903`)(:pull:`904`)
+* Add Python3.8 to Azure Pipelines CI. (:issue:`903`)(:pull:`904`)
 * Minor implemention changes to avoid runtime and deprecation warnings in
   :py:func:`~pvlib.clearsky.detect_clearsky`,
   :py:func:`~pvlib.iam.martin_ruiz_diffuse`,
   :py:func:`~pvlib.losses.soiling_hsu`,
   and various test functions.
 * Fix :py:func:`~pvlib.iotools.read_tmy3` so that when coerced to a single year
-  the TMY3 data will be monotonically increasing (:pull:`910`)
+  the TMY3 index will be monotonically increasing. (:pull:`910`)
 
 Documentation
 ~~~~~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.7.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.2.rst
@@ -39,6 +39,8 @@ Bug fixes
   :py:func:`~pvlib.iam.martin_ruiz_diffuse`,
   :py:func:`~pvlib.losses.soiling_hsu`,
   and various test functions.
+* Fix :py:func:`~pvlib.losses.soiling_hsu` TMY3 data not monotonic exception
+  (:pull:`910`)
 
 Documentation
 ~~~~~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.7.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.2.rst
@@ -8,8 +8,8 @@ API Changes
 * :py:class:`pvlib.forecast.ForecastModel` now requires ``start`` and ``end``
   arguments to be tz-localized. (:issue:`877`, :pull:`879`)
 * :py:func:`pvlib.iotools.read_tmy3` when coerced to a single year now returns
-  indices that are monotonically increasing, for example the last index will be
-  midnight of the *next* year. (:pull:`910`)
+  indices that are monotonically increasing. Therefore, the last index will be
+  January 1, 00:00 of the *next* year. (:pull:`910`)
 
 Enhancements
 ~~~~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.7.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.2.rst
@@ -7,6 +7,9 @@ API Changes
 ~~~~~~~~~~~
 * :py:class:`pvlib.forecast.ForecastModel` now requires ``start`` and ``end``
   arguments to be tz-localized. (:issue:`877`, :pull:`879`)
+* :py:func:`pvlib.iotools.read_tmy3` when coerced to a single year now returns
+  indices that are monotonically increasing, for example the last index will be
+  midnight of the *next* year. (:pull:`910`)
 
 Enhancements
 ~~~~~~~~~~~~
@@ -14,11 +17,6 @@ Enhancements
   the original ``Date (MM/DD/YYYY)`` and ``Time (HH:MM)`` columns that the
   indices were parsed from (:pull:`866`)
 * Add Kimber soiling model :py:func:`pvlib.losses.soiling_kimber` (:pull:`860`)
-* Add :py:func:`~pvlib.iotools.tmy3_monotonic_index` to fix TMY3
-  data coerced to a single year to be monotonically increasing (:pull:`910`)
-* Add ``monotonic_index`` kwarg to :py:func:`~pvlib.iotools.read_tmy3` to fix
-  TMY3 data coerced to a single year to be monotonically increasing, default is
-  ``False`` and warns that it will be changed to ``True`` in v0.8 (:pull:`910`)
 
 Bug fixes
 ~~~~~~~~~
@@ -44,8 +42,8 @@ Bug fixes
   :py:func:`~pvlib.iam.martin_ruiz_diffuse`,
   :py:func:`~pvlib.losses.soiling_hsu`,
   and various test functions.
-* Fix :py:func:`~pvlib.losses.soiling_hsu` TMY3 data not monotonic exception
-  (:pull:`910`)
+* Fix :py:func:`~pvlib.iotools.read_tmy3` so that when coerced to a single year
+  the TMY3 data will be monotonically increasing (:pull:`910`)
 
 Documentation
 ~~~~~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.7.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.2.rst
@@ -34,7 +34,11 @@ Bug fixes
   0.6.4 what's new file. (:issue:`898`)
 * Compatibility with cftime 1.1. (:issue:`895`)
 * Add Python3.8 to Azure Pipelines CI (:issue:`903`)(:pull:`904`)
-
+* Minor implemention changes to avoid runtime and deprecation warnings in
+  :py:func:`~pvlib.clearsky.detect_clearsky`,
+  :py:func:`~pvlib.iam.martin_ruiz_diffuse`,
+  :py:func:`~pvlib.losses.soiling_hsu`,
+  and various test functions.
 
 Documentation
 ~~~~~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.7.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.2.rst
@@ -16,6 +16,9 @@ Enhancements
 * Add Kimber soiling model :py:func:`pvlib.losses.soiling_kimber` (:pull:`860`)
 * Add :py:func:`~pvlib.iotools.tmy3_monotonic_index` to fix TMY3
   data coerced to a single year to be monotonically increasing (:pull:`910`)
+* Add ``monotonic_index`` kwarg to :py:func:`~pvlib.iotools.read_tmy3` to fix
+  TMY3 data coerced to a single year to be monotonically increasing, default is
+  ``False`` and warns that it will be changed to ``True`` in v0.8 (:pull:`910`)
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.7.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.2.rst
@@ -14,7 +14,7 @@ Enhancements
   the original ``Date (MM/DD/YYYY)`` and ``Time (HH:MM)`` columns that the
   indices were parsed from (:pull:`866`)
 * Add Kimber soiling model :py:func:`pvlib.losses.soiling_kimber` (:pull:`860`)
-* Add :py:func:`~pvlib.iotools.fix_tmy3_coerce_year_monotonicity` to fix TMY3
+* Add :py:func:`~pvlib.iotools.tmy3_monotonic_index` to fix TMY3
   data coerced to a single year to be monotonically increasing (:pull:`910`)
 
 Bug fixes
--- a/docs/sphinx/source/whatsnew/v0.7.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.7.2.rst
@@ -16,6 +16,9 @@ Enhancements
 * TMY3 dataframe returned by :py:func:`~pvlib.iotools.read_tmy3` now contains
   the original ``Date (MM/DD/YYYY)`` and ``Time (HH:MM)`` columns that the
   indices were parsed from. (:pull:`866`)
+* Add :py:func:`~pvlib.pvsystem.PVSystem.faiman` and added 
+  ``temperature_model='faiman'`` option to :py:class:`~pvlib.modelchain.ModelChain`
+  (:pull:`897`) (:issue:`836`).
 * Add Kimber soiling model :py:func:`pvlib.losses.soiling_kimber`. (:pull:`860`)
 
 Bug fixes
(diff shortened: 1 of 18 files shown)


mikofski:
OK @CameronTStark  I think this is ready. @kanderso-nrel any comments?

wholmgren:
> if tmy is coerced to year, then the last hour is in the same year, before the other timestamps, so timestamps are no monotonic

Should this be a public, general purpose function in `pvlib.iotools.tmy`? I don't think it makes sense to force the soiling functions to deal with the quirks of the tmy format and pvlib's implementation.

pvlib/losses.py
```diff
@@ -9,13 +9,20 @@
 import pandas as pd
 from pvlib.tools import cosd
 
+_RAIN_ACC_PERIOD = pd.Timedelta('1h')
```
cwhanse: What is the argument for these new constants? They aren't reused and less explicit at the API than e,g.  `pd.Timedelta('1h'). 

Nit: _RAIN_ACCUM_PERIOD to keep the constant the same as the kwarg

pvlib/losses.py
```diff
@@ -9,13 +9,20 @@
...
 
 def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
-                depo_veloc={'2_5': 0.004, '10': 0.0009},
-                rain_accum_period=pd.Timedelta('1h')):
+                depo_veloc=None, rain_accum_period=_RAIN_ACC_PERIOD,
+                is_tmy=False):
     """
     Calculates soiling ratio given particulate and rain data using the model
-    from Humboldt State University [1]_.
+    from Humboldt State University (HSU).
```
cwhanse: Thanks for also improving the docstring

pvlib/losses.py
```diff
@@ -89,14 +115,26 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
     return soiling_ratio
 
 
+def _fix_tmy_monotonicity(rain_index, rain_vals, timestep_interval, rain_tz,
+                          rain_name):
+    # fix TMY to be monotonically increasing by rolling index by 1 interval
```
cwhanse: General question: what is the reasoning for replacing the time index here for soiling functions, rather than incrementing the year for the last time step in `read_tmy3(<tmy file>, coerce_year=<year>)`?

pvlib/losses.py
```diff
@@ -89,14 +115,26 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
...
+
+
 def soiling_kimber(rainfall, cleaning_threshold=6, soiling_loss_rate=0.0015,
                    grace_period=14, max_soiling=0.3, manual_wash_dates=None,
-                   initial_soiling=0, rain_accum_period=24, istmy=False):
+                   initial_soiling=0, rain_accum_period=24, is_tmy=False):
     """
-    Calculate soiling ratio with rainfall data and a daily soiling rate using
-    the Kimber soiling model [1]_.
+    Calculates fraction of energy lossed due to soiling given rainfall data and
```
cwhanse:
```suggestion
    Calculates fraction of energy lost due to soiling given rainfall data and
```

pvlib/losses.py
mikofski: For the deposition velocity, I believe having a mutable default input arg is a Python no-no, I've seen linters ding this, and there are blog posts that also say to avoid this pattern.

For the default rain accumulation period, I'm not sure, and I'm probably wrong, but I feel like having a function/class as a input arg default is an extra burden. Why not just call it once? Although, maybe behind the scenes that's exactly what Python does when it imports a module? I don't know. I agree with you on readability tho, so I'm okay with reverting this, it's not a blocker for me, just a preference. Also, "premature optimisation ..." Sorry

cwhanse: The dict as default was probably my idea, out of ignorance. Are you willing to separate `depo_veloc` into two separate inputs? That would be more explicit and avoids hiding the default in a private constant.

pvlib/losses.py
mikofski: Because it will change the existing, expected API for `read_tmy3()` so I thought it deserved a bigger discussion. I think your comment is similar to Will's above. I'm okay with either approach, although I have a preference for breaking the API and changing `read_tmy3()` to be monotonic when coercing the year. All in favor?

pvlib/losses.py
kanderso-nrel: > maybe behind the scenes that's exactly what Python does 

Yep, default arguments are evaluated and stored when the function is defined, meaning default arguments are created once and re-used forever after.  The argument against mutable defaults is that changes in one function invocation persist to the next: 

```
In [18]: def foo(lst=[], ts=datetime.datetime.now()):
    ...:     lst.append(5)
    ...:     print(ts, lst)
    ...: 
    ...: 
    ...: foo()
    ...: time.sleep(1)
    ...: foo()
    ...: 
2020-02-24 10:52:10.714337 [5]
2020-02-24 10:52:10.714337 [5, 5]
```

So strictly speaking it's okay to have mutable defaults as long as you never mutate them.  But avoiding them at all will ensure you never accidentally mutate them by accident.

mikofski:
FYI: AFAIK, Pandas indices are immutable, therefore I could not figure out how to change just the last index, so the only fix that seemed to work was to replace the whole index.

I don't know if this question was asked, as I could not tell from the context of some comments if there was concern over
- where in the codebase of the fix was located
- or how the fix was implemented

I think it was the former, but just in case it was the latter.

pvlib/losses.py
wholmgren: I used to be a stickler about no mutable default arguments but I now think they're ok if used responsibly. My preferences are in order:

1. split dict into two different kwargs
2. leave the dict alone
3. set kwarg to `None`, set default value within function
4. set kwarg to `None`, use default value from private constant

I'm -1 on the change to the `rain_accum_period` kwarg.

mikofski: Thanks Kevin for the explanation!

>Are you willing to separate `depo_veloc` into two separate inputs? That would be more explicit and avoids hiding the default in a private constant.

The pattern I've seen implemented most often is to put `None` as the default or some other appropriate constant, then check for it and set the default inside the scope of the function. IMO having a lot of arguments makes the function harder to use and there's a linter warning for that too.

I don't think we should depend on the function signature to inform the user about the API, that should be the docstring's job, although ideally they are consistent, and arguably having explicit defaults is more readable. So these are just my opinions, I don't really have a strong feeling about what's best. Paint the bike shed white ... Sorry

pvlib/losses.py
wholmgren: I'm in favor, but should it be done in 0.8 instead of 0.7.2?

pvlib/losses.py
cwhanse: > put None as the default or some other appropriate constant, then check for it and set the default inside the scope of the function. 

This. Was just reading a blog post which recommended this approach.

mikofski:
oops!

mikofski:
Note the codecov/proj/tests failure is spurious. Otherwise, I think I've addressed most of the ideas/feedback from above:

- moved the fix for tmy3 coerced to year to `iotools.tmy` and added a test for it, also removing it and the `is_tmy` arg and all references to tmy from losses or `test_losses`.
- I reverted the changes to `soiling_hsu`, and added a test for default deposition velocity arguments to keep coverage up
- fixed the spelling typo in the Kimber docstring to say, "lost" not "lossed"

Let me know if there's any more feedback

pvlib/iotools/tmy.py
```diff
@@ -526,3 +527,44 @@ def _read_tmy2(string, columns, hdr_columns, fname):
         columns=columns.split(',')).tz_localize(int(meta['TZ'] * 3600))
 
     return data, meta
+
+
+def fix_tmy3_coerce_year_monotonicity(tmy3):
+    """
+    Fix TMY3 coerced to a single year to be monotonically increasing by
```
cwhanse:
```suggestion
    Fix the time index of TMY3 coerced to a single year to be monotonically increasing by
```

pvlib/iotools/tmy.py
```diff
@@ -526,3 +527,44 @@ def _read_tmy2(string, columns, hdr_columns, fname):
...
+    changing the last record to be in January 1st of the next year.
+
+    Parameters
+    ----------
+    tmy3 : pandas.DataFrame
+        TMY3 data frame from :func:`pvlib.iotools.read_tmy3` with year coearced
```
cwhanse:
```suggestion
        TMY3 data frame from :func:`pvlib.iotools.read_tmy3` with year coerced
```

mikofski:
great feedback Cliff, thanks!

>would `fix_tmy3_` be simpler if it creates a new last index value, adds a row to the dataframe with the new last index, and then drop the old last index? Not a big deal, but I wonder if there's a performance cost of creating a whole new index.

Great idea, I checked it out, and surprisingly, although more lines of code, replacing the index is about 10x faster.

### setup
get Greensboro TMY3, check the last few indices, note last index is midnight, Jan 1, 1990
```python
>>> import numpy as np
>>> import pandas as pd
>>> from pvlib.tests.conftest import DATA_DIR
>>> from pvlib.iotools import read_tmy3

>>> greensboro, _ = read_tmy3(DATA_DIR / '723170TYA.CSV', coerce_year=1990)
>>> greensboro[['DryBulb', 'DewPoint', 'RHum', 'Pressure', 'Wspd']].tail()

#                            DryBulb  DewPoint  RHum  Pressure  Wspd
# 1990-12-31 20:00:00-05:00      2.8       1.7    92       981   2.6
# 1990-12-31 21:00:00-05:00      2.8       1.1    89       981   2.6
# 1990-12-31 22:00:00-05:00      2.8       1.1    89       981   2.1
# 1990-12-31 23:00:00-05:00      2.8       0.6    85       981   2.6
# 1990-01-01 00:00:00-05:00      2.2       0.6    89       980   2.6
```

### method 1
drop last row, replace with new "fixed" last row
```python
>>> %%timeit
... last_row = greensboro.iloc[-1]
... last_row.name = last_row.name.replace(year=1991)
... greensboro_1 = greensboro.drop(index=greensboro.index[-1]).append(last_row )

# 28.4 ms ± 813 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
# 29 ms ± 1.27 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

>>> greensboro_1[['DryBulb', 'DewPoint', 'RHum', 'Pressure', 'Wspd']].tail()

#                            DryBulb  DewPoint  RHum  Pressure  Wspd
# 1990-12-31 20:00:00-05:00      2.8       1.7    92       981   2.6
# 1990-12-31 21:00:00-05:00      2.8       1.1    89       981   2.6
# 1990-12-31 22:00:00-05:00      2.8       1.1    89       981   2.1
# 1990-12-31 23:00:00-05:00      2.8       0.6    85       981   2.6
# 1991-01-01 00:00:00-05:00      2.2       0.6    89       980   2.6
```

### method 2
replace index with corrected index
```python
>>> %%timeit
... ind_tz = greensboro.index.tz
... ind_val = greensboro.index.values
... dt = ind_val [1] - ind_val [0]
... ind_val = np.roll(ind_val , 1) + dt
... new_ind = pd.DatetimeIndex(ind_val , tz='UTC').tz_convert(ind_tz )
... greensboro_2 = greensboro.copy()
... greensboro_2 .index = new_ind

# 3.29 ms ± 129 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# 3.16 ms ± 76.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

>>> greensboro_2[['DryBulb', 'DewPoint', 'RHum', 'Pressure', 'Wspd']].tail()

#                            DryBulb  DewPoint  RHum  Pressure  Wspd
# 1990-12-31 20:00:00-05:00      2.8       1.7    92       981   2.6
# 1990-12-31 21:00:00-05:00      2.8       1.1    89       981   2.6
# 1990-12-31 22:00:00-05:00      2.8       1.1    89       981   2.1
# 1990-12-31 23:00:00-05:00      2.8       0.6    85       981   2.6
# 1991-01-01 00:00:00-05:00      2.2       0.6    89       980   2.6
```

mikofski:
>I think a new kwarg for `read_tmy3`, `monotonic_index=True`, makes sense - it may become tiresome having to use `read_tmy3` and `fix_tmy3_` as a pair.

I agree, and was thinking the same thing, but shouldn't the default be false to preserve the existing API? Or is it such a mild change that it doesn't require a deprecation period? We could leave it as `False` for now, and then change it in version 0.8 -- any thoughts?

wholmgren:
Do we want to support a `monotonic_index` kwarg or a `fix_monotonic_index` function in the long term? I'd say no. If no, the next question is do we want to deprecate and warn the user now, or just make the breaking change in an upcoming release? I have a mild preference for make the breaking change and don't introduce temporary workarounds for deprecation. If we just make the change, can we do it in 0.7.2 or should we wait for 0.8? I have a mild preference for 0.8.

mikofski:
>If we just make the change, can we do it in 0.7.2 or should we wait for 0.8? I have a mild preference for 0.8.

I volunteer to maintain this until v0.8 and after. There's already the `is_tmy` kwarg in `soiling_kimber` to maintain until v0.8, sorry. My preference would be to add the monotonic_index, warn the user that the default will change to true in v0.8, and get rid of the hacky `is_tmy` kwarg in `soiling_kimber` and make it easier for folks to use `soiling_hsu`.

mikofski:
* the [Travis test failure](https://travis-ci.org/pvlib/pvlib-python/jobs/655133852#L2361) is unrelated - couldn't retrieve a tmy3 file
* the coverage patch failure is a mystery, this PR doesn't touch `pvsystem.py` or `test_pvsystem.py`
* coverge of project tests randomly shows up as zero for me, don't know why

mikofski:
Hi Cliff, I think I've address your concerns, although I set the `monotonic_index` default to `False` and added a `RuntimeWarning` if the user doesn't change it to `True`. I also added to the documentation.

pvlib/tests/iotools/test_tmy.py
```diff
@@ -77,3 +80,34 @@ def test_gh865_read_tmy3_feb_leapyear_hr24():
     # hour so check that the 1st hour is 1AM and the last hour is midnite
     assert data.index[0].hour == 1
     assert data.index[-1].hour == 0
+
+
+def test_fix_tmy_coerce_year_monotonicity():
```
cwhanse:
```suggestion
def test_tmy3_monotonic_index():
```

wholmgren:
From what I can gather, @mikofski and @cwhanse are in favor of labeling this a bug and fixing it now without any hacks. I'm split between calling it a bug and an API change, but leaning towards bug, especially because I really dislike the awkward procedure of introducing a kwarg in 0.7.2, then changing the behavior and deprecating the kwarg in 0.8, and removing the kwarg in 0.9. That sounds like too much to ask of pvlib.

So I guess I'm wondering where you really stand on this @mikofski? You have a PR that's mergeable, but is this really how you want to proceed?

mikofski:
If we were to agree that it's a bug, then how would we proceed? Fix it and then announce it as a breaking change with no deprecation?

If you all agree, then I'm okay with that approach because I honestly don't think fixing this bug now will negatively affect anyone.

I'll make some changes and push a proposal for you all to look at ok?

mikofski:
@wholmgren and @cwhanse thanks for your patience and guidance thinking this one all the way through - I really feel this is a much cleaner, long term solution that _does the right thing_. I'm ready for your comments, if you have any. Otherwise, feel free to merge

mikofski:
FYI: @wholmgren do you approve this PR now?

docs/sphinx/source/whatsnew/v0.7.2.rst
```diff
@@ -39,6 +42,8 @@ Bug fixes
   :py:func:`~pvlib.iam.martin_ruiz_diffuse`,
   :py:func:`~pvlib.losses.soiling_hsu`,
   and various test functions.
+* Fix :py:func:`~pvlib.iotools.read_tmy3` so that when coerced to a single year
+  the TMY3 data will be monotonically increasing (:pull:`910`)
```
cwhanse:
```suggestion
  the TMY3 index will be monotonically increasing (:pull:`910`)
```

pvlib/iotools/tmy.py
```diff
@@ -28,10 +28,12 @@ def read_tmy3(filename=None, coerce_year=None, recolumn=True):
         a relative file path, absolute file path, or url.
 
     coerce_year : None or int, default None
-        If supplied, the year of the data will be set to this value.
+        If supplied, the year of the data will be set to this value, except for
```
cwhanse:
```suggestion
        If supplied, the year of the index will be set to this value, except for
```

pvlib/iotools/tmy.py
```diff
@@ -28,10 +28,12 @@ def read_tmy3(filename=None, coerce_year=None, recolumn=True):
...
+        the last index which will be set to the *next* year so that the indices
```
cwhanse:
```suggestion
        the last index value which will be set to the *next* year so that the index```

File: pvlib/iotools/tmy.py [hunk 11]
```diff
@@ -28,10 +28,12 @@ def read_tmy3(filename=None, coerce_year=None, recolumn=True):
...
+        increase monotonically.
```
cwhanse:
```suggestion
        increases monotonically.
```

mikofski:
@CameronTStark or @wholmgren the one test failure is unrelated - I don't know why codecov randomly does this to me - so I think this is ready to merge. Thanks Cliff and Will for the great feedback and patience figuring out what the best way forward is here. I'm really happy with how this turned out 🎉

pvlib/losses.py
```diff
@@ -91,12 +98,12 @@ def soiling_hsu(rainfall, cleaning_threshold, tilt, pm2_5, pm10,
 
 def soiling_kimber(rainfall, cleaning_threshold=6, soiling_loss_rate=0.0015,
                    grace_period=14, max_soiling=0.3, manual_wash_dates=None,
-                   initial_soiling=0, rain_accum_period=24, istmy=False):
```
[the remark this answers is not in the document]
mikofski: I don't think we have to because the Kimber soiling model will be new in v0.7.2 when released. It was _not_ in v0.7.1. Does that make sense?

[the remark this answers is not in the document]
wholmgren: My mistake!

pvlib/iotools/tmy.py
```diff
@@ -139,15 +140,16 @@ def read_tmy3(filename=None, coerce_year=None, recolumn=True):
...
-    .. warning:: TMY3 irradiance data corresponds to the previous hour, so the
-        first hour is 1AM, corresponding to the net irradiance from midnite to
-        1AM, and the last hour is midnite of the *next* year, unless the year
-        has been coerced. EG: if TMY3 was 1988-12-31 24:00:00 this becomes
-        1989-01-01 00:00:00
+    .. warning:: TMY3 irradiance data corresponds to the *previous* hour, so
+        the first index is 1AM, corresponding to the irradiance from midnight
+        to 1AM, and the last index is midnight of the *next* year. For example,
+        if the last index in the TMY3 file was 1988-12-31 24:00:00 this becomes
+        1989-01-01 00:00:00 after calling :func:`~pvlib.iotools.read_tmy3`.
```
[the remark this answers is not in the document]
mikofski: good catch, thanks!

wholmgren:
Thanks again @mikofski for the fix and for your patience!

PR #933, opened 2020-03-11
Create plot_ghi_transposition.py example

ericf900:

Example showing transposition from GHi to POA, and comparing winter transposition gain to summer.


kanderso-nrel:
Hey @ericf900, cool example!  This PR's version of the docs are built here, so you can check out how your example looks when it's live (the build takes a few minutes after you push an update): https://pvlib-python--933.org.readthedocs.build/en/933/

A couple notes:
- The docs build system only build files whose names start with "plot_", that's why your example isn't showing up like the others.  If you rename the file and then push that commit, the docs will get rebuilt. 
- Typically we strip out the spyder header (`# -*- coding: utf-8 -*-` / `Created on...` etc) before merging into pvlib
- You should add an entry about this example under the Documentation header and (if you're comfortable with it) your name to the Contributors section in the version 0.7.2 whatsnew file: https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/whatsnew/v0.7.2.rst

kanderso-nrel:
Also can you merge upstream/master into your branch?  Some fixes were recently made to clean up the documentation build log and it would make reviewing this a little easier.

ericf900:
@kanderso-nrel Thanks for the suggestions! I made those updates and updated the PR

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,75 @@
+"""
+GHI to POA Transposition
+=========================
+
+Example of transposing clearsky GHI to POA
```
kanderso-nrel:
```suggestion
Example of generating clearsky GHI and POA irradiance.
```

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,75 @@
...
+"""
+
+# %%
+# This example shows how to use the get_clearsky method to generate clearsky
```
kanderso-nrel:
```suggestion
# This example shows how to use the :py:meth:`pvlib.location.Location.get_clearsky` method to generate clearsky
```

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,75 @@
...
+# GHI data as well as how to use the get_total_iradiance function to transpose
```
kanderso-nrel:
```suggestion
# GHI data as well as how to use the :py:meth:`pvlib.irradiance.get_total_irradiance` function to transpose
```

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,75 @@
...
+# GHI data to Plane of Array (POA) irradiance.
+
+from pvlib import location
+from pvlib.irradiance import get_total_irradiance
+import pandas as pd
+from matplotlib import pyplot as plt
+
+# For this example, we will be using Golden, Colorado
+tz = 'MST'
+lat, lon = 39.755, -105.221
```
kanderso-nrel: very nice choice of location, +1

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,75 @@
...
+
+# Create location object to store lat, lon, timezone
+site = location.Location(lat, lon, tz=tz)
+
+
+# Define a function to handle the transposition
+def get_irradiance(site_location, date, tilt, surface_azimuth):
+    # Creates one day's worth of 10 min intervals
+    times = pd.date_range(date, freq='10min', periods=6*24, tz=tz)
+    # Generate cleaersky data using the Ineichen model, which is the default
```
kanderso-nrel:
```suggestion
    # Generate clearsky data using the Ineichen model, which is the default
```

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,75 @@
...
+    # Get solar azimuth and zenith to pass to the transposition function
+    solar_position = site_location.get_solarposition(times=times)
+    # Use the get_total_irradiance function to transpose the GHI to POA
+    POA_irradiance = get_total_irradiance(
+        surface_tilt=tilt,
+        surface_azimuth=surface_azimuth,
+        dni=clearsky_ghi['dni'],
+        ghi=clearsky_ghi['ghi'],
+        dhi=clearsky_ghi['dhi'],
+        solar_zenith=solar_position['zenith'],
```
kanderso-nrel:
```suggestion
        solar_zenith=solar_position['apparent_zenith'],
```
FYI I'd like someone else to confirm that using apparent solar position is appropriate before making this change

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,75 @@
...
+# Plot GHI vs. POA for winter and summer
+fig, (ax1, ax2) = plt.subplots(1, 2)
+summer_irradiance['GHI'].plot(ax=ax1, label='GHI')
+summer_irradiance['POA'].plot(ax=ax1, label='POA')
+winter_irradiance['GHI'].plot(ax=ax2, label='GHI')
+winter_irradiance['POA'].plot(ax=ax2, label='POA')
+ax1.set_xlabel('Time of day (Summer)')
+ax2.set_xlabel('Time of day (Winter)')
+ax1.set_ylabel('Irradiance (W/m2)')
+ax2.set_ylabel('Irradiance (W/m2)')
```
kanderso-nrel: Maybe leave this y-label off?  The graphs are a little smushed together on readthedocs and it's the same ylabel in both cases.

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,75 @@
...
+ax1.legend()
+ax2.legend()
+plt.show()
+
+# %%
+# Note that in Summer, there is not much gain when comparing POA irradiance to
+# GHI. In the winter, however, POA irradiance is signifiacntly higher than
```
kanderso-nrel:
```suggestion
# GHI. In the winter, however, POA irradiance is significantly higher than
```

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,75 @@
...
+                         'POA': POA_irradiance['poa_global']})
+
+
+# Get irradiance data for summer and winter solstice, assuming 25 degree tilt
+# and a south facing array
+summer_irradiance = get_irradiance(site, '06-20-2020', 25, 180)
+winter_irradiance = get_irradiance(site, '12-21-2020', 25, 180)
+
+# Plot GHI vs. POA for winter and summer
+fig, (ax1, ax2) = plt.subplots(1, 2)
```
mikofski: This example is really great! Thanks!

I have a minor suggestion, can you [vertically stack the plots and share the x-axis](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.subplots.html)?
```python
fig, ax = plt.subplots(2, 1, sharex=True)  # stack plots (2, 1) and share the x axis
ax1, ax2 = ax  # or you can just use ax[0] instead of ax1, and ax[1] instead ax2, minor preference
```

I'm having a hard time reading the dates on the x-axis because they're a bit crowded, so I thought trying them vertical and sharing them might look nicer, but your call - not a blocker for me - also fine as is THANKS!

docs/examples/plot_ghi_transposition.py
cwhanse: According to the parameters for each transposition model option in `get_total_irradiance`, `apparent_zenith` should be passed through, rather than `zenith`.

It's a different matter to verify that each model actually expects `apparent_zenith` rather than `zenith`. I'll look at the underlying references when I can find the opportunity.

docs/examples/plot_ghi_transposition.py
kanderso-nrel: Unfortunately the indexes are 6 months apart so I don't think `sharex=True` will look good.  Could convert the datetime index to a nicer string form with `df.index.strftime("%H:%M")`.  Or could drop `sharex=True`, but then the x-label for the upper axes might overlap the bottom axes.

ericf900: Hi Mike and Kevin! I am looking at this now- I agree the x-axis is a bit cluttered. I think the benefit of keeping the plots side-by-side is that it highlights that, while there is not much of a gain for POA compared to GHI in the summer, overall irradiance is higher. 

I'll make some edits now and push up the new version!

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,76 @@
...
+
+# For this example, we will be using Golden, Colorado
+tz = 'MST'
+lat, lon = 39.755, -105.221
+
+# Create location object to store lat, lon, timezone
+site = location.Location(lat, lon, tz=tz)
+
+
+# Define a function to handle the transposition
```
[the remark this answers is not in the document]
wholmgren: I see what you're getting at, but I also think it's a nice example for novices to see how they can define a function and avoid copy/paste.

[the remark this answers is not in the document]
cwhanse: I wasn't clear - I'm saying keep the function but explain why it's here in the comment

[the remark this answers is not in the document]
ericf900: Hi Cliff- I incorporated these changes into the new version!

docs/examples/plot_ghi_transposition.py
mikofski: You're both right! Sorry for the distraction! Great work

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,82 @@
...
+# Create location object to store lat, lon, timezone
+site = location.Location(lat, lon, tz=tz)
+
+
+# Calculate clear-sky GHI and transpose to plane of array
+# Define a function so that we can re-use the sequence of operations with
+# Different locations
+def get_irradiance(site_location, date, tilt, surface_azimuth):
+    # Creates one day's worth of 10 min intervals
+    times = pd.date_range(date, freq='10min', periods=6*24, tz=tz)
```
mikofski: small tiny nit: could you change `tz` to all caps?
```python
# earlier
TZ = 'MST'

# here
times = pd.date_range(date, freq='10min', periods=6*24, tz=TZ)
```
my reasoning is that to me at least `tz` kinda comes out of no where, it looks like it's in the same scope as the function, but really it's not, it's from the module, so that's why making it all caps might be better?

Or as another option, pass it into the function as an argument, or is it possible to retrieve `tz` from the 1st arg, `site_location`, like `site_location.tz`?

Thanks!

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,82 @@
...
+tz = 'MST'
+lat, lon = 39.755, -105.221
+
+# Create location object to store lat, lon, timezone
+site = location.Location(lat, lon, tz=tz)
+
+
+# Calculate clear-sky GHI and transpose to plane of array
+# Define a function so that we can re-use the sequence of operations with
+# Different locations
```
mikofski: don't capitalize "Different"

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,82 @@
...
+def get_irradiance(site_location, date, tilt, surface_azimuth):
+    # Creates one day's worth of 10 min intervals
+    times = pd.date_range(date, freq='10min', periods=6*24, tz=tz)
+    # Generate clearsky data using the Ineichen model, which is the default
+    # The get_clearsky method returns a dataframe with values for GHI, DNI,
+    # and DHI
+    clearsky_ghi = site_location.get_clearsky(times)
```
mikofski: if `get_clearsky` returns a "dataframe with values for GHI, DNI, and DHI" then IMO maybe don't call it `clearsky_ghi`? How about just `clearsky` or `clearsky_irrad`?
```python
clearsky = site_location.get_clearsky(times)

# then below
dni = clearsky['dni']
ghi = clearsky['ghi']
dhi = clearsky['dhi']
```
what do you think?

docs/examples/plot_ghi_transposition.py
kanderso-nrel: +1 for `site_location.tz`

docs/examples/plot_ghi_transposition.py
```diff
@@ -0,0 +1,82 @@
...
+
+# Plot GHI vs. POA for winter and summer
+fig, (ax1, ax2) = plt.subplots(1, 2, sharey=True)
+summer_irradiance['GHI'].plot(ax=ax1, label='GHI')
+summer_irradiance['POA'].plot(ax=ax1, label='POA')
+winter_irradiance['GHI'].plot(ax=ax2, label='GHI')
+winter_irradiance['POA'].plot(ax=ax2, label='POA')
+ax1.set_xlabel('Time of day (Summer)')
+ax2.set_xlabel('Time of day (Winter)')
+ax1.set_ylabel('Irradiance (W/m2)')
```
mikofski: you can use latex here for W/m^2 if you want, it might look a little bit nicer?
```python
ax1.set_ylabel('Irradiance ($W/m^2$)')
```
image
I also added gridlines, but I noticed other examples don't have them, so just my personal preference, dk what the convention is.

ericf900: Thanks for the feedback Mike! Made these changes and am uploading a new version now :)

PR #934, opened 2020-03-15
Upgrade bifacial functions to work with pvfactors v1.4.1

### LINKED ISSUE #902: compability with pvfactors 1.4
The `pvfactors_timeseries` function does not work with versions of `pvfactors >= 1.3`. 

First reported in the [pvlib google group](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/pvlib-python/07zVOZJApus/nFAlvgF_FwAJ)

@anomam do you think it's worth continuing to maintain this wrapper function? It seemed like it made sense in the early days of `pvfactors` but you've done a lot since then it's not clear to me why people shouldn't just use `pvfactors` directly.

anomam:
- [x] Closes #902
 - [x] Tests added

**Brief description**
- update functions to work with pvfactors v1.4.1
- remove parallel simulations as pvfactors is now fast enough

### DIFF
--- a/pvlib/bifacial.py
+++ b/pvlib/bifacial.py
@@ -63,16 +63,16 @@ def pvfactors_timeseries(
 
     Returns
     -------
-    front_incident_poa_irradiance: numeric
+    poa_front: numeric
         Calculated incident irradiance on the front surface of the PV modules
         (W/m2)
-    back_incident_poa_irradiance: numeric
+    poa_back: numeric
         Calculated incident irradiance on the back surface of the PV modules
         (W/m2)
-    front_absorbed_poa_irradiance: numeric
+    poa_front_absorbed: numeric
         Calculated absorbed irradiance on the front surface of the PV modules
         (W/m2), after AOI losses
-    back_absorbed_poa_irradiance: numeric
+    poa_back_absorbed: numeric
         Calculated absorbed irradiance on the back surface of the PV modules
         (W/m2), after AOI losses
 
--- a/pvlib/bifacial.py
+++ b/pvlib/bifacial.py
@@ -5,8 +5,6 @@ plane-of-array irradiance under various conditions.
 
 import pandas as pd
 import numpy as np
-import warnings
-from pvlib._deprecation import pvlibDeprecationWarning
 
 
 def pvfactors_timeseries(
--- a/pvlib/bifacial.py
+++ b/pvlib/bifacial.py
@@ -9,12 +9,10 @@ import numpy as np
 
 def pvfactors_timeseries(
         solar_azimuth, solar_zenith, surface_azimuth, surface_tilt,
-        axis_azimuth,
-        timestamps, dni, dhi, gcr, pvrow_height, pvrow_width, albedo,
-        n_pvrows=3, index_observed_pvrow=1,
+        axis_azimuth, timestamps, dni, dhi, gcr, pvrow_height, pvrow_width,
+        albedo, n_pvrows=3, index_observed_pvrow=1,
         rho_front_pvrow=0.03, rho_back_pvrow=0.05,
-        horizon_band_angle=15.,
-        run_parallel_calculations=True, n_workers_for_parallel_calcs=2):
+        horizon_band_angle=15.):
     """
     Calculate front and back surface plane-of-array irradiance on
     a fixed tilt or single-axis tracker PV array configuration, and using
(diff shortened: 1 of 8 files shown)


wholmgren:
Thanks @anomam! 

Should we deprecate the `run_parallel_calculations, n_workers_for_parallel_calcs` kwargs or just go ahead and remove them? Usually I'd vote for deprecating in 0.7.2 and removing in 0.8, but in this case it might be better to just go ahead and remove them.

pvlib/bifacial.py
```diff
@@ -62,28 +60,21 @@ def pvfactors_timeseries(
...
-        to run calculations in parallel (recommended) or not.
-    n_workers_for_parallel_calcs: int, default 2
-        Number of workers to use in the case of parallel calculations. The
-        '-1' value will lead to using a value equal to the number
-        of CPU's on the machine running the model.
 
     Returns
     -------
-    front_poa_irradiance: numeric
+    front_incident_poa_irradiance: numeric
```
cwhanse: Are these names tied to pvfactors?  If not, maybe there are names that are more aligned with pvlib patterns, e.g., 
`poa_front ` in place of `front_incident_poa_irradiance`, `poa_front_absorbed` in place of `front_absorbed_poa_irradiance`, or even `effective_irradiance_front` if the spectral adjustment has been built in.  Tossing the idea out.

anomam: Hi @cwhanse , thanks for the feedback. The names in the docstrings are not tied to anything specific in pvfactors, so I updated it as suggested in f740dea

anomam:
> Thanks @anomam!
> 
> Should we deprecate the `run_parallel_calculations, n_workers_for_parallel_calcs` kwargs or just go ahead and remove them? Usually I'd vote for deprecating in 0.7.2 and removing in 0.8, but in this case it might be better to just go ahead and remove them.

Hi @wholmgren , thanks for the message and sorry for the delay.
I felt very inclined to remove it because I find the reporting classes very messy in parallel mode, and also because there's really not much of a reason to keep the parallel mode around in pvlib when the normal mode runs in a few seconds (parallel mode is not faster in most cases).
I keep the parallel mode in pvfactors mostly bc of ideas of future development I have for the package, but nothing that will happen in the foreseeable future.

On the other hand I do want to follow pvlib's best practices and I can bring it back and add a deprecation message to it if that's what the team needs. Please let me know!

wholmgren:
Sorry I wasn't clear. I'm all for removing the parallel mode, but wondering if we should keep those kwargs in pvlib until 0.8. The kwargs would become non-functional and only serve to emit a warning. The purpose would be to avoid breaking user code. But I don't have a strong feeling about whether or not it is worth the effort in this case.

anomam:
> Sorry I wasn't clear. I'm all for removing the parallel mode, but wondering if we should keep those kwargs in pvlib until 0.8. The kwargs would become non-functional and only serve to emit a warning. The purpose would be to avoid breaking user code. But I don't have a strong feeling about whether or not it is worth the effort in this case.

@wholmgren , thanks a lot for clarifying. I added a couple deprecation warnings in 7ba10f8. I hope this works!

@cwhanse , I think that this PR is ready for another review. Please let me know if you need any additional changes.

pvlib/bifacial.py
```diff
@@ -62,35 +63,43 @@ def pvfactors_timeseries(
...
-        instance the shapely geometries, the irradiance components incident on
-        all surfaces of the PV array (for all timestamps), etc.
-        In the pvfactors documentation, this is refered to as the "surface
-        registry".
+    poa_front_absorbed: numeric
+        Calculated absorbed irradiance on the front surface of the PV modules
+        (W/m2), after AOI losses
+    poa_back_absorbed: numeric
+        Calculated absorbed irradiance on the back surface of the PV modules
+        (W/m2), after AOI losses
```
wholmgren, line 84: If I understand the diff correctly, this change to the output will break user code. Ok with me but we should note it in the API Changes section. I guess there's not much point in deprecating the parallel kwargs if we're just changing the return type, so maybe best to just make users deal with the pain in the next release. I'm guessing that users of the bifacial module are relatively tolerant of api changes.

anomam: Thanks for the feedback @wholmgren :
* I could revert the output change and return some kind of placeholder for ``df_registries`` (pvfactors v1.4.1 does not use registries anymore). Or I can just update the API changes section. Just let me know what you prefer!
* I don't mind removing the deprecation warnings as well

wholmgren: In this case I think we're better off immediately conforming to the pvfactors 1.4 API, provided the API change is documented. Any objections?

anomam: Hi @wholmgren , sorry for the delay. I removed the deprecated inputs and warnings, and updated the API changes section of the newest whatsnew file. Please let me know if you need any other fixes

PR #943, opened 2020-03-27
Run ModelChain from POA irradiance

### LINKED ISSUE #536: Using ModelChain with Plane of Array (POA) as weather input
I really love the concept of the ModelChain in pvlib. However, I would like to be able to use the ModelChain to model DC and AC output of a given system (more than one module and inverter) using Plane of Array irradiance as the weather input, rather than inputting the GHI, DNI and DHI components (which I don't have measurements of) and then specifying a transposition model as part of the ModelChain.  Could this functionality of being able to specify:
`mc=pvlib.modelchain.ModelChain(transposition_model='None')`

and then inputing `poa_global` data as an input to `weather` data frame  in `mc.run_model`?

cwhanse:
- [x] Closes #536
 - [x] Tests added

Provides methods to run a ModelChain starting with broadband plane-of-array irradiance or effective irradiance, and with cell temperature or module temperature. Creates a number of helper functions for these two methods.


mikofski:
Does this close #536 ?

cwhanse:
> Does this close #536 ?

Not for the use case of a broadband instrument in the plane of array (e.g. pyranometer). Yes for the use case of a plane-of-array reference cell (with similar spectrum and surface reflectance).

mikofski:
>Not for the use case of a broadband instrument in the plane of array (e.g. pyranometer). Yes for the use case of a plane-of-array reference cell (with similar spectrum and surface reflectance).

For the case of POA measured by pyranometer should we recommend using GTI-DIRINT to back-estimate the DNI & DHI components before using `ModelChain` or should there be a model chain that calls GTI-DIRINT under the hood given pyranometer-POA input?

Either way, I believe the documentation for `ModelChain` should clarify the point you make regarding POA from reference cell vs. pyranometer.

wholmgren:
Thoughts on 1. the current PR's approach of adding e.g. `run_model_from_poa` methods vs. 2. subclassing `ModelChain` into e.g. `ModelChainPOA` and overriding `run_model` ?

cwhanse:
> Not for the use case of a broadband instrument in the plane of array (e.g. pyranometer). Yes for the use case of a plane-of-array reference cell (with similar spectrum and surface reflectance).
> 
> For the case of POA measured by pyranometer should we recommend using GTI-DIRINT to back-estimate the DNI & DHI components before using ModelChain or should there be a model chain that calls GTI-DIRINT under the hood given pyranometer-POA input?

I'm hesitant to build that POA to DNI etc. to POA components translation into the method. We can inform the user that the GTI-DIRINT function is available, but they may have other information (e.g. DHI from a nearby weather station) and can find their own path to POA components.
 
> Either way, I believe the documentation for ModelChain should clarify the point you make regarding POA from reference cell vs. pyranometer.

Absolutely. I think the docstrings for these methods should have an Example section.

pvlib/modelchain.py
```diff
@@ -961,6 +1035,97 @@ def run_model(self, weather, times=None):
         self.aoi_model()
         self.spectral_model()
         self.effective_irradiance_model()
+
+        self.run_model_from_effective_irradiance()
+
+        return self
+
+    def run_model_from_poa(self, total_irrad):
```
wholmgren, line 1043: To keep the API consistent with `run_model`, I think we should keep the full `weather` DataFrame as the input here and the method below.

pvlib/modelchain.py
```diff
@@ -961,6 +1035,97 @@ def run_model(self, weather, times=None):
         self.aoi_model()
         self.spectral_model()
         self.effective_irradiance_model()
```
wholmgren, line 1037: Maybe refactor these 3 methods into another method because the same call sequence is used here and below.

pvlib/modelchain.py
```diff
@@ -926,17 +952,65 @@ def prepare_inputs(self, weather, times=None):
             self.weather['temp_air'] = 20
         return self
 
+    def prepare_inputs_from_poa(self, total_irrad):
```
wholmgren, line 955: as discussed below, maybe stick with `weather` and the input argument. Note the `if` statement at the end of the method should fail because `weather` is not defined.

pvlib/modelchain.py
```diff
@@ -926,17 +952,65 @@ def prepare_inputs(self, weather, times=None):
...
+
+        self.total_irrad = total_irrad
+
+        self._prep_inputs_solar_pos()
+        self._prep_inputs_airmass()
+
+        # PVSystem.get_irradiance and SingleAxisTracker.get_irradiance
+        # and PVSystem.get_aoi and SingleAxisTracker.get_aoi
+        # have different method signatures. Use partial to handle
+        # the differences.
```
wholmgren, line 990: copied comment not relevant in this case

pvlib/modelchain.py
```diff
@@ -926,17 +952,65 @@ def prepare_inputs(self, weather, times=None):
...
+        if isinstance(self.system, SingleAxisTracker):
+            self._prep_inputs_tracking()
+        else:
+            self._prep_inputs_fixed()
+
+        if self.weather.get('wind_speed') is None:
+            self.weather['wind_speed'] = 0
+        if self.weather.get('temp_air') is None:
+            self.weather['temp_air'] = 20
```
wholmgren, line 999: if you move this to immediately after `self._prep_inputs_airmass()` then you have another common sequence with `prepare_inputs`.

pvlib/modelchain.py
```diff
@@ -961,6 +1035,97 @@ def run_model(self, weather, times=None):
...
+        Returns
+        -------
+        self
+
+        Assigns attributes: ``cell_temperature``, ``dc``, ``ac``,
+        ``losses``, ``diode_params`` (if dc_model is a single diode model)
+        """
+
+        # assign poa_global column for the temperature model
+        if 'poa_global' not in self.total_irrad.columns:
```
wholmgren, line 1118: no need for the `if` statement

pvlib/modelchain.py
```diff
@@ -961,6 +1035,97 @@ def run_model(self, weather, times=None):
...
+            try:
+                self.total_irrad['poa_global'] = \
+                    self.effective_irradiance
+            except KeyError:
```
wholmgren, line 1122: I don't understand what KeyError this is supposed to catch.

pvlib/modelchain.py
cwhanse: As an input in addition to `total_irrad`? Or assume that `poa_global` etc are in `weather`? That would be different from what `ModelChain.prepare_inputs` does currently (input is `weather`, output is `total_irrad` with the POA components)

pvlib/modelchain.py
cwhanse: The cell temperature models expect `ModelChain.total_irrad['poa_global']` to be present, and the DC models expect `ModelChain.effective_irradiance` to be present. The `if` block will use `effective_irradiance` in place of `total_irrad['poa_global']` if that latter is missing, but if `poa_global` is present it should be used.

Maybe we need to rethink the content of the different `ModelChain` attributes:
- `weather` is a Dataframe containing ghi, dni, dhi, air temperature, wind speed
- `total_irrad` is a Dataframe containing poa_global, poa_direct, poa_diffuse
- `effective_irradiance` is an array or Series, contains irradiance reaching the cells (POA broadband adjusted for reflection, spectrum, soiling, etc.)

pvlib/modelchain.py
cwhanse: Nor do I

wholmgren:
Good points. ModelChain is really a mess. 

The primary ModelChain entry points (`run_model`, `prepare_inputs`, and `complete_irradiance`) all expect a single DataFrame with all relevant information. These methods don't require any other time series information from ModelChain attributes. I think we should keep this design pattern. Some historical context may help: `ModelChain.run_model` originally included three keyword arguments: `times`, `irradiance`, and `weather`. Over several releases, we've reduced it to just weather. Let's not go in the opposite direction. Let's also not start to rely on users setting explicitly attributes such as `weather` - no other pvlib methods require this and it's generally not a pattern I want to encourage.

Keeping that single input DataFrame design pattern while allowing for the input of were previously intermediate modeling results certainly raises issues with how data is grouped into attributes and how those attributes are named. My proposed solution for that is to generalize the DataFrame argument name to something like `data`, and then assign parts of `data` to the relevant ModelChain attributes. I'm open to a broader rethink, but it's difficult to see how that could be done cleanly without breaking a lot of user code.

cwhanse:
@wholmgren where would we retain the intermediate outputs such as `effective_irradiance`? As new columns in `data` and as `ModelChain` attributes?  I'm not sure I see the value in adding columns to a dataframe that is just passed as input to other functions, maybe collect all the attributes into a `data` attribute which is the Dataframe? We would reserve `weather` for external data that is passed into ModelChain methods, separate from quantities generated by ModelChain methods.

`weather` would contain ghi, dni, dhi, temp_air, wind_speed, precipitable_water, relative_humidity.
`ModelChain.data` would be a dataframe with everything else?

wholmgren:
Sorry for being unclear. I am not in favor of modifying the input DataFrame. I'll try to be more concrete:

* `run_model(weather)` --> No changes. We could, for consistency with new methods, rename `weather` to `data`.

* `run_model_from_poa(data)` --> ModelChain selects known weather data (air temperature, wind speed, ghi, dni, dhi, precipitable_water, relative_humidity, more?) from `data`, stores on `self.weather`. Selects required POA columns from `data` and saves on `self.total_irrad`. Compute effective irradiance, module temperature, power as normal.

* `run_model_from_effective_irradiance(data)` --> ModelChain selects known weather data stores on `self.weather`. Extracts known poa data and stores on `self.total_irrad`. Selects required `effective_irradiance` from data, stores on `effective_irradiance`. Compute module temperature, power as normal.

We could also assign a `self.data` but it's not required and I'm not sure if it's a good or bad idea.

cwhanse:
> * `run_model_from_poa(data)` --> ModelChain selects known weather data (air temperature, wind speed, ghi, dni, dhi, precipitable_water, relative_humidity, more?) from `data`, stores on `self.weather`. Selects required POA columns from `data` and saves on `self.total_irrad`. Compute effective irradiance, module temperature, power as normal.

Working on this again. "selects known weather data" implies that we hardcode a list of keys that will comprise the weather dataframe attribute. This strikes me as difficult to maintain; when a new function is added somewhere (e.g. spectral model) which requires a new weather column, we'll have to add a key to this this list.

What about `weather` always, and only, containing weather-related quantities external to pvlib functions? Assign to `ModelChain.weather` to preserve the input. Then `run_model_from_poa` would have two arguments: `data`, from which `poa_global` etc. is pulled, and `weather` which gets stored on `ModelChain` for downstream functions.

To follow the rule "don't modify input dataframes" we shouldn't be filling in `temp_air` and `wind_speed` with default values if they aren't found in `weather`.

wholmgren:
>"selects known weather data" implies that we hardcode a list of keys that will comprise the weather dataframe attribute. This strikes me as difficult to maintain; when a new function is added somewhere (e.g. spectral model) which requires a new weather column, we'll have to add a key to this this list.

This is already the case. For example, air pressure in solar position

https://github.com/pvlib/pvlib-python/blob/49da0318256d8b46f90d4b29a7023de680f8410b/pvlib/modelchain.py#L884-L888

precipitable water in the first solar correction

https://github.com/pvlib/pvlib-python/blob/49da0318256d8b46f90d4b29a7023de680f8410b/pvlib/modelchain.py#L646-L650

I don't see how two DataFrame inputs improves the situation. I think it makes it worse since now I have to worry about if I am putting the data in the right place too.

> To follow the rule "don't modify input dataframes" we shouldn't be filling in temp_air and wind_speed with default values if they aren't found in weather. Not sure I see how to work around this yet, looking for feedback about weather first.

Correct. I think we can fix this with a `weather.copy()`

cwhanse:
Would you favor a new constant listing the permissible keys for `weather`?

wholmgren:
> Would you favor a new constant listing the permissible keys for weather? Similar to pvlib.pvsystem._DC_MODEL_PARAMS etc.

Yes, good idea.

cwhanse:
@wholmgren see if this is closer to what you had in mind. Since the SPI use case also anticipates entering ModelChain with different temperature data (air, back-of-module, or cell), I've added a `prepare_temperature` function to deliver `cell_temperature` to a `ModelChain` instance. Feels rather hacky with the various try's and if's.

pvlib/modelchain.py
```diff
@@ -972,7 +1140,88 @@ def run_model(self, weather, times=None):
...
+
+        self.aoi_model()
+        self.spectral_model()
+        self.effective_irradiance_model()
+
+        self.run_model_from_effective_irradiance(data)
+
+        return self
+
+    def run_model_from_effective_irradiance(self, data=None):
```
wholmgren, line 1195: does this method work as a standalone entry point? It appears we need something like `prepare_inputs_from_effective_irradiance`.

pvlib/modelchain.py
```diff
@@ -931,23 +996,125 @@ def prepare_inputs(self, weather, times=None):
...
+        """
+        try:
+            self.cell_temperature = data['cell_temperature']
+            if self.temperature_model is not None:
+                warnings.warn('using ''cell_temperature'' provided in input',
+                              ' rather than specified temperature model: {0}'
+                              .format(self.temperature_model.__name__))
+            return self
+        except:
+            pass
```
wholmgren, line 1073: We don't currently allow `self.temperature_model` to be `None`. In any case, I don't see why this warning would be helpful.

pvlib/modelchain.py
```diff
@@ -931,23 +996,125 @@ def prepare_inputs(self, weather, times=None):
...
+        # If module_temperature is in input data we can use the SAPM cell
+        # temperature model.
+        if ('module_temperature' in data) and \
+            (self.temperature_model.__name__ == 'sapm'):
+            # use SAPM cell temperature model only
+            self.cell_temperature = pvlib.temperature.sapm_cell_from_module(
+                module_temperature=data['module_temperature'],
+                poa_global=self.total_irrad['poa_global'],
+                deltaT=self.temperature_model_parameters['deltaT'])
+            return self
```
wholmgren, line 1086: If I understand correctly, we can remove this `return self` and add `else: self.temperature_model()`

wholmgren: If `self.temperature_model` needs some more data then that method can complain about it.

pvlib/modelchain.py
```diff
@@ -840,85 +850,140 @@ def complete_irradiance(self, weather, times=None):
...
+    def _assign_total_irrad(self, data):
+        key_list = [k for k in POA_DATA_KEYS if k in data]
+        self.total_irrad = data[key_list].copy()
+        return self
+
     def prepare_inputs(self, weather, times=None):
         """
         Prepare the solar position, irradiance, and weather inputs to
-        the model.
+        the model, starting with GHI, DNI and DHI.
```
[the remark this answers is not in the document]
cwhanse: To parallel the docstrings for `prepare_inputs_from_poa` etc.

pvlib/modelchain.py
```diff
@@ -931,23 +996,125 @@ def prepare_inputs(self, weather, times=None):
...
+                poa_global=self.total_irrad['poa_global'],
+                deltaT=self.temperature_model_parameters['deltaT'])
+            return self
+
+        # Calculate cell temperature from weather data. Cell temperature models
+        # expect total_irrad['poa_global']. If not found, substitute
+        # self.effective_irradiance.
+        if 'poa_global' not in self.total_irrad:
+            try:
+                self.total_irrad['poa_global'] = self.effective_irradiance
```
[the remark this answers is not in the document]
cwhanse: The problem is that the ModelChain cell temperature methods reference `ModelChain.total_irrad['poa_global']` as the irradiance input to the cell temperature function.  I could change `ModelChain.prepare_temperature` to expect that `ModelChain.total_irrad['poa_global']` has already been assigned, but, this means that `ModelChain.run_from_effective_irradiance` requires data (poa_global) that is upstream of `effective_irradiance` and which the user may not have (poa_global).

I would prefer not to assign `effective_irradiance` to `total_irrad['poa_global']` because that's not what 'poa_global' means. Maybe have the cell temperature methods fall back to `ModelChain.effective_irradiance` if 'poa_global' is not found?

pvlib/modelchain.py
cwhanse: It doesn't stand alone in this iteration. I wanted `run_model_from_effective_irradiance` to encapsulate the temperature/DC/AC/losses models, and use this function at the end of e.g. `run_model_from_poa` to avoid duplicate code. Doing so means that `run_model_from_effective_irradiance` can't set `ModelChain.total_irrad` for example.

I think a way forward is to make the current function private and make a separate public function 1) sets weather, total_irrad etc. then 2) runs the model forward from effective_irradiance.

pvlib/modelchain.py
cwhanse: I agree but we need to resolve the [issue](https://github.com/pvlib/pvlib-python/pull/943#discussion_r482330580) with the effective irradiance entrypoint first

pvlib/modelchain.py
```diff
@@ -18,6 +18,16 @@
 from pvlib._deprecation import pvlibDeprecationWarning
 from pvlib.tools import _build_kwargs
 
+WEATHER_KEYS = set(['ghi', 'dhi', 'dni', 'wind_speed', 'temp_air',
```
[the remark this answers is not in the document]
wholmgren: sticking with comments is fine but we could consider docstrings once #1036 is merged

pvlib/modelchain.py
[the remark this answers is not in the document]
wholmgren: > Maybe have the cell temperature methods fall back to ModelChain.effective_irradiance if 'poa_global' is not found?

Good idea. Would that fallback need to emit a warning? I'd lean towards no but I don't have a strong opinion about it.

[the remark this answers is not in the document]
cwhanse: Were it my analysis I would appreciate the warning that pvlib swapped the irradiance inputs for me.

Let's take this up in a follow-on PR since it will rework the cell temperature methods, rather than the `run_model` methods.

cwhanse:
Test is unrelated. To do, perhaps in follow-on PRs:
- [ ]  docstrings for new ModelChain constants
- [ ]  implement fallback to effective irradiance for cell temperature models, if `ModelChain.total_irrad['poa_global']` is absent

pvlib/tests/test_modelchain.py
```diff
@@ -327,6 +334,90 @@ def test_run_model_tracker(sapm_dc_snl_ac_system, location, weather, mocker):
...
+    data = weather.copy()
+    data[['poa_global', 'poa_diffuse', 'poa_direct']] = total_irrad
+    mc = ModelChain(sapm_dc_snl_ac_system, location)
+    mc.prepare_inputs_from_poa(data)
+    # weather attribute
+    for k in weather:
+        assert_series_equal(mc.weather[k], weather[k])
+    # total_irrad attribute
+    for k in total_irrad:
+        assert_series_equal(mc.total_irrad[k], total_irrad[k])
```
wholmgren, line 357: why the loop instead of assert_frame_equal?

docs/sphinx/source/whatsnew/v0.8.0.rst
```diff
@@ -96,6 +96,11 @@ Enhancements
 * Add :py:func:`pvlib.pvsystem.combine_loss_factors` as general purpose
   function to combine loss factors with a common index.
   Partialy addresses :issue:`988`. Contributed by Brock Taute :ghuser:`btaute`
+* Add capability to run a ModelChain starting with plane-of-array or effective
+  irradiance, or with back-of-module or cell temperature data. New methods are
+  :py:meth:`pvlib.modelchain.ModelChain.run_model_from_poa`,
+  :py:meth:`pvlib.modelchain.ModelChain.run_model_from_effective_irradiance`,
+  and :py:meth:`pvlib.modelchain.ModelChain.prepare_inputs_from_poa` (:pull:943)
```
wholmgren, line 103: also issue 536

wholmgren:
@cwhanse can you make issues for the follow up work?

docs/sphinx/source/whatsnew/v0.8.0.rst
cwhanse: Actually this will close #536, since we moved the entrypoint for `ModelChain.run_model_from_poa` back to include the spectral and reflection models.

cwhanse:
@srlightfoote FYI if you are still watching.

slightfoote:
thanks all!  I think this feature will help ease the burden for folks wanting to use pvlib to model expected performance characteristics of operational plants where pyranometer sensors mounted at plane of array are the norm.

cwhanse:
> thanks all! I think this feature will help ease the burden for folks wanting to use pvlib to model expected performance characteristics of operational plants where pyranometer sensors mounted at plane of array are the norm.

@slightfoote that's the motivating use case. You may be interested in one of our current projects to build a modeling service for O&M use cases. There's some information at solarperformanceinsight.github.io

docs/sphinx/source/whatsnew/v0.8.0.rst
wholmgren:
```suggestion
  and :py:meth:`pvlib.modelchain.ModelChain.prepare_inputs_from_poa` (:issue:`536`, :pull:`943`)
```

wholmgren:
Thanks @cwhanse!

pvlib/tests/test_modelchain.py
```diff
@@ -327,6 +334,88 @@ def test_run_model_tracker(sapm_dc_snl_ac_system, location, weather, mocker):
...
+
+
+def test_prepare_inputs_from_poa(sapm_dc_snl_ac_system, location,
+                                 weather, total_irrad):
+    data = weather.copy()
+    data[['poa_global', 'poa_diffuse', 'poa_direct']] = total_irrad
+    mc = ModelChain(sapm_dc_snl_ac_system, location)
+    mc.prepare_inputs_from_poa(data)
+    # weather attribute
+    assert_frame_equal(mc.weather, weather)
```
wholmgren, line 353: @cwhanse this `assert_frame_equal` fails when I run the tests locally but it seems to pass on the CI. Looking at the code, I expect that it should fail because `ModelChain._assign_weather` adds `wind_speed` and `temp_air` to `mc.weather`. Is my understanding correct? If so, why don't we see test failures on the CI?

cwhanse: Odd, that tested OK locally for me. Your understanding of `_assign_weather` is correct. I question my understanding of `pytest.fixture`

wholmgren: https://github.com/pvlib/pvlib-python/pull/1052

PR #960, opened 2020-04-30
clarify units for U1 and Uv temperature model coeff

### LINKED ISSUE #945: BUG:units for Uv & U1, convective thermal coefficient for Faiman and PVsyst, m/s should be in denominator
**Describe the bug**
the units are unclear, m/s should be in the denominator, but looks like numerator

https://github.com/pvlib/pvlib-python/blob/e526b55365ab0f4c473b40b24ae8a82c7e42f892/pvlib/temperature.py#L216

and 

https://github.com/pvlib/pvlib-python/blob/e526b55365ab0f4c473b40b24ae8a82c7e42f892/pvlib/temperature.py#L306

**Expected behavior**
There probably isn't a standard, and we could argue, but IMHO the units should always be as explict as possible. I like the way [wikipedia](https://en.wikipedia.org/wiki/Heat_transfer_coefficient#Composition) makes them very explicit using parentheses
image

**Screenshots**
pvsyst:
image
Faiman:
image

**Versions:**
 - ``pvlib.__version__``: e526b55365ab0f4c473b40b24ae8a82c7e42f892

**Additional context**
IMHO these units should uses parentheses like Wikipedia does for clarity:
* Uv & U1: [W/(m^2 K)/(m/s)] or [W/m^2/K/(m/s)] or [(W/m^2)/K/(m/s)]
* Uc & U0: [W/(m^2 K)] or [W/m^2/K] or [(W/m^2)/K]

mikofski:
* closes #945 
* group denominator together, same way it was done for U0 and Uc, ie: `W/(m^2 C)` instead of `(W/m^2 C)` which is ambiguous because it could mean `W/m^2 * C = W * C / m^2` which is not the same thing
* add missing division symbol before 2nd denominator `(m/s)` to get `W / (m^2 C) / (m/s)` instead of `(W/m^2 C) (m/s)` which looks like `WmC/ (m^2 s)` so it's ambiguous
* pvsyst help docs (https://www.pvsyst.com/help/thermal_loss.htm) say  `W/m²·k / m/s` which is also ambiguous but you can't win em all

 - [x] Closes #xxxx
 - [x] ~Tests added~
 - [x] ~Updates entries to [`docs/sphinx/source/api.rst`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes.~
 - [x] ~New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.~

### DIFF
--- a/docs/sphinx/source/whatsnew.rst
+++ b/docs/sphinx/source/whatsnew.rst
@@ -6,7 +6,7 @@ What's New
 
 These are new features and improvements of note in each release.
 
-.. include:: whatsnew/v0.7.3.rst
+.. include:: whatsnew/v0.8.0.rst
 .. include:: whatsnew/v0.7.2.rst
 .. include:: whatsnew/v0.7.1.rst
 .. include:: whatsnew/v0.7.0.rst
--- a/docs/sphinx/source/whatsnew/v0.8.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.8.0.rst
@@ -1,4 +1,4 @@
-.. _whatsnew_0730:
+.. _whatsnew_0800:
 
 v0.8.0 (Month day, year)
 -------------------------
@@ -17,7 +17,12 @@ Testing
 
 Documentation
 ~~~~~~~~~~~~~
-* Improved formatting and content of docstrings in `pvlib.atmosphere`
+* Improved formatting and content of docstrings in :py:mod:`pvlib.atmosphere`.
+  (:pull:`969`)
+* Fix LaTeX rendering in :py:func:`pvlib.singlediode.bishop88`. (:pull:`967`)
+* Clarify units for heat loss factors in
+  :py:func:`pvlib.temperature.pvsyst_cell` and
+  :py:func:`pvlib.temperature.faiman`. (:pull:`960`)
 
 Requirements
 ~~~~~~~~~~~~
@@ -25,3 +30,5 @@ Requirements
 Contributors
 ~~~~~~~~~~~~
 * Cliff Hansen (:ghuser:`cwhanse`)
+* Kevin Anderson (:ghuser:`kanderso-nrel`)
+* Mark Mikofski (:ghuser:`mikofski`)
--- a/pvlib/temperature.py
+++ b/pvlib/temperature.py
@@ -293,7 +293,8 @@ def pvsyst_cell(poa_global, temp_air, wind_speed=1.0, u_c=29.0, u_v=0.0,
 
     u_v : float, default 0.0
         Combined heat loss factor influenced by wind. Parameter :math:`U_{v}`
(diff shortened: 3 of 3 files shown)


kanderso-nrel:
In this and other cases with combined units, I wonder if it might be better to use a `:math:` tag to improve the rendering.  The plaintext display is a little difficult to parse:

image

mikofski:
>In this and other cases with combined units, I wonder if it might be better to use a :math: tag to improve the rendering.

Everywhere? what about units that are simpler to parse? Probably best to be consistent, right?
https://github.com/pvlib/pvlib-python/blob/3ce8e1a5e5db42d09982e0153b65b4a8547180fa/pvlib/temperature.py#L46

I'm okay with making the shift to use `:math:` to render all units, but maybe that should be a separate issue? It would be a bit of an undertaking, good first timer maybe?

cwhanse:
> Everywhere? what about units that are simpler to parse? Probably best to be consistent, right?

Perhaps in the few instances where the unit involves a compound fraction. I'm -1 for using :math: for units otherwise.

mikofski:
Quick poll, so we've decided to use `: math:` to render these ugly confusing units here only, EG:
```python
in :eq:`pvsyst` :math:`\left[\frac{W/{\left(m^2 C\right)}{m/s}\right]`.
```

Looks like this
![latex] (https://latex.codecogs.com/png.latex?\left[&space;\frac{W/{\left(m^2&space;C\right)}}{m/s}&space;\right])

adriesse:
Could we do:

```
 W/m2
------
C(m/s)
```
This would be more easily associated with the variables going in.

mikofski:
>Could we do: `W/m^2 / (C(m/s))`

I'm fine with

    :math:`\left[\frac{W/{m^2 }}{C \left( m/s \right )}\right]`

which looks like this:
LaTeX equation}\right])
[Powered by CodeCogs](https://www.codecogs.com/eqnedit.php?latex=\left[\frac{W/{m^2}}{C\left(m/s\right)}\right])

But this makes me wonder if the `U0` and `U_c` should also be changed to look the same?
LaTeX equation
[Powered by CodeCogs](https://www.codecogs.com/eqnedit.php?latex=\left[\frac{W/{m^2}}{C}\right])

What do y'all think?

mikofski:
Also should we be using Kelvin instead of Celsius?

adriesse:
Sure, but I'd also be ok with the plain text:```[W/m^2/C]``` and ```[W/m^2/C/(m/s)]```.
My preference is for Celsius.

cwhanse:
> My preference is for Celsius.

+1

mikofski:
sorry stickler by inline math don't render if it's on two lines 😕

mikofski:
What do you think @CameronTStark? Is this good?

CameronTStark:
LGTM @mikofski, thanks!

PR #968, opened 2020-05-19
Add partial shading example using new reverse bias functionality

kanderso-nrel:

I was checking out the new reverse bias functionality to teach myself about cell-level modeling and thought it might make a good example.  Here is the first pass.  In addition to whether people think this is a good addition to the example gallery, I'd also be interested in any feedback about the validity of the approach in general just for my own learning. 

Built version: https://pvlib-python--968.org.readthedocs.build/en/968/auto_examples/plot_partial_module_shading_simple.html

### DIFF
--- a/docs/examples/plot_partial_module_shading_simple.py
+++ b/docs/examples/plot_partial_module_shading_simple.py
@@ -43,11 +43,10 @@ import matplotlib.pyplot as plt
 
 from scipy.constants import e as qe, k as kB
 
-kb = 1.380649e-23  # J/K
-qe = 1.602176634e-19  # C
-# kb is J/K, qe is C=J/V
-# kb * T / qe -> V
-Vth = kb * (273.15+25) / qe
+# For simplicity, use cell temperature of 25C for all calculations.
+# kB is J/K, qe is C=J/V
+# kB * T / qe -> V
+Vth = kB * (273.15+25) / qe
 
 cell_parameters = {
     'I_L_ref': 8.24,
@@ -81,13 +80,11 @@ cell_parameters = {
 #    so we use the Bishop '88 method here.  This gives us a set of (V, I)
 #    points on the cell's IV curve.
 
-def simulate_full_curve(parameters, Geff, Tcell, method='brentq',
-                        ivcurve_pnts=1000):
+def simulate_full_curve(parameters, Geff, Tcell, ivcurve_pnts=1000):
     """
     Use De Soto and Bishop to simulate a full IV curve with both
     forward and reverse bias regions.
     """
-
     # adjust the reference parameters according to the operating
     # conditions using the De Soto model:
     sde_args = pvsystem.calcparams_desoto(
@@ -112,7 +109,7 @@ def simulate_full_curve(parameters, Geff, Tcell, method='brentq',
         'breakdown_voltage': parameters['breakdown_voltage'],
     }
     v_oc = singlediode.bishop88_v_from_i(
-        0.0, *sde_args, method=method, **kwargs
+        0.0, *sde_args, **kwargs
     )
     # ideally would use some intelligent log-spacing to concentrate points
     # around the forward- and reverse-bias knees, but this is good enough:
@@ -150,8 +147,8 @@ def plot_curves(dfs, labels, title):
     return axes
 
 
-cell_curve_full_sun = simulate_full_curve(cell_parameters, Geff=1000, Tcell=40)
-cell_curve_shaded = simulate_full_curve(cell_parameters, Geff=200, Tcell=40)
+cell_curve_full_sun = simulate_full_curve(cell_parameters, Geff=1000, Tcell=25)
+cell_curve_shaded = simulate_full_curve(cell_parameters, Geff=200, Tcell=25)
 ax = plot_curves([cell_curve_full_sun, cell_curve_shaded],
                  labels=['Full Sun', 'Shaded'],
                  title='Cell-level reverse- and forward-biased IV curves')
@@ -177,7 +174,6 @@ ax = plot_curves([cell_curve_full_sun, cell_curve_shaded],
 # added.  However, because each cell's curve is discretized and the currents
 # might not line up, we align each curve to a common set of current values
 # with interpolation.
-
 
 def interpolate(df, i):
     """convenience wrapper around scipy.interpolate.interp1d"""
@@ -276,7 +272,7 @@ kwargs = {
     'cell_parameters': cell_parameters,
     'poa_direct': 800,
     'poa_diffuse': 200,
-    'Tcell': 40
+    'Tcell': 25
 }
 module_curve_full_sun = simulate_module(shaded_fraction=0, **kwargs)
 module_curve_shaded = simulate_module(shaded_fraction=0.1, **kwargs)
@@ -308,7 +304,7 @@ for diffuse_fraction in np.linspace(0, 1, 11):
         df = simulate_module(cell_parameters,
                              poa_direct=(1-diffuse_fraction)*1000,
                              poa_diffuse=diffuse_fraction*1000,
-                             Tcell=40,
+                             Tcell=25,
                              shaded_fraction=shaded_fraction)
         data.append({
             'fd': diffuse_fraction,
--- a/docs/examples/plot_partial_module_shading_simple.py
+++ b/docs/examples/plot_partial_module_shading_simple.py
@@ -191,8 +191,8 @@ def simulate_module(cell_parameters, poa_direct, poa_diffuse, Tcell,
     include_partial_cell = (shaded_fraction < 1)
     half_substring_curves = (
         [df_lit] * (nrow - nrow_full_shade - 1)
-        + ([df_partial] if include_partial_cell else [])
-        + [df_shaded] * nrow_full_shade
+        + ([df_partial] if include_partial_cell else [])  # noqa: W503
+        + [df_shaded] * nrow_full_shade  # noqa: W503
     )
     df = combine_series(half_substring_curves)
     # all substrings perform equally, so can just scale voltage directly
--- a/docs/examples/plot_partial_module_shading_simple.py
+++ b/docs/examples/plot_partial_module_shading_simple.py
@@ -58,6 +58,12 @@ cell_parameters = {
     'breakdown_voltage': -15,
 }
 
+# %%
+# Simulating a cell's IV curve
+# ----------------------------
+#
+# First, calculate IV curves for individual cells:
+
 
 def simulate_full_curve(parameters, Geff, Tcell, method='brentq',
                         ivcurve_pnts=1000):
@@ -99,6 +105,34 @@ def simulate_full_curve(parameters, Geff, Tcell, method='brentq',
         'i': ivcurve_i,
         'v': ivcurve_v,
     })
+
+
+def plot_curves(dfs, labels):
+    """plot the forward- and reverse-bias portions of an IV curve"""
+    fig, axes = plt.subplots(1, 2)
+    for df, label in zip(dfs, labels):
+        forward = df.loc[df['v'] > 0, :]
+        reverse = df.loc[df['v'] < 0, :]
+        reverse.plot('v', 'i', label=label, ax=axes[0])
+        forward.plot('v', 'i', label=label, ax=axes[1])
+    return axes
+
+
+cell_curve_full_sun = simulate_full_curve(cell_parameters, Geff=1000, Tcell=40)
+cell_curve_shaded = simulate_full_curve(cell_parameters, Geff=200, Tcell=40)
+plot_curves([cell_curve_full_sun, cell_curve_shaded], ['Full Sun', 'Shaded'])
+
+
+# %%
+# Combining cell IV curves to create a module IV curve
+# ----------------------------------------------------
+#
+# To combine the individual cell IV curves and form a module's IV curve,
+# the cells in each substring must be added in series and the substrings
+# added in parallel.  To add in series, the voltages for a given current are
+# added.  However, because each cell's curve is discretized and the currents
+# might not line up, we align each curve to a common set of current values
+# with interpolation.
 
 
 def interpolate(df, i):
@@ -188,6 +222,7 @@ for diffuse_fraction in np.linspace(0, 1, 11):
 results = pd.DataFrame(data)
 results['pmp'] /= results['pmp'].max()  # normalize power to 0-1
 results_pivot = results.pivot('fd', 'fs', 'pmp')
+plt.figure()
 plt.imshow(results_pivot, origin='lower', aspect='auto')
 plt.xlabel('shaded fraction')
 plt.ylabel('diffuse fraction')
--- a/docs/examples/plot_partial_module_shading_simple.py
+++ b/docs/examples/plot_partial_module_shading_simple.py
@@ -1,0 +1,196 @@
+"""
+Calculating the effect of partial module shading
+================================================
+
+Example of modeling cell-to-cell mismatch loss from partial module shading.
+"""
+
+# %%
+# Even though the PV cell is the primary power generation unit, PV modeling is
+# often done at the module level for simplicity because module-level parameters
+# are much more available and it significantly reduces the computational scope
+# of the simulation.  However, module-level simulations are too coarse to be
+# able to model effects like cell to cell mismatch or partial shading.  This
+# example calculates cell-level IV curves and combines them to reconstruct
+# the module-level IV curve.  It uses this approach to find the maximum power
+# under various shading and irradiance conditions.
+#
+# The primary functions used here are:
+# - :py:meth:`pvlib.pvsystem.calcparams_desoto` to estimate the SDE parameters
+#   at the operating conditions
+# - :py:meth:`pvlib.singlediode.bishop88` to calculate the full cell IV curve,
+#   including the reverse bias region.
+#
+# .. note::
+#
+#     This example requires the reverse bias functionality added in pvlib 0.7.2
+#
+# .. warning::
+#
+#     Modeling partial module shading is complicated and depends significantly
+#     on the module's electrical topology.  This example makes some simplifying
+#     assumptions that are not generally applicable.  For instance, it assumes
+#     that all of the module's cell strings perform identically, making it
+#     possible to ignore the effect of bypass diodes.
+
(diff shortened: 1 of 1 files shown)


docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,217 @@
...
+import pandas as pd
+import numpy as np
+from scipy.interpolate import interp1d
+import matplotlib.pyplot as plt
+
+kb = 1.380649e-23  # J/K
+qe = 1.602176634e-19  # C
+Vth = kb * (273.15+25) / qe
+
+parameters = {
```
cwhanse, line 49: suggest calling this `cell_parameters` to emphasize that these aren't module parameters

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,217 @@
...
+        'breakdown_factor': parameters['breakdown_factor'],
+        'breakdown_exp': parameters['breakdown_exp'],
+        'breakdown_voltage': parameters['breakdown_voltage'],
+    }
+    v_oc = singlediode.bishop88_v_from_i(
+        0.0, *sde_args, method=method, **kwargs
+    )
+    vd = np.linspace(0.99*kwargs['breakdown_voltage'], v_oc, ivcurve_pnts)
+
+    ivcurve_i, ivcurve_v, _ = singlediode.bishop88(vd, *sde_args, **kwargs)
```
cwhanse, line 97: An alternative is to use `singlediode.bishop88_i_from_v`, which accepts the cell voltage rather than the diode voltage.

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def interpolate(df, i):
+    """convenience wrapper around scipy.interpolate.interp1d"""
+    f_interp = interp1d(np.flipud(df['i']), np.flipud(df['v']), kind='linear',
+                        fill_value='extrapolate')
+    return f_interp(i)
+
+
+def combine(dfs):
```
cwhanse, line 111: I have some code to submit to `pvlib.ivtools` that will include a function `add_series`.

cwhanse:
The calculation process looks correct to me.  If you have the enthusiasm for it, a plot of one of the IV curves would be interesting. You might also consider accounting for a bypass diode on the modeled string.

docs/examples/plot_partial_module_shading_simple.py
kanderso-nrel: I admit I didn't think carefully about this line.  What is the reasoning to use one over the other in this context?  Strictly speaking, is the breakdown voltage a cell voltage or a diode voltage?

cwhanse: Breakdown voltage is compared with diode (or junction) voltage.  Cell voltage is after series resistance etc.

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+        Tcell)
+    # build a list of IV curves for a single column of cells (half a substring)
+    include_partial_cell = (shaded_fraction < 1)
+    half_substring_curves = (
+        [df_lit] * (nrow - nrow_full_shade - 1)
+        + ([df_partial] if include_partial_cell else [])  # noqa: W503
+        + [df_shaded] * nrow_full_shade  # noqa: W503
+    )
+    df = combine_series(half_substring_curves)
+    # all substrings perform equally, so can just scale voltage directly
```
cwhanse, line 198: Here is where a bypass diode would be applied, substring_voltage = np.clip(lower=bypass_diode_voltage)  0.5V is a typical value.  A module without bypass diodes would be quite unusual. The module-level IV curve looks correct and illustrates why bypass diodes are used.

Cool example.

mikofski:
Have you considered using [`scipy.constants`](https://docs.scipy.org/doc/scipy/reference/constants.html) for elementary charge (qe) and Boltzmann constant (kB)?

```python
from scipy.constants import e as qe, k as kB

>>> qe
# 1.602176634e-19

>>>: kB
# 1.380649e-23
```

Alternately, you can take the temperature insensitive approximation and just keep the thermal voltage (Vth) at 26[mV]. Even though Vth depends on temperature, the diode equation itself is much less sensitive to Vth than it is the cubic temperature dependence of the dark current I0.

IMHO the temperature dependence of Vth is misleading to new PV students and leads them in the wrong direction because it's effect is opposite that of I0.

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+
+from pvlib import pvsystem, singlediode
+import pandas as pd
+import numpy as np
+from scipy.interpolate import interp1d
+import matplotlib.pyplot as plt
+
+kb = 1.380649e-23  # J/K
+qe = 1.602176634e-19  # C
+Vth = kb * (273.15+25) / qe
```
mikofski, line 47: units of `V` for thermal voltage to be consistent?

mikofski: Also maybe giving charge in units of Joules per volt `[J/V]` makes this conversion more intuitive to the reader?

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+    ivcurve_i, ivcurve_v, _ = singlediode.bishop88(vd, *sde_args, **kwargs)
+    return pd.DataFrame({
+        'i': ivcurve_i,
+        'v': ivcurve_v,
+    })
+
+
+def plot_curves(dfs, labels):
+    """plot the forward- and reverse-bias portions of an IV curve"""
+    fig, axes = plt.subplots(1, 2)
```
mikofski, line 112: Maybe add `sharey=True` to subplots to eliminate the redundant y-axis labels in between the two subplots?

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+    for df, label in zip(dfs, labels):
+        df.plot('v', 'i', label=label, ax=axes[0])
+        df.plot('v', 'i', label=label, ax=axes[1])
+        axes[0].set_xlim(right=0)
+        axes[1].set_xlim(left=0)
+
```
mikofski, line 118: Perhaps call `fig.tight_layout()` just to tidy up the plot before showing it?

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+    vd = np.linspace(0.99*kwargs['breakdown_voltage'], v_oc, ivcurve_pnts)
+
+    ivcurve_i, ivcurve_v, _ = singlediode.bishop88(vd, *sde_args, **kwargs)
+    return pd.DataFrame({
+        'i': ivcurve_i,
+        'v': ivcurve_v,
+    })
+
+
```
mikofski, line 109: I found this code snippet a little long and verbose. Maybe splitting it int 2 or 3 sections with some text in between would help the reader follow along with the tutorial better?

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+def plot_curves(dfs, labels):
+    """plot the forward- and reverse-bias portions of an IV curve"""
+    fig, axes = plt.subplots(1, 2)
+    for df, label in zip(dfs, labels):
+        df.plot('v', 'i', label=label, ax=axes[0])
+        df.plot('v', 'i', label=label, ax=axes[1])
+        axes[0].set_xlim(right=0)
+        axes[1].set_xlim(left=0)
```
mikofski, line 117: there's a lot of whitespace in this plot I think you can tailor the x & y limits to better use the space
```python
axes[0].set_ylim([0, 25])  # Imax = 25[A] is enough
axes[1].set_ylim([0, 25])  # to make the point to me
axes[1].set_xlim([0.0, 1.0])  # Vmax = 1.0[V] seems enough to me
```

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+
+    return axes
+
+
+cell_curve_full_sun = simulate_full_curve(cell_parameters, Geff=1000, Tcell=40)
+cell_curve_shaded = simulate_full_curve(cell_parameters, Geff=200, Tcell=40)
+plot_curves([cell_curve_full_sun, cell_curve_shaded], ['Full Sun', 'Shaded'])
+plt.gcf().suptitle('Cell-level reverse- and forward-biased IV curves')
+
```
mikofski, line 126: Maybe add some text here that explains that the reverse bias on the left side has a different scale than the forward bias on the right? Also perhaps summarize, something like, "The plots show the effect of shade decreases the current roughly proportionally, while the voltage decreases much less and the shape of the curve is about the same as if the entire curve were just shifted down by the fraction of shade."

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+
+# %%
+# Combining cell IV curves to create a module IV curve
+# ----------------------------------------------------
+#
+# To combine the individual cell IV curves and form a module's IV curve,
+# the cells in each substring must be added in series and the substrings
+# added in parallel.  To add in series, the voltages for a given current are
```
mikofski, line 134: >"... substrings added in parallel"

not necessarily, in most mono/poly-silicon modules, the substrings are in series as well. But each substring is in parallel with a bypass diode, perhaps that's what you meant?

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+    The current range is based on the first curve's current range.
+    """
+    df1 = dfs[0]
+    imin = df1['i'].min()
+    imax = df1['i'].max()
+    i = np.linspace(imin, imax, 1000)
+    v = 0
+    for df2 in dfs:
+        v_cell = interpolate(df2, i)
+        v += v_cell
```
mikofski, line 159: Or you could put the bypass diode here using `v = v.clip(-bypass_voltage)` like [Cliff commented](#r427429840) _except note the **negative** sign_, where `bypass_voltage` would be an argument of `combine_series(..., bypass_voltage=0.5)` and 0.5[V] is a typical diode trigger voltage (_aka_ [forward voltage drop or _threshold_ voltage](https://en.wikipedia.org/wiki/Diode#Forward_bias)). 

From this it's clearer that at the substring/submodule level reverse bias is not important because the substring is either bypassed or not. And the entire substring tends to behave near Vmp like the most shaded cell in the substring.

The reverse bias part of each cell is really only useful to tell us where the substring shifts from the most shaded cell to the rest of the substring, where there's a bump/jump in the current. This inflection or discontinuity occurs where the shaded cells drop into reverse bias because they can't handle the current, _ie_ after the first knee in the IV curve (usu. near Vmp, but not necessarily), the voltage of the shaded cells suddenly drops from positive to negative until the break down voltage, where the second knee starts, but unfortunately if the total breakdown voltage of the shaded cells _plus_ the forward bias of the remaining cells is less than the negative of the bypass diode trigger voltage, `-Vbypass`, then the substring is bypassed, and the voltage doesn't go any lower than `-Vbypass`.

mikofski:
What do you think is going on in this plot?
image
It looks like there are two levels of shading?

kanderso-nrel:
> What do you think is going on in this plot? ...  It looks like there are two levels of shading?

Indeed -- that module has 12 cells per column and is 10% shaded, so the bottom row of cells is fully shaded and the second row is only partially shaded.  Probably worth an explanation in the text.

PS thanks for these wonderful and extensive comments!  I will go through them tonight and make updates.

docs/examples/plot_partial_module_shading_simple.py
```diff
@@ -0,0 +1,273 @@
...
+    The shade is assumed to be coming up from the bottom of the module when in
+    portrait orientation, so it affects all substrings equally.
+    Substrings are assumed to be "down and back", so the number of cells per
+    string is divided between two columns of cells.
+    """
+    # find the number of cells per column that are in full shadow
+    nrow = cells_per_string//2
+    nrow_full_shade = int(shaded_fraction * nrow)
+    # find the fraction of shade in the border row
+    partial_shade_fraction = 1 - (shaded_fraction * nrow - nrow_full_shade)
```
mikofski, line 176: you might like the [`numpy.divmod`](https://numpy.org/doc/stable/reference/generated/numpy.divmod.html) returns the floor-like quotient and remainder, altho maybe not here?
```python
nrow_full_shade, partial_shade_fraction = np.divmod(shaded_fraction, 1/nrow)
partial_shade_fraction  = 1 - nrow*partial_shade_fraction 
```

docs/examples/plot_partial_module_shading_simple.py
kanderso-nrel: Nope, just a brain fart.  That's what comes from writing explanations late at night I suppose.  Good catch!

docs/examples/plot_partial_module_shading_simple.py
kanderso-nrel: Sorry, should have been more clear -- I'm on board with breakdown voltage being related to the junction rather than the cell as a whole, but I was referring to the "breakdown voltage" value that would be included in a cell parameter dictionary.  If it is inferred from measured IV curves and not adjusted for the series resistance voltage drop, it would be a cell voltage no?  Not that it matters here, just curious for myself.

docs/examples/plot_partial_module_shading_simple.py
kanderso-nrel: Could add the bypass diode here, but I added it where @cwhanse suggested so that it is easier to modify this example to use his forthcoming `add_series` function (I assume his function won't consider bypass diodes).

kanderso-nrel:
Thanks again for the reviews @mikofski and @cwhanse.  I very much apologize for this bypass diode fiasco -- I was focused on max power only and since the shading was uniform, I didn't bother including them.  Had I known that it would prompt such a strong reverse-bias review response, I'd have certainly included them from the start :wink: 

I think I've addressed most of the points, but I'm uncertain about these two:
- thermal voltage definition:  this is beyond my depth.  Happy to do whatever someone else decides is a good idea.  I notice I define it at 25C and then run models at 40C anyway.
- `singlediode.bishop88_i_from_v` vs `singlediode.bishop88` -- is there a technical reason to prefer one to the other?  Or a pedagogical reason?

I'll fix the lint issues on the next update.

cwhanse:
> thermal voltage definition:  this is beyond my depth.  Happy to do whatever someone else decides is a good idea.  I notice I define it at 25C and then run models at 40C anyway.

I think it's a good idea to state the cell temperature you assume and then calculate the thermal voltage rather than assume a thermal voltage.

> singlediode.bishop88_i_from_v vs singlediode.bishop88 -- is there a technical reason to prefer one to the other?  Or a pedagogical reason?

`singlediode.bishop88_i_from_v` let's you specify the voltages in the IV curve. `singlediode.bishop88` asks you to provide `diode_voltage` as input, from which it calculates voltages for the IV curve. If trying to compare a model to measurement, `bishop88_i_from_v` is the better choice. Niether one is guaranteed to have a point at maximum power. Hence `bishop88_mpp`.

docs/examples/plot_partial_module_shading_simple.py
cwhanse: > If it is inferred from measured IV curves 

I believe this is the case. 

> and not adjusted for the series resistance voltage drop, it would be a cell voltage no? 

That probably depends on how the value was extracted from the data. If the value was extracted by fitting the single diode equation including the reverse bias term, then the parameter should be viewed as a junction voltage.

kanderso-nrel:
@cwhanse I tried out switching to `singlediode.bishop88_i_from_v` for this example.  It did work, but dramatically increased the runtime -- it's a couple of orders of magnitude slower than `bishop88` and it really adds up from the repeated calls to build the heatmap matrix.  I think it would be an annoyance for documentation builds.  Ok to leave it as `bishop88`?  Alternatively we could remove the heatmap section, or just do a single row of the matrix instead.

mikofski:
>couple of orders of magnitude slower than `bishop88`

`bishop88_i_from_v` uses Brent or Newton to reverse each current in the sequence from given voltage implicitly, which would mean several calculations per point versus `bishop88` which explicitly calculate both current and voltage using vector math from the given array of dude voltages.

Newton will be a little faster than Brent because it is vectorised (operates on arrays), whereas Brent loops over the sequence.

There is an unspoken goal to replace the Brent with a Cythonized version using [`scipy.optimize.cython_optimize`](https://docs.scipy.org/doc/scipy/reference/optimize.cython_optimize.html) which should improve the routine performance of Brent, but I never did it. There are some examples [here in this gist Jupyter notebooks](https://gist.github.com/mikofski/ac30065073d0d32d6ea3569f6e24e5ec)

cwhanse:
> Ok to leave it as bishop88?  

Yes. Too bad that the runtime hit is that large. I do think that specifying cell rather than diode voltage is more familiar to most users, which is why I suggested using `bishop88_i_from_v` for this example.

mikofski:
>Yes. Too bad that the runtime hit is that large.

I will try to make a PR for the cythonized version of Brent soonish. It should greatly improve the performance, benchmarks were between 10-100x faster than [`numpy.vectorize`](https://numpy.org/doc/stable/reference/generated/numpy.vectorize.html) which is really no better than a for-loop, whereas [`scipy.optimize.cython_optimize`](https://docs.scipy.org/doc/scipy/reference/optimize.cython_optimize.html#module-scipy.optimize.cython_optimize) with `nogil` can be compiled to use [`prange`](https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html).

kanderso-nrel:
@mikofski FYI I was actually using the newton method because brentq doesn't work with reverse bias.  Per the [bishop88_i_from_v docstring](https://pvlib-python.readthedocs.io/en/latest/generated/pvlib.singlediode.bishop88_i_from_v.html): 
> method (str, default 'newton') – Either 'newton' or 'brentq'. ‘’method’’ must be 'newton' if breakdown_factor is not 0.

I'm not against speeding up brentq, just wanted to point out that I don't think it would help this example.

PR #982, opened 2020-06-13
Add instructions for building the docs

kanderso-nrel:
- ~Closes #xxxx~
 - ~Tests added~
 - ~New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.~

### DIFF
--- a/docs/sphinx/source/contributing.rst
+++ b/docs/sphinx/source/contributing.rst
@@ -257,13 +257,16 @@ Building the documentation locally is useful for testing out changes to the
 documentation's source code without having to repeatedly update a PR and have
 Read the Docs build it for you.  Building the docs locally requires installing
 pvlib python as an editable library (see :ref:`installation` for instructions).
-First, install the ``optional`` and ``doc`` dependencies specified in the
+First, install the ``doc`` dependencies specified in the
 ``EXTRAS_REQUIRE`` section of
 `setup.py <https://github.com/pvlib/pvlib-python/blob/master/setup.py>`_.
 An easy way to do this is with::
 
-    pip install pvlib[optional]
     pip install pvlib[doc]
+
+Note: Anaconda users may have trouble using the above command to update an
+older version of docutils. If that happens, you can update it with ``conda``
+(e.g. ``conda install docutils=0.15.2``) and run the above command again.
 
 Once the ``doc`` dependencies are installed, navigate to ``/docs/sphinx`` and
 execute::
--- a/docs/sphinx/source/contributing.rst
+++ b/docs/sphinx/source/contributing.rst
@@ -257,9 +257,10 @@ Building the documentation locally is useful for testing out changes to the
 documentation's source code without having to repeatedly update a PR and have
 Read the Docs build it for you.  Building the docs locally requires installing
 pvlib python as an editable library (see :ref:`installation` for instructions).
-First, install the ``doc`` dependencies specified in the
-`setup.py <https://github.com/pvlib/pvlib-python/blob/master/setup.py>`_
-file. An easy way to do this is with::
+First, install the ``doc`` dependencies specified in the ``EXTRAS_REQUIRE``
+section of
+`setup.py <https://github.com/pvlib/pvlib-python/blob/master/setup.py>`_.
+An easy way to do this is with::
 
     pip install pvlib[doc]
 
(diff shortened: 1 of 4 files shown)


docs/sphinx/source/contributing.rst
```diff
@@ -247,12 +247,37 @@ Parameters that specify a specific type require that specific input type.
...
 
-To build the docs locally, install the ``doc`` dependencies specified in the
+Building the documentation
+--------------------------
+
+Building the documentation locally is useful for testing out changes to the
+documentation's source code without having to repeatedly update a PR and have
+Read the Docs build it for you.  Building the docs locally requires installing
+pvlib python as an editable library (see :ref:`installation` for instructions).
+First, install the ``doc`` dependencies specified in the
```
cwhanse, line 260:
```suggestion
First, install the ``doc`` dependencies specified in the `EXTRAS_REQUIRE` section of
```

docs/sphinx/source/contributing.rst
```diff
@@ -247,12 +247,37 @@ Parameters that specify a specific type require that specific input type.
...
 `setup.py <https://github.com/pvlib/pvlib-python/blob/master/setup.py>`_
-file. See :ref:`installation` instructions for more information.
+file. An easy way to do this is with::
```
cwhanse, line 262:
```suggestion
An easy way to do this is with::
```

kanderso-nrel:
Two changes:
- Mention that both `optional` and `doc` need to be installed, previously only said `doc`
- `pillow` was an untracked requirement.  I'm not entirely sure why this never popped up before, but I guess all of us already had pillow installed and RTD automatically provides it by default.  Anyway it is added to the `doc` requirements list now.

wholmgren:
> Mention that both optional and doc need to be installed, previously only said doc

Off the top of my head, I believe this is only true for the 1. the clearsky page's tables requirement and 2. the forecast page's netcdf/siphon requirement. Am I missing anything? If it's just those then maybe add them to the doc requirement. If it's more then maybe make doc an extension of optional instead of instructing users to install both.

> pillow was an untracked requirement. I'm not entirely sure why this never popped up before, but I guess all of us already had pillow installed and RTD automatically provides it by default. Anyway it is added to the doc requirements list now.

What makes you think this is the case? I don't have pillow and can build the docs just fine.

kanderso-nrel:
Good idea about only adding what is needed to the doc requirement.  I'll try it out.

> What makes you think this is the case? I don't have pillow and can build the docs just fine.

In a fresh environment without pillow the build fails with this:
```
generating gallery...
generating gallery for auto_examples... [ 14%] plot_greensboro_kimber_soiling.py
Extension error:
Could not import pillow, which is required to rescale images (e.g., for thumbnails): No module named 'PIL'
```

You can do a build from scratch without pillow or PIL?  Sphinx-gallery's readme says it is a (possibly unnecessary) untracked dependency: https://github.com/sphinx-gallery/sphinx-gallery

wholmgren:
Indeed building the docs does now require pillow. I was confused about environment/git status. Sorry.

kanderso-nrel:
> Off the top of my head, I believe this is only true for the 1. the clearsky page's tables requirement and 2. the forecast page's netcdf/siphon requirement. Am I missing anything? If it's just those then maybe add them to the doc requirement. If it's more then maybe make doc an extension of optional instead of instructing users to install both.

The example gallery adds another layer since those actually run functions instead of just importing modules.  I think the only addition to your list is scipy (for plot_singlediode.py).  It's still less than half of the `optional` list, but that might change as the gallery grows over time.  Do you think it's better to keep a lean `doc` list or just include all of `optional` so that we don't need to keep track of it?

wholmgren:
I think many of us previously agreed (over many issues/prs) that it would be good to add scipy to the base requirements. So that would resolve the concern over the additional dependency. I can go either way on the lean vs. batteries-included doc list. 

In either case do we need to change something about the readthedocs configuration so that we can more reliably test the doc build?

kanderso-nrel:
> In either case do we need to change something about the readthedocs configuration so that we can more reliably test the doc build?

I suppose if `doc` is supposed to set up the complete docs environment, we should tell RTD to install `doc`, not `all`: https://github.com/pvlib/pvlib-python/blob/master/readthedocs.yml#L6

That way the RTD build would presumably fail if we add a new requirement but forget to add it to the `doc` list.  With that in place I think I would support keeping `doc` lean/minimal -- no need to bring in cython and numba just to build the docs.

kanderso-nrel:
Ok, looks like the RTD build worked with only the updated [doc] list.  I also turned off the setting to use preinstalled system packages so that (I think) it only uses the packages specified in setup.py.  I left scipy as an optional/doc requirement for now, since I assume making it base requirement will be a bit of a bigger job than just editing setup.py.  Anything else to do for this PR?

PR #984, opened 2020-06-14
Add diffuse IAM integration and gallery example

### LINKED ISSUE #983: Add Bill Marion's diffuse IAM integration method?
Is there interest in an implementation of Bill Marion's "Numerical method for angle-of-incidence correction factors for diffuse radiation incident photovoltaic modules" https://doi.org/10.1016/j.solener.2017.03.027 ?  

Since the method itself just integrates an arbitrary AOI correction model over the relevant portion of the sky, I'm imagining a function like this that you could pass the other IAM functions into:

```python
def marion_integrate(function, surface_tilt, region, N=180):
    """
    Integrate an incidence angle modifier (IAM) function over solid angle
    to determine a diffuse irradiance correction factor using Marion's method.

    Parameters
    ----------
    function : callable func(aoi)
        The IAM function to evaluate across solid angle.  Must take only one
        parameter, the angle of incidence in degrees.

    surface_tilt : numeric
        Surface tilt angles in decimal degrees.
        The tilt angle is defined as degrees from horizontal
        (e.g. surface facing up = 0, surface facing horizon = 90)
        surface_tilt must be in the range [0, 180]

    region : {'sky', 'horizon', 'ground'}
        The region to integrate over.  See [1]_ for details.

    N : int, default 180
        The number of increments in the zenith integration.

    Returns
    -------
    iam : numeric
        The incident angle modifier for the specified region.

    References
    ----------
    .. [1] B. Marion "Numerical method for angle-of-incidence correction
       factors for diffuse radiation incident photovoltaic modules",
       Solar Energy, Volume 147, Pages 344-348. 2017.
       DOI: 10.1016/j.solener.2017.03.027
    """
```

kanderso-nrel:
- [x] Closes #983 
 - [x] Tests added

I know I didn't wait for feedback on #983, I just wanted to implement it for fun and figured as long as the code exists, might as well open a PR.  This vectorized implementation is a couple orders of magnitude faster than a straightforward translation of the paper's pseudocode.  The gallery example recreates the figures in the paper.

API reference page: https://pvlib-python--984.org.readthedocs.build/en/984/generated/pvlib.iam.marion_integrate.html

Gallery example: https://pvlib-python--984.org.readthedocs.build/en/984/auto_examples/plot_diffuse_aoi_correction.html

### DIFF
--- a/docs/examples/plot_diffuse_aoi_correction.py
+++ b/docs/examples/plot_diffuse_aoi_correction.py
@@ -9,7 +9,7 @@ loss for diffuse irradiance.
 # %%
 # The fraction of light reflected from the front of a module depends on the
 # angle of incidence (AOI) of the light compared to the panel surface.  The
-# steeper the tilt, the larger the reflected fraction is.  The fraction of
+# greater the AOI, the larger the reflected fraction is.  The fraction of
 # transmitted light to incident light is called the incident angle modifier
 # (IAM).  Several models exist to calculate the IAM for a given incidence
 # angle (e.g. :py:func:`pvlib.iam.ashrae`, :py:func:`pvlib.iam.martin_ruiz`,
@@ -31,7 +31,7 @@ loss for diffuse irradiance.
 #     of Thermal Processes.  DOI: 10.1002/9781118671603
 
 
-from pvlib.iam import marion_integrate, physical
+from pvlib.iam import marion_diffuse, physical
 import numpy as np
 import matplotlib.pyplot as plt
 
@@ -40,94 +40,79 @@ import matplotlib.pyplot as plt
 # IAM Model
 # ---------
 #
-# The IAM model used to generate the figures in [1]_ focuses on the air-glass
-# interface.  It uses Snell's, Fresnel's, and Beer's laws to determine the
-# amount of light transmitted through the interface as a function of AOI.
+# The IAM model used to generate the figures in [1]_ uses Snell's, Fresnel's,
+# and Beer's laws to determine the fraction of light transmitted through the
+# air-glass interface as a function of AOI.
 # The function :py:func:`pvlib.iam.physical` implements this model, except it
 # also includes an exponential term to model attenuation in the glazing layer.
 # To be faithful to Marion's implementation, we will disable this extinction
 # term by setting the attenuation coefficient ``K`` parameter to zero.
 # For more details on this IAM model, see [2]_.
 #
-# Marion generated correction factor profiles for two cases:  a standard
-# uncoated glass with n=1.526 and a glass with anti-reflective (AR) coating
-# with n=1.3.  For convenience, we define a helper function for each case.
-# Comparing them across AOI recreates Figure 3 in [1]_:
-
-
-def calc_uncoated(aoi):
-    return physical(aoi, n=1.526, K=0)
-
-
-def calc_ar_coated(aoi):
-    return physical(aoi, n=1.3, K=0)
-
+# Marion generated diffuse irradiance modifiers for two cases:  a standard
+# uncoated glass with index of refraction n=1.526 and a glass with
+# anti-reflective (AR) coating with n=1.3.
+# Comparing the IAM model across AOI recreates Figure 3 in [1]_:
 
 aoi = np.arange(0, 91)
-cf_uncoated = calc_uncoated(aoi)
-cf_ar_coated = calc_ar_coated(aoi)
+iam_no_coating = physical(aoi, n=1.526, K=0)
+iam_ar_coating = physical(aoi, n=1.3, K=0)
 
-plt.plot(aoi, cf_ar_coated, c='b', label='$F_b$, AR coated, n=1.3')
-plt.plot(aoi, cf_uncoated, c='r', label='$F_b$, uncoated, n=1.526')
+plt.plot(aoi, iam_ar_coating, c='b', label='$F_b$, AR coated, n=1.3')
+plt.plot(aoi, iam_no_coating, c='r', label='$F_b$, uncoated, n=1.526')
 plt.xlabel(r'Angle-of-Incidence, AOI $(\degree)$')
-plt.ylabel('Correction Factor')
+plt.ylabel('Diffuse Incidence Angle Modifier')
 plt.legend()
 plt.ylim([0, 1.2])
 plt.grid()
 
 # %%
-# Diffuse sky irradiance (:math:`F_{sky}`)
-# -----------------------------------------
+# Diffuse sky, ground, and horizon irradiance
+# -------------------------------------------
 #
-# Now that we have an AOI model, we use :py:func:`pvlib.iam.marion_integrate`
-# to integrate it across solid angle and determine diffuse irradiance
-# correction factors.  Marion defines three types of diffuse irradiance:
-# sky, horizon, and ground-reflected.  The IAM correction factor is evaluated
+# Now that we have an AOI model, we use :py:func:`pvlib.iam.marion_diffuse`
+# to integrate it across solid angle and determine diffuse irradiance IAM.
+# Marion defines three types of diffuse irradiance:
+# sky, horizon, and ground-reflected.  The diffuse IAM value is evaluated
 # independently for each type.
-# First we recreate Figure 4 in [1]_, showing the dependence of the sky diffuse
-# correction factor on module tilt.
 
 tilts = np.arange(0, 91, 2.5)
 
-iam_uncoated = marion_integrate(calc_uncoated, tilts, 'sky', N=180)
-iam_ar_coated = marion_integrate(calc_ar_coated, tilts, 'sky', N=180)
+# marion_diffuse calculates all three IAM values (sky, horizon, ground)
+iam_no_coating = marion_diffuse('physical', tilts, n=1.526, K=0)
+iam_ar_coating = marion_diffuse('physical', tilts, n=1.3, K=0)
 
-plt.plot(tilts, iam_ar_coated, c='b', marker='^',
+# %%
+# First we recreate Figure 4 in [1]_, showing the dependence of the sky diffuse
+# incidence angle modifier on module tilt.
+
+plt.plot(tilts, iam_ar_coating['sky'], c='b', marker='^',
          label='$F_{sky}$, AR coated, n=1.3')
-plt.plot(tilts, iam_uncoated, c='r', marker='x',
+plt.plot(tilts, iam_no_coating['sky'], c='r', marker='x',
          label='$F_{sky}$, uncoated, n=1.526')
 plt.ylim([0.9, 1.0])
 plt.xlabel(r'PV Module Tilt, $\beta (\degree)$')
-plt.ylabel('Correction Factor')
+plt.ylabel('Diffuse Incidence Angle Modifier')
 plt.grid()
 plt.legend()
 plt.show()
 
-
 # %%
-# Diffuse horizon and ground irradiance (:math:`F_{hor}, F_{grd}`)
-# -----------------------------------------------------------------
-#
-# Now we recreate Figure 5 in [1]_, showing the dependence of the correction
-# factors for horizon and ground diffuse irradiance on module tilt.  Note that
-# we use 1800 points instead of 180 for the horizon case to match [1]_.
+# Now we recreate Figure 5 in [1]_, showing the dependence of the diffuse iam
+# values for horizon and ground diffuse irradiance on module tilt.  Note that
+# :py:func:`pvlib.iam.marion_diffuse` defaults to using 1800 points for the
+# horizon case (instead of 180 like the others) to match [1]_.
 
-iam_uncoated_grd = marion_integrate(calc_uncoated, tilts, 'ground', N=180)
-iam_ar_coated_grd = marion_integrate(calc_ar_coated, tilts, 'ground', N=180)
-
-iam_uncoated_hor = marion_integrate(calc_uncoated, tilts, 'horizon', N=1800)
-iam_ar_coated_hor = marion_integrate(calc_ar_coated, tilts, 'horizon', N=1800)
-
-plt.plot(tilts, iam_ar_coated_hor, c='b', marker='^',
+plt.plot(tilts, iam_ar_coating['horizon'], c='b', marker='^',
          label='$F_{hor}$, AR coated, n=1.3')
-plt.plot(tilts, iam_uncoated_hor, c='r', marker='x',
+plt.plot(tilts, iam_no_coating['horizon'], c='r', marker='x',
          label='$F_{hor}$, uncoated, n=1.526')
-plt.plot(tilts, iam_ar_coated_grd, c='b', marker='s',
+plt.plot(tilts, iam_ar_coating['ground'], c='b', marker='s',
          label='$F_{grd}$, AR coated, n=1.3')
-plt.plot(tilts, iam_uncoated_grd, c='r', marker='+',
+plt.plot(tilts, iam_no_coating['ground'], c='r', marker='+',
          label='$F_{grd}$, uncoated, n=1.526')
 plt.xlabel(r'PV Module Tilt, $\beta (\degree)$')
-plt.ylabel('Correction Factor')
+plt.ylabel('Diffuse Incidence Angle Modifier')
 plt.grid()
 plt.legend()
 plt.show()
--- a/docs/examples/plot_diffuse_aoi_correction.py
+++ b/docs/examples/plot_diffuse_aoi_correction.py
@@ -67,8 +67,8 @@ plt.ylim([0, 1.2])
 plt.grid()
 
 # %%
-# Diffuse sky, ground, and horizon irradiance
-# -------------------------------------------
+# Diffuse sky, ground, and horizon IAM
+# ------------------------------------
 #
 # Now that we have an AOI model, we use :py:func:`pvlib.iam.marion_diffuse`
 # to integrate it across solid angle and determine diffuse irradiance IAM.
--- a/docs/examples/plot_diffuse_aoi_correction.py
+++ b/docs/examples/plot_diffuse_aoi_correction.py
@@ -9,9 +9,10 @@ loss for diffuse irradiance.
 # %%
 # The fraction of light reflected from the front of a module depends on the
 # angle of incidence (AOI) of the light compared to the panel surface.  The
-# greater the AOI, the larger the reflected fraction is.  The fraction of
-# transmitted light to incident light is called the incident angle modifier
-# (IAM).  Several models exist to calculate the IAM for a given incidence
+# greater the AOI, the larger the reflected fraction is.  The incident angle
+# modifier (IAM) is defined as the ratio of light transmitted at the given
+# AOI to transmitted light at normal incidence.
+# Several models exist to calculate the IAM for a given incidence
 # angle (e.g. :py:func:`pvlib.iam.ashrae`, :py:func:`pvlib.iam.martin_ruiz`,
 # :py:func:`pvlib.iam.sapm`, :py:func:`pvlib.iam.physical`).
 # However, evaluating the IAM for diffuse light is
--- a/docs/examples/plot_diffuse_aoi_correction.py
+++ b/docs/examples/plot_diffuse_aoi_correction.py
@@ -1,0 +1,133 @@
+"""
+Diffuse IAM Calculation
+=======================
+
+Integrating an IAM model across angles to determine the overall reflection
+loss for diffuse irradiance.
+"""
+
+# %%
+# The fraction of light reflected from the front of a module depends on the
+# angle of incidence (AOI) of the light compared to the panel surface.  The
+# steeper the tilt, the larger the reflected fraction is.  The fraction of
+# transmitted light to incident light is called the incident angle modifier
+# (IAM).  Several models exist to calculate the IAM for a given incidence
+# angle (e.g. :py:func:`pvlib.iam.ashrae`, :py:func:`pvlib.iam.martin_ruiz`,
+# :py:func:`pvlib.iam.sapm`, :py:func:`pvlib.iam.physical`).
+# However, evaluating the IAM for diffuse light is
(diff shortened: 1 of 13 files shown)


docs/examples/plot_diffuse_aoi_correction.py
```diff
@@ -0,0 +1,133 @@
...
+# The IAM model used to generate the figures in [1]_ focuses on the air-glass
+# interface.  It uses Snell's, Fresnel's, and Beer's laws to determine the
+# amount of light transmitted through the interface as a function of AOI.
+# The function :py:func:`pvlib.iam.physical` implements this model, except it
+# also includes an exponential term to model attenuation in the glazing layer.
+# To be faithful to Marion's implementation, we will disable this extinction
+# term by setting the attenuation coefficient ``K`` parameter to zero.
+# For more details on this IAM model, see [2]_.
+#
+# Marion generated correction factor profiles for two cases:  a standard
```
cwhanse, line 52:
```suggestion
# Marion generated diffuse irradiance modifiers for two cases:  a standard
```

pvlib/iam.py
```diff
@@ -527,3 +527,120 @@ def sapm(aoi, module, upper=None):
...
+        # convert Series to np array for broadcasting later
+        beta = beta.values
+    ai = np.pi/N  # angular increment
+
+    phi_range = np.linspace(0, np.pi, N, endpoint=False)
+    psi_range = np.linspace(0, 2*np.pi, 2*N, endpoint=False)
+
+    # the pseudocode in [1] do these checks at the end, but it's
+    # faster to do this criteria check up front instead of later.
+    if region == 'sky':
```
cwhanse, line 596: 'sky' region contains 'horizon' region?

pvlib/iam.py
```diff
@@ -527,3 +527,120 @@ def sapm(aoi, module, upper=None):
         iam = pd.Series(iam, aoi.index)
 
     return iam
+
+
+def marion_integrate(function, surface_tilt, region, N=180):
```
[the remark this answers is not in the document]
kanderso-nrel: @cwhanse I originally used a keyword to switch between the available IAM functions like you describe, but I switched to passing in a function so that people could use their own model if they wanted.  It seemed useful when I was writing the example, but in hindsight I think I agree it should just be consistent with the rest of pvlib.  

> I would also prefer that the function returned separate IAM modifiers for sky and ground diffuse irradiance.

To be clear, does this mean removing the `region` parameter and calculating all three IAM values instead of just one?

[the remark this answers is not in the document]
cwhanse: > To be clear, does this mean removing the region parameter and calculating all three IAM values instead of just one?

Yes, that's what I meant to say.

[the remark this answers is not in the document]
cwhanse: On second thought the `marion_integrate` function has merit, maybe add `marion_diffuse` as a new function that accepts a kwarg `model` and uses `marion_integrate` to do the calculation. I think the `model` kwarg` would be more familiar to many users than passing a function handle as an argument.

pvlib/iam.py
kanderso-nrel: According to Table 1 in the reference, it does indeed.  Any to-do item in this PR for that?

cwhanse: My inclination is to expose the zenith angle limits that define 'sky' and 'horizon' and set defaults to Marion's values. But I don't want to complicate the interface. At a minimum, the docstring could describe what constitutes 'sky' and 'horizon'

kanderso-nrel: I'm not seeing a clean way of exposing them either.  Added zenith bounds to the docstrings for now at least.

cwhanse:
@wholmgren @mikofski any comments on this PR? I think it is ready to merge.

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+    array([0.96225034, 0.9653219 ])
+    """
+
+    if N is None:
+        if region in ['sky', 'ground']:
+            N = 180
+        elif region == 'horizon':
+            N = 1800
+        else:
+            raise ValueError('Invalid region: {}'.format(region))
```
wholmgren, line 685: line not tested, see codecov/diff

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+    model : str
+        The IAM function to evaluate across solid angle.  Must be one of
+        `'ashrae', 'physical', 'martin_ruiz', 'sapm'`.
+
+    surface_tilt : numeric
+        Surface tilt angles in decimal degrees.
+        The tilt angle is defined as degrees from horizontal
+        (e.g. surface facing up = 0, surface facing horizon = 90).
+
+    regions : list, default ['sky', 'horizon', 'ground']
```
wholmgren, line 549: Is the rationale for `None` to avoid mutable defaults? If so, how about an immutable tuple `('sky', 'horizon', 'ground')`? That would be more clear to me.

wholmgren: I'd also consider removing the kwarg entirely so the function always returns all three components. Users that want a single component can use the lower level function.

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+    models = {
+        'physical': physical,
+        'ashrae': ashrae,
+        'sapm': sapm,
+        'martin_ruiz': martin_ruiz,
+    }
+
+    try:
+        iam_model = models[model]
+        iam_function = functools.partial(iam_model, **kwargs)
```
wholmgren, line 604: move this line out of the try block

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+    return iam
+
+
+def marion_integrate(function, surface_tilt, region, N=None):
+    """
+    Integrate an incidence angle modifier (IAM) function over solid angle
+    to determine a diffuse irradiance correction factor using Marion's method.
+
+    Lower-level function to actually perform the IAM integration for the
+    specified solid angle region.
```
wholmgren, line 621:
```suggestion
    This lower-level function actually performs the IAM integration for the
    specified solid angle region.
```

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+
+def marion_diffuse(model, surface_tilt, regions=None, **kwargs):
+    """
+    Determine diffuse irradiance incidence angle modifiers using Marion's
+    method of integrating over solid angle.
+
+    Parameters
+    ----------
+    model : str
+        The IAM function to evaluate across solid angle.  Must be one of
```
wholmgren, line 541:
```suggestion
        The IAM function to evaluate across solid angle. Must be one of
```

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+
+    regions : list, default ['sky', 'horizon', 'ground']
+        The regions to integrate over.  Options are:
+
+            * 'sky': radiation from the sky dome (zenith <= 90)
+            * 'horizon': radiation from the region of the sky near the horizon
+              (89.5 <= zenith <= 90)
+            * 'ground': radiation reflected from the ground (zenith >= 90)
+
+        See [1]_ for a detailed description of each class.  If not specified,
```
wholmgren, line 557:
```suggestion
        See [1]_ for a detailed description of each class. If not specified,
```

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+            * 'ground': radiation reflected from the ground (zenith >= 90)
+
+        See [1]_ for a detailed description of each class.
+
+    N : int, optional
+        The number of increments in the zenith integration.
+        If not specified, N will follow the values used in [1]_:
+
+            * 'sky' or 'ground': N=180
+            * 'horizon': N=1800
```
wholmgren, line 650:
```suggestion
            * 'sky' or 'ground': N = 180
            * 'horizon': N = 1800
```

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+    Integrate an incidence angle modifier (IAM) function over solid angle
+    to determine a diffuse irradiance correction factor using Marion's method.
+
+    Lower-level function to actually perform the IAM integration for the
+    specified solid angle region.
+
+    Parameters
+    ----------
+    function : callable func(aoi)
+        The IAM function to evaluate across solid angle.  The function must
```
wholmgren, line 626:
```suggestion
        The IAM function to evaluate across solid angle. The function must
```

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+    N : int, optional
+        The number of increments in the zenith integration.
+        If not specified, N will follow the values used in [1]_:
+
+            * 'sky' or 'ground': N=180
+            * 'horizon': N=1800
+
+    Returns
+    -------
+    Fd : numeric
```
wholmgren, line 654:
```suggestion
    iam : numeric
```

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+        raise ValueError('model must be one of: ' + str(list(models.keys())))
+
+    iam = {}
+    for region in regions:
+        iam[region] = marion_integrate(iam_function, surface_tilt, region)
+
+    return iam
+
+
+def marion_integrate(function, surface_tilt, region, N=None):
```
wholmgren, line 615:
```suggestion
def marion_integrate(function, surface_tilt, region, num=None):
```

wholmgren: `num` is consistent with `np.linspace`. I'd also be good with something like `intpnts`.

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+    """
+    Integrate an incidence angle modifier (IAM) function over solid angle
+    to determine a diffuse irradiance correction factor using Marion's method.
+
+    Lower-level function to actually perform the IAM integration for the
+    specified solid angle region.
+
+    Parameters
+    ----------
+    function : callable func(aoi)
```
wholmgren, line 625:
```suggestion
    function : callable(aoi)
```

wholmgren: I've seen scipy documentation like my suggestion, or just `function`, or just `callable`, but I've not seen `callable func`.

scipy uses `f`, `func`, and `function` for the argument name. I like `function` (no surprise).

pvlib/iam.py
```diff
@@ -527,3 +528,221 @@ def sapm(aoi, module, upper=None):
...
+        be vectorized and take only one parameter, the angle of incidence in
+        degrees.
+
+    surface_tilt : numeric
+        Surface tilt angles in decimal degrees.
+        The tilt angle is defined as degrees from horizontal
+        (e.g. surface facing up = 0, surface facing horizon = 90).
+
+    region : {'sky', 'horizon', 'ground'}
+        The region to integrate over.  Must be one of:
```
wholmgren, line 636:
```suggestion
        The region to integrate over. Must be one of:
```

pvlib/iam.py
wholmgren: ok to use fd in the code

kanderso-nrel:
Thanks @wholmgren for this detailed review.  I'm happy to make the suggested changes, although I'm not sure about this one:

> I'd be in favor of using a scipy integration method over a strict implementation of the reference. Is that feasible?

Here's a quick comparison for the 'sky' case:

```python
from pvlib import tools, irradiance, iam
from scipy import integrate
from functools import partial

def scipy_integrate(function, surface_tilt):

    def integrand(phi, theta, beta, function):
        aoi = irradiance.aoi(beta, 0, phi, theta)
        if aoi >= 90:
            return 0
        area_element = tools.sind(phi)
        return function(aoi) * tools.cosd(aoi) * area_element

    numerator = integrate.dblquad(integrand,
                                  0, 360,  # azimuth
                                  lambda theta: 0, lambda theta: 90,  # zenith
                                  args=(surface_tilt, function))
    denominator = integrate.dblquad(integrand,
                                    0, 360,  # azimuth
                                    lambda theta: 0, lambda theta: 90,  # zenith
                                    args=(surface_tilt, lambda *args: 1))
    Fd = numerator[0] / denominator[0]
    return Fd
```

```python
iam_model = partial(iam.physical, n=1.3)

%time scipy_iam = scipy_integrate(iam_model, 20)
Wall time: 1.24 s

%time marion_iam = iam.marion_integrate(iam_model, 20, 'sky')
Wall time: 6.98 ms

print(scipy_iam, marion_iam)
0.9622497940399887 0.9622503413733542
```

So they get the same answer, but this simple scipy version is much slower, presumably for lack of vectorization.  I'm not very familiar with scipy so maybe it's possible to nest two `integrate.quad_vec` calls.  I'm also not sure how to scale it across an array of tilts other than a list comprehension or `np.vectorize`.  What do you think?

wholmgren:
Thanks for exploring this - yikes. Let's stick with the existing implementation.

mikofski:
Sorry for being late to the game. Would [Numpy `trapz`](https://numpy.org/doc/stable/reference/generated/numpy.trapz.html) work? It's vectorised, and I believe it's the preferred way to integrate.

mikofski:
The other, probably slower option would be [SciPy `integrate.quad`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.quad.html) but I've never used it, and I guess I wouldn't water my time, just use `trapz` I promise you won't regret it

kanderso-nrel:
> Sorry for being late to the game. Would [Numpy `trapz`](https://numpy.org/doc/stable/reference/generated/numpy.trapz.html) work? It's vectorised, and I believe it's the preferred way to integrate.

Is the advantage that we would use trapezoidal integration instead of rectangular integration?  I think it would only really be able to replace the two summation lines, so it wouldn't really simplify the code.  OK with me if the change is for accuracy, not simplicity.  It'll be a little weird to handle the dependency of the solid angle element on zenith; I think the best way is to just lump it into the values to be integrated like it is now.

mikofski:
Your call, I haven't had a chance to look at the code yet. But in the 2d infinite sheds PR , which is just a bunch more of Bill Marion's formulas I use `trapz`, I've just always used it, going back to the MATLAB days. I do think trapezoid rule is preferable to rectangle, not only is it more accurate but it avoids the whole left side or right side problem

kanderso-nrel:
Trying out `np.trapz` on the test cases, the difference from the current implementation is on the order of 10^-4 at worst, sometimes 10^-5 or -6.  Although it's not a perfect comparison because I think the integration bounds would need updating -- with an array of length N, `np.sum` will sum N intervals while `np.trapz` integrates only (N-1).  Anyway with so small a difference for the test cases, I think I'd prefer `np.sum` to keep to the reference because the summation is a little nuanced.  But I won't stand in the way if we think summing is for chumps and the bikeshed should be shaped like a trapezoid :)

docs/examples/plot_diffuse_aoi_correction.py
```diff
@@ -0,0 +1,118 @@
...
+Integrating an IAM model across angles to determine the overall reflection
+loss for diffuse irradiance.
+"""
+
+# %%
+# The fraction of light reflected from the front of a module depends on the
+# angle of incidence (AOI) of the light compared to the panel surface.  The
+# greater the AOI, the larger the reflected fraction is.  The fraction of
+# transmitted light to incident light is called the incident angle modifier
+# (IAM).  Several models exist to calculate the IAM for a given incidence
```
mikofski, line 14: >The fraction of transmitted light to incident light is called the incident angle modifier (IAM).

Is this what Marion says? Because strictly speaking, I don't think this is true. At normal incidence, there might be 4% internal reflection, meaning transmitted light is only 96%. You can derive this directly from [Fresnel's equation at normal incidence](https://en.wikipedia.org/wiki/Fresnel_equations#Normal_incidence):

```
R = ((n1 - n2)/(n1 + n2))**2
```

So if you look at the [Sandia Array Performance model](https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2004/043535.pdf) the IAM or f2 is defined as the following ratio:

```
f2_aoi = Isc(aoi=aoi) / Isc(aoi=0)
```

kanderso-nrel: You're right, sloppy and incorrect wording on my part.  Marion's definition matches yours:

> The AOI correction factor is the in-plane irradiance reaching the PV cells divided by the in-plane irradiance reaching the PV cells when the AOI is normal to the PV module surface.

I'll update with your suggestion, although I'll change `fraction` to `ratio`.  Not sure why I used `fraction` originally, it reads a bit awkwardly.

mikofski:
I'll try to finish reviewing this today, but if you all are already happy, don't let me hold you up.

pvlib/iam.py
```diff
@@ -527,3 +528,214 @@ def sapm(aoi, module, upper=None):
...
+    psi_1 = angles[:, [1]]
+    phi_2 = phi_1 + ai
+    # psi_2 = psi_1 + ai  # not needed
+    phi_avg = phi_1 + 0.5*ai
+    psi_avg = psi_1 + 0.5*ai
+    term_1 = np.cos(beta) * np.cos(phi_avg)
+    # we are free to choose any gamma between zero and 2pi, so choose zero
+    # and omit it from the cos(psi_avg) term
+    term_2 = np.sin(beta) * np.sin(phi_avg) * np.cos(psi_avg)
+    aoi = np.arccos(term_1 + term_2)
```
mikofski, line 717: Can you save `term1 + term2` so you don't have to back-calculated it on line 720 below?
_EG_:
```python
cosaoi = term1 + term2
aoi = arcos(cosaoi)
```
then below on L720:
```python
cosaoi_dAs = cosaoi * dAs
```

Also, Bill's paper has a gamma term for the azimuth of the panels, I feel like you mentioned that up above, but I'm forgetting, why isn't it included here?
```python
cosaoi = term1 + term2 - gamma_pvmodule_azimuth
```
sorry, I didn't read through the previous comments, so forgive me if redundant

pvlib/iam.py
```diff
@@ -527,3 +528,214 @@ def sapm(aoi, module, upper=None):
...
+    # simplify Eq 8, (psi_2 - psi_1) is always ai
+    dAs = ai * (np.cos(phi_1) - np.cos(phi_2))
+    cosaoi_dAs = np.cos(aoi) * dAs
+    # apply the final AOI check, zeroing out non-passing points
+    mask = aoi < np.pi/2
+    cosaoi_dAs = np.where(mask, cosaoi_dAs, 0)
+    numerator = np.sum(function(np.degrees(aoi)) * cosaoi_dAs, axis=0)
```
mikofski, line 724: maybe just for posterity put a comment here explaining why we're purposefully choosing not to use trapezoid rule, like:
```python
# use Riemann sum instead of trapezoid rule to be faithful to Marion's paper,
# number of intervals is large enough that errors are between 1e-4 to 1e-5
```

In all honesty, I lied when I said, "your call." because I think the reason Bill used Reimann instead of trapezoid is because he didn't have access to MATLAB or R? You could probably get away with less intervals and possibly improve the speed by 1 or 2 nanoseconds, AND have your bikeshed painted white! :wink: 😛

pvlib/iam.py
kanderso-nrel: +1 for saving `cosaoi`, I'll make that change.

The panel azimuth term is omitted because the system is rotationally invariant -- it's all about isotropic diffuse irradiance, so the panel azimuth can be chosen freely, and I chose to make it zero and leave it out as a tiny speed optimization.  Bill sets gamma to pi in the Section 3 pseudocode but mentions that gamma can be set to any value between 0 and 2pi.  I'll expand that explanatory comment to hopefully make that more clear.

pvlib/iam.py
kanderso-nrel: After thinking about this some more, I think maybe Bill pulled one over on us (or we pulled one over on ourselves) by not actually doing Riemann sums -- he calculates the AOI for the midpoint of each interval (phi_avg and psi_avg), so it's actually a midpoint integration which (if I remember correctly) is even better for smooth functions than the trapezoid rule!  Would you mind taking a second look at the paper when you get a chance and see if you agree?  

Also, I realized my 1e-4/1e-5 differences were from unfair comparisons.  I had just replaced the `np.sum` calls with `np.trapz`, but those numpy arrays are the 2-D integration surface pivoted into a 1-D array, so it doesn't make sense to treat the values as a single sequence of values to integrate.  To do a real comparison we would probably have to redo the method to use the typical meshgrid approach and stick with 2D arrays.  Maybe that's irrelevant now though.

mikofski: fair enough - I agree, midpoint may be better, so no comment required, thanks!

mikofski:
@kanderso-nrel is this ready to go? There are some "conda-linux" tests that are still not passing. Have you looked into them? Do you know if Will and Cliff's reviews are complete? Shall I merge this if everyone is okay with it?

kanderso-nrel:
The test failures are unrelated to this PR:
- iotools/test_psm3.py::test_get_psm3_tmy
  + Seems to be an issue on the API side, see https://github.com/NREL/developer.nrel.gov/issues/159
- iotools/test_tmy.py::test_read_tmy3_remote
  + Hadn't noticed this one before.  Looks like the RREDC website has changed and I'm not sure where to find those files anymore.  Possibly relevant: https://nsrdb.nrel.gov/data-sets/archives.html

I hesitate to speak for Will and Cliff, but I'd guess the changes since their reviews would be fairly uncontroversial.  OK with me to merge.

mikofski:
Okay @wholmgren, @cwhanse, @kanderso-nrel - I will merge this tomorrow morning unless I hear otherwise. Is squash best?

wholmgren:
Yes always squash

mikofski:
Thanks @kanderso-nrel and everyone for reviewing. This is merged now in master.

PR #987, opened 2020-06-17
Correction on the application of losses in pvwatts_losses()

tylunel:
 - [ ] Tests added

Until now, the consideration of losses in a system is made with the pvwatts model which apply a coefficient to all electrical outputs (''i_sc', 'v_oc', 'i_x', 'i_xx', 'v_mp', 'i_mp', 'p_mp') in pvwatts_losses():
`self.dc *= self.losses`
It results in inconsistent figures because after this line, the following equality is not respected anymore:
>  v_mp * i_mp = p_mp

As I think losses can be considered as an increase in R_s (Not 100% sure, we can discuss it), I propose to correct it with:
self.dc[['v_mp', 'p_mp']] *= self.losses

### DIFF
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -731,7 +731,7 @@ class ModelChain(object):
 
     def pvwatts_losses(self):
         self.losses = (100 - self.system.pvwatts_losses()) / 100.
-        self.dc *= self.losses
+        self.dc[['v_mp', 'p_mp']] *= self.losses
         return self
 
     def no_extra_losses(self):


kanderso-nrel:
I agree that applying pvwatts_losses to the output of more detailed DC models ends up double-counting the loss like you describe.  Unfortunately the proposed solution `self.dc[['v_mp', 'p_mp']] *= self.losses` isn't a complete fix because it assumes that `self.dc` is a dataframe, which isn't the case when `dc_model='pvwatts'` (see the failing tests).

That small complication aside, I'm skeptical that dumping the entire loss onto v_mp is the right move since some of the PVWatts loss components affect current.  I think trying to apply the losses at the current and voltage level would require us to significantly extend PVWatts beyond its intended use.  However, the only alternative I can think of is to just forbid using pvwatts_losses unless `dc_model = ac_model = 'pvwatts'`.

I suppose the ideal solution would be to have a more detailed loss mechanism that applies losses at the conceptually correct steps in the chain instead of in one lump sum.  That's not going to be a one-line change though.

PR #993, opened 2020-06-20
Add coverage exclusion comments to certain test suite functions

### LINKED ISSUE #992: codecov/project/tests PR check always fails
I don't really know what the `codecov/project/tests` check is trying to tell me.  Seems like it's always 0.00% out of the 95.00% target.  I see that the target is indeed 95% per: https://github.com/pvlib/pvlib-python/blob/master/.codecov.yml#L23

I'd like to get this test passing and see green checkmarks next to PRs again.  Aside from the 0.00% which I could forgive as a buggy github integration, what I really don't understand is why this check is failing when codecov reports 96.85% coverage for the tests folder: https://codecov.io/gh/pvlib/pvlib-python/tree/minigalleries/pvlib/tests

I can only assume that improving coverage would satisfy the test.  Seems like the coverage is getting dragged down by helper code (test data creation?) that isn't intended to get called in the test suite itself, like some functions here:
- https://codecov.io/gh/pvlib/pvlib-python/src/master/pvlib/tests/test_numerical_precision.py
- https://codecov.io/gh/pvlib/pvlib-python/src/master/pvlib/tests/iotools/test_ecmwf_macc.py

So if we wanted to get some quick coverage points back, we could add a `# pragma: no cover` comment ([docs](https://coverage.readthedocs.io/en/v4.5.x/excluding.html)) to those functions.  I'll open a PR shortly to see if this makes a difference.

In the past that 95% threshold has been lowered to accommodate changes in the test code, so that's another option.

kanderso-nrel:
- [x] Closes #992
 - ~Tests added~
 - ~Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).~
 - ~New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.~

See #992, opening this to see what codecov thinks about it


kanderso-nrel:
Okay, I won't pretend to understand why, but it seems like codecov didn't like the path regex for the `project/library` coverage calculation and that was somehow messing up the `project/tests` calculation.  If I understand the regex in question (`pvlib/(\w+/)?[^/]+\.py$`) correctly, it matches py files that are in pvlib or in a folder in pvlib, but not in a folder in a folder in pvlib.  I don't see why that complexity is necessary so I simplified it.  

FYI this is the test to see if a codecov.yml is valid: `curl --data-binary @codecov.yml https://codecov.io/validate`

wholmgren:
[Pandas](https://github.com/pandas-dev/pandas/blob/master/codecov.yml) and [XArray](https://github.com/pydata/xarray/blob/master/.codecov.yml) use very simple config files. Maybe we should do that too. [MetPy](https://github.com/Unidata/MetPy/blob/master/.codecov.yml) is a bit more complicated and was probably the primary thing I copied from. Indeed there is a similar [remove the regex commit](https://github.com/Unidata/MetPy/commit/5c0be987d24df00e19ef01d48f308f06e93f1e36#diff-1f7b4b85ed8525c5239f741431a72872) to metpy's file. 

The only thing I really care about is failing PRs when code isn't covered.

pvlib/tests/test_numerical_precision.py
```diff
@@ -55,7 +55,7 @@
     sy_exp = NotImplemented
 
 
-def generate_numerical_precision():
+def generate_numerical_precision():  # pragma: no cover
```
wholmgren, line 58: there's a `if __name__ == '__main__'` block at the bottom of this file too

kanderso-nrel:
> Pandas and XArray use very simple config files. Maybe we should do that too. 

I think I'd be in favor of keeping the three coverage metrics that are currently configured -- each of them seems useful to me.

PR #1000, opened 2020-07-08
Added irradiance_loss_pvsyst function to pvsystem

### LINKED ISSUE #988: Adding Loss Model Categories
Following [this](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/pvlib-python/O5Dhrz5mlRc/-zZez4UnBgAJ) conversation on the Google Group, it would be helpful to create loss models divided into different categories.  I think this could be done effectively with 4 categories:

1. Effective Irradiance
  A. All losses already accounted for in the transposition models
  B. Angle of Incidence losses
  C. Soiling
  D. Snow Coverage
  E. Shading -> this one gets a bit more nuanced if you take into account cell stringing and blends with an electrical loss
  F. Bifaciality -> Also a bit more nuanced as it incorporates module-specific parameters
2. Module Performance
  A. Degradation, including LID, LeTID, and annual degradation
  B. Module Quality, or the actual performance of a module installed compared to its rated performance
  C. Module Mismatch?  (This one might fit better in the DC losses bucket, but technically current mismatch within a string and voltage mismatch between parallel strings impact performance at a module level)
3.  DC Electrical Losses (after dc_model)
  A. Voltage drop due to cabling resistance (DC Ohmic Loss in PVSyst)
  B. Clipping when DC output at MPPT is higher than the inverter rating (although I think this is already well accounted for in pvlib)
4. AC Losses (after ac_model)
  A. Medium Voltage Transformer losses
  B. AC Cabling resistance losses
  C. High Voltage Transformer losses
  D. Transmission line losses
  E. Clipping when a plant is capable of producing more power than it is allowed to inject into its point of interconnection
  F. Any additional auxiliary losses should be included (like parasitic trackers or a control house)

Often the Module Performance and DC Electrical Loss categories are combined, but I think being more explicit can offer value, as those losses would be applied before the single diode model instead of after.

Are there any thoughts on this implementation?

btaute:
Added irradiance_loss_pvsyst function to pvsystem module and method to PVSystem class.

Partially addresses Issue #988.

I still need to incorporate tests and update the "what's new" file.  I'm hoping for some feedback on implementation before starting those.

Thank you for your review.

 - [x] Closes #988 
 - [x] Tests added


cwhanse:
The direction this PR goes is fine. The review will be easier if we can clean up the formatting issues first, LMK if you would like some help with that.

I'm fine with the calculations. I'll probably suggest some renaming of the parameters (drop `irradiance_` from each name), and some editing of the docstring.

btaute:
I updated the formatting, but I'm still running into some issues there that I'm not sure how to resolve.  Some insight on what's required for passing formatting would be greatly appreciated.

With respect to the "irradiance_" included in many variable names, I agree that's cumbersome and am happy to pull that.  The reason I included it was to differentiate from the electrical loss models that can also be associated with shading, soiling, and snow coverage.  If you believe that differentiation is obvious given the context of the function, then it's overly specific.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,74 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+    PVSyst methodology [1]_ of linearly reducing the effective irradiance on a
+    module.  This doesn't account for the electrical losses due to irregular
+    irradiance on different solar cell strings but rather reduces the
+    irradiance input to the single diode equation.
+
+    The losses are compounded using the following equation:
+
+    .. math::
+
+        L_{total}(\%) = 100 [ 1 - \Pi_i ( 1 - \frac{L_i}{100} ) ]
```
cwhanse, line 2569: Try `\Prod`? I'm puzzled by the failure of \%, but if it's not a simple fix, can remove that from the equation and define the quantities in the text.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,74 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    snow_loss = irradiance_snow_loss.reindex_like(effective_irradiance,
+                                                  method=fill_method)
+
+    soiling_loss = irradiance_soiling_loss.reindex_like(effective_irradiance,
+                                                        method=fill_method)
+
+    irradiance_losses = effective_irradiance * (1 - (1 - shading_loss)
+                                                  * (1 - snow_loss)
+                                                  * (1 - soiling_loss))
```
cwhanse, line 2619:
```suggestion
    irradiance_losses = effective_irradiance * (
        1 - (1 - shading_loss) * (1 - snow_loss) * (1 - soiling_loss))
```

cwhanse: If it still gives W503 we can silence that lint error

pvlib/pvsystem.py
kanderso-nrel: I think this is solved by adding a leading `r` to the docstring triple quotes so that backslashes aren't interpreted as escape sequences, i.e.:

```python
def irradiance_loss_pvsyst(effective_irradiance, irradiance_shading_loss,
                           irradiance_snow_loss, irradiance_soiling_loss):
    r"""
    Calculates plane of array irradiance losses using a model based on the
    PVSyst methodology [1]_ of linearly reducing the effective irradiance on a
    ...
    """
```

btaute:
I attempted to add tests.  I'm a little unfamiliar with this system, but I don't think all of the test failures happening are from my addition.  If they are, there is clearly something I'm missing and will require assistance.

I think it's at a place for detailed review when a @pvlib/maintainer has the time.

cwhanse:
The test warnings and failures are unrelated to this PR. I'll review the PR soon.

docs/sphinx/source/whatsnew/v0.8.0.rst
```diff
@@ -35,6 +39,8 @@ Testing
   applied to functions that require args or kwargs. (:pull:`973`)
 * Test added for :py:class:`pvlib.modelchain.ModelChain` to confirm ValueError when
   ``ac_model`` is an invalid string. (:pull:`886`)
+* Tests added for :py:func:`pvlib.pvsystem.irradiance_losses_pvsyst` and
+  :py:meth:`pvlib.pvsystem.PVSystem.irradiance_losses_pvsyst`
```
cwhanse, line 43: I'm not opposed to having these lines in whatsnew, just noting that we haven't explicitly called out adding tests for new functions - flagging it in case other maintainers have an opinion here.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
 
 
+def irradiance_loss_pvsyst(effective_irradiance, irradiance_shading_loss,
+                           irradiance_snow_loss, irradiance_soiling_loss):
+    r"""
+    Calculates plane of array irradiance losses using a model based on the
+    PVSyst methodology [1]_ of linearly reducing the effective irradiance on a
+    module.  This doesn't account for the electrical losses due to irregular
+    irradiance on different solar cell strings but rather reduces the
+    irradiance input to the single diode equation.
```
cwhanse, line 2563:
```suggestion
    Calculates irradiance losses using a model based on PVsyst.
    
    The PVsyst method [1]_ linearly reduces the effective irradiance to
    account for the effects of shading, snow and soiling. The separate
    loss factors are compounded to calculate a single loss factor.
```

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    The losses are compounded using the following equation:
```
cwhanse, line 2565:
```suggestion
    The separate losses are compounded using the following equation:
```

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    .. math::
+
+        L_{total}(\%) = 100 [ 1 - \Pi_i ( 1 - \frac{L_i}{100} ) ]
+
```
cwhanse, line 2570: :math:`L_{total}` is the total irradiance loss resulting from the effects of shading, snow and soiling.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+    index of the returned series will be the same as the effective_irradiance
+    index.  All other parameters will be resampled to match this index with
+    a "fill forward" method for filling holes.
+
+    Parameters
+    ----------
+    effective_irradiance : Series
+            The irradiance (W/m2) that is to be reduced.  Plane-of-array
+            irradiance that already accounts for Angle of Incidence and
+            Spectral losses.
```
cwhanse, line 2581: Question for my benefit: PVsyst applies the snow, shade and soiling losses after reflection losses?

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+            The fraction of the solar panel that is covered with snow.
+
+    irradiance_soiling_loss : Series
+            The fraction of the irradiance prevented from reaching the solar
+            cells due to soiling of the solar panel.
+
+    Returns
+    -------
+    losses: Series
+            Irradiance (W/m2) reduction due to compounded losses.
```
cwhanse, line 2596:
```suggestion
            Irradiance reduction due to compounded losses. [%]
```

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    irradiance_shading_loss : Series
+            The fraction of the solar panel that is shaded.
+
+    irradiance_snow_loss : Series
+            The fraction of the solar panel that is covered with snow.
+
+    irradiance_soiling_loss : Series
+            The fraction of the irradiance prevented from reaching the solar
+            cells due to soiling of the solar panel.
```
cwhanse, line 2591: These inputs are in %, from the equation. If that's correct, change "fraction" to "percent", or add [%] at the end of the description.

pvlib/tests/test_pvsystem.py
```diff
@@ -1347,6 +1347,38 @@ def test_PVSystem_pvwatts_ac_kwargs(mocker):
     assert out < pdc
 
 
+def make_irradiance_loss_pvsyst_test_series():
```
cwhanse, line 1350:
```suggestion
@pytest.fixture()
def irradiance_loss_pvsyst():
```

pvlib/tests/test_pvsystem.py
```diff
@@ -1347,6 +1347,38 @@ def test_PVSystem_pvwatts_ac_kwargs(mocker):
...
+                                                 periods=12, freq='MS'))
+    expected = pd.Series(162.1, index=final_index)
+
+    return {'effective_irradiance': effective_irradiance,
+            'shading': shading, 'snow': snow,
+            'soiling': soiling, 'expected': expected}
+
+
+def test_irradiance_loss_pvsyst():
+    params = make_irradiance_loss_pvsyst_test_series()
```
cwhanse, line 1366:
```suggestion
def test_irradiance_loss_pvsyst(irradiance_loss_pvsyst):
```

pvlib/tests/test_pvsystem.py
```diff
@@ -1347,6 +1347,38 @@ def test_PVSystem_pvwatts_ac_kwargs(mocker):
...
+    out = pvsystem.irradiance_loss_pvsyst(params['effective_irradiance'],
+                                          params['shading'], params['snow'],
+                                          params['soiling'])
+    assert_series_equal(params['expected'], out)
```
cwhanse, line 1370:
```suggestion
    out = pvsystem.irradiance_loss_pvsyst(
        irradiance_loss_pvsyst['effective_irradiance'], irradiance_loss_pvsyst['shading'],
        irradiance_loss_pvsyst['snow'], irradiance_loss_pvsyst['soiling'])
    assert_series_equal(irradiance_loss_pvsyst['expected'], out)
```

pvlib/tests/test_pvsystem.py
cwhanse: May be a good idea to use a test fixture here, for future re-use.

pvlib/tests/test_pvsystem.py
cwhanse: Hope I got that right. If so, please change `test_PVSystem_irradiance_loss_pvsyst` to use the fixture.

pvlib/pvsystem.py
cwhanse: I am suggesting a new line for the docstring here.

pvlib/pvsystem.py
btaute: The PVSyst near shading loss model is broken out into beam, diffuse, and albedo components.  (It's far shading loss is only a beam component.)  It totals these components and applies them as a global shading loss.  Then it applies an Incident Angle Modifier to the beam component.  Then the soiling losses (which have snow losses lumped into them) are applied as a constant haircut to effective irradiance for the entire month.  So, PVSyst is: Shading -> Incident Angle Modifier -> Soiling.

An irradiance_shading_loss_model_pvsyst will need to calculate its global shading loss by accessing the total_irrad['poa_direct'] and total_irrad['poa_diffuse'] parameters.  An alternative implementation to this PR would be to include the shading loss factors when calculating the effective_irradiance_model.

As long as the shading_loss_model properly accounts for each component of the global shading loss and applies this loss as a percentage of total irradiance (as PVSyst does), I don't believe the order of the losses affects the final result.  With this in mind, my objective was to make this PR as small as possible and apply the the shading/soiling/snow losses to the effective irradiance without changing the effective_irradiance_model in PVLib.  Though I'm certainly open to different implementation strategies.

cwhanse: No objections to the approach from me, only encouragement. We can plan to add other small pieces (e.g. far shading) for steps in the PVsyst loss chain, and eventually a convenience method that strings the pieces together, to go from beam and diffuse to effective irradiance with losses.

pvlib/pvsystem.py
btaute: I realize now that the math in my docstring doesn't match my code.  With the code, I intended to have the parameters input as fractions (input .02 to mean a 2% reduction in irradiance) and the output as a loss of irradiance in W/m2.  The docstring equation implies an input of percentages (input 2 to mean a 2% reduction in irradiance) and an output as a percentage of irradiance (which means the effective_irradiance parameter isn't needed anymore).

I'm assuming the preference is just that I pick one or the other and stay consistent throughout, but I thought this would be a good thing to ask for an opinion on, as it will guide the way we approach all of the other loss models going forward.  For reference, the current pvlib spectral/aoi loss models output loss coefficients and the pvwatts_loss_model uses the methodology spelled out in the docstring (returns a percentage).  @cwhanse any preference?

cwhanse: I'm sure that the processes models for e.g. reflection loss (in `pvlib.iam`) return fractions, as does the snow model in `pvlib.snow` and the soiling model. 

Loss parameters in `PVSystem.pvwatts_losses` are in percent and the output is an overall loss factor in %. Unless there's a good reason to change to fractions (and modify the existing function) I'm OK staying with % as the unit for loss model parameters and for it's output.

I suspect that the PVsyst loss chain has a lot in common with the PVWatts loss chain. That's OK, one mission of pvlib is to clarify what different models do.  At some point we probably should have a utility function to convolve loss factors that is used by the two loss chains, perhaps private. Doesn't have to be in this PR.

btaute:
I believe I've incorporated all of your comments @cwhanse. The only one outstanding is the What's New addition.  This should be ready for another (maybe final) review.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+    loss factors are compounded to calculate a single loss factor.
+
+    The separate losses are compounded using the following equation:
+
+    .. math::
+
+        L_{total}(\%) = 100 [ 1 - \Pi_i ( 1 - \frac{L_i}{100} ) ]
+
+    :math:L_{total} is the total irradiance loss resulting from the effects of
+    shading, snow, and soiling.
```
wholmgren, line 2572: other terms in the equation are not defined.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    Note the parameters must each be a series with a DatetimeIndex.  The
+    index of the returned series will be the same as the effective_irradiance
+    index.  All other parameters will be resampled to match this index with
+    a "fill forward" method for filling holes.
```
wholmgren, line 2577: Why do we do this in this function? It's non-standard behavior in pvlib. 

Perhaps it's discussed in the PR comments above (which I have not read) but if it's important it should be in the documentation or in a comment.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+    fill_method = 'ffill'
+
+    shading_loss = irradiance_shading_loss.reindex_like(effective_irradiance,
+                                                        method=fill_method)
+
+    snow_loss = irradiance_snow_loss.reindex_like(effective_irradiance,
+                                                  method=fill_method)
+
+    soiling_loss = irradiance_soiling_loss.reindex_like(effective_irradiance,
+                                                        method=fill_method)
```
wholmgren, line 2615: remove pending resolution of comment above

pvlib/tests/test_pvsystem.py
```diff
@@ -1347,6 +1347,39 @@ def test_PVSystem_pvwatts_ac_kwargs(mocker):
     assert out < pdc
 
 
+@pytest.fixture()
+def irradiance_loss_pvsyst():
+    final_index = pd.date_range(start='1/1/1990 12:00', periods=365, freq='D')
```
wholmgren, line 1352: please use ISO 8601

pvlib/pvsystem.py
```diff
@@ -849,6 +849,22 @@ def pvwatts_ac(self, pdc):
         return inverter.pvwatts(pdc, self.inverter_parameters['pdc0'],
                                 **kwargs)
 
+    def irradiance_loss_pvsyst(self, effective_irradiance,
```
wholmgren, line 852: This method does not extract any data from a `PVSystem` attribute, so it's not clear to me that it's worthwhile.

pvlib/pvsystem.py
wholmgren: Reading the test code I now understand the motivation. I'm not convinced that it's the right approach for a general purpose library. Options are:

1. we do nothing and let numpy/pandas broadcasting/indexing do its thing
2. we expose a `fill_method` argument or keyword argument in the function
3. leave it as is
4. something else?

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
     return losses
 
 
+def irradiance_loss_pvsyst(effective_irradiance, irradiance_shading_loss,
+                           irradiance_snow_loss, irradiance_soiling_loss):
```
wholmgren, line 2557: The function would return the same thing if the user permuted the loss arguments. We could consider `*losses` for a more general and extensible approach.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2553,73 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    Note the parameters must each be a series with a DatetimeIndex.  The
+    index of the returned series will be the same as the effective_irradiance
+    index.  All other parameters will be resampled to match this index with
+    a "fill forward" method for filling holes.
+
+    Parameters
+    ----------
+    effective_irradiance : Series
+            The plane-of-array irradiance (W/m2) that is to be reduced.
```
wholmgren, line 2582: Why require a series of effective irradiance when only its DatetimeIndex is required?

pvlib/pvsystem.py
btaute: I believe exposing a fill_method argument makes sense.  The reason for an explicit outline of the broadcasting method is to ease the transition of users coming from PVSyst.  There, you input a soiling loss value for each month.  Ideally, a user could input an array of 12 values for the soiling_loss parameter, and the function would be able to broadcast them to the appropriate month.  However, I wanted this to maintain flexibility for multi-year simulations with different monthly values and ultimately settled on the solution that I did.  I'm always open to other ideas, but I think we will want to make the results of broadcasting obvious.  (Maybe that's just in a tutorial.)

pvlib/pvsystem.py
btaute: It's really just another aspect of the broadcasting discussion happening in comments above.  The irradiance_losses percentages should be multiplied by effective_irradiance to reduce the effective_irradiance that goes into the single diode model.  While previous discussion led to the implementation of a percentage output (so the reduction of effective_irradiance happens outside of this function), we still want to make sure the DatetimeIndex of the percentage losses lines up with those of effective_irradiance.  It's true that the only parameter required to do this is the DatetimeIndex of effective_irradiance, but it felt odd to request that explicitly instead of simply pulling it from the Series.  Although requesting it explicitly would allow for better parameter naming which might reduce some of the broadcasting confusion.

pvlib/pvsystem.py
btaute: I like this idea.  I believe calling out the losses explicitly helps guide new users, but I think an in-depth tutorial can serve that purpose better while allowing for more flexibility here.

If we make this update, I don't think the function name "irradiance_loss_pvsyst" is apt anymore.  The point of the function is that it allows a convenient way to compound any irradiance losses that should be applied to the total POA (not specific to beam, diffuse, or ground POA).  Maybe that means it should be called "compound_effective_irradiance_losses?"  Maybe "combine_effective_irradiance_losses?"

A future implementation more inline with PVSyst should be a "pvsyst_effective_irradiance" function.

docs/sphinx/source/whatsnew/v0.8.0.rst
cwhanse: I agree with @wholmgren we can omit these lines from whatsnew

pvlib/pvsystem.py
cwhanse: +1 to `combine_loss_factors` or similar for a general purpose function, since the PVWatts model does the same calculation on a different set of inputs.

I think it's helpful to also have a function named `irradiance_loss_pvsyst` or similar, with arguments that are easily recognized by someone used to using the PVsyst loss model. If we keep a _pvsyst function, it should return effective_irradiance with losses applied, instead of the loss factor.

`effective_irradiance_pvsyst` implies that the function also calculates and applies reflection and spectral mismatch modifiers, which is a larger scope that this PR.

btaute:
I'm not sure how to address the Read the Docs build failure.  Any advice there would be greatly appreciated.

I changed the direction of the Pull Request based on the conversation.  It's now a general function for combining losses, which will end up getting called by almost all of the specific system loss functions (i.e. PVSyst loss functions).  The math is simple but it also takes into account resampling, which will be beneficial for extrapolating monthly values to hourly values, etc.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2537,48 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+    Series
+        Fractions resulting from the combination of each loss factor
+    """
+  combined_factor = 1
+ 
+  for loss in losses:
+    loss = loss.reindex(index, method=fill_method)
+    combined_factor *= (1 - loss)
+
+  return 1 - combined_factor
```
kanderso-nrel, line 2579: @btaute The RTD failure is just because this function body is indented 2 spaces, which doesn't match the docstring indentation of 4 spaces.  I think indenting to 4 spaces should fix it.

btaute: Thanks!

cwhanse:
@btaute the numerical test failures are unrelated to your PR #1018

docs/sphinx/source/api.rst
```diff
@@ -311,6 +311,13 @@ PVWatts model
    inverter.pvwatts
    pvsystem.pvwatts_losses
 
+Loss models
```
cwhanse, line 314: Let's move this to the "Effects on PV System Output" section, OK to keep it as a subheading there.

If you agree, the outline hierarchy would be;

Effects on PV System Output
    Loss models
        pvsystem.combine_loss_factors
    snow.coverage_nrel

etc.

pvlib/pvsystem.py
```diff
@@ -2537,6 +2537,48 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+    Combines Series loss fractions while setting a common index.
+
+    The separate losses are compounded using the following equation:
+
+    .. math::
+
+        L_{total} = 1 - [ 1 - \Pi_i ( 1 - L_i ) ]
+
+    :math:L_{total} is the total loss returned
+    :math:L_i is each individual loss factor input
```
cwhanse, line 2551:
```suggestion
    :math:`L_{total}` is the total loss returned
    :math:`L_i` is each individual loss factor input
```

pvlib/pvsystem.py
```diff
@@ -2537,6 +2537,48 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+    All losses will be resampled to match the index parameter using
+    the fill method specified (defaults to "fill forward").
+
+    Parameters
+    ----------
+    index : DatetimeIndex
+        The index of the returned loss factors
+
+    *losses : Series
+        Each parameter is a series of fractions to be compounded
```
cwhanse, line 2563:
```suggestion
    *losses : Series
        One or more Series of fractions to be compounded
```

pvlib/pvsystem.py
```diff
@@ -2537,6 +2537,48 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    fill_method : {None, ‘backfill’/’bfill’, ‘pad’/’ffill’, ‘nearest’}
```
cwhanse, line 2565: Are the opening/closing single quotes important here? Looks like a minor issue the pandas docstring is rendered, assuming that is cut/pasted here. I think single quote ' is ok for both, I could be wrong.

kanderso-nrel: +1 to not using "fancy" quotation marks.  Suggest also specifying the default value rather than listing None, e.g.:

`fill_method : {'backfill'/'bfill', 'pad'/'ffill', 'nearest'}, default 'backfill'`

wholmgren:
@btaute it looks to me like this could be merged once the conflicts are resolved.

btaute:
@wholmgren I believe those conflicts have been resolved.

wholmgren:
@btaute looks like the merge brought some older unwanted code back into the modules.

btaute:
Sorry, I misunderstood the conflicts.  Is this better?

docs/sphinx/source/whatsnew/v0.8.0.rst
```diff
@@ -75,23 +75,9 @@ Enhancements
...
-  for fitting the Pvsyst and De Soto models to IV curve data (:issue:`227`)(:pull:`708`)
-* Add factory methods :py:meth:`~pvlib.modelchain.ModelChain.with_pvwatts`
-  :py:meth:`~pvlib.modelchain.ModelChain.with_sapm` to create ``ModelChain``
-  objects configured for the respective modeling paradigms. The
-  configurations are defined in ``modelchain.PVWATTS_CONFIG``, and
-  ``modelchain.SAPM_CONFIG``. (:issue:`1013`, :pull:`1022`)
-* Added *racking_model*, *module_type*, and *temperature_model_parameters* to
-  PVSystem, LocalizedPVSystem, SingleAxisTracker, and
-  LocalizedSingleAxisTracker repr methods. (:issue:`1027`)
-* Added ability for :py:func:`pvlib.soiling.hsu` to accept arbitrary time intervals. (:pull:`980`)
```
wholmgren, line 94: @btaute one more merge issue here and then we'll be done. sorry for the hassle

btaute: Okay.  I took one more stab at it.

wholmgren:
Thanks @btaute!

PR #1017, opened 2020-08-04
Add diffuse self-shading functions

kanderso-nrel:
- ~Closes #xxxx~
 - [x] Tests added

This implements a modified version of the diffuse self-shading model in Passias and Kallbach (1984) "SHADING EFFECTS IN ROWS OF SOLAR CELL PANELS", which I believe has an error in the equation for the average masking angle (Eq 9).  Here is a gist comparing the paper's equation with the one I derived, along with a numerical integration for comparison: https://gist.github.com/kanderso-nrel/2c6c3a1853338cdef5b4bbc67092ccc8.  Note that I implemented two versions of the paper's equation to account for an ambiguous minus sign.

If anyone has a Mathematica license and can produce a cleaner version of that equation, I'd be happy to replace it -- that quick and dirty Maxima script in the docstring is about as much as I can do with symbolic solvers.


kanderso-nrel:
Test failures appear to be unrelated, a change in pandas: #1018

pvlib/shading.py
```diff
@@ -0,0 +1,140 @@
...
+    References
+    ----------
+    .. [1] Gilman, P. et al., (2018). "SAM Photovoltaic Model Technical
+       Reference Update", NREL Technical Report NREL/TP-6A20-67399.
+       Available at https://www.nrel.gov/docs/fy18osti/67399.pdf
+    .. [2] D. Passias and B. Källbäck, "Shading effects in rows of solar cell
+       panels", Solar Cells, Volume 11, Pages 281-291.  1984.
+       DOI: 10.1016/0379-6787(84)90017-6
+    """
+    return 1 - np.cos(np.radians(masking_angle)/2)**2
```
cwhanse, line 140: Could import `pvlib.tools.cosd` here

pvlib/shading.py
```diff
@@ -0,0 +1,140 @@
...
+    # when beta=0, divide by zero makes psi_avg NaN.  replace with 0:
+    psi_avg = np.where(np.isfinite(psi_avg), psi_avg, 0)
+
+    if isinstance(surface_tilt, pd.Series):
+        psi_avg = pd.Series(psi_avg, index=surface_tilt.index)
+
+    return np.degrees(psi_avg)
+
+
+def passias_sky_diffuse(masking_angle):
```
cwhanse, line 101: Wondering if this signature creates space for other approaches to calculate self-shading. What if the arguments are `tilt` and 'gcr' and `passias_masking_angle` was private and called within this function? I doubt that the average "masking angle" is general among other models, I could be wrong.

A thought: how is this visible sky dome calculation done in `bifacial_vf`? Maybe there's some commonality that could inform the interface here, although we wouldn't re-use code from `bifacial_vf`. If I recall right, that calculation is done with view factors over small increments of angle.

I have a mild preference for `sky_diffuse_passias` as the function name.

pvlib/shading.py
```diff
@@ -0,0 +1,140 @@
...
+        Average angle from horizontal where diffuse light is blocked by the
+        preceding row [degrees].
+
+    See Also
+    --------
+    passias_sky_diffuse
+
+    Notes
+    -----
+    The pvlib-python authors believe that Eqn. 9 in [1]_ is incorrect.
```
cwhanse, line 39: The equation below isn't equivalent to the published equation? Or, you couldn't reproduce the published equation [9]?

kanderso-nrel: The latter.  The published equation [9] also gives different results from evaluating the integral numerically: https://gist.github.com/kanderso-nrel/2c6c3a1853338cdef5b4bbc67092ccc8

pvlib/shading.py
kanderso-nrel: I separated the two functions in anticipation of people wanting to use the "worst-case" angle like SAM does instead of Passias et al.'s average angle.  That said, a function that just does `1 - cos(angle/2)^2` does seem strange.  

Good idea about bifacial_vf; I haven't had an excuse to get familiar with it before now but am happy to have one.

pvlib/shading.py
cwhanse: I'm persuaded. Your solution looks correct to me. I computed a few points on the k=1 and k=2 curves to check.

pvlib/shading.py
```diff
@@ -0,0 +1,140 @@
...
+    Parameters
+    ----------
+    masking_angle : numeric
+        The elevation angle below which diffuse irradiance is blocked
+        [degrees].
+
+    Returns
+    -------
+    derate : numeric
+        The fraction [0-1] of blocked diffuse horizontal irradiance.
```
cwhanse, line 125: sky diffuse irradiance? Or does the fraction account for blocking ground reflected irradiance also (I don't think it does).

pvlib/shading.py
cwhanse: I looked at the SAM documentation. I think that the low level functions that calculates the average `masking_angle` is useful, as would be a function that returns `masking_angle` at a point along the row's surface. Maybe someone would want to use the `masking_angle` in combination with other sky diffuse models.

Still chewing on the name `passias_sky_diffuse` for the effects function.

pvlib/shading.py
```diff
@@ -0,0 +1,193 @@
...
+    angle [1]_.  SAM uses a "worst-case" approach where the masking angle
+    is calculated for the bottom of the array (i.e. ``height=0``) [2]_.
+
+    Parameters
+    ----------
+    surface_tilt : numeric
+        Panel tilt from horizontal [degrees].
+
+    gcr : float
+        The ground coverage ratio of the array [unitless].
```
kanderso-nrel, line 26: I have a notion that `gcr` is sometimes defined as `gcr := width * cos(tilt) / pitch` for fixed-tilt arrays.  Is that a common usage?  If so, this parameter description and the one in `masking_angle_passias` should clarify that `width / pitch` is meant.

pvlib/shading.py
kanderso-nrel: I added a `masking_angle` function that calculates the angle for a given point and renamed `passias_sky_diffuse` to `sky_diffuse_passias`, but am happy to rename if you have a better alternative.

> Maybe someone would want to use the masking_angle in combination with other sky diffuse models.

It occurs to me that the masking angle is also useful for comparing with solar position to determine if direct shading is relevant.

pvlib/shading.py
cwhanse: In my experience the usual definition of ground coverage ratio is the ratio of module area to ground area, where the "ground area" is left somewhat ambiguous. That's the definition used in [Pvsyst](https://forum.pvsyst.com/viewtopic.php?t=2379), [Helioscope ](https://www.folsomlabs.com/content/resources/module-spacing.pdf) and `pvlib.tracking.SingleAxisTracker`. I think we should stick with `gcr = width / pitch`. I'd be interested to see where the ratio of projected horizontal width to pitch is used.

pvlib/shading.py
```diff
@@ -0,0 +1,193 @@
...
+    Even when the sun is high in the sky, a row's view of the sky dome will
+    be partially blocked by the row in front. This causes a reduction in the
+    diffuse irradiance incident on the module. The reduction depends on the
+    masking angle, the elevation angle from a point on the shaded module to
+    the top of the shading row.  SAM assumes the "worst-case" loss where the
+    masking angle is calculated for the bottom of the array [1]_. In [2]_
+    the masking angle is calculated as the average across the module height.
+
+    This function, as in [2]_, makes the assumption that sky diffuse
+    irradiance is isotropic.
```
cwhanse, line 166: Entirely editorial - I'd swap the order of the last and next-to-last sentences, so that the focus stays on the Passias approach.

pvlib/shading.py
```diff
@@ -0,0 +1,193 @@
...
+
+    Parameters
+    ----------
+    surface_tilt : numeric
+        Panel tilt from horizontal [degrees].
+
+    gcr : float
+        The ground coverage ratio of the array [unitless].
+
+    height : numeric
```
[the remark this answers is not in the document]
kanderso-nrel: agreed

cwhanse:
@wholmgren @mikofski any objections to merge?

pvlib/shading.py
```diff
@@ -0,0 +1,193 @@
...
+    # wrap it in an array so that division by zero is handled well
+    beta = np.radians(np.array(surface_tilt))
+    sin_b = np.sin(beta)
+    cos_b = np.cos(beta)
+    X = 1/gcr
+
+    with np.errstate(divide='ignore', invalid='ignore'):  # ignore beta=0
+        term1 = -X * sin_b * np.log(np.abs(2 * X * cos_b - (X**2 + 1))) / 2
+        term2 = (X * cos_b - 1) * np.arctan((X * cos_b - 1) / (X * sin_b))
+        term3 = (1 - X * cos_b) * np.arctan(cos_b / sin_b)
```
mikofski, line 140: @kanderso-nrel  I believe that `arctan(cotangent(beta)) = beta-pi/2` right?

kanderso-nrel: Yeah it's silly to have `arctan(cot(x))`.  The problem is that I wanted the output of that Maxima snippet, the latex equation, and the python code to all be identical but I couldn't figure out what assumptions I needed to give Maxima to simplify that term.  I figured the inefficiency of that python term was negligible compared with making the connections as obvious as possible, but maybe there's a clearer way to tie the three pieces together?

mikofski: See [trig identities: inverse trig funcs](https://en.wikipedia.org/wiki/List_of_trigonometric_identities#Inverse_trigonometric_functions)

    arctan x + arccot x = pi / 2
    arctan x = pi / 2 - arccot x

so

    arctan(cot(beta)) = pi / 2 - arccot(cot(beta)) = pi / 2 - beta

QED

mikofski:
LGTM too. I made one comment on a trig identity that might be useful, but probably not essential. This is a great addition. Thanks Kevin!

pvlib/shading.py
mikofski: Don't sweat it. Let's merge.

PR #530, opened 2018-08-13
add airspeed velocity config and performance tests

wholmgren:
This pull request adds configuration and test files for the [airspeed velocity](https://asv.readthedocs.io/en/stable/index.html) performance benchmarking tool. 

This is new to me, but I'm putting it up here for feedback. I developed this while benchmarking #528 and #529. It seems useful. I'm not sure what the standard should be for merging this.

@mikofski do you have any experience using asv or similar tools? Anyone else?

In the long run, it would be ideal to set up a cloud service to automatically run these tests on the same (possibly virtual) hardware each time. As of now, I only want to consider running these tests locally.

 - [x] Closes #419
 - [x] I am familiar with the [contributing guidelines](http://pvlib-python.readthedocs.io/en/latest/contributing.html).
 - [ ] Fully tested. Added and/or modified tests to ensure correct behavior for all reasonable inputs. Tests (usually) must pass on the TravisCI and Appveyor testing services.
 - [x] Updates entries to `docs/sphinx/source/api.rst` for API changes.
 - [x] Adds description and name entries in the appropriate `docs/sphinx/source/whatsnew` file for all changes.
 - [ ] Code quality and style is sufficient. Passes ``git diff upstream/master -u -- "*.py" | flake8 --diff``
 - [ ] New code is fully documented. Includes sphinx/numpydoc compliant docstrings and comments in the code where necessary.
 - [ ] Pull request is nearly complete and ready for detailed review.

other:
 - [x] add a simple readme like scipy's -- rely on asv's documentation for all but the most basic commands and instructions
 - [ ] consider ``benchmarks/benchmarks`` rather than ``asv_bench/benchmarks`` (copied from xarray)

PR #1037, opened 2020-08-29
Add pvlib.temperature.fuentes

### LINKED ISSUE #1032: Add Fuentes temperature model
PVWatts uses a temperature model, Fuentes, that's not implemented in pvlib python.

Add the Fuentes model described in this pdf: [850330.pdf](https://github.com/pvlib/pvlib-python/files/5124949/850330.pdf)

Continue to use `sapm`.

**Additional context**
Originally discussed in https://github.com/pvlib/pvlib-python/pull/1022#discussion_r473559969

kanderso-nrel:
- [x] Partially addresses #1032 (this is only the low-level function, doesn't expose it in the high-level APIs)
 - [x] Tests added

This implementation is a port of the original FORTRAN77 code with some minor refactoring.  As @wholmgren mentioned in #1032, numba might be a good option for speeding up this non-vectorized code.

If it's okay, I'd prefer to integrate it with PVSystem and ModelChain in a follow-on PR and focus this one on the model implementation.  Also, I put the whatsnew entry in v0.8.0 but OK with me to push it to v0.8.1.


kanderso-nrel:
Ready for review.  Test failures seem unrelated (`test_read_midc_raw_data_from_nrel`).

wholmgren:
Thanks @kanderso-nrel this looks great! 

Can you comment on the performance? I'm not so concerned about the absolute performance, just that we have a benchmark for future reference. This is a case where it would be nice to have #530.

What is the motivation for using a DataFrame within the function? My main concern is that if we try to speed this up with numba then we might want to replace the DataFrame with numpy arrays or lists. I suspect the vast majority of the time is spent in the inner loop and that's fine to jit as is. But the outer loop requires the GIL and I worry, perhaps prematurely, that the pandas `__getitem__` and especially the `__setitem__` will become bottlenecks. In any case, it would probably easier to numba-ify the whole loop instead of just the inner loop.

pvlib/temperature.py
```diff
@@ -423,3 +424,206 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    Notes
+    -----
+    This function returns slightly different values from PVWatts at night
+    and just after dawn. This is because the SAM SSC assumes that module
+    temperature equals ambient temperature when irradiance is zero so it can
+    skip the heat balance calculation at night.
+
+    References
+    ----------
+    .. [1] Fuentes, M. K. A Simplifed Thermal Model for Flat-Plate
```
cwhanse, line 513: I would add "Sandia National Laboratories" as the publishing source on [1])

pvlib/temperature.py
```diff
@@ -423,3 +424,206 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    # sky temperature -- Equation 24
+    df['tsky'] = 0.68 * (0.0552 * df['tamb']**1.5) + 0.32 * df['tamb']
+    # wind speed at module height -- Equation 22
+    # not sure why the 1e-4 factor is included -- maybe the equations don't
+    # behave well if wind == 0?
+    df['windmod'] = df['wind_speed'] * (module_height/wind_height)**0.2 + 1e-4
+
+    tmod0 = 293.15
+
+    for idx, row in df.iterrows():
```
cwhanse, line 584: Not a recommendation, but a question: consider `scipy.integrate.solve_ivp` here? The RHS of the diff. eq. in Fuentes is easily vectorized, might a performance gain over the loops.

pvlib/temperature.py
```diff
@@ -423,3 +424,206 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    poa_global : pandas Series
+        Total incident irradiance [W/m^2]
+
+    temp_air : pandas Series
+        Ambient dry bulb temperature [C]
+
+    wind_speed : pandas Series
+        Wind speed [m/s]
+
+    inoct : float
```
cwhanse, line 474: editorial: 'noct_installed' maybe?

pvlib/temperature.py
kanderso-nrel: Interesting idea.  I'm a little rusty on ODE problems but... isn't the approach taken here not only doing IVP integration (outer loop) but also fixed point iteration (inner loop)?  Is it possible to do both with `solve_ivp`?  If so it would be a good alternative to this manual looping business, but it would take me some time to figure out.

kanderso-nrel:
> What is the motivation for using a DataFrame within the function? 

The only motivation is because I like the `df.iterrows()` interface.  Turns out that `df.loc[i, 'colname'] = value` is dramatically slower than I would have predicted.  I switched to looping through a `zip` of the individual series and storing results in a numpy array, reducing the runtime from 4-5 seconds to ~1 second for an 8760. 

For whether numba is worthwhile, by commenting out the inner loop, runtime goes down to 7ms on my computer, so it is true that 99% of the runtime is now spent in the inner loop.  Extracting the body of the inner loop to its own function and decorating it and `_fuentes_hconv` with `@numba.jit` brings down runtime to 728ms on the first run and ~72ms for subsequent runs.  I'm impressed that even with the overhead of compiling on the first invocation, it still manages to be faster.

If we were to use numba here, would we centralize the ["use numba if available method"](https://github.com/pvlib/pvlib-python/blob/master/pvlib/spa.py#L16-L43) from spa.py and use it for both `spa.py` and `fuentes`?

wholmgren:
Yeah, pandas indexing can be much slower than numpy or list operations. You don't get all that fancy pandas behavior for free! Usually doesn't matter in practice but in a loop it can be costly.

I wasn't expecting a numba implementation in this PR but great to hear of the progress!

> If we were to use numba here, would we centralize the "use numba if available method" from spa.py and use it for both spa.py and fuentes?

That works for me. Also worth noting that spa numba code is ~5 years old, which is most of numba's lifetime. So it might be worth checking if there's a more modern pattern for conditional numba usage.

kanderso-nrel:
After taking a closer look, the way numba is used with `spa.py` and `solarposition.py` is more complicated than I thought, it's not just swapping in a no-op decorator if numba isn't available.  Now that the pure python implementation is not so slow anymore, could we push the numba idea to a different PR?  I think it would be worth chewing on it for a bit.

Test failures not related.

wholmgren:
Yes, I think that's a good plan.

pvlib/temperature.py
cwhanse: I think `scipy.integrate.solve_ivp` _could_ be used here but not without taking apart and reworking some of the supporting calculations. The two-loop approach in Fuentes is a customized solver for the ODE in his paper, taking advantage of an approximation that linearizes the equation. I don't see any conceptual reason why `solve_ivp` with defaults couldn't solve the equation. But, OK to use the Fuentes technique unless performance becomes unsatisfactory.

pvlib/temperature.py
```diff
@@ -423,3 +424,199 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    # the boundary between laminar and turbulent is modeled as an abrupt
+    # change at Re = 1.2e5:
+    if check_reynold and reynold > 1.2e5:
+        # turbulent convection
+        hforce = 0.0282 / reynold**0.2 * densair * windmod * 1007 / 0.71**0.4
+    else:
+        # laminar convection
+        hforce = 0.8600 / reynold**0.5 * densair * windmod * 1007 / 0.71**0.67
+    # free convection via Grashof number
+    # NB: the 0.5 factor is from assuming tilt=30; should tilt be a parameter?
```
cwhanse, line 445: Agree that tilt should be exposed, tilt=30 as a default.

pvlib/temperature.py
```diff
@@ -423,3 +424,199 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+           doi:10.2172/1158421.
+    """
+    # ported from the FORTRAN77 code provided in Appendix A of Fuentes 1987;
+    # nearly all variable names are kept the same for ease of comparison.
+
+    boltz = 5.669e-8
+    emiss = emissivity
+    absorp = absorption
+    xlen = hydraulic_diameter
+    cap0 = 11000
```
cwhanse, line 528: `cap0` accounts for module area and mass, correct?

pvlib/temperature.py
```diff
@@ -423,3 +424,199 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+        PVWatts defauls is 9.144 [m]
+
+    emissivity : float, default 0.84
+        The effectiveness of the module at radiating thermal energy. [unitless]
+
+    absorption : float, default 0.83
+        The fraction of incident irradiance that is converted to thermal
+        energy in the module. [unitless]
+
+    hydraulic_diameter : float, default 0.5
```
cwhanse, line 495: Should have a modern default for a 60 or 72 cell, 6-inch cell module. Fuentes leaves out a lot of detail. Maybe replace `hydraulic_diameter` with `module_width`, `module_length` and calculate the diameter?

cwhanse:
@kanderso-nrel do you have access to the PVWatts code? The reason for adding this model to pvlib-python is to better emulate what PVWatts is doing. It would be good to check that this port from Fuentes's FORTRAN is consistent with the PVWatts code.

kanderso-nrel:
Here's the implementation of the Fuentes model in the SSC: https://github.com/NREL/ssc/blob/c6e41d7d946723e3cc4111262abcab081c151621/shared/lib_pvwatts.cpp#L173-L290

I haven't gone through it detail but a quick skim suggests it's a straightforward port of the F77 code too.  Note that pvwatts.nrel.gov uses PVWatts v5 in the SSC, but I think SAM uses v7. Both [v5](https://github.com/NREL/ssc/blob/develop/ssc/cmod_pvwattsv5.cpp#L30) and [v7](https://github.com/NREL/ssc/blob/develop/ssc/cmod_pvwattsv7.cpp#L30) import the above code, so the difference doesn't matter here. 

Also in case you missed it, the data for the tests were fetched from pvwatts.nrel.gov, so that's a consistency check.  But that doesn't mean there isn't additional value in comparing the code too.

pvlib/temperature.py
kanderso-nrel: That would make the API nicer, but using modern defaults would make the default arguments inconsistent with PVWatts, which as far as I can tell follows the Fuentes defaults.  Which should take precedence here?

cwhanse: I guess, keep the default to mimic PVWatt's behavior now, and add to the docstring how to calculate hydraulic diameter for a rectangular module.

cwhanse:
interesting - the tilt is also hardwired at 30 degrees. I'm persuaded to keep the defaults to match PVWatts/SAM, but inform the user how to change the defaults for tilt and hydraulic_diameter

pvlib/temperature.py
kanderso-nrel: Oh, I missed the last comment here -- I changed it to take `module_width` and `module_length` as parameters and calculate the hydraulic diameter internally.  Happy to change it back and add a Notes section with the hydraulic diameter equation if you think that's the better option, let me know.

pvlib/temperature.py
kanderso-nrel: My understanding is that `cap0` is the product `m c` where `m` is the mass per unit area [kg/m^2] and `c` is the specific heat of the whole module [J / (kg K)], so `cap0` has units [J / (m^2 K)].  The 11,000 comes from Table 3 on page 8.

pvlib/temperature.py
```diff
@@ -423,3 +424,217 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+        Wind speed [m/s]
+
+    noct_installed : float
+        The "installed" nominal operating cell temperature as defined in [1]_.
+        PVWatts assumes this value to be 45 C for rack-mounted arrays and
+        49 C for roof mount systems with restricted air flow around the
+        module.  [C]
+
+    module_height : float, default 5.0
+        The height above ground of the module above the ground. The PVWatts
```
cwhanse, line 487: Should say at what position on the module the height is determined. I couldn't find that Fuentes specified that point, did you? I'm guessing center of module is appropriate, to represent the average wind speed along the module's face..

pvlib/temperature.py
cwhanse:
```suggestion
    cap0 = 11000  # units of [J / (m^2 K)], equal to mass per unit area times specific heat of the module.
```

pvlib/temperature.py
cwhanse: No need to change it back, it's was OK before and better now.

pvlib/temperature.py
kanderso-nrel: The most specific description I found was "average array height above ground" in the second paragraph of the introduction.  Not sure if that's referring to the change in height across a module or if it's trying to accommodate variation in racking height, but center of module makes sense to me too.

pvlib/temperature.py
```diff
@@ -423,3 +424,219 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    Parameters
+    ----------
+    poa_global : pandas Series
+        Total incident irradiance [W/m^2]
+
+    temp_air : pandas Series
+        Ambient dry bulb temperature [C]
+
+    wind_speed : pandas Series
+        Wind speed [m/s]
```
wholmgren, line 478: seems like this function would work just as well with any kind of array input, right? If so, we should make the type requirements here and in the return line `array_like` and change the last line of the function to only return a series if `poa_global` is a series.

kanderso-nrel: Unfortunately the timedelta between measurements is needed for the thermal capacitance/time lag part of the model, so I think it has to either be Series inputs or another parameter would be added for the measurement timestamps.

wholmgren: ah, right. let's leave it alone then. new merge conflict (sorry)

wholmgren:
thanks @kanderso-nrel

PR #1039, opened 2020-09-01
fix github action for publishing releases

wholmgren:
#1038 didn't work because pypi (and test pypi) don't allow local versions (tag+commits.hash). With some work (e.g. [here](https://github.com/yeatmanlab/pyAFQ/blob/master/setup.py#L62)) it would be possible to upload distributions for each commit to test pypi, but I don't think it's worth the effort. So this PR simplifies the build to tags that start with "v" and only pushes to pypi.

I don't understand why the PyPA guide suggests uploading each commit to test pypi if the standard tools don't make it easy to do so. I checked a hadnful of PyPA's own packages and none of them follow this practice either.

Once this PR is merged I'll test it by pushing an 0.8.0 alpha tag.


wholmgren:
✅ https://github.com/pvlib/pvlib-python/runs/1058438049?check_suite_focus=true

✅ https://pypi.org/project/pvlib/0.8.0a0/

PR #1062, opened 2020-09-20
Add numpy SPECTRL2 implementation

kanderso-nrel:
- ~Closes #xxxx~
 - [x] Tests added

Additional to-do:
- [x] Get it to broadcast across vector inputs
- [x] remove debugging code

Calculating the spectral distribution for a single timestamp takes under a millisecond on my laptop.  I've been comparing against the solar_utils wrapper around the NREL C implementation (thanks @mikofski for making it so easy to use!) and have discovered several discrepancies between the equations in the report, the report code appendix, and the C implementation.  It doesn't help that the C implementation uses its own (non-SPA) solar position calculator.  By matching the python code to the C implementation for each of the differences, the difference in output spectra is small enough for me to start blaming the accumulated error from the C implementation's single-precision floats:

Discrepancies:
| Equation | Report | Appendix | spectrl2_2.c |
|----------|--------|----------|--------------|
| 2-4      | 1.335  | 1.335    | 1.3366       |
| 2-11     | 118.93 | 118.93   | 118.3        |
| 3-8      | To'    | Tu'      | Tu'          |
| 3-5, 3-6, 3-7, 3-1 | double-count Cs | single-count Cs | single-count Cs |
| 2-5 | kasten1966 | kasten1966 | kastenyoung1989 |

image


kanderso-nrel:
I think this is ready for review whenever folks get a chance.  For the sake of ease of testing I've aligned all the inconsistencies (see table in original comment) with `spectrl2_2.c` but I'm open to alternatives.  See https://gist.github.com/kanderso-nrel/0dc7e7ebff7b5a9c13b52f8b48eb71ee for the validation scratchpad I've been using.

I created a new `spectrum` module for this.  Would `atmosphere` be a better home?  Could consider moving the other  spectral functions (`pvlib.atmosphere.first_solar_spectral_correction`, `pvlib.pvsystem.sapm_spectral_loss`) here if not.

Thanks!

cwhanse:
> I created a new `spectrum` module for this. Would `atmosphere` be a better home? Could consider moving the other spectral functions (`pvlib.atmosphere.first_solar_spectral_correction`, `pvlib.pvsystem.sapm_spectral_loss`) here if not.

I'm in favor the new `spectrum.py` and, related to #436, eventually moving functions that calculate spectral modifiers from `pvsystem.py` and `atmosphere.py` into `spectrum.py`.

adriesse:
Nice work.

pvlib/spectrum.py
```diff
@@ -0,0 +1,318 @@
...
+
+    # ET spectral irradiance correction for earth-sun distance seasonality
+    day_angle = 2 * np.pi * (dayofyear - 1) / 365.0  # Eq 2-3
+    earth_sun_distance_correction = (
+        1.000110
+        + 0.034221 * np.cos(day_angle)
+        + 0.001280 * np.sin(day_angle)
+        + 0.000719 * np.cos(2*day_angle)
+        + 0.000077 * np.sin(2*day_angle)
+    )  # Eq 2-2
```
wholmgren, line 102: call to `irradiance.get_extra_radiation(method='spencer')`

pvlib/spectrum.py
```diff
@@ -0,0 +1,318 @@
...
+
+    # Rayleigh scattering
+    airmass = relative_airmass * surface_pressure / 101300
```
wholmgren, line 105: hm, not quite the same as `atmosphere.absolute_airmass`. worth adding a comment about that

wholmgren (approved):
Looks good, but I didn't compare the implementation to the reference.

pvlib/spectrum.py
```diff
@@ -0,0 +1,318 @@
...
+    aerosol_asymmetry_factor : float or numpy array, default 0.65
+        Aeorosol asymmetry factor (mean cosine of scattering angle) [unitless]
+
+    Returns
+    -------
+    spectra : dict
+        A dict of arrays of length 122. All values are spectral irradiance
+        with units W/m^2/um except for `wavelength`, which is in microns.
+            * wavelength
+            * et
```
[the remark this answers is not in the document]
kanderso-nrel: Tossing out some options:
- `extraterrestrial`
- `dni_extra` to match `irradiance.get_extra_radiation`
- `extra_radiation` to match `irradiance.clearness_index`
- other?

pvlib/spectrum.py
```diff
@@ -0,0 +1,318 @@
...
+             surface_pressure, relative_airmass, precipitable_water, ozone,
+             dayofyear, aerosol_turbidity_500nm, scattering_albedo_400nm=0.945,
+             alpha=1.14, wavelength_variation_factor=0.095,
+             aerosol_asymmetry_factor=0.65):
+    """
+    Estimate spectral irradiance using the Bird Simple Spectral Model
+    (SPECTRL2).
+
+    The Bird Simple Spectral Model [1]_ produces terrestrial spectra between
+    0.3 and 4 um with a resolution of approximately 10 nm. Direct and diffuse
```
cwhanse, line 164: Mild preference to use nm as the unit throughout, at least here in the docstring.  Not sure if nm is practical for the outputs and tests.

cwhanse (commented):
I'll volunteer to check this against the reference.

Is there a reason to exclude pandas.Series from the input types?

pvlib/spectrum.py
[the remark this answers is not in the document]
wholmgren: I thought we'd discussed this in another issue in the last year but I can't find it. I think we use `dni_extra` in more functions so I'd lean that way for now.

Or avoid the problem entirely and return a tuple.

pvlib/spectrum.py
```diff
@@ -0,0 +1,318 @@
...
+The ``spectrum`` module contains functions that implement models for the solar
+irradiance spectrum.
+"""
+
+import pvlib
+from pvlib.tools import cosd
+import numpy as np
+
+# SPECTRL2 extraterrestrial spectrum and atmospheric absorption coefficients
+_SPECTRL2_COEFFS = np.zeros(122, dtype=np.dtype([
```
adriesse, line 11: Any thoughts about putting these into a csv/txt file?

pvlib/spectrum.py
```diff
@@ -0,0 +1,318 @@
...
+        + 0.034221 * np.cos(day_angle)
+        + 0.001280 * np.sin(day_angle)
+        + 0.000719 * np.cos(2*day_angle)
+        + 0.000077 * np.sin(2*day_angle)
+    )  # Eq 2-2
+
+    # Rayleigh scattering
+    airmass = relative_airmass * surface_pressure / 101300
+    rayleigh_transmittance = np.exp(
+        # Note: the report uses 1.335 but spectrl2_2.c uses 1.3366
```
adriesse, line 107: I think this information should appear in a doc string (also).

pvlib/spectrum.py
```diff
@@ -0,0 +1,318 @@
...
+
+    The Bird Simple Spectral Model [1]_ produces terrestrial spectra between
+    0.3 and 4 um with a resolution of approximately 10 nm. Direct and diffuse
+    spectral irradiance are modeled for horizontal and tilted surfaces under
+    cloudless skies. SPECTRL2 models radiative transmission, absorption, and
+    scattering due to atmospheric aerosol, water vapor, and ozone content.
+
+    Parameters
+    ----------
+    surface_tilt : float or numpy array
```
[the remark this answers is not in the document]
adriesse: Do you need to say something about the dimensions of the array arguments?

pvlib/spectrum.py
```diff
@@ -0,0 +1,318 @@
...
+    rs = sky_reflectivity
+    rg = ground_albedo
+    Ig = (Id * cosZ + Ir + Ia) * rs * rg / (1 - rs * rg)  # Eq 3-7
+
+    # total scattered irradiance
+    # Note: see discussion about Cs above.
+    # Is = Ir + Ia + Ig  # Eq 3-1
+    Is = (Ir + Ia + Ig) * Cs  # Eq 3-1
+
+    # calculate spectral irradiance on a tilted surface, Eq 3-18
```
adriesse, line 292: Could transposition be a separate function?  Or could the use be expected to do it using the existing transposition functions?

kanderso-nrel: It could be a separate function, but I'm inclined to implement the complete model.

If users want to do their own transposition, is there any issue with them passing `surface_tilt=0` and transposing afterwards (other than the inefficiency of doing an unnecessary transposition in this function)?

pvlib/spectrum.py
kanderso-nrel: Good point, I'll add a notes section for these differences.

pvlib/spectrum.py
kanderso-nrel: Minor preference from me to keep them close to the code, but I can't deny that it's 70 lines of junk that don't really need to be here.  Happy to dump to a csv if people don't want it in the code.

kanderso-nrel (commented):
> Is there a reason to exclude pandas.Series from the input types?

Probably, but I'll have to remind myself what it is.  I'd guess to make broadcasting easier.  Whatever the reason, re-wrapping in a Series before returning would probably not be an issue.

cwhanse:
> > Is there a reason to exclude pandas.Series from the input types?
> 
> Probably, but I'll have to remind myself what it is. I'd guess to make broadcasting easier. Whatever the reason, re-wrapping in a Series before returning would probably not be an issue.

I'm thinking of users who would bring e.g. `surface_tilt` or `apparent_zenith` into this function as a Series.

pvlib/spectrum.py
cwhanse: Unless we see a use for these values independent of the SPECTRL2 function I'd prefer to keep the "junk" in the source code.

kanderso-nrel:
> I'm thinking of users who would bring e.g. surface_tilt or apparent_zenith into this function as a Series.

Yeah, the gallery example is a good example of that.  I think the guts of the function should stick with numpy arrays for broadcasting purposes.  Is there a better way to allow Series inputs than manually unwrapping and re-wrapping the arrays like this?  https://github.com/pvlib/pvlib-python/blob/master/pvlib/bifacial.py#L86-L109

I don't think the RTD failure is relevant; seems like they have some configuration issue that makes the installation go crazy (example log: https://readthedocs.org/api/v2/build/12563636.txt)

cwhanse:
> I think the guts of the function should stick with numpy arrays for broadcasting purposes. 

OK with me, and OK to leave the signature in ndarray. I'll note that one advantage of accepting Series input (if practical) would be to to default `dayofyear=None` and recover dayofyear from the index of one of the inputs.

wholmgren:
> Is there a better way to allow Series inputs than manually unwrapping and re-wrapping the arrays like this? https://github.com/pvlib/pvlib-python/blob/master/pvlib/bifacial.py#L86-L109

Unfortunately, every approach that I'm aware of involves tedious (e.g. `is_pandas` as shown below), repetitive (e.g. `bifacial.py`), or complicated (previously discussed decorators) code. In the past I've searched stack overflow and various google groups for good solutions and I've not found any.

In my opinion, the bifacial input checking is too much. We've used a pattern like this in other functions:

```
is_pandas = isinstance(apparent_zenith, pd.Series)
if is_pandas:
    # coerce all inputs to np.array using e.g.
    a, b, c = tuple(map(np.array, (a, b, c))
    dayofyear = apparent_zenith.index.day_of_year
# the rest of the function
if is_pandas:
    return dict_with_appropriate_values_coerced_to_series_with_index_equal_to_apparent_zenith
    # or
    return pd.DataFrame(the_dict, index=apparent_zenith.index)
else:
    return the_dict
```

I'd also change the signature order to `apparent_zenith, aoi, surface_tilt,...`

I'm still fine with leaving the input requirement to be ndarray, but we should avoid setting ourselves up for breaking API changes in the future.

kanderso-nrel:
Inputs now allowed to be `pd.Series`.  I'm having trouble articulating why, but I'm less enthusiastic about returning pandas objects -- housing the spectra in DataFrames feels like it's misusing DataFrames in some way.  Maybe it's because the column names would be floats (one per wavelength).  To me it seems cleaner to always return arrays, but I wouldn't mind being overruled.

pvlib/spectrum.py
```diff
@@ -0,0 +1,363 @@
...
+        in [1]_ for a rural aerosol model. [unitless]
+    wavelength_variation_factor : numeric, default 0.095
+        Wavelength variation factor [unitless]
+    aerosol_asymmetry_factor : numeric, default 0.65
+        Aeorosol asymmetry factor (mean cosine of scattering angle) [unitless]
+
+    Returns
+    -------
+    spectra : dict
+        A dict of arrays of length 122. All values are spectral irradiance
```
cwhanse, line 214: Except for `wavelength` would each array's shape be (len(apparent_zenith), 122)? Maybe I don't understand.

adriesse:
This module is already quite big isn't it?  What about creating `pvlib/spectrum/spectral2.py`?  or `pvlib/spectral/spectral2.py`?  Then if anyone ever translates smarts they'll know where to put it.  :)

adriesse:
Regarding the conversion from Series to arrays, I find that numpy usually plays pretty well with Series, so I would be interested to learn where it causes problems for you.  Maybe you don't really need to convert them all?

pvlib/spectrum.py
```diff
@@ -0,0 +1,366 @@
...
+    """
+    # values need to be np arrays for broadcasting, so unwrap Series if needed:
+    is_pandas = isinstance(apparent_zenith, pd.Series)
+    if is_pandas:
+        original_index = apparent_zenith.index
+        (apparent_zenith, aoi, surface_tilt, ground_albedo, surface_pressure,
+         relative_airmass, precipitable_water, ozone, aerosol_turbidity_500nm,
+         scattering_albedo_400nm, alpha, wavelength_variation_factor,
+         aerosol_asymmetry_factor) = \
+            tuple(map(np.array, [
```
adriesse, line 263: Have you considered using np.asanyarray?

pvlib/spectrum.py
```diff
@@ -0,0 +1,366 @@
...
+    return (
+        earth_sun_distance_correction,
+        rayleigh_transmittance,
+        aerosol_transmittance,
+        vapor_transmittance,
+        ozone_transmittance,
+        mixed_transmittance,
+        aerosol_scattering,
+        aerosol_absorption,
+    )
```
adriesse, line 154: Although it's a private function, I think would be worth commenting the shapes of the arguments and return values.

pvlib/spectrum.py
```diff
@@ -0,0 +1,366 @@
...
+def _spectrl2_transmittances(apparent_zenith, relative_airmass,
+                             surface_pressure, precipitable_water, ozone,
+                             optical_thickness, scattering_albedo, dayofyear):
+    """
+    Calculate transmittance factors from Section 2 of Bird and Riordan 1984.
+    """
+    wavelength = _SPECTRL2_COEFFS['wavelength'][:, np.newaxis]
+    vapor_coeff = _SPECTRL2_COEFFS['water_vapor_absorption'][:, np.newaxis]
+    ozone_coeff = _SPECTRL2_COEFFS['ozone_absorption'][:, np.newaxis]
+    mixed_coeff = _SPECTRL2_COEFFS['mixed_absorption'][:, np.newaxis]
```
adriesse, line 96: Explain code, e.g. (if I understood this correctly):
`# make vectors into single-column arrays by adding a dimension/axis`
Also consider using:
`.reshape(-1,1)`

cwhanse:
> This module is already quite big isn't it? What about creating `pvlib/spectrum/spectral2.py`? or `pvlib/spectral/spectral2.py`? Then if anyone ever translates smarts they'll know where to put it. :)

I could support this, even if a pvlib SMARTS won't appear any time soon.

kanderso-nrel:
Thanks @adriesse for the numpy lesson :)  The numpy vs pandas issue is that I want to cleanly produce the outer product of two vectors where one axis indexes time and the other indexes wavelength, for example:

```python
N = 2
airmass = pd.Series(range(N), index=pd.date_range('2019-01-01', freq='h', periods=N))
vapor_coeff = np.array([1, 2, 3])[:, np.newaxis]
airmass * vapor_coeff  # fails, but airmass.values * vapor_coeff works and returns an array with shape (3, 2)
```

Is there a nice way to get this behavior when one is a Series?

pvlib/spectrum.py
```diff
@@ -0,0 +1,366 @@
...
+            * poa_direct
+            * poa_global
+
+    Notes
+    -----
+    NREL's C implementation ``spectrl2_2.c`` [2]_ of the model differs in
+    several ways from the original report [1]_.  The report itself also has
+    a few differences between the in-text equations and the code appendix.
+    The list of known differences is shown below.  Note that this
+    implementation follows ``spectrl2_2.c``.
```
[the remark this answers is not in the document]
kanderso-nrel: I didn't compare with the Excel version in any meaningful way.  I agree it would be nice to be able to switch between the two (if there is indeed a difference), but I hesitate to add additional complexity to this PR.  Future work?

adriesse:
I guess I haven't run into this exact situation often.  When one argument is a Series, the `pandas.__mul__` function takes over, and it doesn't play along.  (Might even warrant raising an issue.)

A clean way to get the desired behavior would be to use `np.outer`.  This would also make the intent entirely obvious, which is not the case when you rely on broadcasting with `*`.  It would require you to review each multiplication to see which one needs it, which is obviously a bit of work.  If you need a lot of them, it might indeed encumber the code and be considered "not nice".  On the other hand making the code intent obvious is "nice" in another way.

I won't hold anything up though. I just provide my thoughts on what is overall a very good contribution to pvlib!

pvlib/spectrum/spectrl2.py
```diff
@@ -0,0 +1,384 @@
...
+import pandas as pd
+
+# SPECTRL2 extraterrestrial spectrum and atmospheric absorption coefficients
+_SPECTRL2_COEFFS = np.zeros(122, dtype=np.dtype([
+    ('wavelength', 'float64'),
+    ('spectral_irradiance_et', 'float64'),
+    ('water_vapor_absorption', 'float64'),
+    ('ozone_absorption', 'float64'),
+    ('mixed_absorption', 'float64'),
+]))
```
adriesse, line 17: There seems to be little benefit to putting the constants into a structured array.  How about 5 constant column vectors like this?

```
WAVELENGTH = np.transpose(np.array([[
                   300.0, 305..0 etc.
]]))
```
A side benefit is that one can recognize the use of the constants in the code more easily because they are CAPS.

pvlib/spectrum.py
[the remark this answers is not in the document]
cwhanse: maybe someone at NREL would know if the C code or the Excel version is in some sense authoritative. I agree with deferring any kind of switch.

wholmgren (approved):

pvlib/spectrum/spectrl2.py
```diff
@@ -0,0 +1,384 @@
...
+    # spectrum of direct irradiance, Eq 2-1
+    Id = spectrum_et_adj * Tr * Ta * Tw * To * Tu
+
+    cosZ = cosd(apparent_zenith)
+    # Eq 3-17
+    Cs = np.where(wavelength <= 450, ((wavelength + 550)/1000)**1.8, 1.0)
+    ALG = np.log(1 - aerosol_asymmetry_factor)  # Eq 3-14
+    BFS = ALG * (0.0783 + ALG * (-0.3824 - ALG * 0.5874))  # Eq 3-13
+    AFS = ALG * (1.459 + ALG * (0.1595 + ALG * 0.4129))  # Eq 3-12
+    cosZ = cosd(apparent_zenith)
```
cwhanse, line 320: duplicate line

pvlib/spectrum/spectrl2.py
```diff
@@ -0,0 +1,384 @@
...
+    sky_reflectivity = (
+        # Top * Twp * Taap * (0.5 * (1-Trp) + (1-Fsp) * Trp * (1-Tasp))
+        Tup * Twp * Taap * (0.5 * (1-Trp) + (1-Fsp) * Trp * (1-Tasp))
+    )  # Eq 3-8
+
+    # a common factor for 3-5 and 3-6
+    common_factor = spectrum_et_adj * cosZ * To * Tu * Tw * Taa
+    # Note: spectrl2_2.c differs from the report in how the Cs value is used.
+    # The two commented out lines match the report, while the following match
+    # spectrl2_2.c:
```
cwhanse, line 343:
```suggestion
    # spectrl2_2.c. With regard to Cs, the equations in the report and spectrl12_2.c are algebraically equivalent.
```

cwhanse (approved):
Nice work @kanderso-nrel easily traceable to the reference and well-organized.

kanderso-nrel:
FYI I updated the comparison notebook using the current HEAD of this branch.  No significant changes from the original plots at the top of this thread.  https://gist.github.com/kanderso-nrel/0dc7e7ebff7b5a9c13b52f8b48eb71ee

adriesse:
What's a good and convenient source for precipitable_water, ozone, aerosol_turbidity_500nm data? Preferably multiple values per day, USA and/or worldwide.

mikofski:
> What's a good and convenient source for precipitable_water, ozone, aerosol_turbidity_500nm data? Preferably multiple values per day, USA and/or worldwide.

ECMWF, try [get_ecmwf_macc](https://pvlib-python.readthedocs.io/en/stable/generated/pvlib.iotools.get_ecmwf_macc.html)

adriesse:
I will!  There's a lot of stuff in iotools.  Where have I been?

wholmgren:
> ECMWF, try get_ecmwf_macc

plus [angstrom_aod_at_lambda](https://pvlib-python.readthedocs.io/en/stable/generated/pvlib.atmosphere.angstrom_aod_at_lambda.html#pvlib.atmosphere.angstrom_aod_at_lambda)

Are we ready to merge this?

kanderso-nrel:
> There seems to be little benefit to putting the constants into a structured array. How about 5 constant column vectors like this?

I never responded to @adriesse's last suggestion -- I had the vague notion that someone might appreciate the constants already being in tabular form so that they can easily export it for another purpose.  Other than that, you're right that columns only get used individually, so not much need for a structured array.  For this and using `np.outer` instead of implicit broadcasting, I don't really feel strongly either way and have a mild preference for wrapping this up, so +1 to merge from me.  But happy to make continue making changes if anyone else feels differently.

wholmgren:
I liked the structured array when I first read the code, but I also see @adriesse's point. I don't think there's a wrong choice here so I'm going to go ahead and merge. Thanks all!

Issue #1064, opened 2020-09-23
Third option to remove low solar elevation artifact in clearsky.ineichen


martinherrerias:
See #435 and #459 for context.

One way to remove the artifacts while introducing less bias is to clip the `airmass_absolute` values to the point where the exponential correction becomes problematic. Something like:

`airmass_critical = (cg2*(fh1+fh2*(tl-1))/0.018).^1.25`
`airmass_clipped = np.fmin(airmass_absolute, airmass_critical)`
`ghi = (np.exp(-cg2*airmass_clipped*(fh1 + fh2*(tl - 1)) + 0.01*airmass_clipped**1.8))`
`...`
`ghi = cg1 * dni_extra * cos_zenith * tl / tl * np.fmax(ghi, 0)`

`airmass_critical` is a minimum for the combined exponential terms, which shoot up afterwards. Loosely:
`airmass_critical = argmin(f(x)), for f = exp(-cg2*x*(fh1 + fh2*(tl - 1)) + 0.01*x^1.8))` 

This threshold starts at AM = 2.6 (for turbidity = 1 and altitude = 0), equivalent to ~22.5° solar elevation, and goes up to O(10) for higher turbidities and altitudes. For most cases, it corresponds to solar elevation angles below the 4° cutting threshold of the CIE 1994 quality control guidelines that Ineichen & Perez seem to have used, i.e. they probably didn't see beyond it.

Clipping at the minimum makes the adjustment smooth (C1), and leaves the results of the model untouched for most operating conditions. 

![clearsky_ineichen](https://user-images.githubusercontent.com/7872449/93875695-c0bbfb00-fcd5-11ea-8fb3-b512fb1228ab.png)

_Originally posted by @martinherrerias in https://github.com/pvlib/pvlib-python/issues/435#issuecomment-696662259_

martinherrerias:
@mikofski, there is no reference for this fix, and I also had no plans of publishing it. I take comfort in the fact that the proposed correction falls between the two (already published) versions of the model, but @adriesse is right to point out that actually none of the models is validated for very low solar elevations.

PR #1066, opened 2020-09-23
Issue #1064: use (clipped-air-mass) Perez enhancement factor...

martinherrerias:
... in clearsky.ineichen, allowing explicit enable/disable of original correction.

The clipped correction is set as default, while preserving the functionality of calls with explicit use of ..., perez_enhancement=True/False/0/1

I haven't run the full test suite, or edited any form of documentation.

### DIFF
--- a/pvlib/clearsky.py
+++ b/pvlib/clearsky.py
@@ -14,7 +14,7 @@ from pvlib import atmosphere, tools
 
 
 def ineichen(apparent_zenith, airmass_absolute, linke_turbidity,
-             altitude=0, dni_extra=1364., perez_enhancement=False):
+             altitude=0, dni_extra=1364., perez_enhancement=None):
     '''
     Determine clear sky GHI, DNI, and DHI from Ineichen/Perez model.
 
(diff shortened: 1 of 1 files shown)


pvlib/clearsky.py
```diff
@@ -14,7 +14,7 @@
 
 
 def ineichen(apparent_zenith, airmass_absolute, linke_turbidity,
-             altitude=0, dni_extra=1364., perez_enhancement=False):
+             altitude=0, dni_extra=1364., perez_enhancement=None):
```
cwhanse, line 17: IMO I don't see broad value in having three options:

1. no Perez term
2. Perez term without air mass clipping
3. Perez term with air mass clipping

I'm OK dropping the second.

wholmgren:
Thanks @martinherrerias for your efforts!

I think we need some performance validation against more recent clear sky models before we should accept further modifications to a published algorithm.

cwhanse:
@wholmgren what about an option to clip the air mass as @martinherrerias suggests? Default to no clipping.

wholmgren:
The part of me that wants pvlib to be pragmatic and distribute the best code and algorithms as possible says we should change `perez_enhancement=False|True` to `perez_enhancement='none'|'perez'|'herrerias'`.

But the part of me that wants pvlib to remain true to the source material feels like this is a non-trivial modification to a widely used and cited algorithm. We could consider a new function. The `perez_enhancement` kwarg was developed after consultation with one of the developers of the algorithm, so it seems more justifiable to keep that modification with the original function.

If I read a paper that said "we used pvlib pythons `ineichen` function to calculate clear sky" then I'd want to ask a) where did you get turbidities from? b) did you use the "perez enhancement"? and now c) did you use the "herrerias modification to the perez enhancement"? Feels messy.

adriesse:
Is the clipping published?  What problems does not clipping cause?

PR #1068, opened 2020-09-25
WIP: PVSystem attributes as iterables

### LINKED ISSUE #1067: Multiple arrays in a PV system
I'm proposing to add capability to model more than one array in a PV system. Defining the term "array", I mean a collection of PV modules at the same orientation (tilt and azimuth). Many actual PV systems have several arrays, distinguished by different orientation and/or number of modules and strings of modules.

The current `PVSystem` defines the DC side of a PV system using `modules_per_string` and `strings_per_inverter` and applies attributes `surface_tilt` and `surface_azimuth` to all modules. A PV system may have a single inverter. This model implicitly describes a PV system with a single array.

@wholmgren have discussed this a bit offline and think a way forward would be to make `PVSystem` attributes iterable. For example, `surface_tilt` could contain a list or array, with each element being an array of tilts to be applied to one of the arrays comprising the PV system. `PVSystem` and `ModelChain` methods would be extended to perform calculations for each array in the PV system, broadcasting where needed.

Two principles guiding this proposal:

1. maintain the current interface for modeling systems with a single array. It is simple but effective, and modeling a single array system shouldn't become more complicated.
2. methods should return the type indicated by the input attributes, so that a user doesn't need to deal with lists of length 1 when modeling a single array system.

We discussed, and disfavor other approaches:
- a new class for PV systems with multiple arrays, which not appealing when considering the duplication of code, and the likely  complication of `ModelChain` methods.
- subclassing `PVsystem` or adding attributes, either of which look to be confusing for users.

Open for comments here, before I start a branch to work on this feature.

#457 is relevant. Having multiple arrays lays the groundwork for modeling inverters with multiple MPPTs.

cwhanse:
- [ ] Closes #1067 
 - [ ] Tests added

Notes:
 - changes the signature of PVSystem.sapm_effective_irradiance to accept Dataframe instead of two Series. Up for discussion.
 - need to back out the commits for the Ross temperature model.

To do:

- [ ] PVSystem methods to allow surface_tilt and surface_azimuth to be iterable, representing more than one array. 
- [ ] PVSystem methods to allow modules_per_string and strings_per_inverter to be iterable.
- [ ] Changes to PVSystem to allow for multiple inverters
- [ ] Tests representing fixed-tilt systems with multiple arrays
- [ ] Changes to SingleAxisTracker to implement multiple arrays.


wholmgren:
> changes the signature of PVSystem.get_irradiance to accept Dataframe instead of two Series. Up for discussion.

What's the idea here? The different arrays have different illumination? 

It's hard to tell if you're committed to the current implementation with loops/list comprehensions or if it's still experimentation. If so, let's discuss before you continue with it. I'd like to think there are better options along the lines of @mikofski's https://github.com/pvlib/pvlib-python/issues/1067#issuecomment-698780275

cwhanse:
Ack. Mistake in my comment: the change to Dataframe is in `PVSystem.sapm_effective_irradiance` not in `PVSystem.get_irradiance`.

Two arrays at different orientation have different plane-of-array irradiance. The upstream method `PVSystem.get_irradiance` calculates plane-of-array components by calling `pvlib.irradiance.get_total_irradiance` which returns a Dataframe. So in this code's pattern, `PVSystem.get_irradiance` returns a list of Dataframes. 

The downstream method `PVSystem.sapm_effective_irradiance` currently requests 'poa_direct' and 'poa_diffuse' as separate Series. Rather than unpack the list of Dataframes to create two lists of Series (could be done), I thought putting 'poa_direct' and 'poa_diffuse' together in a Dataframe was simpler.

I'm not committed to this pattern, putting some code that runs out for discussion.  And I think this code is duck-ish. The `try - except TypeError` pattern handles the case where the attribute is iterable (try) or not (except TypeError).  One place where that doesn't work is PVSystem.calcparams_desoto, where for a single array `effective_irradiance` will be a Series.

cwhanse:
Closed in favor of #1076

PR #1074, opened 2020-10-05
Refactor detect_clearsky

cwhanse:
 - [x] Tests added

Refactor of `clearsky.detect_clearsky` in preparation for adding BrightSun detect clearsky algorithm. Major changes:
- uses pd.Series.rolling instead of Hankel matrix to compute criteria over sliding windows. This change accomodates centered windows, i.e., aligning the criteria for a window to a midpoint rather than to the left endpoint (default behavior in this code).
- calculation of criteria delegated to helper functions.
- new kwarg `align` (default 'left') otherwise function signature is unchanged.


cwhanse:
if `clearsky.detect_clearsky` is of interest please take a look, in particular, at `_calc_stats` and the other helper functions.

I need to add some tests for the non-trivial helper functions, and I'm sure there's room for performance improvement. The `align` feature is added anticipating the BrightSun method which uses centered windows. I think that the BrightSun method could be implemented using numpy arrays and the Hankel matrix, but using pandas.Series.rolling seems more transparent to me.

We could drop `align='right'` as an option (the current code's behavior is `align='left'` but `pandas.rolling` only offers `center=False (default, means right-aligned) and `center=True` (centered). Maintaining the current behavior with `align='left'` by default wasn't too hard. I'm not sure how much value the `align` options offer, but there they are. `align='center'` is more in line (sorry) with what Matt and I had in mind when we wrote the algorithm, but we didn't code it that way in Matlab.

wholmgren:
Is there a compelling use case for `align='left'`? It sounds to me like we should consider it a bug in the initial implementation, and always use `center=True` (without exposing a kwarg). 

I will carefully review when I have the time. I want to understand why this implementation works but the previous attempts to use pandas did not work.

cwhanse:
`align='left'` is here to preserve the old behavior, personally I agree with `align='center'` as the better choice for a default.

I also think we could drop support for `array` inputs and lose the `times` argument.

wholmgren:
To put it differently: why should we preserve the old behavior? It sounds like a bug. People that want the old behavior can use an old, buggier version of pvlib.

> I also think we could drop support for array inputs and lose the times argument.

agree

cwhanse:
I guess I'm unwilling to call the left alignment a bug - it's not wrong, rather, it's an undesirable behavior. I'm OK changing the default to 'center'. And open to not offering 'left', 'right' as options.

wholmgren:
Other than consistency with previous versions, when might a user prefer 'left' or 'right'? Without a compelling reason, I'd hard-code center and then let users ask/contribute if they come up with a good reason for a kwarg.

cwhanse:
> Other than consistency with previous versions, when might a user prefer 'left' or 'right'? 

Consistency is the only reason I came up with. The option to align 'left', 'center', 'right' falls out of the combination of numpy and pandas defaults.

> I'd hard-code center 

I'm OK with this.

pvlib/clearsky.py
```diff
@@ -597,6 +599,264 @@ def _calc_d(aod700, p):
...
+
+def _get_sample_intervals(times, window_length):
+    """ Calculates timeinterval and samples per window for Reno-style clear
+    sky detection functions
+    """
+    window = pd.Timedelta(minutes=window_length)
+    temp = pd.Series(data=1, index=times)
+
+    # determine if we can proceed
+    if times.inferred_freq and len(temp.unique()) == 1:
```
kanderso-nrel, line 810: This condition is opaque to me.  Won't `temp.unique()` always be `array([1])` if `times` is not empty?

pvlib/clearsky.py
```diff
@@ -597,6 +599,264 @@ def _calc_d(aod700, p):
...
+        sample_interval = times[1] - times[0]
+        sample_interval = sample_interval.seconds / 60  # interval in minutes
```
kanderso-nrel, line 812: Is it worth using `.total_seconds()` for sub-second precision here?  The `seconds` attribute doesn't consider fractions of a second, for example with `times = pd.to_datetime(['2019-01-01 00:00:00 000000', '2019-01-01 00:01:00 500000'], format='%Y-%m-%d %H:%M:%S %f')`

pvlib/clearsky.py
```diff
@@ -597,6 +599,264 @@ def _calc_d(aod700, p):
...
+    # if align == 'right':
+    #     shift = 1 - samples_per_window
+    # elif align == 'center':
+    #     shift = - (samples_per_window // 2)
+    # else:
+    #     shift = 0
+    shift = -(samples_per_window // 2)
+    idx = clear_windows.shift(shift)
+    # drop rows at the end corresponding to windows past the end of data
+    idx.drop(clear_windows.tail(samples_per_window - 1).index, inplace=True)
```
kanderso-nrel, line 854: Is the argument here equivalent to `clear_windows.index[samples_per_window-1:]`?

pvlib/clearsky.py
```diff
@@ -782,147 +1029,13 @@ def rmse(alpha):
...
-        components['slope_max'] = (np.max(
-            meas_slope - alpha * clear_slope, axis=0))
-
+        components['mean_diff'] = c1
+        components['max_diff'] = c2
+        components['line_length'] = c3
+        components['slope_nstd'] = c4
+        components['slope_max'] = c5
+        components['mean_nan'] = c6
+#        components['windows'] = clear_windows
```
kanderso-nrel, line 1038: If some keys are disappearing from this dict we should consider mentioning it in the whatsnew

pvlib/tests/test_clearsky.py
```diff
@@ -611,8 +611,62 @@ def test_detect_clearsky_irregular_times(detect_clearsky_data):
...
+                                      sample_interval=1, align=align)
+        res_mean, res_max, res_line_length, res_slope_nstd, res_slope = result
+        assert_series_equal(res_mean, expected[align]['mean'])
+        assert_series_equal(res_max, expected[align]['max'])
+        assert_series_equal(res_line_length, expected[align]['line_length'])
+        assert_series_equal(res_slope_nstd, expected[align]['slope_nstd'])
+        assert_series_equal(res_slope, expected[align]['slope'])
+
+
+def test_calc_c5():
```
kanderso-nrel, line 657: Nit, double underscore for private function: `test__calc_c5`

kanderso-nrel (commented):
a partial review

pvlib/clearsky.py
```diff
@@ -597,6 +599,264 @@ def _calc_d(aod700, p):
...
+    #     shift = 1 - window
+    # else:
+    #     shift = 0
+    shift = 0
+    return shift
+
+
+def _count_in_window(ts, window):
+    # return int type instead of default int64, to avoid error in pandas
+    return int(ts.resample(window).agg('sum')[0])
```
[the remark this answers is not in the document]
cwhanse: Thanks, I forgot that pandas resample assumes an alignment which isn't always what's expected.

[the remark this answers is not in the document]
cwhanse: Reverting back to the current calculation, I think I was working towards accomodating unequal timestamp intervals.

pvlib/clearsky.py
cwhanse: yes....a mistake

pvlib/clearsky.py
cwhanse: Equivalent to `clear_windows.index[1 - samples_per_window:]` is there a reason to prefer this over `tail`?

pvlib/clearsky.py
cwhanse: That is unintentional, at one point I was thinking that `clear_windows` would go away, but it isn't. I'll include it in `components`.

pvlib/clearsky.py
cwhanse: One would hope that `seconds` and `total_seconds()` are interchangeable when the data are on the minute. But they are not.
```
import pandas as pd
idx = pd.date_range(start="2012-04-01 10:30:00-07:00",
                    end="2012-04-01 10:31:00-07:00",
                    freq="1T")
td = idx[1] - idx[0]
print(td.seconds)
print(td.total_seconds())
```
That 1ns roundoff messes up the calculation of `samples_per_window`

kanderso-nrel: I don't think I follow about the 1ns roundoff.  I tried your snippet on a few pandas versions and always got this output:
```
60
60.0
```

That said, probably not worth putting any more effort into splitting this hair.

cwhanse: pandas 0.22.0, python3.6, sorry for not including that in my earlier reply.

cwhanse:
> To put it differently: why should we preserve the old behavior? It sounds like a bug. People that want the old behavior can use an old, buggier version of pvlib.
> 
> > I also think we could drop support for array inputs and lose the times argument.
> 
> agree

@wholmgren prefer to deprecate removing the `times` argument and restricting input type to Series?

wholmgren:
Ok with me but how do we gracefully deprecate a required argument from the middle of a signature?

cwhanse:
> > @wholmgren prefer to deprecate removing the `times` argument and restricting input type to Series?
> 
> Ok with me but how do we gracefully deprecate a required argument from the middle of a signature? I think in the past we've only done this kind of change when moving a function to a new name or module.


"gracefully" idk, maybe accept array types and the `times` argument, and use the `times` argument but raise a warning if it is supplied.

Or release it as a breaking API change.

My guess is that the array-type users are in the minority and the Series-type users would appreciate losing the `times` parameter.

wholmgren:
I'm not opposed to delaying this to 0.9 and making it a breaking change. We can do a non-breaking change if we change both `times` and `window_length` to kwargs. Maybe someone else has a better idea.

cwhanse:
Let's leave the parameters unchanged. In v0.9 I'd want to rename this function to `detect_clearsky_reno` to make room for `detect_clearsky_brightsun`. We can adjust the parameters in the renamed function.

pvlib/clearsky.py
kanderso-nrel: A bit more efficient (py3.7, pandas 0.25.1) but I won't raise a fuss about it.
```python
s = pd.Series(range(8760), index=pd.date_range('2019-01-01', freq='h', periods=8760))

%timeit _ = s.tail(10).index
241 µs ± 10.3 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit _ = s.index[-10:]
134 µs ± 4.23 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

kanderso-nrel (approved):
LGTM, a couple suggestions here and below:
- Consider dropping the newly unnecessary `times` argument on line 643 in `clearsky.rst`. 
- I notice that the reference performed the rescaling "using only clear periods on days when greater than 50% of the day was classified as clear".  Just pointing it out in case it's worth doing the same here or documenting the difference.

pvlib/clearsky.py
```diff
@@ -782,147 +1034,13 @@ def rmse(alpha):
...
-        components['line_length_flag'] = c3
-        components['slope_nstd_flag'] = c4
-        components['slope_max_flag'] = c5
-        components['mean_nan_flag'] = c6
+        components['mean_diff'] = c1
+        components['max_diff'] = c2
+        components['line_length'] = c3
+        components['slope_nstd'] = c4
+        components['slope_max'] = c5
+        components['mean_nan'] = c6
```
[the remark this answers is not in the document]
cwhanse: This may have been unintentional. In the current PR, only the Boolean series for each criteria are returned. In v0.8, the Boolean series are returned in the '_flag' columns, and the values for each criteria are returned in the columns named for each criteria e.g. 'line_length'

I'm happy to put it back as in v0.8, if that's agreed.

cwhanse:
> * Consider dropping the newly unnecessary `times` argument on line 643 in `clearsky.rst`.

We decided not to drop the 'times' argument in this refactor, but rather to rename this function in v0.9 to `detect_clearsky_reno` and drop the 'times' argument then.

> * I notice that the reference performed the rescaling "using only clear periods on days when greater than 50% of the day was classified as clear".  Just pointing it out in case it's worth doing the same here or documenting the difference.

Adding a comment to this effect.

kanderso-nrel:
> 
> 
> > * Consider dropping the newly unnecessary `times` argument on line 643 in `clearsky.rst`.
> 
> We decided not to drop the 'times' argument in this refactor, but rather to rename this function in v0.9 to `detect_clearsky_reno` and drop the 'times' argument then.

I'm on board with that plan for 0.9 -- to clarify, this suggestion was about this [documentation page](https://pvlib-python.readthedocs.io/en/stable/clearsky.html#detect-clearsky) where the inputs are Series and passing `cs.index` isn't needed anymore.  OK with me to just leave it until 0.9 when the parameter goes away for good.

pvlib/clearsky.py
```diff
@@ -597,7 +599,258 @@ def _calc_d(aod700, p):
     return d
 
 
-def detect_clearsky(measured, clearsky, times, window_length,
+def bird(zenith, airmass_relative, aod380, aod500, precipitable_water,
```
wholmgren, line 602: I'd prefer to see this function stay at the end of the module for this PR. The changes in this PR are not trivial so it helps to minimize the number of lines changed. I'm fine with a follow up PR to reorganize the module to your liking.

pvlib/clearsky.py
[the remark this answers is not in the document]
wholmgren: -1 on the API change for 0.8

pvlib/clearsky.py
```diff
@@ -597,7 +599,132 @@ def _calc_d(aod700, p):
...
+    #     shift = 1 - samples_per_window
+    # elif align == 'center':
+    #     shift = - (samples_per_window // 2)
+    # else:
+    #     shift = 0
+    shift = -(samples_per_window // 2)
+    idx = clear_windows.shift(shift)
+    # drop rows at the end corresponding to windows past the end of data
+    idx.drop(clear_windows.index[1 - samples_per_window:], inplace=True)
+    idx = idx.astype(bool)  # shift added nan, changed type to object
```
wholmgren, line 722: Seems like pandas makes this more complicated instead of easier. Is this equivalent to the 3 lines above?

```python
idx = clear_windows.iloc[shift:1-samples_per_window]
```

pvlib/clearsky.py
```diff
@@ -750,20 +883,22 @@ def detect_clearsky(measured, clearsky, times, window_length,
         c2 = np.abs(meas_max - alpha*clear_max) < max_diff
         c3 = (line_diff > lower_line_length) & (line_diff < upper_line_length)
         c4 = meas_slope_nstd < var_diff
-        c5 = np.max(np.abs(meas_slope -
-                           alpha * clear_slope), axis=0) < slope_dev
+        c5 = _calc_slope_max_diff(
+            meas, alpha * clear, samples_per_window, 'center') < slope_dev
```
wholmgren, line 887: nit: `alpha * clear` is computed here and on line 877, so use `scaled_clear = alpha * clear` or similar at the top of the loop

pvlib/clearsky.py
cwhanse: No, because the index isn't shifted. Since window statistics are aligned to center by default, the 'shift' changes alignment to 'left' so that the clear index values can be extracted with the Hankel matrix.  I don't recall the circumstances which changed the type to object from bool, but it came up.

wholmgren: I assumed the line below dropped the index, but fine to keep it this way if not.

pvlib/clearsky.py
```diff
@@ -692,56 +827,54 @@ def detect_clearsky(measured, clearsky, times, window_length,
...
+    # factor considers all identified clear times, which is different from [1]
+    # where the scaling factor was determined from clear times on days with
+    # at least 50% of the day being identified as clear.
     alpha = 1
     for iteration in range(max_iterations):
-        clear_line_length = np.sum(np.sqrt(
-            alpha * alpha * clear_diff * clear_diff +
-            sample_interval * sample_interval), axis=0)
+        _, _, clear_line_length, _, _ = _calc_stats(
+            alpha * clear, samples_per_window, sample_interval, 'center')
```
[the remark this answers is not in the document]
cwhanse: The bottlenecks in `_calc_stats` are the two uses of `.apply` which operate custom functions on each window. I could factor the line length calculation out to it's own helper function since setting up the roller isn't the bottleneck. That would avoid calling `.apply` twice in the alpha loop.

wholmgren:
https://github.com/cwhanse/pvlib-python/pull/6 adds an asv test for this function. Results on my machine for the latest commit 5927d73 using `asv run -b 'detect_clearsky.DetectClear.time_detect_clearsky' HEAD`

```
[  0.14%] · For pvlib-python commit 5927d738 <pr/cwhanse/1074>:
[  0.14%] ·· Building for conda-py3.6-ephem3.7.6.0-numba0.36.1-numpy1.12.0-pandas0.22.0-pytables3.6.1-scipy1.2.0....
[  0.14%] ·· Benchmarking conda-py3.6-ephem3.7.6.0-numba0.36.1-numpy1.12.0-pandas0.22.0-pytables3.6.1-scipy1.2.0
[  0.17%] ··· Running (detect_clearsky.DetectClear.time_detect_clearsky--).
[  0.21%] ··· detect_clearsky.DetectClear.time_detect_clearsky                                                                                                                                                                      1.27±0.02s
[  0.21%] ·· Building for conda-py3.8-ephem-numba-numpy-pandas-pytables-scipy....
[  0.21%] ·· Benchmarking conda-py3.8-ephem-numba-numpy-pandas-pytables-scipy
[  0.24%] ··· Running (detect_clearsky.DetectClear.time_detect_clearsky--).
[  0.28%] ··· detect_clearsky.DetectClear.time_detect_clearsky                                                                                                                                                                      13.5±0.03s
```

disconcerting result for the python 3.8 environment.

cwhanse:
thanks @wholmgren  I profiled the code and nearly all the time is being spent in those two '.apply' methods. That's the good news, I know where to focus. I don't think this should be merged with such a drastic loss of performance.

kanderso-nrel:
I would guess that the order of magnitude slowdown between the two asv environments is from pandas introducing the optional `raw` parameter in v0.23.0, making `Rolling.apply()` switch from passing arrays to passing Series by default.

wholmgren:
> I would guess that the order of magnitude slowdown between the two asv environments is from pandas introducing the optional raw parameter in v0.23.0, making Rolling.apply() switch from passing arrays to passing Series by default.

That's a good guess. If true, we should bump the min pandas requirement so that we can pass `raw=True` without concern for the version. 

@cwhanse I'm fine with holding up 0.8.1 for a little bit while you experiment. I just want it released by the time we'd want to merge the Array PR.

cwhanse:
I looked at passing 'raw' (for later pandas versions) and didn't see a speedup. I could have done it wrong. I think the issue is that if the function being applied isn't in the set of precompiled methods, that `.apply` loops through the windows.  This changes from using `pd.rolling.apply` to `np.apply_along_axis` which is ~x10 faster.

cwhanse:
@wholmgren I'm not able to run the asv benchmark, running into some arcane bug in Windows 10. Could you check on the performance of the latest version?

wholmgren:
```
asv compare 6e936f5 HEAD                                                                                                                                    1 ↵

All benchmarks:

       before           after         ratio
     [6e936f57]       [c3c1f9e3]
     <pr/cwhanse/1074~5^2>       <detect_clearsky>
+      14.5±0.3ms       2.26±0.04s   155.72  detect_clearsky.DetectClear.time_detect_clearsky [holmgren-mbp.local/conda-py3.6-ephem3.7.6.0-numba0.36.1-numpy1.12.0-pandas0.22.0-pytables3.6.1-scipy1.2.0]
+      16.1±0.4ms       2.30±0.06s   142.84  detect_clearsky.DetectClear.time_detect_clearsky [holmgren-mbp.local/conda-py3.8-ephem-numba-numpy-pandas-pytables-scipy]
```

cwhanse:
back to work on it I guess.

cwhanse:
@wholmgren this looks better on the profiler

wholmgren:
```
asv compare 6e936f5 HEAD

All benchmarks:

       before           after         ratio
     [6e936f57]       [983bd6f8]
     <pr/cwhanse/1074~5^2>       <detect_clearsky>
+      14.5±0.3ms       55.6±0.8ms     3.83  detect_clearsky.DetectClear.time_detect_clearsky [holmgren-mbp.local/conda-py3.6-ephem3.7.6.0-numba0.36.1-numpy1.12.0-pandas0.22.0-pytables3.6.1-scipy1.2.0]
+      16.1±0.4ms         54.2±2ms     3.36  detect_clearsky.DetectClear.time_detect_clearsky [holmgren-mbp.local/conda-py3.8-ephem-numba-numpy-pandas-pytables-scipy]
```

I'm willing to accept this performance decrease if we think it's a net win for readability and ease of reuse in future algorithms.

cwhanse:
@wholmgren I removed the use of `pandas.roller`, it is a performance hit. I don't think this is back to the speed of the original because it uses `Series` as the internal type rather than `array`. It could go back to `array` but that would take some work to maintain the centered interval alignments.

pvlib/clearsky.py
```diff
@@ -773,14 +897,15 @@ def rmse(alpha):
...
-        warnings.warn('failed to converge after %s iterations'
+        warnings.warn('rescaling failed to converge after %s iterations'
                       % max_iterations, RuntimeWarning)
 
     # be polite about returning the same type as was input
     if ispandas:
         clear_samples = pd.Series(clear_samples, index=times)
 
     if return_components:
+        components = OrderedDict()
```
wholmgren, line 908:
```suggestion
```

wholmgren (commented):
```
asv compare pvlib/master cwhanse/detect_clearsky

All benchmarks:

       before           after         ratio
     [926d2f95]       [e53572d9]
                      <detect_clearsky>
+      14.7±0.4ms       54.0±0.5ms     3.66  detect_clearsky.DetectClear.time_detect_clearsky [holmgren-mbp.local/conda-py3.6-ephem3.7.6.0-numba0.36.1-numpy1.12.0-pandas0.22.0-pytables3.6.1-scipy1.2.0]
+      15.1±0.1ms         52.4±1ms     3.47  detect_clearsky.DetectClear.time_detect_clearsky [holmgren-mbp.local/conda-py3.8-ephem-numba-numpy-pandas-pytables-scipy]
```

wholmgren:
@cwhanse I took the liberty of parameterizing the test in the latest commit. Results below with `ndays` in parenthesis.  

```
asv compare pvlib/master cwhanse/detect_clearsky

All benchmarks:

       before           after         ratio
     [926d2f95]       [e53572d9]
                      <detect_clearsky>
+     2.08±0.06ms       36.2±0.8ms    17.45  detect_clearsky.DetectClear.time_detect_clearsky(1) [holmgren-mbp.local/conda-py3.6-ephem3.7.6.0-numba0.36.1-numpy1.12.0-pandas0.22.0-pytables3.6.1-scipy1.2.0]
+     2.63±0.09ms       32.2±0.6ms    12.27  detect_clearsky.DetectClear.time_detect_clearsky(1) [holmgren-mbp.local/conda-py3.8-ephem-numba-numpy-pandas-pytables-scipy]
+      14.2±0.4ms       55.1±0.6ms     3.87  detect_clearsky.DetectClear.time_detect_clearsky(10) [holmgren-mbp.local/conda-py3.6-ephem3.7.6.0-numba0.36.1-numpy1.12.0-pandas0.22.0-pytables3.6.1-scipy1.2.0]
+      15.2±0.2ms       52.0±0.6ms     3.41  detect_clearsky.DetectClear.time_detect_clearsky(10) [holmgren-mbp.local/conda-py3.8-ephem-numba-numpy-pandas-pytables-scipy]
+         165±1ms          231±2ms     1.40  detect_clearsky.DetectClear.time_detect_clearsky(100) [holmgren-mbp.local/conda-py3.6-ephem3.7.6.0-numba0.36.1-numpy1.12.0-pandas0.22.0-pytables3.6.1-scipy1.2.0]
+         164±2ms          222±1ms     1.35  detect_clearsky.DetectClear.time_detect_clearsky(100) [holmgren-mbp.local/conda-py3.8-ephem-numba-numpy-pandas-pytables-scipy]
```

wholmgren (approved):
Ready to merge?

cwhanse:
> Ready to merge?

Yes for me. @kanderso-nrel ?

pvlib/tests/test_clearsky.py
```diff
@@ -569,19 +579,19 @@ def test_detect_clearsky_iterations(detect_clearsky_data):
     alpha = 1.0448
     with pytest.warns(RuntimeWarning):
         clear_samples = clearsky.detect_clearsky(
-            expected['GHI'], cs['ghi']*alpha, cs.index, 10, max_iterations=1)
-    assert (clear_samples[:'2012-04-01 10:41:00'] == True).all()
-    assert (clear_samples['2012-04-01 10:42:00':] == False).all()
```
kanderso-nrel, line 574: De Morgan wonders if this should use `.any()` instead of `.all()`?

kanderso-nrel (approved):
I didn't check calculation details this time but LGTM.  One small question below.

pvlib/tests/test_clearsky.py
kanderso-nrel: Oops, this comment is about the line `    assert not clear_samples['2012-04-01 10:42:00':].all()  # expected False`

cwhanse: You are correct, should be `.any()`

Issue #1069, opened 2020-09-25
Applied PVsyst-type losses in ModelChain


ncroft-b4:
I would like to start a discussion around developing features that implement PVsyst style losses that act as direct physical operators rather than as percentage factors on just the dc section of the `ModelChain`. For instance, implementing soiling losses as monthly percentage inputs that operate on the irradiance input leading to effective irradiance, DC ohmic losses as a percentage at STC loss from which a resistance value is calculated and operates on the current of the DC output, and/or AC ohmic Losses as a STC loss percentage that operates on the output of the ac_model. 

It is my understanding that loss models in the ModelChain are currently implemented as a single operation after the `dc_model` has been run and before the `ac_model` and only affect the dc outputs. Therefore "pvsyst_losses" functionality would not be an additional loss model at the `losses_model` part of the chain, but would require extra steps added to the chain. 

I wanted to bring this up and open it for discussion about preferred ways to implement this. For instance , a soiling_loss function could be added that executes before the `aoi_model.` This potentially depreciates the losses_model as a single self-contained function.

I suppose what I am suggesting is to remove the losses_model in favor of new functions like soiling_losses, dc_losses, ac_losses. What to run for each of these could then be defined per model strategy whether it is pvwatts, sapm, or pvsyst, etc,. If this strategy seems valid I would like to attempt a pull request to begin implementing it.

cwhanse:
The only legacy interface for loss models is in `ModelChain.losses_model`, which sits between `ModelChain.dc_model` and `ModelChain.ac_model` in the various `run_model` methods. The current options for `losses_model` are `PVSystem.pvwatts_losses` and `ModelChain.no_extra_losses`.  We could rename `losses_model` to something more specific `dc_losses_model` perhaps, and preserve the PVWatts function at that stage. That creates space to name and insert other loss model methods, e.g., `poa_losses` for soiling and snow.

ncroft-b4:
Or perhaps instead of renaming `losses_model`, it could be kept as a `ModelChain` attribute where a user could specify it to be either 'pvwatts' or 'pvsyst'? Then instead of the `ModelChain.losses_model` being a function, it is a label that gets used to decide which functions to run at various losses methods.

And if it is 'pvsyst' then there would be 'pvsyst' specific keys in the 'losses_parameters' that are used by `PVSystem`.

[...]

PR #1084, opened 2020-10-26
Change setter for losses_model to allow it to set multiple loss funct…

ncroft-b4:
…ions, anticipating more complex loss models. Adds self._dc_losses

 - [ ] Closes #xxxx
 - [x] Tests added

### DIFF
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -328,8 +328,8 @@ class ModelChain:
         user-defined function.
 
     losses_model: str, default 'no_loss'
-        Valid strings are 'pvwatts', 'no_loss'. The loss functions will
-        be set according to this parameter.
+        Valid strings are 'pvwatts', 'no_loss', 'pvsyst'. The loss functions
+        will be set according to this parameter.
 
     name: None or str, default None
         Name of ModelChain instance.
@@ -940,6 +940,9 @@ class ModelChain:
             elif model == 'no_loss':
                 self._losses_model = model
                 self._dc_losses = self.no_extra_losses
+            elif model == 'pvsyst':
+                self._losses_model = model
+                self._dc_losses = self.pvsyst_dc_losses
             else:
                 raise ValueError(model + ' is not a valid losses model')
         else:
@@ -955,6 +958,19 @@ class ModelChain:
 
     def no_extra_losses(self):
         self.losses = 1
+        return self
+
+    def pvsyst_dc_losses(self):
+        """
+        Calculate time series of ohmic losses and apply those to the mpp power
+        output of the `dc_model` based on the pvsyst equivalent resistance
+        method. Uses a `dc_ohmic_percent` parameter in the `losses_parameters`
+        of the PVsystem.
+        """
+        Rw = self.system.pvsyst_dc_losses_eq_ohms()
+        ohmic_loss_fraction = Rw * self.dc['i_mp'] / self.dc['v_mp']
+        self.dc_ohmic_losses = ohmic_loss_fraction * self.dc['p_mp']
+        self.dc['p_mp'] = self.dc['p_mp'] - self.dc_ohmic_losses
         return self
 
     def effective_irradiance_model(self):
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -327,9 +327,9 @@ class ModelChain:
         The ModelChain instance will be passed as the first argument to a
         user-defined function.
 
-    losses_model: str, default 'no_loss'
-        Valid strings are 'pvwatts', 'no_loss'. The loss functions will
-        be set according to this parameter.
+    losses_model: str or function, default 'no_loss'
+        Valid strings are 'pvwatts', 'no_loss'. The ModelChain instance
+        will be passed as the first argument to a user-defined function.
 
     name: None or str, default None
         Name of ModelChain instance.
@@ -935,15 +935,13 @@ class ModelChain:
         elif isinstance(model, str):
             model = model.lower()
             if model == 'pvwatts':
-                self._losses_model = model
-                self._dc_losses = self.pvwatts_losses
+                self._losses_model = self.pvwatts_losses
             elif model == 'no_loss':
-                self._losses_model = model
-                self._dc_losses = self.no_extra_losses
+                self._losses_model = self.no_extra_losses
             else:
                 raise ValueError(model + ' is not a valid losses model')
         else:
-            raise ValueError('losses_model must be a string')
+            self._losses_model = partial(model, self)
 
     def infer_losses_model(self):
         raise NotImplementedError
@@ -1376,7 +1374,7 @@ class ModelChain:
         """
         self._prepare_temperature(data)
         self.dc_model()
-        self._dc_losses()
+        self.losses_model()
         self.ac_model()
 
         return self
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -328,8 +328,8 @@ class ModelChain:
         user-defined function.
 
     losses_model: str, default 'no_loss'
-        Valid strings are 'pvwatts', 'no_loss', 'pvsyst'. The loss functions
-        will be set according to this parameter.
+        Valid strings are 'pvwatts', 'no_loss'. The loss functions will
+        be set according to this parameter.
 
     name: None or str, default None
         Name of ModelChain instance.
@@ -940,9 +940,6 @@ class ModelChain:
             elif model == 'no_loss':
                 self._losses_model = model
                 self._dc_losses = self.no_extra_losses
-            elif model == 'pvsyst':
-                self._losses_model = model
-                self._dc_losses = self.pvsyst_dc_losses
             else:
                 raise ValueError(model + ' is not a valid losses model')
         else:
@@ -958,19 +955,6 @@ class ModelChain:
 
     def no_extra_losses(self):
         self.losses = 1
-        return self
-
-    def pvsyst_dc_losses(self):
-        """
-        Calculate time series of ohmic losses and apply those to the mpp power
-        output of the `dc_model` based on the pvsyst equivalent resistance
-        method. Uses a `dc_ohmic_percent` parameter in the `losses_parameters`
-        of the PVsystem.
-        """
-        Rw = self.system.pvsyst_dc_losses_eq_ohms()
-        ohmic_loss_fraction = Rw * self.dc['i_mp'] / self.dc['v_mp']
-        self.dc_ohmic_losses = ohmic_loss_fraction * self.dc['p_mp']
-        self.dc['p_mp'] = self.dc['p_mp'] - self.dc_ohmic_losses
         return self
 
     def effective_irradiance_model(self):
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -960,8 +960,7 @@ class ModelChain:
         of the PVsystem.
         """
         Rw = self.system.dc_ohms_from_percent()
-        ohmic_loss_fraction = Rw * self.dc['i_mp'] / self.dc['v_mp']
-        self.dc_ohmic_losses = ohmic_loss_fraction * self.dc['p_mp']
+        self.dc_ohmic_losses = pvsystem.dc_ohmic_losses(Rw, self.dc['i_mp'])
         self.dc['p_mp'] = self.dc['p_mp'] - self.dc_ohmic_losses
         return self
 
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -327,6 +327,11 @@ class ModelChain:
         The ModelChain instance will be passed as the first argument to a
         user-defined function.
 
+    dc_ohmic_model: None, str or function, default None
+        Valid strings are 'dc_ohms_from_percent', 'no_loss'. The ModelChain
+        instance will be passed as the first argument to a user-defined
+        function.
+
     losses_model: str or function, default 'no_loss'
         Valid strings are 'pvwatts', 'no_loss'. The ModelChain instance
         will be passed as the first argument to a user-defined function.
@@ -343,6 +348,7 @@ class ModelChain:
                  airmass_model='kastenyoung1989',
                  dc_model=None, ac_model=None, aoi_model=None,
                  spectral_model=None, temperature_model=None,
+                 dc_ohmic_model=None,
                  losses_model='no_loss', name=None, **kwargs):
 
         self.name = name
@@ -361,7 +367,10 @@ class ModelChain:
         self.spectral_model = spectral_model
         self.temperature_model = temperature_model
 
+        # losses
+        self.dc_ohmic_model = dc_ohmic_model
         self.losses_model = losses_model
+
         self.orientation_strategy = orientation_strategy
 
         self.weather = None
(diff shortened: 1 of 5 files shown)


ncroft-b4:
The strategy I used here is to use the setter on `losses_model` to setup more than one loss functions attached to `ModelChain`. To start, `self._dc_losses`. In the future, a different `losses_model` input could setup multiple loss functions, such as `self._soil_losses`, `self._ac_losses`, etc.

It might be more appropriate to have a separate setter for each loss function, instead. But I thought it makes sense to keep a single user parameter that gets passed when setting up a `ModelChain`, i.e. 'pvwatts' or 'pvsyst', instead of multiple parameters for each loss function, and so it made sense to keep the single setter for the single parameter. 

Would appreciate any comments on this direction before continuing further with possible future PRs to add on other loss functions, as discussed in #1069.

ncroft-b4:
I added in some trial functions to `pvsystem` and `modelchain` in the second and third commits that add capability for calculating dc ohmic losses from a percentage input, that input being the percentage dc loss at STC. The idea would be that other loss functions could be added in a similar fashion. My next step for this push will be to add a test in `test_pvsystem` for the equivalent resistance function.

pvlib/pvsystem.py
```diff
@@ -2334,6 +2355,35 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
     return losses
 
 
+def pvsyst_dc_losses_eq_ohms(V_mp_ref, I_mp_ref, dc_ohmic_percent=0,
```
cwhanse, line 2358: I think we can drop 'pvsyst' from the name and credit Pvsyst in a reference. Or, if we keep 'pvsyst' I'd prefer that the source (pvsyst) be towards the end of the name. Some suggestions:

eq_ohms_pvsyst (do we need to categorize the function as dc_losses?)
dc_loss_eq_ohms_pvsyst
dc_loss_pvsyst_eq_ohms (if we expect a collection of Pvsyst loss functions)

Some style/format edits are needed:
- don't capitalize the parameters V_mp_ref and I_mp_ref
- add a Returns block
- %_loss_at_stc won't display as intended. Suggest adding a symbol L_{stc} and defining it in the text.

pvlib/pvsystem.py
```diff
@@ -2334,6 +2355,35 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+                             modules_per_string=1,
+                             strings_per_inverter=1):
+    """
+    Calculates the equivalent resistance of the wires from a percent
+    ohmic loss at STC, defined by the user as an input to loss_parameters.
+    Equivalent resistance is calculated with the fucntion:
```
cwhanse, line 2364:
```suggestion
    Calculates the equivalent resistance of the wires from a percent
    ohmic loss at STC.
    
    Equivalent resistance is calculated with the fucntion:
```

pvlib/modelchain.py
```diff
@@ -955,6 +960,19 @@ def no_extra_losses(self):
         self.losses = 1
         return self
 
+    def pvsyst_dc_losses(self):
```
cwhanse, line 963: This is the only DC loss applied by pvsyst?  If no, is the plan to add to this method when other DC loss models are implemented?

cwhanse (commented):
I'm OK with the direction this PR is taking. `ModelChain.losses_model` becomes an identifier for a set of loss models. Although here only a DC loss model is added, a set can grow to include irradiance, shading, soiling, etc. The set approach seems to be a practical strategy since loss models can overlap or conflict in different implementations (e.g., shading can be modeled as a reduction in irradiance, or could be accounted for by a factor applied to DC power).

pvlib/pvsystem.py
wholmgren: `ohmic_loss_to_equivalent_resistance`?

wholmgren: shouldn't `dc_ohmic_percent` be a required parameter for the function?

pvlib/modelchain.py
```diff
@@ -955,6 +960,19 @@ def no_extra_losses(self):
...
+        """
+        Calculate time series of ohmic losses and apply those to the mpp power
+        output of the `dc_model` based on the pvsyst equivalent resistance
+        method. Uses a `dc_ohmic_percent` parameter in the `losses_parameters`
+        of the PVsystem.
+        """
+        Rw = self.system.pvsyst_dc_losses_eq_ohms()
+        ohmic_loss_fraction = Rw * self.dc['i_mp'] / self.dc['v_mp']
+        self.dc_ohmic_losses = ohmic_loss_fraction * self.dc['p_mp']
```
wholmgren, line 972: I generally want to avoid doing any computations in `ModelChain`, so I'd put these calculations in another `pvsystem` function. I don't see a need for a corresponding `PVSystem` method.

pvlib/pvsystem.py
```diff
@@ -2334,6 +2355,35 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+        Rw = (%_loss_at_stc / 100) * (Varray / Iarray)
+
+    Parameters
+    ----------
+    V_mp_ref: numeric
+    I_mp_ref: numeric
+    dc_ohmic_percent: numeric, default 0
+    modules_per_string: numeric, defualt 1
+    strings_per_inverter: numeric, defualt 1
+    """
```
wholmgren, line 2377: add returns section

pvlib/pvsystem.py
```diff
@@ -2334,6 +2355,35 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+        Rw = (%_loss_at_stc / 100) * (Varray / Iarray)
+
+    Parameters
+    ----------
+    V_mp_ref: numeric
+    I_mp_ref: numeric
+    dc_ohmic_percent: numeric, default 0
+    modules_per_string: numeric, defualt 1
+    strings_per_inverter: numeric, defualt 1
```
wholmgren, line 2376: spelling

pvlib/modelchain.py
```diff
@@ -935,13 +935,18 @@ def losses_model(self, model):
...
-                self._losses_model = self.no_extra_losses
+                self._losses_model = model
+                self._dc_losses = self.no_extra_losses
+            elif model == 'pvsyst':
+                self._losses_model = model
+                self._dc_losses = self.pvsyst_dc_losses
             else:
                 raise ValueError(model + ' is not a valid losses model')
         else:
-            self._losses_model = partial(model, self)
```
wholmgren, line 944: -1 on this change. It's a powerful feature (albeit likely seldom-used) and the change would make `losses_model` inconsistent with the other model options.

wholmgren (commented):
I'm generally good with the basic function and method. I'm concerned about how this works with `ModelChain`. I think we're going to need to start breaking up the `losses_model` into a handful of options. From the `pvwatts_losses` kwargs, snow, soiling, wiring, and shading could all be treated with their own `ModelChain` kwargs.

pvlib/pvsystem.py
ncroft-b4: I do think we can drop pvsyst from the name and add it as a reference. maybe `dc_ohms_from_percent`? tough to balance descriptiveness with conciseness for this one. @cwhanse I incorporated your comment into the latest commit. Thanks.

pvlib/modelchain.py
ncroft-b4: In PVsyst you can also input ohms directly, which we could support pretty easily I think with a different loss parameter that uses the same model chain function but bypasses the equivalent resistance function. I want to investigate that more to confirm the method and then will work to implement.

pvlib/pvsystem.py
ncroft-b4: > shouldn't dc_ohmic_percent be a required parameter for the function?

I was trying to anticipate that a user might choose PVsyst losses for a `ModelChain` that would set all the PVsyst-style loss functions but not have explicit losses set for each loss. In that case they would not need to have `dc_ohmic_percent` defined and the function could still run but but defaulted to 0.

pvlib/modelchain.py
ncroft-b4: I was trying to model this off of the `pvwatts_losses` function, which uses a `PVSystem` method to get a parameter for the loss equation and then calculates the losses from the operating values attached to the `ModelChain`. Since this operates on `ModelChain.dc` I felt it was appropriate to have the calculation with the object it is operating on. Otherwise we will need to pass `ModelChain.dc` into a `pvsystem` function that would only be used by a `ModelChain`. That said, I'd be happy to make the change and make a new `pvsystem` function if that's the preference.

pvlib/modelchain.py
ncroft-b4: I went back a forth about how to set this up, and agreed that this makes it inconsistent. How about treating `losses_model` somewhat like `clearsky_model` or `transposition_model` where it is always just a string and then that string can trigger different behavior at the various losses steps in the chain? The difference there though is that those tags are strings because they get passed to other functions that way, and losses_model would behave differently if we did it like that. I liked the functionality of the setter as a way both to initialize appropriate functions and work as a check for a compatible model. 

But, we could set it up that if `losses_model` does not have a setter and is a string, there is a `ModelChain._dc_losses` function that runs, and that calls different functions (`pvwatts`, `pvsyst_dc_losses`, `no_extra_losses`, etc) that get triggered based on what string `ModelChain.losses_model` is. And similarly for other loss functions like a `ModelChain._shading_losses`.

pvlib/pvsystem.py
ncroft-b4: > don't capitalize the parameters V_mp_ref and I_mp_ref

Actually, I just realized that because V_mp_ref and I_mp_ref are coming from `PVSystem.module_parameters` where they are keyed with capitals, then `_build_kwargs` will key them into the `kwargs` with capitals and pass them into the function that way, that is why left them capitalized as parameters. This is similar to the parameters for `calcparams_cec`, some of which remain capitalized. Stylistically, do you prefer to keep everything lower case for the function parameters and force the `kwargs` into lower-cased keys after `_build_kwargs`? I think that would clash with other methods in the module, so I might go back to having these remain capitalized.

pvlib/modelchain.py
ncroft-b4: The more I think about it, the more I do lean towards having each loss (snow, shading, ohmic, etc) be its own `ModelChain` kwarg. It increases the amount of setup a user has to do, because they have to pass in each of those if they want them, instead of just say passing in "pvsyst", or "pvwatts" and loss models, but it would allow a user to mix and match methods which is how all the rest of the methods are setup. Then we can add in helpers kind of like the `with_sapm` one that can setup default configurations. 

This way will add in more code overall, I think, but it will keep it consistent an flexible. I will work to build that out a bit and then submit it for review.

pvlib/pvsystem.py
wholmgren: > I was trying to anticipate that a user might choose PVsyst losses for a ModelChain that would set all the PVsyst-style loss functions but not have explicit losses set for each loss. In that case they would not need to have dc_ohmic_percent defined and the function could still run but but defaulted to 0.

That kind of decision is best left for the PVSystem/ModelChain layer. All of the base functions should make sense on their own.

pvlib/modelchain.py
```diff
@@ -924,6 +933,41 @@ def fuentes_temp(self):
...
+            self._dc_ohmic_model = partial(model, self)
+
+    def dc_ohms_from_percent(self):
+        """
+        Calculate time series of ohmic losses and apply those to the mpp power
+        output of the `dc_model` based on the pvsyst equivalent resistance
+        method. Uses a `dc_ohmic_percent` parameter in the `losses_parameters`
+        of the PVsystem.
+        """
+        Rw = self.system.dc_ohms_from_percent()
```
cwhanse, line 962: I don't think we want to embed a model in a ModelChain method. I would put this calculation in a new function `pvsystem.dc_ohmic_loss`, and have the `ModelChain` method call that function.

cwhanse (commented):
@ncroft-b4 lmk if you'd like some help writing tests

pvlib/modelchain.py
ncroft-b4: I thought about that and @wholmgren had the same comment, but there are other calculations that run as `ModelChain` methods, such as 'pvwatts_losses' (though the calculations there are slightly simpler). I modeled this off of that, where it calls a `'pvsystem` method to retrieve a parameter based on system specifications, but then operates on `ModelChain` attributes. This method only makes sense in the context of a `ModelChain` where you have dc data produced by the chain. It felt awkward to pass a `ModelChain` attribute (`dc`) into a `pvsystem` function, and have that function only be used by `ModelChain`. But if that is the preference I am happy to make the change.

ncroft-b4:
@cwhanse I made an attempt at the tests. I would appreciate a review that I met the intention there.

pvlib/modelchain.py
wholmgren: Definitely right to say it would be awkward to pass a `ModelChain` attribute into a `pvsystem` function or `PVSystem` method, but I don't think that's required. A couple of ideas...

First, I think we should add this trivial function:

```python
# in pvsystem.py
def dc_ohmic_losses(ohms, current):
    return ohms * current * current
```

Then your `ModelChain` method could be:

```python
    def dc_ohms_from_percent(self):
        Rw = self.system.dc_ohms_from_percent()
        self.dc_ohmic_losses = pvsystem.dc_ohmic_losses(Rw, self.dc['i_mp'])
        self.dc['p_mp'] = self.dc['p_mp'] - self.dc_ohmic_losses
        return self
```

Or add a new `PVSystem` method:

```python
# pvsystem.py
    def dc_ohmic_losses_from_percent(self, current):
        Rw = self.dc_ohms_from_percent()
        return pvsystem.dc_ohmic_losses(Rw, current)

# modelchain.py
    def dc_ohms_from_percent(self):
        self.dc_ohmic_losses = self.system.dc_ohmic_losses(self.dc['i_mp'])
        self.dc['p_mp'] = self.dc['p_mp'] - self.dc_ohmic_losses
        return self
```

pvlib/tests/test_modelchain.py
```diff
@@ -704,6 +704,41 @@ def constant_losses(mc):
...
+
+    assert isinstance(mc.dc_ohmic_losses, pd.Series)
+
+
+def test_dc_ohmic_model_no_dc_ohmic_loss(cec_dc_snl_ac_system,
+                                         location,
+                                         weather,
+                                         mocker):
+
+    m = mocker.spy(pvsystem, 'dc_ohms_from_percent')
```
cwhanse, line 731: I'd change this to spy on `ModelChain.no_dc_ohmic_loss`

cwhanse (commented):
@ncroft-b4 tests look OK so far. For coverage of line L953, the [aoi method test](https://github.com/pvlib/pvlib-python/blob/1844e3d30069abdf43cbf85e55c41f3b3e72f738/pvlib/tests/test_modelchain.py#L664) is a model.

ncroft-b4:
Would like to close this out in favor of #1168 which includes compatibility for arrays in the latest updates.

PR #1076, opened 2020-10-06
Support for PV systems with multiple arrays

wfvining:
- [x] Closes #1067 
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [x] Tests added
 - [x] Updates entries to [`docs/sphinx/source/api.rst`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes.
 - [x] Update `docs/sphinx/source/pvsystem.rst`
 - [x] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels and Milestone are assigned to the Pull Request and linked Issue.
 - [x]  Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
- [x] Accept `tuple` or `list` input
- [x] ~Update `modelchain.rst`~ (See #1115)
- [x] ~Address unused `kwargs` in `PVSystem.calcparams_xxxx` and `PVSystem.sapm`~ (punting on this: #1118)

Add a `pvsystem.Array` class and update API to support a single PVSystem instance with multiple arrays. Incorporates the multi-array API in to `modelchain.ModelChain`. Changes aim to be transparent to users who are modeling single-array systems, while the types of some fields (POA, DC power, etc) will change to a tuple for those who use the a PVSystem with multiple arrays.

PR #1085, opened 2020-10-27
implement multi-MPPT inverter model

### LINKED ISSUE #457: Inverter with multiple MPPTs
It's been a couple years since I have used pvlib (and the progress is amazing) so forgive me if this has already been covered.  Is there an implementation of inverters with multiple MPPTs?

cwhanse:
- [x] Closes #457
 - [x] Tests added

An approach that leverages the existing function `pvlib.inverter.sandia` and the model parameters available in the SAM file. Conceptually, this is the same approach as in Pvsyst, although it appears that Pvsyst offers more options for adjusting the inverter's behavior (and Pvsyst uses an internally-developed inverter output model, which isn't fully documented in the public files).


pvlib/inverter.py
```diff
@@ -91,29 +119,75 @@ def sandia(v_dc, p_dc, inverter):
...
+    '''
+    if dc_limit is None:
+        dc_limit = inverter['Paco']
+
+    power_ac = np.zeros_like(p_dc[0])
+    power_dc = np.zeros_like(p_dc[0])
+    for pdc in p_dc:
+        power_dc += pdc
+
+    for vdc, pdc in zip(v_dc, p_dc):
```
wholmgren, line 185: We should be careful about zip behavior and mismatched length. I suspect we either need to handle it with `itertools.zip_longest` or disallow it.

pvlib/tests/test_inverter.py
```diff
@@ -95,6 +95,27 @@ def test_sandia_Pnt_micro():
     assert_series_equal(pacs, pd.Series([-0.043, 132.545914746, 240.0]))
 
 
+def test_sandia_multi(cec_inverter_parameters):
+    vdcs = pd.Series(np.linspace(0, 50, 3))
+    idcs = pd.Series(np.linspace(0, 11, 3)) / 2
```
wholmgren, line 100: The implementation `test_sandia_multi` appears to allow for a variety of type inputs, so I think it's worth adding regression tests for float and array input. I can imagine several other implementations that would not be so accommodating.

pvlib/inverter.py
```diff
@@ -91,29 +119,75 @@ def sandia(v_dc, p_dc, inverter):
...
+    --------
+    pvlib.inverter.sandia
+    '''
+    if dc_limit is None:
+        dc_limit = inverter['Paco']
+
+    power_ac = np.zeros_like(p_dc[0])
+    power_dc = np.zeros_like(p_dc[0])
+    for pdc in p_dc:
+        power_dc += pdc
```
wholmgren, line 183: Any disadvantage to `power_dc = sum(p_dc)`?

```python
In [4]: import numpy as np

In [5]: a = np.array([1, 1])

In [6]: b = np.array([2, 2])

In [7]: sum((a, b))
Out[7]: array([3, 3])

In [8]: import pandas as pd

In [9]: c = pd.Series([1, 1])

In [10]: d = pd.Series([2, 2])

In [11]: sum((c, d))
Out[11]:
0    3
1    3
dtype: int64
```

pvlib/inverter.py
cwhanse: v_dc and p_dc need to be paired, one entry per array. Entry could be float, Series or numeric.

If v_dc and p_dc are different length then I would raise an Error. OK?

wholmgren: That works. I believe ValueError is most appropriate.

mikofski:
I found these pages in the pvsyst help docs interesting:
* [Power sharing](https://www.pvsyst.com/help/powersharing.htm(
* Grid Inverters: main parameters] (https://www.pvsyst.com/help/inverters_general.htm)

>**Be carefu**l: as in the PVsyst inverter database, there is no information about the maximum power (or current) for each individual MPPT input, it is your responsibility to check in the data sheets whether the allocated power is compatible with the manufacturer's specifications.  PVsyst puts a  general limit to the sharing ratio, which is modifiable in the Hidden parameters.

Has anyone checked the _hidden parameters_?

cwhanse:
> Has anyone checked the _hidden parameters_?

I don't have a Pvsyst license, I'd be interested to know what parameters are hidden that relate to multi-MPPT inverters.

I've read those pages and also exchanged email with Bruno at Pvsyst. I don't feel confident that I accurately understand how inverter efficiency is modeled in Pvsyst, so am not proposing a `inverter.pvsyst` function. 

Inverter datasheets typically list limits on DC current per MPP input, but these limits appear be related to safety (e.g. fuse ratings) rather than limits on DC-AC conversion.

For this model, I had proposed a parameter 'dc_limit' which would act as a cap on DC power at each MPP input. But after some discussions, I've removed that parameter for two, related reasons:
- there's no reliable source for its value (the safety limits in datasheets are not what is needed here)
- it is not clear that inverters actually impose limits per MPP input.

cwhanse:
Bumping this PR as I think it's ready for review, and is needed to move #1076 forward.

pvlib/inverter.py
```diff
@@ -91,29 +124,80 @@ def sandia(v_dc, p_dc, inverter):
...
+
+    Uses Sandia's Grid-Connected PV Inverter model [1]_.
+
+    Parameters
+    ----------
+    v_dc : numeric or tuple of numeric
+        DC voltage on each MPPT input of the inverter. [V]
+
+    p_dc : numeric or tuple of numeric
+        DC power on each MPPT input of the inverter. [W]
```
wholmgren, line 152: what's the use case for allowing numeric? Can we always require tuple/list of numeric?

pvlib/inverter.py
```diff
@@ -91,29 +124,80 @@ def sandia(v_dc, p_dc, inverter):
...
+    .. [1] D. King, S. Gonzalez, G. Galbraith, W. Boyson, "Performance Model
+       for Grid-Connected Photovoltaic Inverters", SAND2007-5036, Sandia
+       National Laboratories.
+
+    See also
+    --------
+    pvlib.inverter.sandia
+    '''
+
+    if isinstance(p_dc, tuple):
```
wholmgren, line 182: this will fail if a user provides a list instead of a tuple. I think most python users expect functions to treat lists and tuples as interchangeable. I'd argue a 1-d array should be acceptable in most cases, too.

pvlib/inverter.py
```diff
@@ -91,29 +124,80 @@ def sandia(v_dc, p_dc, inverter):
...
+    else:
+        power_dc = sum((p_dc,))  # handle Series, array or float
+
+    power_ac = 0. * power_dc
+
+    if isinstance(p_dc, tuple):
+        for vdc, pdc in zip(v_dc, p_dc):
+            power_ac += pdc / power_dc * _sandia_eff(vdc, power_dc, inverter)
+    else:
+        power_ac = _sandia_eff(v_dc, power_dc, inverter)
```
wholmgren, line 195: I don't understand why we'd include this line in the `sandia_multi` function. Isn't it duplicating functionality in `sandia`? Maybe I don't understand the use cases well enough.

pvlib/inverter.py
```diff
@@ -16,6 +16,39 @@
...
+def _sandia_limits(power_ac, p_dc, Paco, Pnt, Pso):
+    r'''
+    Applies minimum and maximum power limits to `power_ac`
+    '''
+    power_ac = np.minimum(Paco, power_ac)
+    try:
+        power_ac[p_dc < Pso] = -1.0 * abs(Pnt)
+    except TypeError:  # float
+        if power_ac < Pso:
+            power_ac = -1.0 * abs(Pnt)
```
wholmgren, line 48: I was going to say the `np.where` in the original function does all of this in one line, but the conditional in the array case is different from the conditional in the float case, so I'm just confused.

pvlib/inverter.py
cwhanse: to have this function behave appropriately when only one MPPT input is being used. In that case I'm not opposed to requiring the input to be `list or tuple of numeric` of length 1.

pvlib/inverter.py
cwhanse: I was thinking that `inverter.sandia_multi(600., 1500., *params)` should work, but can't really say that's a required use.

Eventually I'd see `inverter.sandia` and `inverter.sandia_multi` merging, once the latter is better documented, with validation, and has broader acceptance.

pvlib/tests/test_inverter.py
```diff
@@ -95,6 +95,45 @@ def test_sandia_Pnt_micro():
...
+
+def test_sandia_multi_array(cec_inverter_parameters):
+    vdcs = np.linspace(0, 50, 3)
+    idcs = np.linspace(0, 11, 3)
+    pdcs = idcs * vdcs
+    pacs = inverter.sandia_multi(vdcs, pdcs, cec_inverter_parameters)
+    assert_allclose(pacs, np.array([-0.020000, 132.004278, 250.000000]))
+
+
+def test_sandia_multi_float(cec_inverter_parameters):
```
cwhanse, line 123: @wholmgren do you think the use case in this test should be retained?

pvlib/inverter.py
cwhanse: Line 47 is a typo, should be `p_dc < Pso:`.  The `try/except` is an attempt to handle the case when `power_ac` is a float. `np.where` converts `Series` to `ndarray`, and I was hoping to avoid another use of `isinstance` to preserve input type. I'm OK using the technique in `inverter.sandia` if you prefer that.

pvlib/tests/test_inverter.py
wholmgren: No

pvlib/tests/test_inverter.py
```diff
@@ -95,6 +95,45 @@ def test_sandia_Pnt_micro():
...
+    pdcs = idcs * vdcs
+    with pytest.raises(ValueError, match='p_dc and v_dc have different'):
+        inverter.sandia_multi((vdcs,), (pdcs, pdcs), cec_inverter_parameters)
+
+
+def test_sandia_multi_array(cec_inverter_parameters):
+    vdcs = np.linspace(0, 50, 3)
+    idcs = np.linspace(0, 11, 3)
+    pdcs = idcs * vdcs
+    pacs = inverter.sandia_multi(vdcs, pdcs, cec_inverter_parameters)
```
wholmgren, line 119: I'd also fail this unless changed to `...(vdcs,), (pdcs,)...`

pvlib/tests/test_inverter.py
cwhanse: That will simplify things.

pvlib/inverter.py
wholmgren: > np.where converts Series to ndarray, and I was hoping to avoid another use of isinstance to preserve input type.

Ah, I forgot about that. Perhaps your method is preferable because it works with any array-like input (e.g. dask). I suggest assigning the condition and the limit value to variables before the try/except block to ensure there's no inconsistency.

wholmgren:
@cwhanse you can decorate that failing psm test function with `@pytest.mark.xfail(strict=True, reason='github 1091')`

pvlib/inverter.py
wfvining: +1 to requiring an Iterable of length 1

pvlib/inverter.py
cwhanse: The problem with 1-d arrays is that this will run and produce garbage results:
```
inverter.sandia_multi(v_dc=np.array([500., 500.]), p_dc=np.array([1500., 1500.]), ...)
```
Because `p_dc` is iterable
```
power_dc = sum(p_dc)
```
sums over the array and produces a float instead of an array.

This doesn't work either since numpy doesn't create 1d arrays containing a single 1d array of any length:
```
inverter.sandia_multi(v_dc=np.array(np.array([500., 500.])), p_dc=np.array(np.array([1500., 1500.])), ...)
```
I'd just not list `array` as an accepted type for these inputs.

pvlib/inverter.py
cwhanse: Do we want to test for and enforce `tuple` or `list` type? Duck-typing risks returning garbage results if the input type is array.

pvlib/inverter.py
wholmgren: I'd hope it works with a 2D array

```
>>> np.array((np.array([500., 500.]), ))
array([[500., 500.]])
```

pvlib/inverter.py
wholmgren: I'd prefer a garbage in, garbage out approach in this situation because it feels like we're still figuring out this function.

pvlib/inverter.py
```diff
@@ -91,29 +125,74 @@ def sandia(v_dc, p_dc, inverter):
...
+    .. [1] D. King, S. Gonzalez, G. Galbraith, W. Boyson, "Performance Model
+       for Grid-Connected Photovoltaic Inverters", SAND2007-5036, Sandia
+       National Laboratories.
+
+    See also
+    --------
+    pvlib.inverter.sandia
+    '''
+
+    if len(p_dc) == len(v_dc):
```
wholmgren, line 183: change to `!=` and put the `raise` in the `if` block, then move the rest of the code out of the `if/else` structure. no need for an `else`.

wholmgren (approved):

cwhanse:
@pvlib/pvlib-maintainer will merge on 12/2 unless there are additional reviews

kanderso-nrel (approved):
LGTM, minor comments below

cwhanse:
@pvlib/pvlib-maintainer Will merge today if no further reviews

wholmgren:
let's move on with it - thanks @cwhanse!

jhfatehi:
Sorry I am a bit late to the game on this.  My understanding of the multi-MPPT modeling method currently being worked on is that all arrays are operating at the same DC voltage.  The focus is address clipping per MPPT.  When I first brought up this question I was modeling a residential system with 2 arrays on different planes and of different sizes.  My solution at the time was to average the DC voltage between the 2 arrays and feed that into the inverter efficiency loss function.  Given the huge variations in insolation because of shading on a typical residential system, any accuracy loss because of the inverter efficiency was not a concern to me at the time.  No suggestion here.  I just wanted to circle back with the original use case.

PR #1137, opened 2021-01-20
add testpaths to setup.cfg pytest section

wholmgren:

I had been running into odd behavior with vscode no longer able to automatically discover pvlib tests and with strange errors when trying to run `pytest` from the root pvlib directory - sometimes I'd get numpy import errors in good environments and sometimes I'd see errors with old pvlib version numbers. But `pytest pvlib` worked just fine. It turns out that the presence of conda environments in `benchmarks/` led to bad numpy imports and the presence of old pvlib versions in `dist/` led to other issues. Adding these directories to the `setup.cfg` pytest section solved the issues, and I added a bunch of extra directories for safety.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,3 +21,4 @@ exclude = pvlib/_version.py docs dist
 
 [tool:pytest]
 junit_family=xunit2
+norecursedirs = benchmarks .git* dist pvlib.egg-info junit htmlcov paper ci .vscode .ipynb_checkpoints
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,4 +21,4 @@ exclude = pvlib/_version.py docs dist
 
 [tool:pytest]
 junit_family=xunit2
-norecursedirs = benchmarks .git* dist pvlib.egg-info junit htmlcov paper ci .vscode .ipynb_checkpoints
+testpaths = pvlib/tests


wholmgren:
@kanderso-nrel @mikofski can you merge this if you have no concerns with it? It feels like just a bit more than I should do without a second opinion.

kanderso-nrel (commented):
If all of our tests are in one place (`/pvlib/tests`), would it make more sense to tell pytest where to look with `testpaths` ([docs](https://docs.pytest.org/en/stable/reference.html#confval-testpaths)) instead of where not to look with `norecursedirs`?  

Also the pytest docs [recommend not using setup.cfg](https://docs.pytest.org/en/stable/customize.html#setup-cfg) except for very simple cases.  So if at some point in the future we find ourselves looking at this config section again, might be worth switching to one of the recommended options.

wholmgren:
> If all of our tests are in one place (/pvlib/tests), would it make more sense to tell pytest where to look with testpaths (docs) instead of where not to look with norecursedirs?

that is more clear and I think it might have been faster to collect too.

> Also the pytest docs recommend not using setup.cfg except for very simple cases. So if at some point in the future we find ourselves looking at this config section again, might be worth switching to one of the recommended options.

I saw that too. I'm not sure what qualifies as too complicated for setup.cfg. Maybe we could condense a few configuration files into a pyproject.toml file.

kanderso-nrel:
I think the config in the most recent commit qualifies as simple.  No objections from me to merging, but I don't have benchmark envs handy so I didn't reproduce the issue locally to confirm the fix works.

PR #1143, opened 2021-01-24
Adds Location.get_aoi()

dcambie:
 - [x] Tests added

Add a Location method to get the angle of incidence for a series of time points.

### DIFF
--- a/docs/sphinx/source/api.rst
+++ b/docs/sphinx/source/api.rst
@@ -146,6 +146,7 @@ Decomposing and combining irradiance
 .. autosummary::
    :toctree: generated/
 
+   location.Location.get_aoi
    irradiance.get_extra_radiation
    irradiance.aoi
    irradiance.aoi_projection
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -87,6 +87,10 @@ Enhancements
   by ``pvsystem.PVSystem.modules_per_strings`` and
   ``pvsystem.PVSystem.strings_per_inverter``. Note that both attributes still
   default to 1. (:pull:`1138`)
+* In :py:class:`~pvlib.location.Location`, the method
+  :py:meth:`~pvlib.location.Location.get_aoi` has been added, returning the
+  angle of incidence for a time series.
+  (:pull:`1143`)
 
 Bug fixes
 ~~~~~~~~~
@@ -113,3 +117,4 @@ Contributors
 * Mark Mikofski (:ghuser:`mikofski`)
 * Nate Croft (:ghuser:`ncroft-b4`)
 * Kevin Anderson (:ghuser:`kanderso-nrel`)
+* Dario Cambié (:ghuser:`dcambie`)
--- a/pvlib/location.py
+++ b/pvlib/location.py
@@ -5,12 +5,12 @@ This module contains the Location class.
 # Will Holmgren, University of Arizona, 2014-2016.
 
 import datetime
-import warnings
 
 import pandas as pd
 import pytz
 
 from pvlib import solarposition, clearsky, atmosphere, irradiance
+
 
 class Location:
     """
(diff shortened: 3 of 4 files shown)


kanderso-nrel (commented):
Hi @dcambie, thanks for the PR.  I think I'm not in favor of adding an AOI method to `Location` because it would blur the line between `Location` and the new `Array` class (see [Array.get_aoi](https://pvlib-python.readthedocs.io/en/latest/generated/pvlib.pvsystem.Array.get_aoi.html)).  For example if `Location` has a `get_aoi` method, why not methods for IAM and irradiance transposition too?

Suggested improvements below in case there's a good use case I'm not thinking of.

dcambie:
Oh cool, i was not aware of `Array`! Sorry, my bad, nevermind :)

cwhanse:
@dcambie `Array` is brand new and isn't yet released. I'm with @kevinsa5 that adding a `get_aoi` method to `Location` risks confusion. An AOI depends on system properties such as tilt and azimuth which aren't specific to a location. Can you describe the use cases which would benefit from `Location.get_aoi`?

dcambie:
My use case is the following.

I am using pvlib to generate solar position and irradiance data for a solar photochemical reactor[1]. Those data are then used to initialize a pvtrace[2] simulation.

As we intend to simulate and compare the reactor performance in different locations, we need to find the optimal tilt angle for each of them.

So currently, given site (pvlib.location.Location) and an year of datapoint every 30min we use Location.get_solarposition() to find the azimuth and altitude for solar position in the simulation, and location.get_clearsky() for ghi/dhi (for raytracing we use the solar position only for the direct component, the diffuse component is assumed to have random origin). Then we use spectral2 for the spectrum.
However, to compensate for the reactor tilt angle we needed to do some calculation (i.e. getting the dni (as (ghi-dhi)/cos(90deg-elevation) and then multiplying it by the dot product of the normal to the reactor with the normal of solar vector).
Since I discovered aoi already performed some of the calculations we needed I thought of adding it to Locations so we get from pvlib a pandas dataframe with solar positions, ghi/dhi and aoi for all our timepoints/location combination.


[1] Luminescent Solar Concentrator Photomicroreactor see https://onlinelibrary.wiley.com/doi/pdf/10.1002/anie.201908553 
[2] https://github.com/dcambie/pvtrace

wholmgren:
@dcambie are you familiar with pvlib's `ModelChain` ([user guide](https://pvlib-python.readthedocs.io/en/stable/modelchain.html), [api reference](https://pvlib-python.readthedocs.io/en/stable/api.html#modelchain))?

It sounds to me like a `ModelChain` would be a good fit for your application but you might need to define custom, no-op functions for a few steps. Here's a some code that might help:

```python
times = pd.date_range(start='20201201', end='20201203', freq='5min', tz='America/Phoenix')

loc = location.Location(32.2, -110.9, altitude=700)

weather = loc.get_clearsky(times)

system = pvsystem.PVSystem(surface_tilt=32.2, surface_azimuth=180)

# see https://pvlib-python.readthedocs.io/en/stable/modelchain.html#user-defined-models
def no_op(mc):
    return mc


mc = modelchain.ModelChain(
    system, 
    loc,
    aoi_model='physical',
    spectral_model='no_loss',
    dc_model=no_op,
    ac_model=no_op,
    temperature_model=no_op
)
mc.run_model(weather)
```

With pvlib < 0.9, inspect results like:

```python
mc.aoi
2020-12-01 00:00:00-07:00    157.902125
2020-12-01 00:05:00-07:00    158.039237
2020-12-01 00:10:00-07:00    158.109274
2020-12-01 00:15:00-07:00    158.111590
2020-12-01 00:20:00-07:00    158.046165
                                ...    
2020-12-02 23:40:00-07:00    156.380563
2020-12-02 23:45:00-07:00    156.772334
2020-12-02 23:50:00-07:00    157.105987
2020-12-02 23:55:00-07:00    157.378941
2020-12-03 00:00:00-07:00    157.588975
Freq: 5T, Name: aoi, Length: 577, dtype: float64
```

With pvlib >= 0.9 (0.9.0 alpha 1 is available on pypi), you need to use the `mc.results` attribute 

```python
mc.results.aoi
```

Maybe use `pd.concat` to stitch together the most relevant attributes for your application into a single DataFrame.

You can also define methods that are useful for your application instead of using the suggested `no_op` function.

dcambie:
Thanks!! I'm not familiar with Model Chain, I'll look into it

PR #1145, opened 2021-01-24
Add read_bsrn function

### LINKED ISSUE #1015: Add BSRN format reader to iotools
I need a parser for the NASA Langley CAPABLE BSRN site. Given the importance and the quality of the BSRN sites, I expect that others would benefit from this parser as well. 

Does anyone have experience with this site or have a parser for the format?

https://capable.larc.nasa.gov/data/

https://cove.larc.nasa.gov/BSRN/LRC49/

Data from December 2014 to present.

1 month of data per file. Appears to be uploaded in the first few days of the following month. 1 minute intervals.

Another fun fixed width file. Entries like:

```
  1  987   1003 -99.9 -999 -999    975 -99.9 -999 -999
             72 -99.9 -999 -999    287 -99.9 -999 -999     19.9  37.2 1026
```
```
  1  988   1006 -99.9 -999 -999    977 -99.9 -999 -999
             72 -99.9 -999 -999    290 -99.9 -999 -999     19.8  36.9 1026
```
```
  1 1438     22 -99.9 -999 -999      0 -99.9 -999 -999
             21 -99.9 -999 -999    307 -99.9 -999 -999     17.7  57.9 1023
```
```
  1 1439     21 -99.9 -999 -999      0 -99.9 -999 -999
             20 -99.9 -999 -999    307 -99.9 -999 -999     17.6  56.8 1023
```

The first number is the day of the month. The second is the minute of the day. Times appear to be in UTC.

I believe the ordering is:

* CM22 pyranometer GHI (upper left)
* CM31 pyranometer DHI (lower left)
* CH1 pyrheliometer DNI (upper right)
* PIR infrared (lower right)

I'd probably read the file into a DataFrame without meaningful columns, split it into two DataFrames using `.iloc[::2]` and `.iloc[1::2]`, parse the date time information into and index, then stitch the data back together.

CAPABLE Site Coordinates:
Latitude: 37.1038
Longitude: -76.3872
Elevation: 3 m ASL

cross post from https://github.com/SolarArbiter/solarforecastarbiter-core/issues/541

AdamRJensen:
- [x] Closes #1015
 - [x] Tests added


AdamRJensen:
@wholmgren I've changed it such that it now commits to pvlib:master. Hope it's correct.

kanderso-nrel:
Hi @AdamRJensen, could you update this PR to resolve the code style issues raised by `Stickler`?  It's easier to review PRs once stickler is satisfied.  The style issues are listed here: https://github.com/pvlib/pvlib-python/pull/1145/files

wholmgren:
Thanks @AdamRJensen! We will need some tests. Do you want me to push a test to your branch or do you want to take it on yourself? It seemed like you might have been disinclined to add a test in your first PR, but I am happy to help since this will be a big help to me.

AdamRJensen:
Hi @kanderso-nrel, thanks for the heads up! 

I wasn't sure if PVlib is very strict on style issues, e.g. line length<79 characters? I have corrected all instances of this, but in some places, it makes the code a little harder to follow / less compact.

AdamRJensen:
@wholmgren I am definitely up for attempting to make tests. Can you point me in a direction to get started?

wholmgren:
Check out [`pvlib/tests/iotools/test_surfrad.py`](https://github.com/pvlib/pvlib-python/blob/master/pvlib/tests/iotools/test_surfrad.py) and [`pvlib/tests/iotools/test_surfrad.py`](https://github.com/pvlib/pvlib-python/blob/master/pvlib/tests/iotools/test_surfrad.py) [`pvlib/tests/iotools/test_solrad.py`](https://github.com/pvlib/pvlib-python/blob/master/pvlib/tests/iotools/test_solrad.py). 

Add a test file to `pvlib/data`. Here's [an example for the SURFRAD reader](https://github.com/pvlib/pvlib-python/blob/master/pvlib/data/surfrad-slv16001.dat). You'll find similar for the other iotools. 

Start by making your test as simple as possible. A single data file, no remote io capabilities. Probably ok to only assert the index values and the columns for now.

wholmgren:
> I wasn't sure if PVlib is very strict on style issues, e.g. line length<79 characters? I have corrected all instances of this, but in some places, it makes the code a little harder to follow / less compact.

If you think it makes the code less readable then it's acceptable to add `  # noqa: E501` to that line.

kanderso-nrel:
If the test file is a proper BSRN data file, we should make sure we're staying within their [conditions of data release](https://bsrn.awi.de/data/conditions-of-data-release/).

AdamRJensen:
> If the test file is a proper BSRN data file, we should make sure we're staying within their [conditions of data release](https://bsrn.awi.de/data/conditions-of-data-release/).

@kanderso-nrel and @wholmgren any opinion on if I should create a pseudo-file or ask Amelie from BSRN if we can upload a single file for test purposes?

pvlib/iotools/bsrn.py
```diff
@@ -0,0 +1,141 @@
...
+
+    The BSRN (Baseline Surface Radiation Network) is a world wide network
+    of high-quality solar radiation monitoring stations as described in [1]_.
+    The function only parses the basic measurements (LR0100), which include
+    global, diffuse, direct and downwelling long-wave radiation [2]_. Future
+    updates may include parsing of additional data and meta-data.
+
+    Required username and password are easily obtainable by writing an email to
+    Amelie Driemel (Amelie.Driemel@awi.de) [3]_ on condition that users follow
+    BSRN's Data Release Guidelines [4]_.
```
cwhanse, line 33:
```suggestion
    Required username and password are easily obtainable as described in the
    BSRN's Data Release Guidelines [4]_.

```

I wouldn't hardcode someone's email into the function docstring. A link to their page should be enough.

cwhanse:
@AdamRJensen I think we need to get explicit OK from the WRMC, because pvlib-python isn't a traditional "publication", but (in my view) qualifies as a "project for wide release." What matters, though, is WRMC's view of pvlib-python.

The way I read the Data Release Guidelines, assuming WRMC concurrence pvlib-python can include a file of BSRN data provided:
- the file is clearly identified as originated from the WRMC (use the file name for this purpose if there isn't a free-form header)
- users are provided with information on how to obtain the original data (function docstring does this, but we could also add a short paragraph to the LICENSE file about using the BSRN file).

wholmgren:
My reading is that BSRN data downloaded directly from NASA websites is subject to NASA's [Data and Information Policy](https://earthdata.nasa.gov/collaborate/open-data-services-and-software/data-information-policy). So I don't see any issue with including a test file from either of the GIM or LaRC sites.

AdamRJensen:
@cwhanse, @wholmgren I have received the green light from BSRN management to add a single test file, however, it requires permission from the specific station manager.

Unless you have any objections, I ask permission from the station manager of the Payerne, Schwitzerland (PAY) station as they provide the largest number of parameters (logical records) in the files and will therefore also be suitable if additional capabilities are added in the future.

wholmgren:
@AdamRJensen thanks, that works for me.

AdamRJensen:
@wholmgren Permission has been granted from the station manager at PAY!

I have added a zipped test file, which is how BSRN provides them. However, I cannot add the same file unzipped as it is 26 MB, which cannot be uploaded to GitHub. Can this be circumvented or is it sufficient only to test the function on a zipped file?

wholmgren:
@AdamRJensen a zipped file is fine with me. I checked out the branch and I see that the zipped file is 4.1 MB. This is slightly larger than other test files but not by enough that I'm concerned. An alternative is to reduce the length of the file by only including the first couple of days of data (see file below for example - added `.txt` to get around GitHub comment requirements). Other than convenience, is there a reason to include the entire data record? 

[bsrn-pay0616-2days.dat.txt](https://github.com/pvlib/pvlib-python/files/5952590/bsrn-pay0616-2days.dat.txt)

pvlib/iotools/bsrn.py
```diff
@@ -0,0 +1,146 @@
...
+                if line.startswith('*'):  # Find start of all logical records
+                    line_no_dict[line[2:6]] = num  # key is 4 digit LR number
+    else:
+        with open(filename, 'r') as f:
+            for num, line in enumerate(f):
+                if num == 1:  # Get month and year from the 2nd line
+                    start_date = pd.Timestamp(year=int(line[7:11]),
+                                              month=int(line[3:6]), day=1)
+                if line.startswith('*'):  # Find start of all logical records
+                    line_no_dict[line[2:6]] = num
```
wholmgren, line 107: refactor into something like

```python
if str(filename).endswith('.gz'):
    open_func, mode = gzip.open, 'rt'
else:
    open_func, mode = open, 'r'
with open_func(filename, mode) as f:
    for...
```

pvlib/iotools/bsrn.py
```diff
@@ -0,0 +1,146 @@
...
+    data['minute'] = data['minute'].astype('Int64')
+
+    # Set datetime index and localize to UTC
+    data.index = (start_date
+                  + pd.to_timedelta(data['day']-1, unit='d')
+                  + pd.to_timedelta(data['minute'], unit='min'))
+
+    try:
+        data.index = data.index.tz_localize('UTC')  # BSRN timestamps are UTC
+    except TypeError:
```
wholmgren, line 143: what's the point of this TypeError?

pvlib/iotools/bsrn.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+
+    # Read file and store the starting line number for each logical record (LR)
+    line_no_dict = {}
+    if str(filename).endswith('.gz'):  # check if file is a gzipped (.gz) file
+        with gzip.open(filename, 'rt') as f:
+            for num, line in enumerate(f):
+                if num == 1:  # Get month and year from the 2nd line
+                    start_date = pd.Timestamp(year=int(line[7:11]),
+                                              month=int(line[3:6]), day=1)
```
wholmgren, line 97: Should the `Timestamp` include `tz='UTC'`?

pvlib/iotools/bsrn.py
wholmgren: also no need for operating on the index: `data = data.tz_localize('UTC')`

wholmgren (commented):
A quick review below. Looks great overall! I'll try it out on some other data too and might have more comments then.

pvlib/iotools/bsrn.py
AdamRJensen: Indeed a much cleaner solution. Thanks!

pvlib/iotools/bsrn.py
AdamRJensen: The try statement was inspired from another pvlib io function. However, with the tz='utc' in the start date as per your suggestion (better solution) this is no longer relevant.

pvlib/iotools/bsrn.py
AdamRJensen: Yes, this is indeed a much nicer solution than the try statement and localizing later

AdamRJensen (commented):
Thanks for the feedback. All of it has been implemented.

AdamRJensen:
> @AdamRJensen a zipped file is fine with me. I checked out the branch and I see that the zipped file is 4.1 MB. This is slightly larger than other test files but not by enough that I'm concerned. An alternative is to reduce the length of the file by only including the first couple of days of data (see file below for example - added `.txt` to get around GitHub comment requirements). Other than convenience, is there a reason to include the entire data record?
> 
> [bsrn-pay0616-2days.dat.txt](https://github.com/pvlib/pvlib-python/files/5952590/bsrn-pay0616-2days.dat.txt)

@wholmgren No there is no particular reason for including all data at the moment. But if it is not a problem and we're okay just testing on a zip file, then I am in favor of just keeping the entire file as it is.

AdamRJensen:
Closed by accident.

pvlib/iotools/bsrn.py
```diff
@@ -0,0 +1,137 @@
...
+    # If LR0100 is the last logical record, then read rest of file
+    if start_row-1 == max(line_no_dict.values()):
+        end_row = num  # then parse rest of the file
+    else:  # otherwise parse until the beginning of the next logical record
+        end_row = min([i for i in line_no_dict.values() if i > start_row])
+    nrows = end_row-start_row
+
+    # Read file as a fixed width file (fwf)
+    data = pd.read_fwf(filename, skiprows=start_row, nrows=nrows, header=None,
+                       colspecs=COL_SPECS, na_values=[-999.0, -99.9])
```
wholmgren, line 116: It should be possible to avoid reading the file twice by accumulating the first read of just the '0100' section into an IO buffer, stopping the read at the end of the section, and then passing that buffer into `pd.read_fwf`. 

I think this might result in meaningful performance improvements when reading many files. Or maybe we need to profile the code before we put the effort into it.

In any case, I'm ok saving this idea for future work.

AdamRJensen: @wholmgren The initial opening of the file and looping through all of the lines takes ~414 ms, whereas the pd.read_fwf takes 1.15 s. Tested in a Jupyter Notebook using the %%timeit command.

Other parts of the code take on the order of µs (except setting the datetime index 3 ms).

So there are potential savings, however, after searching for a while I still have no idea of how to write to a buffer line by line. I'm also very fine with saving this idea for future work though.

kanderso-nrel:
A clue for some of the test failures: https://github.com/pandas-dev/pandas/pull/37909

I get failures locally with b5ed6ee and pandas==1.2.2, but adding `compression='infer'` to the `pd.read_fwf` call gets them passing.

pvlib/iotools/bsrn.py
wholmgren: Thanks for putting in the effort to profile the code! The potential speed up seems modest so let's save it for future work.

wholmgren:
@AdamRJensen please review changes on my `bsrnfixes` branch: 

https://github.com/wholmgren/pvlib-python/blob/bsrnfixes/pvlib/iotools/bsrn.py
https://github.com/wholmgren/pvlib-python/blob/bsrnfixes/pvlib/tests/iotools/test_bsrn.py

The function didn't quite produce the expected results when running it against this NASA LRC file that only contains the 0100 measurements: https://cove.larc.nasa.gov/BSRN/LRC49/2021/lrc0121.dat

In the process of trying to fix it, I changed the line number convention to make it make more sense to me. Fine with me if you want to keep the original convention so long as it's fixed to work correctly for files with only 0100. 

I also refactored the test to use both files. I didn't actually add the NASA LRC file to git due to the data license considerations above, so you'll need to download it and put it in the right place if you want that test to run correctly. Alternatively, we could make a second test file from the PAY file that only contains the 0100 record.

If you can, try adding my fork as another git remote, fetching my branch, and comparing against yours. Normally I would make a pull request to your fork, but for some reason github won't give me the option to select it as the base repository.

Finally, it appears the `bsrn.py` file uses windows line breaks. Can you change that to unix line breaks?

AdamRJensen:
@kanderso-nrel Thank you very much for solving the issue regarding the compression. Cleared a lot of issues! There are still 3 issues left... any ideas on these?

@wholmgren I prefer keeping it consistent with Python starting the line numbers at zero, hence I have added start=2 to the enumerate function. Also, I attempted to change the file to Unix line ending.. but I have no way of checking if it is correct now.

kanderso-nrel:
Older pandas versions had inconsistent behavior for offset strings... there's a relevant pandas issue or PR about this but I couldn't quickly find it.  Anyway I'd try `T` as the unit for the Timedelta because it means the same as `minute` in pandas lingo but might be supported by our minimum pandas version.  If `python36-min` doesn't like that either then I'll try to find that pandas page I'm remembering and figure out what will work across all versions.

For the codecov failure, it's because not 100% of your lines of code get run by the test suite: https://github.com/pvlib/pvlib-python/pull/1145/files#annotation_969052144.  If @wholmgren is okay with leaving the non-gzip branch untested then 96 can be left alone.  I haven't read the code in-depth enough to know whether it would be easy to engineer a test that runs 112.

wholmgren:
the bsrn.py file in this zip should have the right line breaks: [bsrn.py.zip](https://github.com/pvlib/pvlib-python/files/5962270/bsrn.py.zip)

wholmgren (approved):
I can fix line breaks in a follow up PR.

@kanderso-nrel any other comments?

pvlib/iotools/bsrn.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Read a BSRN station-to-archive file into a DataFrame.
+
+    The BSRN (Baseline Surface Radiation Network) is a world wide network
+    of high-quality solar radiation monitoring stations as described in [1]_.
+    The function only parses the basic measurements (LR0100), which include
+    global, diffuse, direct and downwelling long-wave radiation [2]_. Future
+    updates may include parsing of additional data and meta-data.
+
+    BSRN files are freely avaiable and can be accessed via FTP [3]_. Required
```
kanderso-nrel, line 30:
```suggestion
    BSRN files are freely available and can be accessed via FTP [3]_. Required

```

kanderso-nrel (approved):
I didn't run any code myself, but docs LGTM with a minor typo.

pvlib/iotools/bsrn.py
kanderso-nrel: I accidentally suggested an extraneous newline here -- I tried to edit the suggestion in time, but looks like I wasn't quick enough.  It's small enough that I don't care if it gets fixed here.  Thanks @AdamRJensen for this contribution!

AdamRJensen:
I'm unsure if there's anything else I need to do, but if that's the case please let me know. Thank you all for your patience and help!

wholmgren:
Thanks @AdamRJensen for the new feature and for your patience during the review! 

I expect it'll be a month or so before we release 0.9.0, but I'll make another pre-release by the end of next week.

PR #1147, opened 2021-01-25
Implement PVSystem.get_ac

### LINKED ISSUE #998: Consider replacing inverter methods in PVSystem with a single method with a kwarg
#886 moved inverter-related functions from `pvsystem.py` to `inverter.py`. Methods `PVSystem.snlinverter` and `PVsystem.adrinverter` currently wrap `inverter.sandia` and `inverter.adr`, respectively, which previously were `pvsystem.snlinverter` and `pvsystem.adrinverter`.

Perhaps renaming the functions creates potential for confusion.  

It may be an improved API to replace `PVSystem.snlinverter`, `PVSystem.adrinverter` and `PVSystem.pvwatts_ac` with a single method `PVSystem.dc_to_ac` or similar, with a kwarg `model='sandia'` for example.

Opening for discussion.

cwhanse:
- [x] Closes #998
 - [x] Tests added

Anticipates deprecation of the separate inverter methods PVSystem.snlinverter, .adrinverter, .sandia_multi, .pvwatts, and .pvwatts_multi after `ModelChain.ac_model` is reworked.

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -852,6 +852,40 @@ class PVSystem:
         return i_from_v(resistance_shunt, resistance_series, nNsVth, voltage,
                         saturation_current, photocurrent)
 
+    def get_ac(self, model, p_dc, v_dc=None):
+        r"""Calculates AC power from p_dc using the inverter model indicated
+        by model and self.inverter_parameters.
+
+        Parameter model must be one of 'sandia', 'adr', or 'pvwatts'.
+
+        Parameter v_dc is required for model='sandia' or model='adr'.
+
+        """
+        model = model.lower()
+        if model == 'sandia':
+            if self.num_arrays > 1:
+                p_dc = self._validate_per_array(p_dc)
+                v_dc = self._validate_per_array(v_dc)
+                inv_fun = inverter.sandia_multi
+            else:
+                inv_fun = inverter.sandia
+            return inv_fun(v_dc, p_dc, self.inverter_parameters)
+        elif model == 'pvwatts':
+            kwargs = _build_kwargs(['eta_inv_nom', 'eta_inv_ref'],
+                                   self.inverter_parameters)
+            if self.num_arrays > 1:
+                p_dc = self._validate_per_array(p_dc)
+                inv_fun = inverter.pvwatts_multi
+            else:
+                inv_fun = inverter.pvwatts
+            return inv_fun(p_dc, self.inverter_parameters['pdc0'], **kwargs)
+        elif model == 'adr':
+            return inverter.adrinverter(v_dc, p_dc, self.inverter_parameters)
+        else:
+            raise ValueError(
+                model + ' is not a valid AC power model.',
+                ' model must be one of "sandia", "adr" or "pvwatts"')
+
     # inverter now specified by self.inverter_parameters
     def snlinverter(self, v_dc, p_dc):
         """Uses :py:func:`pvlib.inverter.sandia` to calculate AC power based on
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -852,7 +852,6 @@ class PVSystem:
         return i_from_v(resistance_shunt, resistance_series, nNsVth, voltage,
                         saturation_current, photocurrent)
 
-    @_unwrap_single_value
     def get_ac(self, model, p_dc, v_dc=None):
         r"""Calculates AC power from p_dc using the inverter model indicated
         by model and self.inverter_parameters.
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -907,7 +907,7 @@ class PVSystem:
                 inv_fun = inverter.pvwatts
             return inv_fun(p_dc, self.inverter_parameters['pdc0'], **kwargs)
         elif model == 'adr':
-            return inverter.adrinverter(v_dc, p_dc, self.inverter_parameters)
+            return inverter.adr(v_dc, p_dc, self.inverter_parameters)
         else:
             raise ValueError(
                 model + ' is not a valid AC power model.',
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -922,28 +922,33 @@ class PVSystem:
                 ' model must be one of "sandia", "adr" or "pvwatts"')
 
     def snlinverter(self, v_dc, p_dc):
+        """Uses :py:func:`pvlib.inverter.sandia` to calculate AC power based on
+        ``self.inverter_parameters`` and the input voltage and power.
+
+        See :py:func:`pvlib.inverter.sandia` for details
         """
-        Deprecated. Use ``PVSystem.get_ac`` instead.
-        """
-        warnings.warn('PVsystem.sandia_multi is deprecated and will be',
-                      ' removed in v0.10. Use PVSystem.get_ac instead')
-        return self.get_ac('sandia', p_dc, v_dc)
+        return inverter.sandia(v_dc, p_dc, self.inverter_parameters)
 
     def sandia_multi(self, v_dc, p_dc):
+        """Uses :py:func:`pvlib.inverter.sandia_multi` to calculate AC power
+        based on ``self.inverter_parameters`` and the input voltage and power.
+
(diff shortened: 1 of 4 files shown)


pvlib/pvsystem.py
```diff
@@ -852,6 +852,67 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
         return i_from_v(resistance_shunt, resistance_series, nNsVth, voltage,
                         saturation_current, photocurrent)
 
+    def get_ac(self, model, p_dc, v_dc=None):
```
wholmgren, line 855: I believe this needs `@_unwrap_single_value` but I could be wrong.

pvlib/pvsystem.py
```diff
@@ -852,6 +852,67 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
...
+        pvlib.inverter.sandia
+        pvlib.inverter.sandia_multi
+        pvlib.inverter.adr
+        pvlib.inverter.pvwatts
+        pvlib.inverter.pvwatts_multi
+        """
+        model = model.lower()
+        if model == 'sandia':
+            if self.num_arrays > 1:
+                p_dc = self._validate_per_array(p_dc)
```
wholmgren, line 894: in other PVSystem methods we perform this validation independent of how many arrays exist, so I think it should move above the `if model` block.

The `v_dc` test should also move above the model block, but within a `if v_dc is not None` block.

pvlib/pvsystem.py
```diff
@@ -852,6 +852,67 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
...
+                v_dc = self._validate_per_array(v_dc)
+                inv_fun = inverter.sandia_multi
+            else:
+                inv_fun = inverter.sandia
+            return inv_fun(v_dc, p_dc, self.inverter_parameters)
+        elif model == 'pvwatts':
+            kwargs = _build_kwargs(['eta_inv_nom', 'eta_inv_ref'],
+                                   self.inverter_parameters)
+            if self.num_arrays > 1:
+                p_dc = self._validate_per_array(p_dc)
```
wholmgren, line 904: remove this line assuming we move the validation call above the `if model` block

pvlib/pvsystem.py
```diff
@@ -852,6 +852,67 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
...
+        --------
+        pvlib.inverter.sandia
+        pvlib.inverter.sandia_multi
+        pvlib.inverter.adr
+        pvlib.inverter.pvwatts
+        pvlib.inverter.pvwatts_multi
+        """
+        model = model.lower()
+        if model == 'sandia':
+            if self.num_arrays > 1:
```
wholmgren, line 893: this test is used in 2 (and probably needs to be used in 3) of the blocks, so maybe it would be more clear if it gets its own variable like `multiple_arrays = self.num_arrays > 1` above the `if model` block.

pvlib/pvsystem.py
```diff
@@ -852,6 +852,67 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
...
+        elif model == 'pvwatts':
+            kwargs = _build_kwargs(['eta_inv_nom', 'eta_inv_ref'],
+                                   self.inverter_parameters)
+            if self.num_arrays > 1:
+                p_dc = self._validate_per_array(p_dc)
+                inv_fun = inverter.pvwatts_multi
+            else:
+                inv_fun = inverter.pvwatts
+            return inv_fun(p_dc, self.inverter_parameters['pdc0'], **kwargs)
+        elif model == 'adr':
```
wholmgren, line 909: we need to handle the case that `model == 'adr' and self.num_arrays > 1` with another `ValueError`.

pvlib/pvsystem.py
cwhanse: I think it should. Now I'm wondering why the tests didn't fail.

pvlib/pvsystem.py
cwhanse: I can try that, but `_validate_per_array` always returns a tuple, which the inverter functions aren't set up to handle, e.g. `inverter.sandia` can't accept `p_dc` as a tuple.

I'm viewing `get_ac` as replacing the 5 inverter methods, not as a wrapper around those methods. Would you prefer that `get_ac` calls the `PVSystem` methods instead of the `inverter` functions?

wholmgren: Ok, I misunderstood. Your method is good.

pvlib/pvsystem.py
cwhanse: `@_unwrap_single_value` shouldn't be used for the `get_ac` method, since every inverter function returns a numeric. There are no "per-array" outputs.

wholmgren: ah, of course. the docstring incorrectly says `power_ac : numeric, or tuple of numeric`

pvlib/pvsystem.py
```diff
@@ -852,6 +852,75 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
...
+        See also
+        --------
+        pvlib.inverter.sandia
+        pvlib.inverter.sandia_multi
+        pvlib.inverter.adr
+        pvlib.inverter.pvwatts
+        pvlib.inverter.pvwatts_multi
+        """
+        model = model.lower()
+        mulitple_arrays = self.num_arrays > 1
```
wholmgren, line 894: mulitple --> multiple

pvlib/pvsystem.py
```diff
@@ -852,6 +852,75 @@ def i_from_v(self, resistance_shunt, resistance_series, nNsVth, voltage,
...
+            if mulitple_arrays:
+                p_dc = self._validate_per_array(p_dc)
+                inv_fun = inverter.pvwatts_multi
+            else:
+                inv_fun = inverter.pvwatts
+            return inv_fun(p_dc, self.inverter_parameters['pdc0'], **kwargs)
+        elif model == 'adr':
+            if mulitple_arrays:
+                raise ValueError(
+                    'The adr inverter function cannot be used for an inverter',
```
wholmgren, line 915:
```suggestion
                    'The adr inverter function cannot be used for an inverter ',
```

pvlib/pvsystem.py
cwhanse: Corrected now.

cwhanse:
With `PVSystem.get_ac` do we want to deprecate e.g. `PVSystem.snlinverter`?  The separate inverter methods aren't used by `get_ac`.

wholmgren (commented):
The new method looks good aside from trivial things. I do think we should deprecate the other methods in this PR.

cwhanse:
> I do think we should deprecate the other methods in this PR.

They can't be deprecated until after `ModelChain.ac_model` is reworked to use `PVSystem.get_ac`.

wholmgren (approved):
Ok, I can work the ModelChain changes in a follow up PR.

pvlib/tests/test_pvsystem.py
```diff
@@ -1374,6 +1374,21 @@ def test_PVSystem_multi_scale_voltage_current_power(mocker):
...
+def test_PVSystem_get_ac_sandia(cec_inverter_parameters, mocker):
+    inv_fun = mocker.spy(inverter, 'sandia')
+    system = pvsystem.PVSystem(
+        inverter=cec_inverter_parameters['Name'],
+        inverter_parameters=cec_inverter_parameters,
+    )
+    vdcs = pd.Series(np.linspace(0, 50, 3))
+    idcs = pd.Series(np.linspace(0, 11, 3))
+    pdcs = idcs * vdcs
+    pacs = system.get_ac('sandia', vdcs, pdcs)
```
wholmgren, line 1386: I don't understand why this test passed because `vdcs` and `pdcs` are swapped from their correct order. Fixed in #1150

cwhanse: `get_ac` orders the arguments `p_dc, v_dc` because `v_dc=None` is needed to support `pvwatts`. The inverter functions generally order the arguments `v_dc`, `p_dc` which is unfortunate.

PR #1150, opened 2021-01-27
refactor ModelChain inverter methods to use PVSystem.get_ac

wholmgren:
- follow up to #1147 
 - [x] Tests added

One could argue there should be a deprecation period for the `ModelChain.snlinverter` and `ModelChain.adrinverter` methods, but I'd be shocked if anyone is directly calling them.

### DIFF
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -783,29 +783,21 @@ class ModelChain:
     def infer_ac_model(self):
         """Infer AC power model from system attributes."""
         inverter_params = set(self.system.inverter_parameters.keys())
-        if self.system.num_arrays > 1:
-            return self._infer_ac_model_multi(inverter_params)
         if _snl_params(inverter_params):
             return self.sandia_inverter
         if _adr_params(inverter_params):
-            return self.adr_inverter
+            if self.system.num_arrays > 1:
+                raise ValueError(
+                    'The adr inverter function cannot be used for an inverter',
+                    ' with multiple MPPT inputs')
+            else:
+                return self.adr_inverter
         if _pvwatts_params(inverter_params):
             return self.pvwatts_inverter
         raise ValueError('could not infer AC model from '
                          'system.inverter_parameters. Check '
                          'system.inverter_parameters or explicitly '
                          'set the model with the ac_model kwarg.')
-
-    def _infer_ac_model_multi(self, inverter_params):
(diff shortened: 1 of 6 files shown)


pvlib/modelchain.py
```diff
@@ -790,9 +786,9 @@ def infer_ac_model(self):
         if self.system.num_arrays > 1:
```
cwhanse, line 786: Do we still need both `infer_ac_model` and `infer_ac_model_multi`?  Maybe not.

wholmgren: The alternative is something like the nested if statements of `PVSystem.get_ac`. We need some way of raising a `ValueError` if `num_arrays > 1` and the parameters are *not* consistent with pvwatts or sandia (essentially what we have now in `ModelChain`). Equivalently, for now, we could raise a `ValueError` if `num_arrays > 1` and the parameters are consistent with adr (essentially what we have now in `PVSystem.get_ac`). This behavior is tested in `test_ModelChain_invalid_inverter_params_arrays`. I have a slight preference to leave it alone, but I'll change it if you prefer it more like `PVSystem.get_ac`.

cwhanse: > We need some way of raising a ValueError if num_arrays > 1 and the parameters are not consistent with pvwatts or sandia (essentially what we have now in ModelChain).

Agree. It would make more sense to me to do the validation in `PVSystem.get_ac` so that it catches both ModelChain and PVSystem users. The `_infer` methods are only used when the inverter model isn't named, and the parameters for either `_multi` inverter model are the same as for the single MPPT version.

Assuming you agree, I'm OK deferring to later work

wholmgren: One more trade off here... The validation in the infer method helps avoid errors in PVSystem.get_ac during run_model. Traditionally ModelChain.run_model is nearly guaranteed to run if ModelChain can infer all methods at construction. Of course, as you point out, it only works if the user does not pass ac_model.

cwhanse: Raise ValueError in `_infer_ac_model` when parameters indicate 'adr' and num_arrays > 1. Otherwise `get_ac` should be fine.

cwhanse (approved):

mikofski:
Every time I see this I think it says `PVsyst.get_ac` 😆

PR #1162, opened 2021-02-05
Make sure times are assigned with ModelChain.run_from_poa

alorenzo175:
- [ ] Closes #xxxx
 - [x] Tests added

When running `ModelChain.run_from_poa`, `ModelChain.times` is not assigned so that the solar position results are incorrect and have a `NaT` index.

The issue was introduced in #1076 here https://github.com/pvlib/pvlib-python/blob/b0d29f550ea14d49499a97d60fa80c32a2a883d6/pvlib/modelchain.py#L1196-L1203, so I'm not sure if any what's new update is needed.

### DIFF
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -1324,6 +1324,7 @@ class ModelChain:
                 _build_weather(weather) for weather in data
             )
         self._configure_results()
+        self._assign_times()
         return self
 
     def _assign_total_irrad(self, data):
@@ -1392,7 +1393,6 @@ class ModelChain:
         self._check_multiple_input(weather, strict=False)
         self._verify_df(weather, required=['ghi', 'dni', 'dhi'])
         self._assign_weather(weather)
-        self._assign_times()
 
         self._prep_inputs_solar_pos(weather)
         self._prep_inputs_airmass()
@@ -1488,7 +1488,7 @@ class ModelChain:
         data = _to_tuple(data)
         self._check_multiple_input(data)
         self._assign_weather(data)
-        self._assign_times()
+
         self._verify_df(data, required=['poa_global', 'poa_direct',
                                         'poa_diffuse'])
         self._assign_total_irrad(data)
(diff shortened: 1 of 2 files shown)


wholmgren:
looks like the same issue exists in `ModelChain.run_model_from_effective_irradiance`. `times` is not separable from `weather`, so maybe it should be called within `_assign_weather`.

alorenzo175:
Yes it is missing there too, but I guess not a problem since it doesn't compute and assign solar position. So move it to `_assign_weather`?

cwhanse:
@wholmgren it appears that the only use of `ModelChain.times` is as input to a solar position calculation [here](https://github.com/pvlib/pvlib-python/blob/5a5d4aa7e5e5942b170b05f6acdad8830e7dc6a7/pvlib/modelchain.py#L1185) and [here](https://github.com/pvlib/pvlib-python/blob/5a5d4aa7e5e5942b170b05f6acdad8830e7dc6a7/pvlib/modelchain.py#L1236). I image those calls used `weather.index` before the multi-array capability. I'm not sure that `ModelChain.times` is consistent with the rest of `ModelChain` attributes, and if it persists is likely to get baked into user applications because it is convenient.

I'm OK patching this now so that pvlib behaves consistently. But down the road I think we'll want to remove `ModelChain.times` in favor of perhaps `ModelChain.results.times`?

wholmgren:
I agree. A year ago or so I wanted to remove `ModelChain.times` entirely because it no longer served any purpose, but refactorings since then have brought it back to life. `ModelChain.weather` should also be moved to the results. 

Let's move it to `_assign_weather` as a short term fix.

wholmgren (approved):

cwhanse:
> I agree. A year ago or so I wanted to remove `ModelChain.times` entirely because it no longer served any purpose, but refactorings since then have brought it back to life. `ModelChain.weather` should also be moved to the results.

Worth opening an issue so we don't forget about this

PR #1167, opened 2021-02-12
Optional requirements installed using pvlib[optional] between quotes

EnriqueCepeda:
It is only a typo I have discovered after trying to install the optional dependencies :)

### DIFF
--- a/docs/sphinx/source/installation.rst
+++ b/docs/sphinx/source/installation.rst
@@ -75,7 +75,7 @@ non-editable way, use one of the following commands to install pvlib-python::
 
     # get pvlib and optional dependencies from the Python Package Index
     # another option if you know what you are doing
-    pip install pvlib[optional]
+    pip install "pvlib[optional]"
 
 .. note::
 


kanderso-nrel:
Hi @EnriqueCepeda!  I'm curious, can you give some details about what OS/environment you're using where the quotation marks were necessary?  I tried it on win10 and the anaconda prompt and it worked fine unquoted:

```
(base) C:\Users\KANDERSO\projects>conda create -n pvlib_optional python=3.7
[output removed for brevity]

(base) C:\Users\KANDERSO\projects>conda activate pvlib_optional

(pvlib_optional) C:\Users\KANDERSO\projects>pip install pvlib[optional]
[output removed for brevity]
Successfully installed NREL-PySAM-stubs-2.2.0 beautifulsoup4-4.9.3 cftime-1.4.1 chardet-4.0.0 cycler-0.10.0 cython-0.29.21 ephem-3.7.7.1 future-0.18.2 idna-2.10 kiwisolver-1.3.1 llvmlite-0.35.0 matplotlib-3.3.4 netcdf4-1.5.5.1 nrel-pysam-2.2.0 numba-0.52.0 numexpr-2.7.2 numpy-1.20.1 pandas-1.2.2 patsy-0.5.1 pillow-8.1.0 protobuf-3.15.0rc1 pvfactors-1.4.1 pvlib-0.8.1 pyparsing-2.4.7 python-dateutil-2.8.1 pytz-2021.1 requests-2.25.1 scipy-1.6.0 shapely-1.7.1 siphon-0.8.0 six-1.15.0 soupsieve-2.2 statsmodels-0.12.2 tables-3.6.1 urllib3-1.26.3
```

EnriqueCepeda:
I've discovered why. I use Zsh as a shell and this shell uses square brackets for pattern matching. Quotation marks are necessary, otherwise Zsh interprets the patterns as part of a pattern instead of a text. 
https://stackoverflow.com/questions/30539798/zsh-no-matches-found-requestssecurity

kanderso-nrel:
Well I was going to express a reluctance to cater our docs to non-standard shells, but wikipedia tells me that zsh is the default shell in MacOS starting from Catalina.  As I imagine other people will run into this issue too, @EnriqueCepeda what do you think about something like `pip install pvlib[optional]   #  or pip install "pvlib[optional]" for MacOS with zsh` so it's a small learning moment for the reader?  Or we could just keep it simple like you've done.

Whatever we do, it probably makes sense to make the same change for the other places `pip install pvlib[...]` appears in the docs. Here's a list, although I'd say only `contributing.rst` and `installation.rst` need the edit -- no need for the old whatsnew files.

```
(pvlib_optional_quotes) C:\Users\KANDERSO\projects\pvlib-python>git grep -nF pvlib[
docs/sphinx/source/contributing.rst:263:    pip install pvlib[doc]
docs/sphinx/source/installation.rst:78:    pip install pvlib[optional]
docs/sphinx/source/installation.rst:86:    `pip install pvlib[optional]`, or you can install pvlib from conda-forge
docs/sphinx/source/installation.rst:248:    pip install pvlib[optional]
docs/sphinx/source/whatsnew/v0.6.0.rst:126:* pvlib now ships with a pvlib[optional] installation option to automatically
docs/sphinx/source/whatsnew/v0.6.0.rst:128:  ``pip install pvlib[optional]``. Additional installation options include
```

wholmgren:
I use oh-my-zsh on both mac and linux machines. I usually use `\` to escape the brackets e.g. `pip install pvlib\[all\]`. I'm fine with a comment in the code or a parenthetical remark in the text.

EnriqueCepeda:
@kanderso-nrel I would go with the KISS option if it works anyways, is simpler for the user

cwhanse:
@kanderso-nrel @wholmgren what's the disposition of this PR?

wholmgren:
I oscillate between -0 and +0 so I'll leave to others to decide the right approach here.

mikofski:
Sorry, I vote no.

It seems like if one uses zsh then they will run into this issue anytime they have square brackets, not just with pvlib. For example if installing some other project with pip extras. So I sorta feel like it's not pvlib's place to take this on in the docs, and it might be a distraction to others who might ask why they need to use quotation marks. Looking at other packages, I did notice that IPython uses single quotation marks in [this example: `pip install '.[test]'`](https://ipython.readthedocs.io/en/stable/install/install.html#installation-from-source). I didn't look very hard but I couldn't find very many other examples. Pip itself does _not_ use any quotation marks in [their example see no.7)](https://pip.pypa.io/en/stable/cli/pip_install/#examples).

kanderso-nrel:
I agree that the pip extras syntax isn't unique to pvlib, but I bet the subset of users reading `installation.rst` are more likely to be seeing it for the first time.  If I had to vote something that isn't ±0 it'd be -1 because google already has plenty of helpful links for this error message: https://www.google.com/search?q=zsh%3A+no+matches+found%3A+pvlib[optional]

Maybe we should add an FAQ page for things like this.

mikofski:
Votes to close, all in favor? (Use the approve button)

wholmgren:
@EnriqueCepeda thank you for the suggestion!

PR #1168, opened 2021-02-15
DC Ohmic Losses for pvsystem and modelchain

ncroft-b4:
- [x] partially addresses #1069 
 - [x] Tests added

Adds functionality for array level losses and adds functions for dc ohmic losses to `pvsystem` and `modelchain`. Replaces #1084 for compatibility with `Array` update.

### DIFF
--- a/docs/sphinx/source/api.rst
+++ b/docs/sphinx/source/api.rst
@@ -331,6 +331,7 @@ Pvsyst model
    temperature.pvsyst_cell
    pvsystem.calcparams_pvsyst
    pvsystem.singlediode
+   pvsystem.dc_ohms_from_percent
 
 PVWatts model
 ^^^^^^^^^^^^^
@@ -389,6 +390,8 @@ Loss models
    :toctree: generated/
 
    pvsystem.combine_loss_factors
+   pvsystem.dc_ohms_from_percent
+   modelchain.ModelChain.dc_ohmic_model
 
 Snow
 ----
@@ -615,6 +618,7 @@ ModelChain properties that are aliases for your specific modeling functions.
    modelchain.ModelChain.aoi_model
    modelchain.ModelChain.spectral_model
    modelchain.ModelChain.temperature_model
+   modelchain.ModelChain.dc_ohmic_model
    modelchain.ModelChain.losses_model
    modelchain.ModelChain.effective_irradiance_model
 
@@ -645,6 +649,8 @@ ModelChain model definitions.
    modelchain.ModelChain.pvsyst_temp
    modelchain.ModelChain.faiman_temp
    modelchain.ModelChain.fuentes_temp
+   modelchain.ModelChain.dc_ohmic_model
+   modelchain.ModelChain.no_dc_ohmic_loss
    modelchain.ModelChain.pvwatts_losses
    modelchain.ModelChain.no_extra_losses
 
--- a/docs/sphinx/source/api.rst
+++ b/docs/sphinx/source/api.rst
@@ -332,6 +332,7 @@ Pvsyst model
    pvsystem.calcparams_pvsyst
    pvsystem.singlediode
    pvsystem.dc_ohms_from_percent
+   pvsystem.dc_ohmic_losses
 
 PVWatts model
 ^^^^^^^^^^^^^
@@ -391,7 +392,6 @@ Loss models
 
    pvsystem.combine_loss_factors
    pvsystem.dc_ohms_from_percent
-   modelchain.ModelChain.dc_ohmic_model
 
 Snow
 ----
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -355,6 +355,11 @@ class ModelChain:
         The ModelChain instance will be passed as the first argument to a
         user-defined function.
 
+    dc_ohmic_model: None, str or function, default None
+        Valid strings are 'dc_ohms_from_percent', 'no_loss'. The ModelChain
+        instance will be passed as the first argument to a user-defined
+        function.
+
     losses_model: str or function, default 'no_loss'
         Valid strings are 'pvwatts', 'no_loss'. The ModelChain instance
         will be passed as the first argument to a user-defined function.
@@ -377,6 +382,7 @@ class ModelChain:
                  airmass_model='kastenyoung1989',
                  dc_model=None, ac_model=None, aoi_model=None,
                  spectral_model=None, temperature_model=None,
+                 dc_ohmic_model=None,
                  losses_model='no_loss', name=None):
 
         self.name = name
@@ -395,7 +401,9 @@ class ModelChain:
         self.spectral_model = spectral_model
         self.temperature_model = temperature_model
 
+        self.dc_ohmic_model = dc_ohmic_model
         self.losses_model = losses_model
+
         self.orientation_strategy = orientation_strategy
 
         self.weather = None
@@ -1057,6 +1065,40 @@ class ModelChain:
 
     def fuentes_temp(self):
         return self._set_celltemp(self.system.fuentes_celltemp)
+
+    @property
+    def dc_ohmic_model(self):
+        return self._dc_ohmic_model
+
+    @dc_ohmic_model.setter
+    def dc_ohmic_model(self, model):
+        if model is None:
+            self._dc_ohmic_model = self.no_dc_ohmic_loss
+        elif isinstance(model, str):
+            model = model.lower()
+            if model == 'dc_ohms_from_percent':
+                self._dc_ohmic_model = self.dc_ohms_from_percent
+            elif model == 'no_loss':
+                self._dc_ohmic_model = self.no_dc_ohmic_loss
+            else:
+                raise ValueError(model + ' is not a valid losses model')
+        else:
+            self._dc_ohmic_model = partial(model, self)
+
+    def dc_ohms_from_percent(self):
+        """
+        Calculate time series of ohmic losses and apply those to the mpp power
+        output of the `dc_model` based on the pvsyst equivalent resistance
+        method. Uses a `dc_ohmic_percent` parameter in the `losses_parameters`
+        of the PVsystem.
+        """
+        Rw = self.system.dc_ohms_from_percent()
+        self.dc_ohmic_losses = pvsystem.dc_ohmic_losses(Rw, self.dc['i_mp'])
+        self.dc['p_mp'] = self.dc['p_mp'] - self.dc_ohmic_losses
+        return self
+
+    def no_dc_ohmic_loss(self):
+        return self
 
     @property
     def losses_model(self):
@@ -1728,6 +1770,7 @@ class ModelChain:
         """
         self._prepare_temperature(data)
         self.dc_model()
+        self.dc_ohmic_model()
         self.losses_model()
         self.ac_model()
 
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -1103,8 +1103,19 @@ class ModelChain:
         of the PVsystem.
         """
         Rw = self.system.dc_ohms_from_percent()
-        self.dc_ohmic_losses = pvsystem.dc_ohmic_losses(Rw, self.dc['i_mp'])
-        self.dc['p_mp'] = self.dc['p_mp'] - self.dc_ohmic_losses
+        if isinstance(self.results.dc, tuple):
+            self.dc_ohmic_losses = tuple(
+                pvsystem.dc_ohmic_losses(Rw, df['i_mp'])
+                for Rw, df in zip(Rw, self.results.dc)
+            )
+            for df, loss in zip(self.results.dc, self.dc_ohmic_losses):
+                df['p_mp'] = df['p_mp'] - loss
+        else:
+            self.dc_ohmic_losses = pvsystem.dc_ohmic_losses(
+                Rw, self.results.dc['i_mp']
+            )
+            self.results.dc['p_mp'] = (self.results.dc['p_mp']
+                                       - self.dc_ohmic_losses)
         return self
 
     def no_dc_ohmic_loss(self):
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -382,7 +382,7 @@ class ModelChain:
         The ModelChain instance will be passed as the first argument to a
         user-defined function.
 
-    dc_ohmic_model: None, str or function, default None
+    dc_ohmic_model: str or function, default 'no_loss'
         Valid strings are 'dc_ohms_from_percent', 'no_loss'. The ModelChain
         instance will be passed as the first argument to a user-defined
         function.
@@ -1082,9 +1082,7 @@ class ModelChain:
 
     @dc_ohmic_model.setter
     def dc_ohmic_model(self, model):
-        if model is None:
-            self._dc_ohmic_model = self.no_dc_ohmic_loss
-        elif isinstance(model, str):
+        if isinstance(model, str):
             model = model.lower()
             if model == 'dc_ohms_from_percent':
                 self._dc_ohmic_model = self.dc_ohms_from_percent
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -429,8 +429,6 @@ class ModelChain:
         self.dc_ohmic_model = dc_ohmic_model
         self.losses_model = losses_model
 
-        self.orientation_strategy = orientation_strategy
-
         self.weather = None
         self.times = None
 
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -1082,9 +1082,9 @@ class ModelChain:
 
     @dc_ohmic_model.setter
     def dc_ohmic_model(self, model):
-        # if model is None:
-        #     self._dc_ohmic_model = self.no_dc_ohmic_loss
-        if isinstance(model, str):
+        if model is None:
+            self._dc_ohmic_model = self.no_dc_ohmic_loss
+        elif isinstance(model, str):
             model = model.lower()
             if model == 'dc_ohms_from_percent':
                 self._dc_ohmic_model = self.dc_ohms_from_percent
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -273,7 +273,8 @@ class ModelChainResult:
     _singleton_tuples: bool = field(default=False)
     _per_array_fields = {'total_irrad', 'aoi', 'aoi_modifier',
                          'spectral_modifier', 'cell_temperature',
-                         'effective_irradiance', 'dc', 'diode_params'}
+                         'effective_irradiance', 'dc', 'diode_params',
+                         'dc_ohmic_losses'}
 
     # system-level information
     solar_position: Optional[pd.DataFrame] = field(default=None)
@@ -293,6 +294,7 @@ class ModelChainResult:
     dc: Optional[PerArray[Union[pd.Series, pd.DataFrame]]] = \
         field(default=None)
     diode_params: Optional[PerArray[pd.DataFrame]] = field(default=None)
+    dc_ohmic_losses: Optional[PerArray[pd.Series]] = field(default=None)
 
     def _result_type(self, value):
         """Coerce `value` to the correct type according to
@@ -407,7 +409,7 @@ class ModelChain:
                  airmass_model='kastenyoung1989',
                  dc_model=None, ac_model=None, aoi_model=None,
                  spectral_model=None, temperature_model=None,
-                 dc_ohmic_model=None,
+                 dc_ohmic_model='no_loss',
                  losses_model='no_loss', name=None):
 
         self.name = name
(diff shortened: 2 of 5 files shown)


pvlib/pvsystem.py
```diff
@@ -1149,6 +1166,9 @@ class Array:
         Valid strings are 'open_rack', 'close_mount', and 'insulated_back'.
         Used to identify a parameter set for the SAPM cell temperature model.
 
+    array_losses_parameters: None, dict or Series, default None.
+        Supported keys are dc_ohmic_percent.
```
cwhanse, line 1170:
```suggestion
        Supported keys are 'dc_ohmic_percent'.
```

pvlib/pvsystem.py
```diff
@@ -2810,6 +2858,71 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    .. math::
+        Rw = (L_{stc} / 100) * (Varray / Iarray)
+
+    :math:`L_{stc}` is the input dc loss as a percent, e.g. 1.5% loss is
+    input as 1.5
+
+    Parameters
+    ----------
+    V_mp_ref: numeric
```
cwhanse, line 2878: Each parameter needs a definition and a unit. E.g.,
```
v_mp_ref: numeric
    Voltage at maximum power in reference conditions [V]
```

I have a slight preference for lower case parameter names here. `V_mp_ref` is the key for the appropriate field in the CEC module database but I don't see that we need to use the exact same name for the function's parameter. The pvsyst module model in pvlib doesn't use that key value, but V_mp_ref appears in the Pvsyst documentation. SAPM calls it `Vmpo`.

pvlib/pvsystem.py
```diff
@@ -2810,6 +2858,71 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    Rw = (dc_ohmic_percent / 100) * (vmp / imp)
+
+    return Rw
+
+
+def dc_ohmic_losses(ohms, current):
+    """
+    Returns ohmic losses in in units of power from the equivalent
+    resistance of of the wires and the operating current.
```
cwhanse, line 2906:
```suggestion
    Returns ohmic losses in units of power from the equivalent
    resistance of the wires and the operating current.
```

pvlib/pvsystem.py
```diff
@@ -2810,6 +2858,71 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    Parameters
+    ----------
+    ohms: numeric, float
+    current: numeric, float or array-like
```
cwhanse, line 2911:
```suggestion
    ohms: numeric
        Equivalent resistance of wires [ohm]
    current: numeric, float or array-like
        Operating current [A]
```

pvlib/pvsystem.py
```diff
@@ -2810,6 +2858,71 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    Returns
+    ----------
+    numeric
+        Single or array-like value of the losses in units of power
```
cwhanse, line 2916:
```suggestion
    loss: numeric
        Power loss [W]
```

pvlib/pvsystem.py
```diff
@@ -1372,6 +1398,28 @@ def get_iam(self, aoi, iam_model='physical'):
...
+        `self.module_parameters["V_mp_ref"]`,
+        `self.module_parameters["I_mp_ref"]`,
+        `self.modules_per_string`, and `self.strings_per_inverter`.
+        See :py:func:`pvlib.pvsystem.dc_ohms_from_percent` for details.
+        """
+
+        kwargs = _build_kwargs(['dc_ohmic_percent'],
+                               self.array_losses_parameters)
+
+        kwargs.update(_build_kwargs(['V_mp_ref', 'I_mp_ref'],
```
cwhanse, line 1415: I wonder if there is a way to extend this to the other DC module models. If `module_parameters` is taken from the CEC database, it will include `V_mp_ref` and `I_mp_ref`.

The other cases are Pvsyst, SAPM and PVWatts.

For Pvsyst, pvlib provides no parameter database. The pvsyst DC power functions don't require V_mp_ref or I_mp_ref, so the user would need to know that these must be added.

For SAPM, the expected names are `Vmpo` and `Impo`.

For PVWatts, the ohmic losses model can't be applied since PVWatts only deals with power (not voltage and current separately).

`V_mp_ref`, `I_mp_ref` and `dc_ohmic_percent` are required by the downstream method. Maybe raise a ValueError here if any are missing at this step.

In the ModelChain `dc_ohms_from_percent` method, we could check `dc_model` and raise a more informative message how to correct the ValueError.

pvlib/pvsystem.py
```diff
@@ -2810,6 +2858,71 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+    Calculates the equivalent resistance of the wires from a percent
+    ohmic loss at STC.
+
+    Equivalent resistance is calculated with the function:
+
+    .. math::
+        Rw = (L_{stc} / 100) * (Varray / Iarray)
+
+    :math:`L_{stc}` is the input dc loss as a percent, e.g. 1.5% loss is
+    input as 1.5
```
kanderso-nrel, line 2874: I suggest describing the other three variables here as well (maybe briefly saying `V_array` is `mods_per_string * v_mp_mod`, and the equivalent for `I_array`), and maybe moving the 1.5% example to the parameter description section below.

pvlib/pvsystem.py
```diff
@@ -2810,6 +2858,71 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    Returns
+    ----------
+    Rw: numeric
+        Equivalent resistance in ohms
+
+    References
+    ----------
+    -- [1] PVsyst 7 Help. "Array ohmic wiring loss".
+            https://www.pvsyst.com/help/ohmic_loss.htm
```
kanderso-nrel, line 2892:
```suggestion
    .. [1] PVsyst 7 Help. "Array ohmic wiring loss".
       https://www.pvsyst.com/help/ohmic_loss.htm
```

Renders better on readthedocs: https://pvlib-python--1168.org.readthedocs.build/en/1168/generated/pvlib.pvsystem.dc_ohms_from_percent.html

pvlib/pvsystem.py
```diff
@@ -2810,6 +2858,71 @@ def pvwatts_losses(soiling=2, shading=3, snow=0, mismatch=2, wiring=2,
...
+
+    Returns
+    ----------
+    numeric
+        Single or array-like value of the losses in units of power
+
+    References
+    ----------
+    -- [1] PVsyst 7 Help. "Array ohmic wiring loss".
+            https://www.pvsyst.com/help/ohmic_loss.htm
```
kanderso-nrel, line 2921: Same formatting here

kanderso-nrel (commented):
A few quick comments about docstring formatting

pvlib/pvsystem.py
ncroft-b4: Ok. I need to sort out the relations between the DC models and the module parameters, which have always been a bit confusing, I admit. I would like to make this applicable for CEC, SAPM, and PVsyst models. Sounds like for SAPM, I can add the option to use Vmpo and Impo if those are present in the module parameters. But for PVsyst, it may require a different workflow, or like you say, to raise a ValueError to say that those parameters need to be added. I will look into this to add an update.

cwhanse: Do you know if the Pvsyst software names these two quantities? If so, I'd use those names. A user will have to add them to `module_parameters` regardless since there's no data source from which to inherit the names.

pvlib/modelchain.py
```diff
@@ -1068,6 +1076,40 @@ def faiman_temp(self):
...
+
+    def dc_ohms_from_percent(self):
+        """
+        Calculate time series of ohmic losses and apply those to the mpp power
+        output of the `dc_model` based on the pvsyst equivalent resistance
+        method. Uses a `dc_ohmic_percent` parameter in the `losses_parameters`
+        of the PVsystem.
+        """
+        Rw = self.system.dc_ohms_from_percent()
+        self.dc_ohmic_losses = pvsystem.dc_ohmic_losses(Rw, self.dc['i_mp'])
```
ncroft-b4, line 1106: I just realized this might need to be rewritten to handle the case where there are multiple arrays and so `dc_ohms_from_percent` returns a tuple instead of a single value. In that case what will `self.dc[i_mp]` be? Commenting to mark to investigate and fix.

cwhanse: `self.results.dc` will be a tuple of `DataFrame`.

pvlib/pvsystem.py
ncroft-b4: Added definitions and units and changed to lowercase generic parameters vmp_ref and imp_ref.

pvlib/pvsystem.py
ncroft-b4: Changed the parameter from ohms to resistance, as suggested.

pvlib/pvsystem.py
ncroft-b4: I added a section that checks from and adds the parameters in sets of either `V_mp_ref`, `I_mp_ref`; `Vmpo`, `Impo`; or `Vmpp`, Impp` (pvsyst-like, from the documentation and pan file parameters). It should raise a value error if these are not found. If this looks like a workable solution, should more tests be added for each case?

pvlib/modelchain.py
ncroft-b4: Thanks. I added functionality for multiple arrays in the latest commit. I think all of the comments are addressed now and it is ready for the review and push forward.

pvlib/modelchain.py
```diff
@@ -1068,6 +1076,51 @@ def faiman_temp(self):
...
+    def dc_ohms_from_percent(self):
+        """
+        Calculate time series of ohmic losses and apply those to the mpp power
+        output of the `dc_model` based on the pvsyst equivalent resistance
+        method. Uses a `dc_ohmic_percent` parameter in the `losses_parameters`
+        of the PVsystem.
+        """
+        Rw = self.system.dc_ohms_from_percent()
+        if isinstance(self.results.dc, tuple):
+            self.dc_ohmic_losses = tuple(
```
wholmgren, line 1107: Do we want to assign this to `self.results.dc_ohmic_losses`? If yes, we'll need to add it to the `ModelChainResults` class. If not, we should drop `self.` and leave it as a variable that's just used within this function.

docs/sphinx/source/api.rst
```diff
@@ -389,6 +390,8 @@ Loss models
    :toctree: generated/
 
    pvsystem.combine_loss_factors
+   pvsystem.dc_ohms_from_percent
+   modelchain.ModelChain.dc_ohmic_model
```
wholmgren, line 394: no need for the modelchain entry here

docs/sphinx/source/api.rst
```diff
@@ -331,6 +331,7 @@ Pvsyst model
    temperature.pvsyst_cell
    pvsystem.calcparams_pvsyst
    pvsystem.singlediode
+   pvsystem.dc_ohms_from_percent
```
wholmgren, line 334: also `pvsystem.dc_ohmic_losses`

pvlib/pvsystem.py
```diff
@@ -1372,6 +1398,67 @@ def get_iam(self, aoi, iam_model='physical'):
...
+            `self.module_parameters["Vmpo"]`,
+            `self.module_parameters["Impo"]`,
+        PVsyst-like or other:
+            `self.module_parameters["Vmpp"]`,
+            `self.module_parameters["Impp"]`,
+
+        Other array parameters that are used are:
+        `self.losses_parameters["dc_ohmic_percent"]`,
+        `self.modules_per_string`, and
+        `self.strings`.
```
wholmgren, line 1421: Can you add some blank lines to improve how this is rendered? rst is finicky about this kind of thing. I think you'll want a blank line after every `:`. You'll also need `*` if you want the parameters to appear in a list. For example:

```
        Makes use of array module parameters according to the
        following DC models:

        CEC:

            * `self.module_parameters["V_mp_ref"]`,
            * `self.module_parameters["I_mp_ref"]`,
```

pvlib/pvsystem.py
```diff
@@ -1372,6 +1398,67 @@ def get_iam(self, aoi, iam_model='physical'):
...
+
+        See :py:func:`pvlib.pvsystem.dc_ohms_from_percent` for more details.
+        """
+
+        kwargs = _build_kwargs(['dc_ohmic_percent'],
+                               self.array_losses_parameters)
```
wholmgren, line 1427: Can you change this to `dc_ohmic_percent = self.array_losses_parameters['dc_ohmic_percent']`? Or maybe `dc_ohmic_percent = self.array_losses_parameters.get('dc_ohmic_percent', 0)` if you want to provide a default value. 

The issue is that we have to have a value for `dc_ohmic_percent`, but `_build_kwargs` will return an empty dictionary if it doesn't find the key in the dictionary. That's helpful in other methods that we use `_build_kwargs` so that we avoid needing to provide default values to functions, but it's not what we want here. Sorry it's not clear in the other usages of `_build_kwargs`.

pvlib/pvsystem.py
```diff
@@ -1372,6 +1398,67 @@ def get_iam(self, aoi, iam_model='physical'):
...
+
+        # add relevent Vmp and Imp parameters from CEC parameters
+        if all([elem in self.module_parameters
+                for elem in ['V_mp_ref', 'I_mp_ref']]):
+            kwargs.update({'vmp_ref': self.module_parameters['V_mp_ref'],
+                           'imp_ref': self.module_parameters['I_mp_ref']})
```
wholmgren, line 1433: here I would prefer to see explicit assignment (`vmp_ref = `) instead of bundling in kwargs because these are required parameters.

pvlib/pvsystem.py
```diff
@@ -1372,6 +1398,67 @@ def get_iam(self, aoi, iam_model='physical'):
...
+            raise ValueError('Parameters for Vmp and Imp could not be found '
+                             'in the array module parameters. Module '
+                             'parameters must include one set of '
+                             '{"V_mp_ref", "I_mp_Ref"}, '
+                             '{"Vmpo", "Impo"}, or '
+                             '{"Vmpp", "Impp"}.'
+                             )
+
+        kwargs.update({'modules_per_string': self.modules_per_string,
+                       'strings': self.strings})
```
wholmgren, line 1458: continuing the kwargs theme... this would become `kwargs = {'modules_per_string'...`

pvlib/pvsystem.py
```diff
@@ -1372,6 +1398,67 @@ def get_iam(self, aoi, iam_model='physical'):
...
+
+        return dc_ohms_from_percent(**kwargs)
```
wholmgren, line 1460: and finally `return dc_ohms_from_percent(vmp_ref, imp_ref, dc_ohmic_percent, **kwargs)`

wholmgren (commented):
Thanks @ncroft-b4. Minor items below. Nice job with the tests.

pvlib/modelchain.py
ncroft-b4: I think it should be assigned to the results. Otherwise there is no record of what power was taken out of `p_mp` and no way to confirm, test, or just review the losses after the fact.

It looks like `pvwatts_losses` still creates a `self.losses` attribute. Do we want to think about coordinating the storage of those losses with other losses like `dc_ohmic_losses` now and any others that get added later? Or treat each separately?

wholmgren: > It looks like pvwatts_losses still creates a self.losses attribute. 

Thanks for catching that. `pvwatts_losses` and `no_extra_losses` should assign values to `ModelChainResults`, and `ModelChain._deprecated_attrs` should include `losses`. We can address that in another issue/pr.

> Do we want to think about coordinating the storage of those losses with other losses like dc_ohmic_losses now and any others that get added later? Or treat each separately?

Good question. I'd lean toward treating them separately for now unless we have a good idea of how the losses models will evolve in the future.

ncroft-b4: Sounds good. I will work on adding `dc_ohmic_losses` to `ModelChainResults` in the next commit.

cwhanse: >  I'd lean toward treating them separately for now

I agree. If `losses` attributes pile up in `ModelChainResults` we can consider creating a `Losses` dataclass or other mechanism to group losses together.

pvlib/modelchain.py
```diff
@@ -1068,6 +1076,51 @@ def faiman_temp(self):
...
         return self._set_celltemp(self.system.fuentes_celltemp)
 
+    @property
+    def dc_ohmic_model(self):
+        return self._dc_ohmic_model
+
+    @dc_ohmic_model.setter
+    def dc_ohmic_model(self, model):
+        if model is None:
+            self._dc_ohmic_model = self.no_dc_ohmic_loss
```
wholmgren, line 1086: Can you delete the `None` case here and in the docstring? In the `ModelChain` initialization for most models, a parameter with a value of `None` will result in `ModelChain` attempting to infer the model. Here I think we should follow the losses_model pattern and just not support None:

```
    losses_model: str or function, default 'no_loss'
        Valid strings are 'pvwatts', 'no_loss'. The ModelChain instance
        will be passed as the first argument to a user-defined function.
```

Perhaps we'll add inference methods for these parameters in the future that are kicked off when a user supplies None, but I think that's out of scope here.

pvlib/modelchain.py
```diff
@@ -380,6 +382,11 @@ class ModelChain:
         The ModelChain instance will be passed as the first argument to a
         user-defined function.
 
+    dc_ohmic_model: None, str or function, default None
```
wholmgren, line 385:
```suggestion
    dc_ohmic_model: str or function, default 'no_loss'
```

wholmgren: see more discussion in comment in setter method

pvlib/modelchain.py
```diff
@@ -1068,6 +1076,51 @@ def faiman_temp(self):
     def fuentes_temp(self):
         return self._set_celltemp(self.system.fuentes_celltemp)
 
+    @property
+    def dc_ohmic_model(self):
+        return self._dc_ohmic_model
```
wholmgren, line 1081: any idea why codecov says this line isn't tested? It looks like it should be.

wholmgren (commented):
almost there...

pvlib/modelchain.py
ncroft-b4: That was confusing for me as well. And now it is telling me that blank lines 1082 and 1095 are not covered, and that 1090 is not covered which it should be as there is a test specifically for `dc_ohmic_model='no_loss'`. I'd welcome an assist from someone more familiar with the tests on what might be leading to the failed coverage if anyone has an idea.

wholmgren: I ran a coverage report locally and cannot reproduce the codecov results. Adding exceptions to these lines causes errors, as you would expect.

pvlib/pvsystem.py
```diff
@@ -1149,6 +1166,9 @@ class Array:
         Valid strings are 'open_rack', 'close_mount', and 'insulated_back'.
         Used to identify a parameter set for the SAPM cell temperature model.
 
+    array_losses_parameters: None, dict or Series, default None.
```
wholmgren, line 1169: sorry, should have caught this earlier.... it seems repetitive to call this parameter `array_losses_parameters` since it's already on an `Array`. I suppose there is some risk of user confusion if this and the `PVSystem` `losses_parameters` are called the same thing, but it doesn't hurt to have extra keys in these dictionaries. It's unlikely that would change in the future.

Maybe we should live with the current API on master in an alpha release or two and then maybe take up renaming in a follow up issue. I could see something like #1176 providing more clarity on this.

Thoughts @ncroft-b4 @cwhanse ?

wholmgren (approved):

pvlib/pvsystem.py
cwhanse: I'm OK with either `array_losses_parameters` or `Array.losses_parameters` (and `PVSystem.losses_parameters`) in this PR.

I think as loss models are added to pvlib and we work those models into ModelChain, we'll end up refactoring the parameter containers.

cwhanse (approved):

pvlib/pvsystem.py
ncroft-b4: Yeah I think my concern was making it distinguishable from the `PVSystem` parameter. Also, to keep compatibility with initializing a `PVSystem` directly, there needs to be a difference between the `losses_parameters` that would be passed into the `Array` and the ones that would stay with the `PVSystem`. 

That makes me realize that I added `array_losses_parameters` to the list of keys that get passed to `Array` from `PVSystem` if an `Array` is not specified, but I did not add `array_losses_parameters` explicitly to the documentation for `PVSystem` or to the `init`. That makes sense to be because `array_losses_parameters` is only an `Array` parameter, but it does make it different from the other `Array` parameters that are still part of `PVSystem` by legacy. I'm good with it as it is but if you want to keep it uniform with the legacy parameters we could change it.

wholmgren:
Ok, let's move on for now. Thanks @ncroft-b4 for the contribution and your patience as we worked through it!

wholmgren:
and after I scrolled to the top to go to my notifications I realized that we don't have a whats new entry. Sorry. I can add that in another PR.

adriesse:
I have just quickly scrolled through this and wanted to mention something that may be relevant to someone sometime: PVsyst models DC ohmic losses by increasing the series resistance in the single-diode model parameters.  This seems like a good approach to me, and I suppose one can still do that independently of this new code.

PR #1172, opened 2021-02-20
Add cams.get_mcclear function

AdamRJensen:
- [ ] Closes #xxxx
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [ ] Tests added
 - [x] Updates entries to [`docs/sphinx/source/api.rst`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes.
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [x] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [ ] Pull request is nearly complete and ready for detailed review.
 - [ ] Maintainer: Appropriate GitHub Labels and Milestone are assigned to the Pull Request and linked Issue.

Add a function to retrieve CAMS McClear clear-sky radiation as previously unsuccessfully attempted in #271, #274, #279.

PR #279, opened 2016-12-05
iotools: reader for maccrad

dacoex:
a separate PR for MACC-RAD as requested in https://github.com/pvlib/pvlib-python/pull/274#issuecomment-264486119

- only 1 example file
- renamed to util

Discussion in #271

PR #1175, opened 2021-02-22
Add sodapro.get_cams_radiation

AdamRJensen:
- [ ] Closes #xxxx
 - [x] Tests added

Add a function to retrieve CAMS McClear clear-sky radiation as previously unsuccessfully attempted in #271, #274, #279. This is a new pull request as my previous attempt #1172 erroneously had a long/incorrect commit history.


AdamRJensen:
@mikofski Thanks for the help in making a new pull request. The previous PR #1175 has now been closed.

As previously mentioned: I've asked the developing team of the CAMS radiation and McClear services, and have been informed that there is not and will not be any demo API email/access. So if tests have to be implemented, we need to register a PVLIB email. Only a registered email address is necessary and no password will ever be public.

The function includes some very convenient features, including the label, integrated, and map_variable arguments, making it very intuitive for Python users. I'd very much like a critical look at the function and feedback on how it can be improved.

AdamRJensen:
@mikofski As far as I can tell [CAMS McClear](http://www.soda-pro.com/web-services/radiation/cams-mcclear/info) and ECMWF MACC [pvlib.iotools.ecmwf_macc](https://github.com/pvlib/pvlib-python/blob/master/pvlib/iotools/ecmwf_macc.py) return very different parameters.

wholmgren:
I'm not very familiar with the details of CAMS services, but it appears the major differences between this McClear data reader and the existing MACC reader are in spectral components of AOD. McClear returns AOD at 550 nm for a variety of particles, while MACC returns AOD at a variety of wavelengths. So I think they are complementary and both readers can exist in pvlib.

The CAMS data license (available [here](http://www.soda-pro.com/web-services/radiation/cams-mcclear)) allows us to bundle a file for testing. 

How about one function for parsing data and one function for fetching data? The fetch function can call the parse function so a user only needs to make one function call. But separating the logic lets users load data in a different manner if desired (e.g. from disk or using threads).

AdamRJensen:
@wholmgren I will separate it into a get and read/parse function. Any preference in calling it read or parse? Is there a difference?

The nice part about having it in one function is that you know if the file has 6 (verbose=False) or 23 columns (verbose=True). I suppose this can be solved by reading in the file, and then checking how many columns there are, and based on this assign the 6 or 23 column name header.

The primary output of CAMS McClear is the time-series of broad-band clear sky irradiance. This clear-sky model can be used for a number of purposes, e.g. quality checks, etc., and has the major advantage over most clear-sky models in that it is based on data for the specific location and the specific time and not just seasonal trends. The ECMWF MACC reader only returns AOD for different wavelengths and water vapor, which has very different uses.

wholmgren:
> Any preference in calling it read or parse? Is there a difference?

I'm guessing parse. Check out `psm3.py` for an example of a module with a `get_psm3` (downloads data from remote source, passes it to parse), `parse_psm3` that converts the buffered data into a well-formatted DataFrame, and finally `read_psm3` that simplifies usage for local files. 

#1155 has some recent discussion relevant to the benefits of splitting up fetching and parsing.

> The nice part about having it in one function is that you know if the file has 6 (verbose=False) or 23 columns (verbose=True). I suppose this can be solved by reading in the file, and then checking how many columns there are, and based on this assign the 6 or 23 column name header.

That could work, but you could also keep it simple and add that flag to the read/parse function too. Most people use the higher level function for direct downloads + parsing and it's not a big deal to make users of the lower level parse function repeat a parameter.

AdamRJensen:
@wholmgren Ok, I will attempt to make a similar structure, with parse, read, and get. Or should we stick to just parse and get?
This function differs from psm3, in that arguments are needed to get passed from the get function to the read function and then to the parse function: particularly the integrated, label, and map_variables arguments.

wholmgren:
I don't see that it's an issue to pass any of those arguments from one function to another or to expect that a user with local data already knows something about that data. But I don't want to over complicate it - iotools is supposed to be relatively easy to contribute to. It's really fine with me if you want to keep it all in one function.

AdamRJensen:
@wholmgren I have now created read and parse sub-functions similar to psm3. It is indeed nice that it works with can retrieve files but also work with local files now.

Also, I found a nice way to parse through the meta-data nicely and detect the number of columns and the names from the file itself.

The documentation for the two new functions is not fully developed, but I will do this within the next few days. I'll also be working on making tests. Though you are welcome to look through / test the function now.

AdamRJensen:
@wholmgren I have renamed the function to 'get_cams_radiation' as I have extended the functions to read/parse/get both (CAMS Radiation)[http://www.soda-pro.com/web-services/radiation/cams-radiation-service/info] and (McClear)[http://www.soda-pro.com/web-services/radiation/cams-mcclear/info] as the two services follow the same format (API input is identical, with the exception of one string, specifying whether the users wants Radiation or McClear).

I have been looking into developing tests, but I am very uncertain about which aspects need to be tested? Do we want to register a pvlib email and test the 'get' method by making a real API call?

wholmgren:
The parse and read functions can be tested without remote access, right? For this PR, I'd be ok just testing those as long as there's a follow up issue for testing the get function.

Real API calls to flaky services cause many of our test failures, so I'd like to avoid more of that. But the get function can still be tested using either `pytest-mock` or `requests-mock`. Here's an example of using `pytest-mock` with a pvlib io function:

https://github.com/pvlib/pvlib-python/blob/8b98768818ee5ad85d9479877533651a2e9dc2cd/pvlib/tests/iotools/test_midc.py#L78-L89

Here are a few examples of using `requests-mock` for an io function that accesses the NREL PVDAQ service: https://github.com/SolarArbiter/solarforecastarbiter-core/blob/92197e62eb356aa50a3f65364f04a6b28aae30f2/solarforecastarbiter/io/fetch/tests/test_pvdaq.py

I have a very slight preference for the `requests-mock` pattern, but it's not currently used in pvlib and we'd need to add it to [setup.py](https://github.com/pvlib/pvlib-python/blob/8b98768818ee5ad85d9479877533651a2e9dc2cd/setup.py#L51-L52) and the [ci config files](https://github.com/pvlib/pvlib-python/tree/master/ci).

pvlib/iotools/cams.py
```diff
@@ -0,0 +1,357 @@
...
+    .. [2] `CAMS McClear Service Info
+       <http://www.soda-pro.com/web-services/radiation/cams-mcclear/info>`_
+    .. [3] `CAMS McClear Automatic Access
+       <http://www.soda-pro.com/help/cams-services/cams-mcclear-service/automatic-access>`_
+    """
+    if time_step in TIME_STEPS_MAP.keys():
+        time_step_str = TIME_STEPS_MAP[time_step]
+    else:
+        warnings.warn('Time step not recognized, 1 hour time step used!')
+        time_step, time_step_str = '1h', 'PT01H'
```
wholmgren, line 167:
```suggestion
    try:
        time_step_str = TIME_STEPS_MAP[time_step]
    except KeyError:
        raise ValueError(f'Time step not recognized. Must be one of {list(TIME_STEPS_MAP.keys())}')
```

pvlib/iotools/cams.py
```diff
@@ -0,0 +1,357 @@
...
+
+    if (verbose is True) & ((time_step != '1min') | (time_ref != 'UT')):
+        verbose = False
+        warnings.warn("Verbose mode only supports 1 min. UT time series!")
```
wholmgren, line 171: I'd probably also raise a ValueError here, but that's debatable.

pvlib/iotools/cams.py
```diff
@@ -0,0 +1,357 @@
...
+    service = 'get_{}'.format(service.lower())  # Format CAMS service string
+
+    # Manual format the request url, due to uncommon usage of & and ; in url
+    url = ("http://{}/service/wps?Service=WPS&Request=Execute&"
+           "Identifier={}&version=1.0.0&RawDataOutput=irradiation&"
+           "DataInputs=latitude={};longitude={};altitude={};"
+           "date_begin={};date_end={};time_ref={};summarization={};"
+           "username={};verbose={}"
+           ).format(server, service, latitude, longitude, altitude, start_date,
+                    end_date, time_ref, time_step_str, email, verbose)
```
wholmgren, line 193: can you reformat into one or more f-strings? It's hard to keep track of this many parameters passed to format.

That format is a bit of a pain. What about something like

```python
base_url = f"http://{server}/service/wps"
data_inputs = (
    f"latitude={latitude};longitude={longitude};altitude={altitude};"
    f"date_begin={start_date};date_end={end_date};time_ref={time_ref};"
    f"summarization={time_step_str};username={email};verbose={verbose}"
params = {
    'Service': WPS,
    'Request': 'Execute',
    'Identifier': service,
    'version': 1.0.0,
    'RawDataOutput': 'irradiation',
    'DataInputs': data_inputs
}

res = requests.get(base_url, params=params)
```

pvlib/iotools/cams.py
```diff
@@ -0,0 +1,357 @@
...
+
+    res = requests.get(url)
+
+    # Invalid requests returns helpful XML error message
+    if res.headers['Content-Type'] == 'application/xml':
+        errors = res.text.split('ows:ExceptionText')[1][1:-2]
+        raise requests.HTTPError(errors, response=res)
```
wholmgren, line 200: can we use [`raise_for_status`](https://requests.readthedocs.io/en/master/api/#requests.Response.raise_for_status) instead of the manual error handling here?

pvlib/iotools/cams.py
```diff
@@ -0,0 +1,357 @@
...
+    # Check if returned file is a csv data file
+    elif res.headers['Content-Type'] == 'application/csv':
+        fbuf = io.StringIO(res.content.decode('utf-8'))
+        data, meta = parse_cams_radiation(fbuf, integrated=integrated,
+                                          label=label,
+                                          map_variables=map_variables)
+        return data, meta
+    else:
+        warnings.warn('File content type not recognized.')
```
wholmgren, line 209: I would hope that it's not possible for a user to get here and we don't need it. If we do need it then it should be an exception instead of a warning.

wholmgren (commented):
Looking good!

pvlib/iotools/cams.py
AdamRJensen: @wholmgren Unfortunately, this does not work, even though it is indeed much cleaner and understandable. When the request.get function formats the URL, it converts any equal signs in the parameter values to %3D and semi-colons to %3B. I found this by looking at the generated URL (res.url).

E.g. the first part of the data input section of the URL needs to be "&DataInputs=latitude=2.9;longitude=-54.0;altitude=-999;", but it is instead converted by the request.get function into "&DataInputs=latitude%3D2.9%3Blongitude%3D-54.0%3Baltitude%3D-999%3B" which is not recognized by the API.

However, adding the datainputs manually to the base URL and letting request.get format part of it works:

    base_url = f"http://{server}/service/wps"

    data_inputs = (
        f"latitude={latitude};longitude={longitude};altitude={altitude};"
        f"date_begin={start_date};date_end={end_date};time_ref={time_ref};"
        f"summarization={time_step_str};username={email};verbose={verbose}")

    params = {
        'Service': 'WPS',
        'Request': 'Execute',
        'Identifier': service,
        'version': '1.0.0',
        'RawDataOutput': 'irradiation',
        ####'DataInputs': data_inputs ####  this line will be deleted
        }

    res = requests.get(base_url + '?DataInputs=' + data_inputs, params=params)

What do you think of that?

pvlib/iotools/cams.py
AdamRJensen: Ok, I will remove this.

pvlib/iotools/cams.py
AdamRJensen: The API does not return HTTP errors as you would expect from normal services. If a request is invalid, an XML file is returned instead of a csv file and the service fails at throwing any HTTP errors (invalid request return the HTTP status code 200 and res.ok = True).

Line 199 extracts the error message embedded in the XML message.

    errors = res.text.split('ows:ExceptionText')[1][1:-2]

pvlib/iotools/cams.py
AdamRJensen: I would prefer to only raise a warning, so if this is ok, I will leave it as it is.

AdamRJensen:
@wholmgren Yes, the parse and read functions can be tested without remote access.

How many test files should I upload: cover different time steps (1h, 1M), verbose (false/true), mcclear vs. cams_radiation? Should the label and column remapping function also be tested? Index, column names, dtypes and/or values?

pvlib/iotools/cams.py
wholmgren: looks like a good solution

pvlib/iotools/cams.py
wholmgren: Ah, that is unfortunate. Please expand the comment to state that the API still returns HTTP status code 200 so `res.raise_for_status()` will not work.

wholmgren:
> How many test files should I upload: cover different time steps (1h, 1M), verbose (false/true), mcclear vs. cams_radiation? 

I'm not too concerned about different time steps. Probably worth testing false/true with both mcclear and cams.

> Should the label and column remapping function also be tested? Index, column names, dtypes and/or values?

I'm guessing you'll already be testing these some extent and it's hard for me to say what else needs to be done until another review. If you can keep the data files short then it's pretty reasonable to manually create the expected dataframe and use `assert_frame_equal`.

pvlib/iotools/cams.py
AdamRJensen: This has been done in the new commit "Reworked URL formatting & comments on HTTP status"

AdamRJensen:
@wholmgren I've added test, using the assert_frame_equal method for four different test files (1min and monthly for both verbose True and False).

The pull-request is still failing codecov, which I assume is because the get function is not being tested. Should anything be done to silence codecov?

pvlib/iotools/cams.py
wholmgren: so users will see this warning and then hit `raise requests.HTTPError(errors, response=res)` on line 210?

wholmgren (commented):
> The pull-request is still failing codecov, which I assume is because the get function is not being tested. Should anything be done to silence codecov?

The get function is doing enough parsing that it's probably worth testing it. I am still -0.5 on testing against the live api and +1 on testing using mock or requests-mock. I will probably use this new functionality, so I am willing to write those mock tests in a follow up PR since we haven't yet established the pattern in pvlib.

I'm wondering how other maintainers feel about continuing to test against live APIs versus using mocked requests. See https://github.com/pvlib/pvlib-python/pull/1175/files#issuecomment-787108829 and https://github.com/pvlib/pvlib-python/pull/1175/files#issuecomment-787150267 for more discussion.

AdamRJensen:
> so users will see this warning and then hit raise requests.HTTPError(errors, response=res) on line 210?

@wholmgren just before the warning is printed, verbose is set equal to False. I like having the option of just setting verbose=True in order to get as much info as possible (if time_step='1min' then you get additional columns, otherwise not)

    if (verbose is True) & ((time_step != '1min') | (time_ref != 'UT')):
        verbose = False
        warnings.warn("Verbose mode only supports 1 min. UT time series!")

kanderso-nrel:
> I'm wondering how other maintainers feel about continuing to test against live APIs versus using mocked requests. See https://github.com/pvlib/pvlib-python/pull/1175/files#issuecomment-787108829 and https://github.com/pvlib/pvlib-python/pull/1175/files#issuecomment-787150267 for more discussion.

FYI those links don't go to specific comments for me.  I think there is value in testing against the live APIs, but the tests are rather slow and annoying when they do fail.  What about splitting `iotools` into a separate repository with its own CI configuration?  It could be completely stand-alone, or maybe pvlib imports it as a dependency.  

I'm not totally against mocked requests, but I suspect they'd be a pain to set up, especially for contributors who haven't done it before.

adriesse:
Just some food for thought here: the data is CAMS but the provider is soda-pro.  You can also get this data from ecmwf, although in my experience it is slower, and more importantly the process and output format are different.  So the question I raise is whether that should be reflected in the naming of things somehow, i.e. calling the module "sodapro"...

adriesse:
A thought regarding tests on remote services.  You can pretty much guarantee that these web services to will change or be out of service and break the pvlib code at some point.  What about having a test function for users to call rather than an automated test.  A bit like ping but a bit more sophisticated.  Good practice would be to call the ping function (using your own credentials) before launching into something more complex.  If the service is down, you  try again the next day.  If the service has changed, you open an issue.

wholmgren:
@mikofski @cwhanse thoughts on using the `pvlib-admin@googlegroups.com` for live tests? @mikofski created this for the code of conduct but I don't see why it couldn't be used here too. I assume it still works though I have not recently tested it.

cwhanse:
> @mikofski @cwhanse thoughts on using the `pvlib-admin@googlegroups.com` for live tests? @mikofski created this for the code of conduct but I don't see why it couldn't be used here too. I assume it still works though I have not recently tested it.

Better than using one of our personal emails. We could try it and go back to the current practice if it doesn't work well.

mikofski:
Seems like a good use of `pvlib-admins@googlegroups.com`, it's ok by me

wholmgren:
Thanks. @AdamRJensen please go ahead and write a live test or two using `pvlib-admins@googlegroups.com`.

> So the question I raise is whether that should be reflected in the naming of things somehow, i.e. calling the module "sodapro"...

Good point @adriesse. I'd lean towards sodapro.

> What about having a test function for users to call rather than an automated test. A bit like ping but a bit more sophisticated. 

That is a very interesting idea!

AdamRJensen:
@wholmgren just to make sure I understand correctly, I should rename only the module to sodapro.py, but the function names will remain the same?

Also, pvlib-admin@googlegroups.com needs to be registered with a SoDa user in order for me to use it. Should I do that? (I would imagine that you want to keep track of the passwords associated with the account?)

Lastly, are there any updates on how to make tests of live API's? You previously mentioned requests-mock and pytest-mock.

wholmgren:
> I should rename only the module to sodapro.py, but the function names will remain the same? Or should function names be something like: .get_sodapro_cams()?

I'm +0.5 on renaming the module. If we rename the module, I'm -1 on including sodapro in the function name because it feels unnecessarily repetitive.

> Also, pvlib-admin@googlegroups.com needs to be registered with a SoDa user in order for me to use it. Should I do that? (I would imagine that you want to keep track of the passwords associated with the account?)

I attempted to register the email with SoDa this morning. So far it's not been approved. If it is approved, I will add the password to the Azure CI environment. In the meantime, you could write tests that look up environment variables like `SODAPRO_USERNAME` and `SODAPRO_PASSWORD` and then use your own credentials. You'd also need to add logic to skip the tests if the environment variables are not set. 

> Lastly, are there any updates on how to make tests of live API's? You previously mentioned requests-mock and pytest-mock.

Here's a rough sketch of what I'd do:

```python
def test_get_cams_radiation(requests_mock):
    expected_response = 'your data in text form'  # might need to add .encode() but not sure
    requests_mock.get('www.soda-is.com', content=expected_response)  # uri might need params for finer testing
    out = cams.get_cams_radiation()
    expected = ?
    assert out == expected


def test_get_cams_radiation_bad_request(requests_mock):
    expected_response = 'junky xml'  # might need to add .encode() but not sure
    requests_mock.get('www.soda-is.com', text=expected_response)  # uri might need params for finer testing
    pytest.raises(requests.HTTPError):
        cams.get_cams_radiation()
```

pvlib/tests/iotools/test_sodapro.py
```diff
@@ -0,0 +1,157 @@
+"""
+test iotools for CAMS
+"""
+
+import pandas as pd
+import numpy as np
+import pytest
+
+from pvlib.iotools import sodapro
+from conftest import DATA_DIR, assert_frame_equal
```
kanderso-nrel, line 10: I think the CI failures are because conftest is in the parent directory, so it needs a `..` like this line in `test_psm3.py`:

https://github.com/pvlib/pvlib-python/blob/5b0b15ae2e451ddcd91af5f69f9bfb4595ec57e1/pvlib/tests/iotools/test_psm3.py#L7

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,361 @@
...
+CAMS_INTEGRATED_COLUMNS = [
+    'TOA', 'Clear sky GHI', 'Clear sky BHI', 'Clear sky DHI', 'Clear sky BNI',
+    'GHI', 'BHI', 'DHI', 'BNI',
+    'GHI no corr', 'BHI no corr', 'DHI no corr', 'BNI no corr']
+
+# Dictionary mapping CAMS Radiation and McClear variables to pvlib names
+CAMS_VARIABLE_MAP = {
+    'TOA': 'ghi_extra',
+    'Clear sky GHI': 'ghi_clear',
+    'Clear sky BHI': 'bhi_clear',
```
kanderso-nrel, line 21: I'm curious what other folks think about using `bhi` for pvlib.  I assume it is beam horizontal irradiance?  It's a shame that diffuse and direct both start with "d".

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,361 @@
...
+                                 '0 year 1 month 0 day 0 h 0 min 0 s': '1M'}
+
+
+def get_cams(start_date, end_date, latitude, longitude, email,
+             identifier='mcclear', altitude=None, time_step='1h',
+             time_ref='UT', verbose=False, integrated=False, label=None,
+             map_variables=True, server='www.soda-is.com'):
+    """
+    Retrieve time-series of radiation and/or clear-sky global, beam, and
+    diffuse radiation from CAMS [1]_, [2]_ using the WGET service [3]_.
```
kanderso-nrel, line 50: Better to avoid references in the first sentence of a docstring -- the first sentence gets copied to `api.rst`, where sphinx sees the `[1]_` and complains that it doesn't know what you're talking about (`/home/docs/checkouts/readthedocs.org/user_builds/pvlib-python/checkouts/1175/docs/sphinx/source/api.rst:490:<autosummary>:1: WARNING: Unknown target name: "1".`).

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,361 @@
...
+        daily, or monthly.
+    time_reference: str, {'UT', 'TST'}, default: 'UT'
+        'UT' (universal time) or 'TST' (True Solar Time)
+    verbose: boolean, default: False
+        Verbose mode outputs additional parameters (aerosols). Only available
+        for 1 minute and universal time. See [1] for parameter description.
+    integrated: boolean, default False
+        Whether to return radiation parameters as integrated values (Wh/m^2)
+        or as average irradiance values (W/m^2) (pvlib preferred units)
+    label: {‘right’, ‘left’}, default: None
```
kanderso-nrel, line 90: nitpick, use simple quotes instead of fancy ones (here and in the lines below).  The docstrings of other functions have it too

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,361 @@
...
+    Clear sky DHI, dhi_clear  float   Clear sky diffuse radiation on horizontal
+    Clear sky BNI, dni_clear  float   Clear sky beam radiation normal to sun
+    GHI, ghi*                 float   Global horizontal radiation
+    BHI, bhi*                 float   Beam (direct) radiation on horizontal
+    DHI, dhi*                 float   Diffuse horizontal radiation
+    BNI, dni*                 float   Beam (direct) radiation normal to the sun
+    Reliability*              float   Reliable data fraction in summarization
+    ========================  ======  =========================================
+
+    *Parameters only returned if identifier='cams_radiation'. For description
```
kanderso-nrel, line 131: Sphinx is unhappily trying to interpret these asterisks in a way I don't understand.  Not sure what to do about it.  Maybe replace the asterisks with some other symbol like †, or maybe break out those parameters into their own table?

kanderso-nrel (commented):
Just a quick partial review.  All the codecov comments make it difficult to review the code -- I think they'll go away with the minor `conftest` change below.

pvlib/iotools/sodapro.py
AdamRJensen: @kanderso-nrel correct it is beam horizontal irradiance. It's a common issue without any good solution. CAMS calls it beam instead of direct which is one solution. Have you any suggestions for alternatives?

cwhanse: I'm +0.1 on "beam" as the term here, and -1 on using "direct" which would require us to get cute with the "horizontal" part of the acronym. The only use of "beam" in pvlib is [this function name ](https://github.com/pvlib/pvlib-python/blob/5b0b15ae2e451ddcd91af5f69f9bfb4595ec57e1/pvlib/irradiance.py#L274).  'bhi' doesn't conflict with other terms, and may help users to translate between CAMS and pvlib terms.

pvlib/iotools/sodapro.py
AdamRJensen: Always good to be consistent. I suppose I copied it from another function. Thanks!

pvlib/tests/iotools/test_sodapro.py
AdamRJensen: You were indeed correct! I see that Will changed this in the other functions since my last commit.

pvlib/iotools/sodapro.py
AdamRJensen: As suggested I have used †, instead. It still shows up as an asterisk in the documentation built by readthedocs. Can you see if it still complains (or where can I see that)?

pvlib/iotools/sodapro.py
AdamRJensen: I have corrected this by adding a second sentence.

pvlib/iotools/sodapro.py
kanderso-nrel: I think the source code still uses the asterisk, maybe you didn't push that commit to github?

To see the sphinx build log, navigate to the list of RTD builds either through the version switcher at the bottom left of any RTD page, or directly with [this link](https://readthedocs.org/projects/pvlib-python/builds/).  Then choose the build you want, and on the right side of the page there is a "view raw" link that takes you to the raw log file.  Right now there's a ton of warnings about forecast.py and ModelChain that I think have already been resolved on `master`, so you may want to merge master into your branch just to get rid of those.

AdamRJensen: The error warning was: "WARNING: Inline emphasis start-string without end-string.". 

In reStructuredText an asterisk is the start of italiced text, so Sphinx was probably looking for an ending asterisk. For future reference, an asterisk can be used inline, [but it has to be escaped with a backslash, i.e., \*](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html).

pvlib/iotools/sodapro.py
wholmgren: I'm fine with bhi_clear. Looks like [pv-terms] does not have a relevant recommendation.

AdamRJensen:
@wholmgren @kanderso-nrel I've added mock tests and they run fine locally. However, here they throw the error "fixture 'requests_mock' not found"

There seems to be different work arounds for this, see this [post](https://stackoverflow.com/questions/59796255/why-is-the-requests-mock-decorator-pattern-throwing-a-fixture-m-not-found-er). Any opinion on which method?

kanderso-nrel:
Is it just because the pytest plugin isn't getting installed on the CI?  If so, try adding `requests-mock` to the `TESTS_REQUIRE` list in setup.py, and the various requirements files in the `ci` folder.

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+    .. [3] `CAMS McClear Automatic Access
+       <http://www.soda-pro.com/help/cams-services/cams-mcclear-service/automatic-access>`_
+    """
+    try:
+        time_step_str = TIME_STEPS_MAP[time_step]
+    except KeyError:
+        raise ValueError(f'Time step not recognized. Must be one of '
+                         f'{list(TIME_STEPS_MAP.keys())}')
+
+    if (verbose is True) & ((time_step != '1min') | (time_ref != 'UT')):
```
kanderso-nrel, line 170: Style nit:  I reserve the bitwise operators `&` and `|` for situations where I actually need the difference between the standard logical `and` and `or`, e.g. combining boolean vectors, so that it is marginally easier to tell when something is vectorized vs when it's just a scalar comparison.  Also minor preference to just use `verbose` rather than `verbose is True`, unless it's important here that the value is actually `True` instead of just something truthy?

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+
+    email = email.replace('@', '%2540')  # Format email address
+    identifier = 'get_{}'.format(identifier.lower())  # Format identifier str
+
+    base_url = f"http://{server}/service/wps"
+
+    data_inputs = (
+        f"latitude={latitude};longitude={longitude};altitude={altitude};"
+        f"date_begin={start_date};date_end={end_date};time_ref={time_ref};"
+        f"summarization={time_step_str};username={email};verbose={verbose}")
```
kanderso-nrel, line 195: Perhaps worth a comment that this text all represents a single GET parameter -- I was scratching my head why these weren't just included as pairs in the `params` dict until I saw that they are delimited by semicolons instead of ampersands. 

Idea:  would this be easier to read with something like this?
```python
data_inputs_dict = {
   'latitude': latitude,
   ...
}
data_inputs = ";".join([f"{key}={value}" for key, value in data_inputs_dict.items()])
```

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+              'Identifier': identifier,
+              'version': '1.0.0',
+              'RawDataOutput': 'irradiation',
+              }
+
+    # The DataInputs parameter of the URL has to be manually formatted and
+    # added to the base URL as it contains sub-parameters seperated by
+    # semi-colons, which gets incorrectly formatted by the requests function
+    # if passed using the params argument.
+    res = requests.get(base_url + '?DataInputs=' + data_inputs, params=params,
```
kanderso-nrel, line 208: Might be marginally cleaner to include `DataInputs` in the `params` dict and let `requests` add it to the url instead of adding it manually here.  That might let you avoid manually encoding the email address above as well.

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+                       timeout=timeout)
+
+    # Invalid requests returns an XML error message and the HTTP staus code 200
+    # as if the request was successful. Therefore, errors cannot be handled
+    # automatic (e.g. res.raise_for_status()) and errors are handled manually
+    if res.headers['Content-Type'] == 'application/xml':
+        errors = res.text.split('ows:ExceptionText')[1][1:-2]
+        raise requests.HTTPError(errors, response=res)
+    # Successful requests returns a csv data file
+    elif res.headers['Content-Type'] == 'application/csv':
```
kanderso-nrel, line 218: Should we be safe and add an `else` to this `if/elif` chain?  Or is that in [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) territory

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+    meta = {}
+    # Initial lines starting with # contain meta-data
+    while True:
+        line = fbuf.readline().rstrip('\n')
+        if line.startswith('# Observation period'):
+            # The last line of the meta-data section contains the column names
+            names = line.lstrip('# ').split(';')
+            break  # End of meta-data section has been reached
+        elif ': ' in line:
+            meta[line.split(': ')[0].lstrip('# ')] = line.split(': ')[1]
```
kanderso-nrel, line 271: Not a blocker for me, but just pointing out that this assumes that metadata values are single lines, which isn't true for the `Content` information.  I'm in favor of leaving this as-is instead of building a parser that can handle multiple lines.

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+    # Set index as the stop observation time (right) and localize to UTC
+    # default label for monthly data is 'right' following Pandas' convention
+    elif (label == 'right') | ((label is None) & (time_step == '1M')):
+        data.index = pd.to_datetime(obs_period.str[1], utc=True)
+
+    # Change index for time_step '1d' and '1M' to be date and not datetime
+    if (time_step == '1d') | (time_step == '1M'):
+        data.index = pd.DatetimeIndex(data.index.date)
+    # For monthly data with 'right' label, the index should be the last
+    # date of the month and not the first date of the following month
```
kanderso-nrel, line 302: Either way seems reasonable to me (and I personally prefer left-labeling to avoid questions like this), but I wonder what other people think.  As a point of comparison, TMY3 data are right-labeled but without a sub-unit shift like this.  

I also wonder if all this labeling code is worth it -- maybe we should just return the start observation time and be done with it.

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+    if not integrated:  # Convert radiation values from Wh/m2 to W/m2
+        integrated_cols = [c for c in CAMS_INTEGRATED_COLUMNS
+                           if c in data.columns]
+
+        if time_step == '1M':
+            time_delta = (pd.to_datetime(obs_period.str[1])
+                          - pd.to_datetime(obs_period.str[0]))
+            hours = time_delta.dt.total_seconds()/60/60
+            data[integrated_cols] = data[integrated_cols].\
+                divide(hours.tolist(), axis='rows').round(4)
```
kanderso-nrel, line 315: Is `tolist()` necessary here?  And why `.round(4)`?

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+        else:
+            data[integrated_cols] = (data[integrated_cols] /
+                                     TIME_STEPS_IN_HOURS[time_step]).round(4)
+
+    data.index.name = 'time'  # Set index name to None
```
kanderso-nrel, line 320: I'd leave the index unnamed, but either way the comment doesn't match the code :)

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+    map_variables: bool, default: True
+        When true, renames columns of the Dataframe to pvlib variable names
+        where applicable. See variable CAMS_VARIABLE_MAP.
+
+    Returns
+    -------
+    data: pandas.DataFrame
+        Timeseries data from CAMS Radiation or McClear
+    meta: dict
+        Metadata avaiable in the file.
```
kanderso-nrel, line 249:
```suggestion
        Metadata available in the file.
```

kanderso-nrel (commented):
This is looking pretty good @AdamRJensen!  Some notes below, and let's talk about the tests on our call later.

ci/requirements-py36-min.yml
```diff
@@ -19,3 +19,4 @@ dependencies:
         - scipy==1.2.0
         - pytest-rerunfailures # conda version is >3.6
         - pytest-remotedata # conda package is 0.3.0, needs > 0.3.1
+        - requests-mock
```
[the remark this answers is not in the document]
wholmgren: I have a slight preference for avoiding conda forge unless it simplifies things.

pvlib/iotools/sodapro.py
wholmgren: > maybe we should just return the start observation time and be done with it.

+1

wholmgren (commented):
Nice job with the tests @AdamRJensen.

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+    if (label == 'left') | ((label is None) & (time_step != '1M')):
+        data.index = pd.to_datetime(obs_period.str[0], utc=True)
+    # Set index as the stop observation time (right) and localize to UTC
+    # default label for monthly data is 'right' following Pandas' convention
+    elif (label == 'right') | ((label is None) & (time_step == '1M')):
+        data.index = pd.to_datetime(obs_period.str[1], utc=True)
+
+    # Change index for time_step '1d' and '1M' to be date and not datetime
+    if (time_step == '1d') | (time_step == '1M'):
+        data.index = pd.DatetimeIndex(data.index.date)
```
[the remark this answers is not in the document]
AdamRJensen: I had been mislead by how datetimeindexs with hours other than midnight only show the date in Spyder. Thanks for noticing. I'm chaing the comment to "# For time_steps '1d' and '1M', drop timezone and round to nearest midnight"

pvlib/iotools/sodapro.py
```diff
@@ -0,0 +1,368 @@
...
+
+
+def get_cams(start_date, end_date, latitude, longitude, email,
+             identifier='mcclear', altitude=None, time_step='1h',
+             time_ref='UT', verbose=False, integrated=False, label=None,
+             map_variables=True, server='www.soda-is.com', timeout=30):
+    """
+    Retrieve time-series of radiation and/or clear-sky global, beam, and
+    diffuse radiation from CAMS. Data from CAMS Radiation [1]_ and CAMS McClear
+    [2]_ are retrieved using the WGET service [3]_.
```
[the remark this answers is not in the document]
AdamRJensen: Good point. Thanks for noticing.

pvlib/iotools/sodapro.py
AdamRJensen: Consider it implemented!

pvlib/iotools/sodapro.py
AdamRJensen: Very nice solution. This makes it resemble a traditional params dictionary as much as possible and gives a nice overview.

pvlib/iotools/sodapro.py
AdamRJensen: @wholmgren have previously recommended against this. See this previous [discussion](https://github.com/pvlib/pvlib-python/pull/1175#discussion_r584194688).

pvlib/iotools/sodapro.py
AdamRJensen: I have removed the .round(4) and added check_less_precise=True to assert_frame_equal. This works nicely, without requiring any additional modifications. :D

The .tolist() is necessary, I believe it's because the dataframe and series does not have the same index. Are you interested in an alternative solution?

kanderso-nrel: Ah I see, I agree that there is an index mismatch.  I don't immediately see an alternative that is any cleaner, so `.tolist()` is fine with me 👍

pvlib/iotools/sodapro.py
AdamRJensen: I recognize your concern and would perhaps agree if it the service didn't return monthly data.

My main argument is that Pandas labels minute and daily data by the left interval, however, monthly data is labelled right (i.e., last day of the month). The majority of users of monthly data would be required to manually change the index labels, before being able to merge with any other resample data in pandas.Personally I think it adds a lot of functionality that makes the function a lot easier to use

kanderso-nrel: FWIW here are some one-liners for converting month-begin data to month-end:
```python
In [20]: s = pd.Series([1,2,3], pd.date_range('2019-01-01', periods=3, freq='MS'))

In [21]: s.index
Out[21]: DatetimeIndex(['2019-01-01', '2019-02-01', '2019-03-01'], dtype='datetime64[ns]', freq='MS')

In [22]: s.resample('M').mean().index
Out[22]: DatetimeIndex(['2019-01-31', '2019-02-28', '2019-03-31'], dtype='datetime64[ns]', freq='M')

# or
In [23]: s.to_period('M').to_timestamp('M').index
Out[23]: DatetimeIndex(['2019-01-31', '2019-02-28', '2019-03-31'], dtype='datetime64[ns]', freq='M')
```

How about removing the `label` parameter but adding a docstring example with (a better version of) the above snippet for the benefit of whatever subset of pvlib users want monthly insolation?  E.g. [this](https://github.com/pvlib/pvlib-python/blob/master/pvlib/modelchain.py#L624), which renders as [this](https://pvlib-python.readthedocs.io/en/stable/generated/pvlib.modelchain.ModelChain.with_sapm.html).

wholmgren (approved):
Thanks @AdamRJensen. @kanderso-nrel I'll let you decide when to merge this and refrain from further review unless requested.

AdamRJensen:
@wholmgren @kanderso-nrel I recommend that this first gets merged when a decision on #1245 is made.

kanderso-nrel (approved):
Sorry for the delay here, this looks great.  Thanks @AdamRJensen!  One more idea, what do you think about adding the new-to-pvlib `bhi` acronym to the [Variables and Symbols](https://pvlib-python.readthedocs.io/en/stable/variables_style_rules.html) list?

PR #1177, opened 2021-02-25
NOCT cell temperature function

cwhanse:
- ~~[ ] Closes #xxxx~~
 - [x] Tests added

Implement cell temperature function that is the default in SAM when the CEC module model is chosen.


cwhanse:
Ready for review. check failure is codecov, but all of `temperature.py` is covered.

cwhanse:
> Mostly looks good. Are you planning a follow up issue/pr for integration into `PVSystem` and `ModelChain`?

Yes, I wanted to get consensus on the function first.

pvlib/temperature.py
```diff
@@ -706,3 +706,94 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
...
+                            x >= 3.5], [18., 11., 6., 2., 0.])
+
+
+def noct(poa_global, temp_air, wind_speed, noct, eta_m_ref,
+         effective_irradiance=None, transmittance_absorbtance=0.9,
+         array_height=1, mount_standoff=3.5):
+    '''
+    Cell temperature model from the System Advisor Model (SAM).
+
+    The model is described in [1], Section 10.6.
```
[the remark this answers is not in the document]
cwhanse: I agree with the comment. Let me see if I can locate a copy of Duffie and Beckman. There's an issue with the De Soto reference, I downloaded both (paper and thesis) and there's no discussion in either of the cell temperature model.

pvlib/temperature.py
```diff
@@ -706,3 +706,94 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
         sun0 = sun
 
     return pd.Series(tmod_array - 273.15, index=poa_global.index, name='tmod')
+
+
+def _adj_noct(x):
```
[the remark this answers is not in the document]
cwhanse: SAM code doesn't process a distance, only passes through an integer index to the values. Except for the "x > 3.5" case, the SAM interface text doesn't specify > or >=. So we're filling in a detail here. I'll adjust the default to 3.51 and put in some comments.

pvlib/temperature.py
```diff
@@ -706,3 +706,94 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
...
+        raise ValueError(
+            f'array_height must be 1 or 2, {array_height} was given')
+
+    noct_adj = noct + _adj_noct(mount_standoff)
+    tau_alpha = transmittance_absorbtance * irr_ratio
+
+    cell_temp_init = ross(poa_global, temp_air, noct_adj)
+    heat_loss = 1 - eta_m_ref / tau_alpha
+    wind_loss = 9.5 / (5.7 + 3.8 * wind_adj)
+    return cell_temp_init * heat_loss * wind_loss
```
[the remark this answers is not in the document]
cwhanse: You are correct. We can't re-use `ross` here.

pvlib/temperature.py
```diff
@@ -706,3 +706,100 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
...
+        wind_adj = 0.61 * wind_speed
+    else:
+        raise ValueError(
+            f'array_height must be 1 or 2, {array_height} was given')
+
+    noct_adj = noct + _adj_for_mounting_standoff(mount_standoff)
+    tau_alpha = transmittance_absorbtance * irr_ratio
+
+    # [1] Eq. 10.37 isn't clear on exactly what "G" is. SAM SSC code uses
+    # poa_global where G appears
```
kanderso-nrel, line 801: At the bottom of page 57 it says "The cell temperature is only calculated when the effective transmitted irradiance from Equation 10.25 is greater than zero, G>0.", and tracing back through to Eq 10.20 shows that `G` is after accounting for surface reflection and airmass.

pvlib/temperature.py
```diff
@@ -706,3 +706,100 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
...
+    ----------
+    .. [1] Gilman, P., Dobos, A., DiOrio, N., Freeman, J., Janzou, S.,
+           Ryberg, D., 2018, "SAM Photovoltaic Model Technical Reference
+           Update", National Renewable Energy Laboratory Report
+           NREL/TP-6A20-67399.
+    '''
+    if effective_irradiance is None:
+        irr_ratio = 1.
+    else:
+        irr_ratio = effective_irradiance / poa_global
```
kanderso-nrel, line 787: Not sure this is correct.  Eq 10.22 has the ratio as G0/G, where G0 is irradiance after reflections and G is after reflections and airmass.

kanderso-nrel (commented):
I'm only partially able to reconcile the standoff adjustment values in SAM Tech Ref Eq. 10.36 with their supposed source, Fuentes Table 4.  Not sure where some of these SAM/SSC values came from (e.g. 2inch -> +6C?); it almost seems like they interpolated/projected the Fuentes values.  Maybe that makes sense here too -- to go from discrete to continuous, just interpolate instead of bin?  Although that would be more of a departure from SAM/SSC.

Some helpful links for anyone following along:
- SAM technical reference: https://www.nrel.gov/docs/fy18osti/67399.pdf
- Fuentes: https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/1985/850330.pdf
- SSC implementation: https://github.com/NREL/ssc/blob/055e2847ea1f4a81b4a0062beb972ec45e77d185/shared/lib_cec6par.cpp#L160
- Additional SSC settings: https://github.com/NREL/ssc/blob/055e2847ea1f4a81b4a0062beb972ec45e77d185/shared/lib_pv_io_manager.cpp#L957-L963

pvlib/temperature.py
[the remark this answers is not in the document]
kanderso-nrel: Maybe the reference to De Soto is for the reflection model?  https://github.com/NREL/ssc/blob/055e2847ea1f4a81b4a0062beb972ec45e77d185/shared/lib_cec6par.cpp#L176-L182

pvlib/temperature.py
```diff
@@ -706,3 +706,100 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
...
+    # aren't clear on the precise intervals, the choice of < or <= here is
+    # pvlib's.
+    return np.piecewise(x, [x <= 0, (x > 0) & (x < 0.5),
+                            (x >= 0.5) & (x < 1.5), (x >= 1.5) & (x < 2.5),
+                            (x >= 2.5) & (x <= 3.5), x > 3.5],
+                        [0., 18., 11., 6., 2., 0.])
+
+
+def noct(poa_global, temp_air, wind_speed, noct, eta_m_ref,
+         effective_irradiance=None, transmittance_absorbtance=0.9,
```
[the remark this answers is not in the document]
cwhanse: Should be "absorptance" misspelled in the SAM doc and I didn't check.

cwhanse:
> I'm only partially able to reconcile the standoff adjustment values in SAM Tech Ref Eq. 10.36 with their supposed source, Fuentes Table 4. 

I think we should regard SAM as the source for standoff adjustment values. 

About "G" vs. "G0" in the SAM document: the term "G" is redefined twice in Chapter 10 (Eq. 10.13 and 10.22 are POA broadband, Eq. 10.25 and 10.26 (and 7.8) are effective irradiance absorbed by the cells. I don't read C well, but it looked to me that `G_total` in the [SAM function](https://github.com/NREL/ssc/blob/055e2847ea1f4a81b4a0062beb972ec45e77d185/shared/lib_cec6par.cpp#L160) is POA irradiance, not effective irradiance. 

It would be desirable if this temperature model could be used without requiring both `poa_global` and `effective_irradiance`, or a third quantity falling in between which isn't present in pvlib: absorbed irradiance, POA after reflection but before spectral adjustment, this quantity is 'S' in De Soto's papers.

The SAM implementation is more involved that what is described in Duffie and Beckman. One way forward here is to implement `noct_duffie_beckman` rather than `noct_sam`. In D&B, 'G' is POA irradiance; the model factors aren't adjusted with `effective_irradiance` which is the current complication.

cwhanse:
I spent some time trying to get SAM to give me back cell temperatures, without success. That quantity isn't available in the time series output, and the relevant function (`noct_celltemp_t`) isn't directly accessible through PySAM. I was able to get a 'tcell' value using PySAM's `Pv6parmod` object and of course it's not what I expected to see from the inputs I provided.

At this point, my recommendation is to fall back to implementing the model in Duffie and Beckman, and adding the SAM model in the future (as a different function) when we're clear exactly what the SAM model does. @kanderso-nrel? @wholmgren?

The driver for this PR is the Solar Performance Insight project, where it would be nice to have a pre-packaged "SAM" model chain.

kanderso-nrel:
No preference from me on the way best way forward.  For getting cell temps from SAM:  it is available in the UI, but maybe not under a name you expect (Subarray 1 Cell Temperature), see screenshots below.  You can grab numerical values from the `Data tables` tab.  I think I've pulled it directly from SSC before, but that was before the days of PySAM so I don't know how to do it there. 

<details>

image

---

image


</details>

cwhanse:
@kanderso-nrel thanks for those pointers to SAM output. I am able to reproduce the Subarray 1 Cell Temperature values. To reproduce, I equated the following terms used in the [SAM code](https://github.com/NREL/ssc/blob/055e2847ea1f4a81b4a0062beb972ec45e77d185/shared/lib_cec6par.cpp#L214) with pvlib variables:
- `Geff_total` is equivalent to pvlib's `effective_irradiance`,
- `G_total` is equivalent to pvlib's `poa_global` (plane of array before reflections or spectrum adjustment). 
I will put comments in the pvlib code to clarify why we assign these variables and write one test that reproduces the SAM value.

cwhanse:
Failed check is codecov.  This is ready to go, I think.

pvlib/temperature.py
```diff
@@ -706,3 +706,112 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
...
+
+    Parameters
+    ----------
+    poa_global : numeric
+        Total incident irradiance. [W/m^2]
+
+    temp_air : numeric
+        Ambient dry bulb temperature. [C]
+
+    wind_speed : numeric, default 1.0
```
wholmgren, line 737: no default value in the function signature

pvlib/temperature.py
```diff
@@ -706,3 +706,112 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
...
+
+    noct : float
+        Nominal operating cell temperature [C], determined at conditions of
+        800 W/m^2 irradiance, 20 C ambient air temperature and 1 m/s wind.
+
+    effective_irradiance : numeric, default None.
+        The irradiance that is converted to photocurrent. If None,
+        assumed equal to poa_global. [W/m^2]
+
+    eta_m_ref : float
```
wholmgren, line 750: comes before effective_irradiance in function signature.

Can we call this `module_efficiency` so that more people can easily understand it?

pvlib/temperature.py
```diff
@@ -706,3 +706,112 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
...
+        Module external efficiency at reference conditions of 1000 W/m^2 and
+        20C. Calculate as
+        .. math::
+
+            \eta_{m} = \frac{V_{mp} I_{mp}}{A \times 1000 W/m^2}  # noQA: W605
```
wholmgren, line 755: I'm surprised you need noqa: W605 since the whole docstring is preceded by `r`

wholmgren (commented):
👍 for `test_noct_sam_against_sam`. I don't know what's going on with codecov. Maybe we need to start specifying a target value for that configuration.

pvlib/temperature.py
kanderso-nrel: Suggest changing this to inline with `:math:` as right now it's rendering as a code block (not a math block) on RTD.

kanderso-nrel (commented):
After taking a closer look and correcting my understanding of the SSC implementation, LGTM besides @wholmgren's comments and a formatting note below.

pvlib/temperature.py
cwhanse: I'm ok with `module_efficiency`. I used `eta_m_ref` because it's the same parameter as `eta_m` in `pvsyst_cell`, except that it's not clear if the Pvsyst value is at reference conditions.

pvlib/temperature.py
cwhanse: I had the `noQA` before I put the `r` in the docstring. Should be OK without the `noQA` now.

wholmgren (approved):

PR #1190, opened 2021-03-09
Function to calculate temperature coefficient of power for pvsyst

cwhanse:
- ~~[ ] Closes #xxxx~~
 - [x] Tests added

The temperature coefficient of power is not an input to the pvsyst single diode model. Rather, it is a result of the model, the derivative of power with respect to cell temperature, at the maximum power point, at reference conditions.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -103,6 +103,8 @@ Enhancements
   (:pull:`1147`, :issue:`998`, :pull:`1150`)
 * Added :py:func:`~pvlib.temperature.noct_sam`, a cell temperature model
   implemented in SAM (:pull:`1177`)
+* Added :py:func:`~pvlib.ivtools.sdm.pvsyst_temperature_coeff` to calculate
+  the temperature coefficient of power for the pvsyst module model.
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -102,9 +102,10 @@ Enhancements
   from DC power. Use parameter ``model`` to specify which inverter model to use.
   (:pull:`1147`, :issue:`998`, :pull:`1150`)
 * Added :py:func:`~pvlib.temperature.noct_sam`, a cell temperature model
-  implemented in SAM (:pull:`1177`)
+  implemented in SAM. (:pull:`1177`)
 * Added :py:func:`~pvlib.ivtools.sdm.pvsyst_temperature_coeff` to calculate
   the temperature coefficient of power for the pvsyst module model.
+   (:pull:`1190`)
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -102,11 +102,10 @@ Enhancements
   from DC power. Use parameter ``model`` to specify which inverter model to use.
   (:pull:`1147`, :issue:`998`, :pull:`1150`)
 * Added :py:func:`~pvlib.temperature.noct_sam`, a cell temperature model
-  implemented in SAM. (:pull:`1177`)
+  implemented in SAM. (:pull:`1177`, :pull:`1195`)
 * Added :py:func:`~pvlib.ivtools.sdm.pvsyst_temperature_coeff` to calculate
   the temperature coefficient of power for the pvsyst module model.
    (:pull:`1190`)
-  implemented in SAM (:pull:`1177`, :pull:`1195`)
 
 Bug fixes
 ~~~~~~~~~
--- a/pvlib/ivtools/sdm.py
+++ b/pvlib/ivtools/sdm.py
@@ -1339,5 +1339,6 @@ def pvsyst_temperature_coeff(alpha_sc, gamma_ref, mu_gamma, I_L_ref, I_o_ref,
     args = (irrad_ref, alpha_sc, gamma_ref, mu_gamma, I_L_ref,
             I_o_ref, R_sh_ref, R_sh_0, R_s, cells_in_series, R_sh_exp, EgRef,
             temp_ref)
+    pmp = maxp(temp_ref, *args)
     gamma_pdc = derivative(maxp, temp_ref, args=args)
-    return gamma_pdc
+    return gamma_pdc / pmp
--- a/pvlib/ivtools/sdm.py
+++ b/pvlib/ivtools/sdm.py
@@ -13,8 +13,7 @@ from scipy import optimize
 from scipy.special import lambertw
 from scipy.misc import derivative
 
-from pvlib.pvsystem import singlediode, v_from_i
-from pvlib.pvsystem import calcparams_pvsyst
+from pvlib.pvsystem import calcparams_pvsyst, singlediode, v_from_i
 from pvlib.singlediode import bishop88_mpp
 
 from pvlib.ivtools.utils import rectify_iv_curve, _numdiff
@@ -1316,12 +1315,9 @@ def pvsyst_temperature_coeff(alpha_sc, gamma_ref, mu_gamma, I_L_ref, I_o_ref,
 
     Returns
     -------
-    gamma_pmp : float
+    gamma_pdc : float
         Temperature coefficient of power at maximum power point at reference
-        conditions. [%/C]
-
-    p_mp_ref : float
-        Power at the maximum power point at reference conditions. [W]
+        conditions. [1/C]
 
     References
     ----------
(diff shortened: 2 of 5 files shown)


cwhanse:
Ready for review

pvlib/ivtools/sdm.py
```diff
@@ -11,8 +11,11 @@
 import scipy.constants
 from scipy import optimize
 from scipy.special import lambertw
+from scipy.misc import derivative
 
 from pvlib.pvsystem import singlediode, v_from_i
+from pvlib.pvsystem import calcparams_pvsyst
```
wholmgren, line 17:
```suggestion
from pvlib.pvsystem import calcparams_pvsyst, singlediode, v_from_i
```

pvlib/ivtools/sdm.py
```diff
@@ -1252,3 +1255,94 @@ def _calc_theta_phi_exact(vmp, imp, iph, io, rs, rsh, nnsvth):
...
+
+    temp_ref : float, default 25
+        Reference cell temperature. [C]
+
+
+    Returns
+    -------
+    gamma_pmp : float
+        Temperature coefficient of power at maximum power point at reference
+        conditions. [%/C]
```
wholmgren, line 1321: I was surprised to see %/C since the units of `gamma_pdc` are 1/C in `pvwatts_dc`. But I see we do use %/C in `sdm.py`. So either is fine with me at this point.

pvlib/ivtools/sdm.py
```diff
@@ -1252,3 +1255,94 @@ def _calc_theta_phi_exact(vmp, imp, iph, io, rs, rsh, nnsvth):
...
+
+    p_mp_ref : float
```
wholmgren, line 1323: A little strange to see `gamma_pmp` (p m p) and then `p_mp_ref` (p underscore m p)

wholmgren: Also a little strange to see pmp returned by a function that is ostensibly for finding temperature coefficient. I might argue that it would make more sense to either just return `gamma_pmp` or change the name of the function and also return vmp and imp. But I don't feel strongly about it.

pvlib/tests/ivtools/test_sdm.py
```diff
@@ -379,4 +380,20 @@ def test__calc_theta_phi_exact_vector():
...
+    params = {'alpha_sc': 0., 'gamma_ref': 1.1, 'mu_gamma': 0.,
+              'I_L_ref': 6., 'I_o_ref': 5.e-9, 'R_sh_ref': 200.,
+              'R_sh_0': 2000., 'R_s': 0.5, 'cells_in_series': 60}
+    params_lowT = pvsystem.calcparams_pvsyst(1000, 24, **params)
+    params_hiT = pvsystem.calcparams_pvsyst(1000, 26, **params)
+    res_lowT = pvsystem.singlediode(*params_lowT)
+    res_hiT = pvsystem.singlediode(*params_hiT)
+    expected = (res_hiT['p_mp'] - res_lowT['p_mp']) / 2.
+    # convert to %/C
+    expected = expected * 100 / (0.5 * (res_hiT['p_mp'] + res_lowT['p_mp']))
```
wholmgren, line 397: I'd prefer for this value to be hard coded and for the function calls above to be removed.

pvlib/tests/ivtools/test_sdm.py
```diff
@@ -379,4 +380,20 @@ def test__calc_theta_phi_exact_vector():
...
+    gamma_pdc, pdc_0 = sdm.pvsyst_temperature_coeff(**params)
```
wholmgren, line 398: I'm going to be a stickler again for passing args as args and kwargs as kwargs because it helps guard against unintended API changes.

wholmgren: or we can make them keyword-only arguments without defaults, though I don't think we've introduced that pattern in pvlib.

pvlib/ivtools/sdm.py
cwhanse: +1 to 1/C, more consistent with pvlib and the downstream PV Insight use. I probably cut/pasted from `sdm.fit_cec_sam` then didn't think about the interface again. We use %/C in `sdm.fit_cec_sam` because that's the quantity SAM requests.

pvlib/ivtools/sdm.py
cwhanse: Pvsyst doesn't name this parameter, so I'll call it `gamma_pdc` with units of 1/C, same as is used by the PVWatts model. If the function is returning a value with units of 1/C, then `p_mp_ref` isn't needed (it is if the unit is %/C).

docs/sphinx/source/whatsnew/v0.9.0.rst
```diff
@@ -102,6 +102,10 @@ Enhancements
   from DC power. Use parameter ``model`` to specify which inverter model to use.
   (:pull:`1147`, :issue:`998`, :pull:`1150`)
 * Added :py:func:`~pvlib.temperature.noct_sam`, a cell temperature model
+  implemented in SAM. (:pull:`1177`)
+* Added :py:func:`~pvlib.ivtools.sdm.pvsyst_temperature_coeff` to calculate
+  the temperature coefficient of power for the pvsyst module model.
+   (:pull:`1190`)
   implemented in SAM (:pull:`1177`, :pull:`1195`)
```
wholmgren, line 109: bad merge

pvlib/tests/ivtools/test_sdm.py
```diff
@@ -379,4 +380,23 @@ def test__calc_theta_phi_exact_vector():
...
-    np.testing.assert_allclose(phi[1], 2.2079, atol=.0001)
+    assert_allclose(phi[1], 2.2079, atol=.0001)
+
+
+def test_pvsyst_temperature_coeff():
+    # test for consistency with dP/dT estimated with secant rule
+    params = {'alpha_sc': 0., 'gamma_ref': 1.1, 'mu_gamma': 0.,
+              'I_L_ref': 6., 'I_o_ref': 5.e-9, 'R_sh_ref': 200.,
+              'R_sh_0': 2000., 'R_s': 0.5, 'cells_in_series': 60}
+    expected = -0.7489980887568493
```
wholmgren, line 391: Seems a bit large for a realistic module. I don't have enough intuition with the params values to say more. I just want to double check that this really is the expected value and the function really is returning the right thing.

cwhanse: Good catch. The value was in W/C. Fixed now.

pvlib/tests/ivtools/test_sdm.py
```diff
@@ -379,4 +380,23 @@ def test__calc_theta_phi_exact_vector():
...
-    np.testing.assert_allclose(phi[1], 2.2079, atol=.0001)
+    assert_allclose(phi[1], 2.2079, atol=.0001)
+
+
+def test_pvsyst_temperature_coeff():
+    # test for consistency with dP/dT estimated with secant rule
+    params = {'alpha_sc': 0., 'gamma_ref': 1.1, 'mu_gamma': 0.,
+              'I_L_ref': 6., 'I_o_ref': 5.e-9, 'R_sh_ref': 200.,
+              'R_sh_0': 2000., 'R_s': 0.5, 'cells_in_series': 60}
+    expected = -0.004886706494879083
```
wholmgren, line 391:
```suggestion
    expected = -0.004887
```

wholmgren: In the past @adriesse has gently nudged us towards using reasonable precision when testing expected. While it can be a pain, I do think it makes the tests more understandable and makes it easier to catch issues like the one you fixed in the last commit.

wholmgren (approved):
merging tomorrow morning unless there is further discussion.

pvlib/tests/ivtools/test_sdm.py
cwhanse: I don't object on principle to the suggestion. But here I think more precision is useful, in case the test goes wonky in the future, and someone is reconstructing the expected value to know they have that calculation correct.

adriesse: I only vaguely recall the nudging occasion, but I have a general pet peeve about people producing too many digits of noise (e.g. NSRDB).  You can't even compress it!  

For tests I think it depends on what the test is aiming to accomplish.  In this case you might see what happens if someone passes in float32 or maybe even float16.  I often use float32.

Looking in the code now it seems like lots of potential changes/improvements in the functions that are called could affect the numeric result without breaking the fundamental calculation.  Do you want that detected?

One thing you could perhaps consider would be to not rely on defaults when calling `derivative`.  Or do away with `derivative` and call `pmax` twice yourself.  Then the user doesn't have to go to scipy documentation to figure out what is actually being done.

cwhanse: Good points about the options in `scipy.misc.derivative`. I didn't see an advantage to explicitly choosing options, or alternatively, hardcoding a particular derivative rule, but I'm open to doing either.

Peeve acknowledged and shared. The precision for `expected` is to enable reconstructing this value and being confident that the reconstruction is the same as the test originator's calculation.

I wasn't thinking about detecting small deviations in the output of `pvsyst_temperature_coeff`, only producing a value that is "good enough" in some sense. That's why I set `rtol=0.0005`; I think 4 digits agreement is enough to say two temperature coefficients are the same.

adriesse: Sorry, I hadn't noticed the rtol.  I got mislead by all those digit thinking the must be there because you  wanted them to be there.

wholmgren: @cwhanse do you want to make revisions or go ahead and merge? I agree that @adriesse makes good points, but I am also fine with saying go read the scipy documentation if you want that level of detail.

cwhanse: I'm OK merging this and improving it later, if issues arise. The only options are point spacing and order of the derivative rule, and the defaults (dx=1.0C, order=3) are reasonable for this case.

wholmgren:
thanks @cwhanse and @adriesse

PR #1196, opened 2021-03-13
Deprecate/raise error for PVSystem "pass-through" properties

kanderso-nrel:
- ~Closes #xxxx~
 - [x] Tests added
 - Updates entries to [`docs/sphinx/source/api.rst`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes.
 - New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.

Trying out the idea suggested in https://github.com/pvlib/pvlib-python/pull/1176#discussion_r584745688: for the values that used to be `PVSystem` attributes but have been moved down to `Array`, emit a deprecation warning for systems with only one `Array` and an error if more than one. 

Note to reviewers: I was guided more by test failures than by a solid understanding of the emerging PVSystem+Array architecture.  Wouldn't be surprised if I've missed something here.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -89,16 +89,16 @@ Deprecations
   :py:class:`pvlib.tracking.SingleAxisTracker` have been deprecated in
   favor of the corresponding :py:class:`pvlib.pvsystem.Array` attributes:
 
-  * ``PVSystem.module_parameters``
-  * ``PVSystem.module``
-  * ``PVSystem.module_type``
   * ``PVSystem.albedo``
-  * ``PVSystem.temperature_model_parameters``
+  * ``PVSystem.module``
+  * ``PVSystem.module_parameters``
+  * ``PVSystem.module_type``
+  * ``PVSystem.modules_per_string``
+  * ``PVSystem.racking_model``
+  * ``PVSystem.strings_per_inverter``
   * ``PVSystem.surface_tilt``
   * ``PVSystem.surface_azimuth``
-  * ``PVSystem.racking_model``
-  * ``PVSystem.modules_per_string``
-  * ``PVSystem.strings_per_inverter``
+  * ``PVSystem.temperature_model_parameters``
 
 
 Enhancements
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -65,6 +65,22 @@ Deprecations
   * ``ModelChain.total_irrad``
   * ``ModelChain.tracking``
 
+* The following attributes of :py:class:`pvlib.pvsystem.PVSystem` and
+  :py:class:`pvlib.tracking.SingleAxisTracker` have been deprecated in
+  favor of the corresponding :py:class:`pvlib.pvsystem.Array` attributes:
+
+  * ``PVSystem.module_parameters``
+  * ``PVSystem.module``
+  * ``PVSystem.module_type``
+  * ``PVSystem.albedo``
+  * ``PVSystem.temperature_model_parameters``
+  * ``PVSystem.surface_tilt``
+  * ``PVSystem.surface_azimuth``
+  * ``PVSystem.racking_model``
+  * ``PVSystem.modules_per_string``
+  * ``PVSystem.strings_per_inverter``
+
+
 Enhancements
 ~~~~~~~~~~~~
 * Add :func:`~pvlib.iotools.read_bsrn` for reading BSRN solar radiation data
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -334,23 +334,25 @@ class ModelChain:
         to a user-defined function.
 
     ac_model: None, str, or function, default None
-        If None, the model will be inferred from the contents of
-        system.inverter_parameters and system.arrays[i].module_parameters.
+        If None, the model will be inferred from the parameters that
+        are common to all of system.inverter_parameters.
         Valid strings are 'sandia', 'adr', 'pvwatts'. The
         ModelChain instance will be passed as the first argument to a
         user-defined function.
 
     aoi_model: None, str, or function, default None
-        If None, the model will be inferred from the contents of
-        system.arrays[i].module_parameters. Valid strings are 'physical',
-        'ashrae', 'sapm', 'martin_ruiz', 'no_loss'. The ModelChain instance
-        will be passed as the first argument to a user-defined function.
+        If None, the model will be inferred from the parameters that
+        are common to all of system.arrays[i].module_parameters.
+        Valid strings are 'physical', 'ashrae', 'sapm', 'martin_ruiz',
+        'no_loss'. The ModelChain instance will be passed as the
+        first argument to a user-defined function.
 
     spectral_model: None, str, or function, default None
-        If None, the model will be inferred from the contents of
-        system.arrays[i].module_parameters. Valid strings are 'sapm',
-        'first_solar', 'no_loss'. The ModelChain instance will be passed
-        as the first argument to a user-defined function.
+        If None, the model will be inferred from the parameters that
+        are common to all of system.arrays[i].module_parameters.
+        Valid strings are 'sapm', 'first_solar', 'no_loss'.
+        The ModelChain instance will be passed as the first argument to
+        a user-defined function.
 
     temperature_model: None, str or function, default None
         Valid strings are 'sapm', 'pvsyst', 'faiman', and 'fuentes'.
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -735,11 +735,11 @@ class ModelChain:
         elif {'pdc0', 'gamma_pdc'} <= params:
             return self.pvwatts_dc, 'pvwatts'
         else:
-            raise ValueError('could not infer DC model from '
-                             'system.arrays[i].module_parameters. Check '
-                             'system.arrays[i].module_parameters or '
-                             'explicitly set the model with the dc_model '
-                             'kwarg.')
+            raise ValueError(
+                'Could not infer DC model from the module_parameters '
+                'attributes of system.arrays. Check the module_parameters '
+                'attributes or explicitly set the model with the dc_model '
+                'keyword argument.')
 
     def sapm(self):
         dc = self.system.sapm(self.results.effective_irradiance,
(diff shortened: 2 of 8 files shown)


kanderso-nrel:
I think this is ready for review.  It feels risky to me to use `system.arrays[0]` in places where we expect, but don't verify, that the system has only one Array.  Maybe worth adding a `PVSystem.array` convenience property that returns `self.arrays[0]` if `len(self.arrays) == 1` and raises an error otherwise?

wholmgren:
>  It feels risky to me to use system.arrays[0] in places where we expect, but don't verify, that the system has only one Array. Maybe worth adding a PVSystem.array convenience property that returns self.arrays[0] if len(self.arrays) == 1 and raises an error otherwise?

Can you elaborate? The tests pass and the diff is straightforward. We also have the property `PVSystem.num_arrays`. I'm open to to adding a property like `PVSystem._has_single_array`, but maybe that is a case of [too much indirection](https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection).

I'm +1 on this PR. Would be good to hear any comments/concerns from others before moving forward.

kanderso-nrel:
> 
> 
> > It feels risky to me to use system.arrays[0] in places where we expect, but don't verify, that the system has only one Array. Maybe worth adding a PVSystem.array convenience property that returns self.arrays[0] if len(self.arrays) == 1 and raises an error otherwise?
> 
> Can you elaborate? The tests pass and the diff is straightforward. We also have the property `PVSystem.num_arrays`. I'm open to to adding a property like `PVSystem._has_single_array`, but maybe that is a case of [too much indirection](https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection).
> 

Yes, sorry for not being clear.  My concern is that code like [this](https://github.com/kanderso-nrel/pvlib-python/blob/4d27febd1bfad1e72c60025299f69653b5ce299e/pvlib/modelchain.py#L1490-L1499), which does not check that `len(self.arrays) == 1`, puts the onus on the caller to verify that it is appropriate to call that method.  Right now that contract is fulfilled, but I could see myself missing that requirement in some future refactor and accidentally introducing a subtle bug.  I'd encourage us to put assumption checks like `if self.num_arrays == 1` as close as possible to the code that makes the assumption.  My `PVSystem.array` property suggestion in the previous comment is in the spirit of pandas's [item()](https://pandas.pydata.org/docs/reference/api/pandas.Series.item.html), which is a convenient way to perform the `length==1` check and element retrieval in one step.  If nobody else is particularly concerned about this then I'll drop it :)

pvlib/modelchain.py
```diff
@@ -328,26 +328,26 @@ class ModelChain:
 
     dc_model: None, str, or function, default None
         If None, the model will be inferred from the contents of
-        system.module_parameters. Valid strings are 'sapm',
+        system.arrays[i].module_parameters. Valid strings are 'sapm',
```
cwhanse, line 331:
```suggestion
        If None, the model will be inferred from the parameters that
        are common to all of system.arrays[i].module_parameters. 
        Valid strings are 'sapm',
```

May help a user debug an odd case where the intersection is missing some required parameter, because one of `system.arrays[i].module_parameters` is incomplete.

If acceptable, same comment applies to the description of `ac_model`, `aoi_model`, `spectral_model`

pvlib/modelchain.py
```diff
@@ -328,26 +328,26 @@ class ModelChain:
...
         'desoto', 'cec', 'pvsyst', 'pvwatts'. The ModelChain instance will
         be passed as the first argument to a user-defined function.
 
     ac_model: None, str, or function, default None
         If None, the model will be inferred from the contents of
-        system.inverter_parameters and system.module_parameters. Valid
-        strings are 'sandia', 'adr', 'pvwatts'. The
+        system.inverter_parameters and system.arrays[i].module_parameters.
```
cwhanse, line 337: I don't see that `infer_ac_model` uses any information from `system.arrays`. The check on `num_arrays` is meant to provide a more meaningful error message, if a user chooses the `adr` inverter model that doesn't (presently) support multiple inputs.

cwhanse (approved):

kanderso-nrel:
It's not clear to me why codecov is flagging the diff coverage.  When I calculate coverage locally it shows line 961 (really, 956) in modelchain.py covered as expected.  

@wholmgren ready for you to review when you get a chance.

pvlib/modelchain.py
```diff
@@ -645,9 +650,10 @@ def infer_dc_model(self):
             return self.pvwatts_dc, 'pvwatts'
         else:
             raise ValueError('could not infer DC model from '
-                             'system.module_parameters. Check '
-                             'system.module_parameters or explicitly '
-                             'set the model with the dc_model kwarg.')
+                             'system.arrays[i].module_parameters. Check '
+                             'system.arrays[i].module_parameters or '
```
wholmgren, line 654: I don't know if this notation will make sense to some users. I'm fine with it, but here's an alternative:

Could not infer DC model from the module_parameters attributes of system.arrays. Check the module_parameters attributes or explicitly set the model with the dc_model keyword argument.

pvlib/pvsystem.py
```diff
@@ -64,6 +64,36 @@ def f(*args, **kwargs):
...
+        overrides = {  # some Array attrs aren't the same as PVSystem
+            'strings_per_inverter': 'strings',
+        }
+        array_attr = overrides.get(pvsystem_attr, pvsystem_attr)
+        alternative = f'{class_name}.arrays[i].{array_attr}'
+
+        if len(self.arrays) > 1:
+            raise AttributeError(
+                f'{class_name}.{pvsystem_attr} not supported for multi-array '
+                f'systems. Use {alternative} instead.')
```
wholmgren, line 88: Consistent with my comment in modelchain.py, maybe something like:

{class_name}.{pvsystem_attr} not supported for multi-array systems. Set {array_attr} for each Array in {class_name}.arrays instead.

pvlib/pvsystem.py
```diff
@@ -1019,72 +1049,86 @@ def pvwatts_ac(self, pdc):
...
         return tuple(array.module_type for array in self.arrays)
 
     @property
     @_unwrap_single_value
+    @_check_deprecated_passthrough
     def temperature_model_parameters(self):
         return tuple(array.temperature_model_parameters
                      for array in self.arrays)
 
     @temperature_model_parameters.setter
```
wholmgren, line 1075: The `setter`s were added in #1076, so the deprecation requirement is a little more subtle. I think we either need to implement deprecation setters for each attribute or implement a `__setattr__` like this from modelchain:

https://github.com/pvlib/pvlib-python/blob/7eae1fc58cd693708c65102294a3991e6fbc328a/pvlib/modelchain.py#L503-L510

For that matter we could implement a `__getattr__` instead of properties, like this one:

https://github.com/pvlib/pvlib-python/blob/7eae1fc58cd693708c65102294a3991e6fbc328a/pvlib/modelchain.py#L491-L501

I don't know that one approach is better than another.

docs/sphinx/source/whatsnew/v0.9.0.rst
```diff
@@ -65,6 +65,22 @@ Deprecations
...
+  * ``PVSystem.module_parameters``
+  * ``PVSystem.module``
+  * ``PVSystem.module_type``
+  * ``PVSystem.albedo``
+  * ``PVSystem.temperature_model_parameters``
+  * ``PVSystem.surface_tilt``
+  * ``PVSystem.surface_azimuth``
+  * ``PVSystem.racking_model``
+  * ``PVSystem.modules_per_string``
+  * ``PVSystem.strings_per_inverter``
```
wholmgren, line 81: I like the alphabetization of the list preceding this one and would prefer to see that here too.

wholmgren:
Also a handful of problems in `pvsystem.rst`. I think you'll want to bring your branch up to date before tackling those.

pvlib/pvsystem.py
kanderso-nrel: I'm not following the issue here, can you expand on what's different about the `setter`s?

wholmgren: After trying to come up with an example, I'm realizing the problem is a little different than I first described.

There are currently setters for only 4 properties:

* temperature_model_parameters
* surface_tilt
* surface_azimuth
* racking_model

The remaining properties do not contain setters. At first I thought this meant that for all of the other properties, a user could set something on a `PVSystem` object that's different than what they'd get back from a getter that inspects the Arrays. But I'm now reminded that the user simply will not be able to set them.

```python
In [5]: system = PVSystem([Array()])

In [6]: system.albedo = 0.3
AttributeError                            Traceback (most recent call last)
<ipython-input-6-d89b04b86f95> in <module>
----> 1 system.albedo = 0.3

AttributeError: can't set attribute
```

So the issue is that the user can set some but not all of the deprecated attributes. I'd prefer if they could set none or all.

I brought up the `__getattr__` and `__setattr__` methods only as an alternative to the properties and setters.

kanderso-nrel: > So the issue is that the user can set some but not all of the deprecated attributes. I'd prefer if they could set none or all.

I see, and agree it should be consistent.  I think omitting the setters was a (probably unintentional?) breaking change, so I think being able to set all (with a deprecation warning) makes sense.  I've added setters where they were missing.

wholmgren (approved):
@cwhanse ok to merge?

cwhanse:
> @cwhanse ok to merge?

There one warning in `modelchain.py` from codecov about a ValueError line not being hit, but it looks to me that the ValueError should have been raised by [this test](https://github.com/pvlib/pvlib-python/blob/4619a1eceb3c228fb27a6c2303f1f4793f1b0502/pvlib/tests/test_modelchain.py#L1704).

Other than that, LGTM.

kanderso-nrel:
> There one warning in `modelchain.py` from codecov about a ValueError line not being hit, but it looks to me that the ValueError should have been raised by [this test](https://github.com/pvlib/pvlib-python/blob/4619a1eceb3c228fb27a6c2303f1f4793f1b0502/pvlib/tests/test_modelchain.py#L1704).

I'm not sure what I did wrong when I checked coverage earlier in the thread, but I now agree with codecov that the ValueError was not exercised.  I checked coverage in `pvlib==0.8.1` and it's not covered there either.  The test you linked is for a similar but distinct error.  62a549e makes that test more specific and adds a test for the uncovered error as well.

cwhanse:
Merge when green!

PR #1201, opened 2021-03-24
Calculate the effect of clouds ("cloud opacity factor") on spectral irradiance (spectrl2)

trondkr:
- [ ] Closes #xxxx
 - [ x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [ ] Tests added
 - [ x] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.

This pull request adds a new function to irradiance.py which allows for the effect of cloud opacity to be accounted for when calculating spectral irradiance using spectrl2. I have added an example for how to use this to calculate the effects of clouds on irradiance.


pvlib/irradiance.py
```diff
@@ -31,6 +30,90 @@
...
+
+        irr_ghi_clouds:np.ndarray
+            Total direct irradiance (poa_global) estimated
+            using `pvlib.irradiance.get_total_irradiance` and
+            `pvlib.irradiance.campbell_norman irradiance
+            with clouds (transmittance)
+
+        spectra:np.ndarray
+            Spectral irradiance output from `pvlib.spectrum.spectrl2`
+            under clear-sky conditions
```
kanderso-nrel, line 72: 1. Similar to above, I'm not sure that this input needs a recommended source (note that [1] does not use spectrl2).  
2. I think the output of `spectrl2` is actually a dict, not an `ndarray`.  But would it make sense to take the spectral inputs as individual parameters (i.e. have the user pass in `spectra['poa_sky_diffuse'][:, 0]` etc instead of `spectra`)?  Although pvlib often returns many outputs bundled together for convenience, inputs are usually kept separate.

pvlib/irradiance.py
```diff
@@ -123,7 +206,10 @@ def _handle_extra_radiation_types(datetime_or_doy, epoch_year):
     # a better way to do it.
     if isinstance(datetime_or_doy, pd.DatetimeIndex):
         to_doy = tools._pandas_to_doy  # won't be evaluated unless necessary
-        def to_datetimeindex(x): return x                       # noqa: E306
+
+        def to_datetimeindex(x):
+            return x  # noqa: E306
```
kanderso-nrel, line 211: Unfortunately the unrelated formatting changes here and below probably aren't desirable.  Would it be easy to undo these changes, either through `git revert`, or maybe just by manually pasting over the text with the original version at https://github.com/pvlib/pvlib-python/blob/master/pvlib/irradiance.py?

I think the automated stickler-ci check will only complain about the formatting of code that is different from `pvlib/master`, so if the stuff below here is reverted back to the original, stickler shouldn't complain about it (even if your local linter does).

pvlib/irradiance.py
```diff
@@ -31,6 +30,90 @@
...
+                         irr_dir_clouds: np.ndarray,
+                         irr_ghi_clouds: np.ndarray,
+                         spectra: dict) -> (np.ndarray, np.ndarray):
+    """
+        Calculate the effect of "cloud opacity factor" on spectral
+        irradiance under clear sky.
+
+        First we calculate the rho fraction based on campbell_norman
+        irradiance with clouds converted to POA irradiance. In the
+        paper [1] these values are obtained from observations. The equations
```
kanderso-nrel, line 43:
```suggestion
        paper [1]_ these values are obtained from observations. The equations
```

Adding an underscore like this turns the [1] into a clickable link to the reference text below

pvlib/irradiance.py
```diff
@@ -31,6 +30,90 @@
...
+            `pvlib.irradiance.campbell_norman` irradiance with clouds
+            (transmittance)
+
+        irr_dir_clouds:np.ndarray
+            Total direct irradiance (poa_direct) estimated using
+            `pvlib.irradiance.get_total_irradiance` and
+            `pvlib.irradiance.campbell_norman` irradiance with
+            clouds (transmittance)
+
+        irr_ghi_clouds:np.ndarray
```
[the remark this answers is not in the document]
kanderso-nrel: Also is `np.ndarray` a requirement?  If scalars and `pd.Series` also work, then better to use the catch-all term `numeric` for the type.

pvlib/irradiance.py
```diff
@@ -31,6 +30,90 @@
                    'sea': 0.06}
 
 
+def cloud_opacity_factor(irr_dif_clouds: np.ndarray,
+                         irr_dir_clouds: np.ndarray,
+                         irr_ghi_clouds: np.ndarray,
+                         spectra: dict) -> (np.ndarray, np.ndarray):
```
kanderso-nrel, line 36: Although we might add type hints eventually (see e.g. https://github.com/pvlib/pvlib-python/pull/1146#issuecomment-778386641), I think for now better to keep the signature unannotated (and see below points about types anyway).

pvlib/irradiance.py
```diff
@@ -31,6 +30,90 @@
...
+        Materials and Solar Cells, Volume 157, 2016,
+        Pages 913-922,
+        """
+
+    rho = irr_dif_clouds / irr_ghi_clouds
+
+    wl = spectra['wavelength']
+    irr_diff_s = np.trapz(y=spectra['poa_sky_diffuse'][:, 0], x=wl)
+    irr_dir_s = np.trapz(y=spectra['poa_direct'][:, 0], x=wl)
+    irr_glob_s = np.trapz(y=spectra['poa_global'][:, 0], x=wl)
```
kanderso-nrel, line 95: Indexing with `[:, 0]` seems incorrect as it ignores every spectrum except the first.  Shouldn't all input spectra be considered, or am I thinking about this incorrectly?

Or maybe I'm confused because I assumed that the function is supposed to work on time series inputs but is actually intended to only work on one spectrum at a time?

kanderso-nrel (commented):
Thanks for opening this PR @trondkr, and sorry for the delayed review!  This seems like a good place to start for adding non-clear-sky spectral modeling to pvlib.  I focused just on the new function in `irradiance.py` here and will do another review for the example later.

The new function in `irradiance.py` will need some tests.  How should we create cases of inputs and expected outputs?  Maybe use that pvlighthouse tool?

RDaxini:
@trondkr are you still interested in completing this work by any chance? If not, I'd be happy to take over this PR to complete the implementation.

RDaxini:
@trondkr are you still interested in completing this work? If not, I'd be happy to take over this PR to complete the implementation.

PR #1211, opened 2021-04-15
Add Array.get_cell_temperature, PVSystem.get_cell_temperature

kanderso-nrel:
- ~Closes #xxxx~
 - [x] Tests added

Trying out the idea from https://github.com/pvlib/pvlib-python/pull/1176#issuecomment-803657015.  If merged, I'd then update #1176 to have `Array.get_cell_temperature` grab the model parameters from the `Array`'s `Mount`.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -67,6 +67,17 @@ Deprecations
   * ``ModelChain.weather``
   * ``ModelChain.times``
 
+* The following ``PVSystem`` cell temperature functions have been deprecated
+  and replaced with the wrapper method
+  :py:meth:`~pvlib.pvsystem.PVSystem.get_cell_temperature`:
+
+  * :py:meth:`~pvlist.pvsystem.PVSystem.sapm_celltemp`
+  * :py:meth:`~pvlist.pvsystem.PVSystem.pvsystem.PVSystem.pvsyst_celltemp`
+  * :py:meth:`~pvlist.pvsystem.PVSystem.pvsystem.PVSystem.faiman_celltemp`
+  * :py:meth:`~pvlist.pvsystem.PVSystem.pvsystem.PVSystem.fuentes_celltemp`
+  * :py:meth:`~pvlist.pvsystem.PVSystem.pvsystem.PVSystem.noct_sam_celltemp`
+
+
 Enhancements
 ~~~~~~~~~~~~
 * Add :func:`~pvlib.iotools.read_bsrn` for reading BSRN solar radiation data
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -45,6 +45,10 @@ Breaking changes
   ``surface_azimuth`` are now required parameters for
   :py:func:`pvlib.modelchain.basic_chain` (:issue:`1028`, :pull:`1181`)
 
+* Removed the ``**kwargs`` parameters from :py:meth:`~pvlib.pvsystem.PVSystem.sapm` and
+  :py:meth:`~pvlib.pvsystem.PVSystem.calcparams_desoto` and 
+  :py:meth:`~pvlib.pvsystem.PVSystem.calcparams_cec` (:issue:`1118`, :pull:`1222`)
+
 
 Deprecations
 ~~~~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -120,6 +120,9 @@ Bug fixes
 
 Testing
 ~~~~~~~
+* Added ``__init__.py`` to ``pvlib/tests`` and subdirectories. This follows
+  pytest recommendations for package layout and imports from ``conftest.py``.
+  Solves issue with test discovery in vscode. (:pull:`1204`)
 
 Documentation
 ~~~~~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -71,11 +71,11 @@ Deprecations
   and consolidated into the new wrapper method
   :py:meth:`~pvlib.pvsystem.PVSystem.get_cell_temperature` (:pull:`1211`):
 
-  * :py:meth:`~pvlist.pvsystem.PVSystem.sapm_celltemp`
-  * :py:meth:`~pvlist.pvsystem.PVSystem.pvsystem.PVSystem.pvsyst_celltemp`
-  * :py:meth:`~pvlist.pvsystem.PVSystem.pvsystem.PVSystem.faiman_celltemp`
-  * :py:meth:`~pvlist.pvsystem.PVSystem.pvsystem.PVSystem.fuentes_celltemp`
-  * :py:meth:`~pvlist.pvsystem.PVSystem.pvsystem.PVSystem.noct_sam_celltemp`
+  * :py:meth:`~pvlib.pvsystem.PVSystem.sapm_celltemp`
+  * :py:meth:`~pvlib.pvsystem.PVSystem.pvsyst_celltemp`
+  * :py:meth:`~pvlib.pvsystem.PVSystem.faiman_celltemp`
+  * :py:meth:`~pvlib.pvsystem.PVSystem.fuentes_celltemp`
+  * :py:meth:`~pvlib.pvsystem.PVSystem.noct_sam_celltemp`
 
 
 Enhancements
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -67,8 +67,8 @@ Deprecations
   * ``ModelChain.weather``
   * ``ModelChain.times``
 
-* The following ``PVSystem`` cell temperature functions have been deprecated
-  and replaced with the wrapper method
+* The following ``PVSystem`` cell temperature methods have been deprecated
+  and consolidated into the new wrapper method
   :py:meth:`~pvlib.pvsystem.PVSystem.get_cell_temperature` (:pull:`1211`):
 
   * :py:meth:`~pvlist.pvsystem.PVSystem.sapm_celltemp`
--- a/docs/sphinx/source/whatsnew/v0.9.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.0.rst
@@ -77,6 +77,8 @@ Deprecations
   * :py:meth:`~pvlib.pvsystem.PVSystem.fuentes_celltemp`
   * :py:meth:`~pvlib.pvsystem.PVSystem.noct_sam_celltemp`
 
+* The ``eta_m`` parameter for :py:func:`~pvlib.temperature.pvsyst_cell` is
+  replaced by parameter ``module_efficiency``. (:issue:`1188`, :pull:`1218`)
 
 Enhancements
 ~~~~~~~~~~~~
@@ -128,6 +130,8 @@ Bug fixes
 * Reindl model fixed to generate sky_diffuse=0 when GHI=0.
   (:issue:`1153`, :pull:`1154`)
 * Update GFS product names for GFS v16. (:issue:`1202`, :pull:`1203`)
+* Corrected methodology error in :py:func:`~pvlib.scaling.wvm`. Tracks with
+  fix in PVLib for MATLAB. (:issue:`1206`, :pull:`1213`)
 
 Testing
 ~~~~~~~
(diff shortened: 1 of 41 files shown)


kanderso-nrel:
PSM3 test failures presumably related to https://github.com/NREL/developer.nrel.gov/issues/208

wholmgren (approved):
Overall this looks really good to me. The dancing around `effective_irradiance` is a little awkward, but I don't have a good recommendation for an alternative.

docs/sphinx/source/whatsnew/v0.9.0.rst
```diff
@@ -67,6 +67,17 @@ Deprecations
   * ``ModelChain.weather``
   * ``ModelChain.times``
 
+* The following ``PVSystem`` cell temperature methods have been deprecated
+  and consolidated into the new wrapper method
+  :py:meth:`~pvlib.pvsystem.PVSystem.get_cell_temperature` (:pull:`1211`):
+
+  * :py:meth:`~pvlist.pvsystem.PVSystem.sapm_celltemp`
+  * :py:meth:`~pvlist.pvsystem.PVSystem.pvsystem.PVSystem.pvsyst_celltemp`
```
cwhanse, line 75: duplicate steps in each path

pvlib/modelchain.py
```diff
@@ -1009,26 +1009,26 @@ def _set_celltemp(self, model):
         temp_air = _tuple_from_dfs(self.results.weather, 'temp_air')
         wind_speed = _tuple_from_dfs(self.results.weather, 'wind_speed')
         kwargs = {}
-        if model == self.system.noct_sam_celltemp:
+        if model == 'noct_sam':
```
cwhanse, line 1012: The docstring says that `model` is a function, it's changed to be a string here and elsewhere.

With this change, is `ModelChain.temperature_model` needed?

pvlib/pvsystem.py
```diff
@@ -1473,6 +1449,83 @@ def get_iam(self, aoi, iam_model='physical'):
...
+                'absorption', 'surface_tilt', 'module_width', 'module_length'],
+                self.temperature_model_parameters)
+            if 'surface_tilt' not in params:
+                params['surface_tilt'] = self.surface_tilt
+        elif model == 'noct_sam':
+            func = temperature.noct_sam
+            params = _build_kwargs(['transmittance_absorptance', 'eta_m_ref',
+                                    'noct', 'array_height', 'mount_standoff'],
+                                   self.temperature_model_parameters)
+            if not {'noct', 'eta_m_ref'}.issubset(params):
```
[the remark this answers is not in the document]
kanderso-nrel: I'm not sure how I feel about this.  Without a check like this, the error message isn't as helpful as it could be -- `TypeError: noct_sam() missing 1 required positional argument: 'noct'` makes no mention of `temperature_model_parameters`.  But I'm not really a fan of having all these lists of `pvlib.temperature` parameter names in `pvlib.pvsystem`.

What do you think about letting python raise the `TypeError` and we add a hint to the error message like this:

```python
try:
    temperature_cell = func(poa_global, temp_air, wind_speed, **params)
except TypeError as e:
    msg = e.args[0] + (
        ".  Check temperature_model_parameters to make sure "
        "all required parameters are specified."
    )
    raise TypeError(msg)
```

Downside is that it adds irrelevant noise to TypeErrors caused by other problems.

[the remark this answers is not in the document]
cwhanse: I won't stand in the way of this PR over error messaging. If we want to verify parameters at this stage, as a kindness to users, maybe a dict in `pvlib.temperature` akin to `pvlib.pvsystem._DC_MODEL_PARAMS`.

pvlib/pvsystem.py
```diff
@@ -378,6 +378,72 @@ def get_iam(self, aoi, iam_model='physical'):
         return tuple(array.get_iam(aoi, iam_model)
                      for array, aoi in zip(self.arrays, aoi))
 
+    @_unwrap_single_value
+    def get_cell_temperature(self, poa_global, temp_air, wind_speed, model,
```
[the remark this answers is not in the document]
wholmgren: I'm not concerned about the name here because it's just an argument. I'm more concerned about the column name in the dataframe for `run_model` and variants.

pvlib/pvsystem.py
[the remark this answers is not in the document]
wholmgren: I think the problem is that we're using `_build_kwargs` to build all of the arguments instead of just the optional keyword arguments. I suggest a pattern that performs explicit look ups instead of deferred checks. Fine with me to use the standard python KeyError or break into the kinder message template.

```python
args = (self.temperature_model_par[arg] for arg in required_args)
kwargs = _build_kwargs(optional_args, self.temperature_model_parameters) 
```

required_args and optional_args could come from a private dict like @cwhanse suggests. I'm also ok with hard coding it here (sounds easy). And I'm ok with using `inspect` (I vaguely recall considering this previously and thinking it was too complicated for pvlib, but I currently like it).

pvlib/pvsystem.py
```diff
@@ -1473,6 +1449,83 @@ def get_iam(self, aoi, iam_model='physical'):
...
+                msg = ('Parameters noct and eta_m_ref are required.'
+                       ' Found {} in temperature_model_parameters.'
+                       .format(self.temperature_model_parameters))
+                raise KeyError(msg)
+        else:
+            raise ValueError(f'{model} is not a valid cell temperature model')
+
+        # allow kwargs to override
+        params.update(kwargs)
+        temperature_cell = func(poa_global, temp_air, wind_speed, **params)
```
wholmgren, line 1526: and here I would use `*args` and `**kwargs`

pvlib/pvsystem.py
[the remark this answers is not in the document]
kanderso-nrel: I added something along these lines, let me know what you think.

pvlib/modelchain.py
```diff
@@ -992,12 +992,9 @@ def _set_celltemp(self, model):
...
         ----------
-        model : function
-            A function that takes POA irradiance, air temperature, and
-            wind speed and returns cell temperature. `model` must accept
-            tuples or single values for each parameter where each element of
-            the tuple is the value for a different array in the system
-            (see :py:class:`pvlib.pvsystem.PVSystem` for more information).
+        model : str
+            A cell temperature model name to pass to
+            :py:meth:`pvlib.pvsystem.PVSystem.get_cell_temperature`.
```
cwhanse, line 997:
```suggestion
            :py:meth:`pvlib.pvsystem.PVSystem.get_cell_temperature`.
            Valid names are 'sapm', 'pvsyst', 'faiman', 'fuentes', 'noct_sam'
```

pvlib/pvsystem.py
```diff
@@ -1475,6 +1448,97 @@ def get_iam(self, aoi, iam_model='physical'):
...
+        -------
+        numeric
+            values in degrees C.
+
+        Notes
+        -----
+        Some temperature models have requirements for the input types;
+        see the documentation of the underlying model function for details.
+        """
+        def _build_args(keys):
```
cwhanse, line 1484: maybe put this in `pvlib.tools`? OK if that is deferred until another use arises.

pvlib/pvsystem.py
```diff
@@ -1475,6 +1448,97 @@ def get_iam(self, aoi, iam_model='physical'):
...
+                missing_key = e.args[0]
+                msg = (f"Missing required parameter '{missing_key}'. Found "
+                       f"{self.temperature_model_parameters} in "
+                       "temperature_model_parameters.")
+                raise KeyError(msg)
+            return args
+
+        if model == 'sapm':
+            func = temperature.sapm_cell
+            required = _build_args(['a', 'b', 'deltaT'])
```
cwhanse, line 1497: I like this `required`, `optional` pattern. Easy to read and intuitive to me.

kanderso-nrel:
Checks are green.  Anything left on this one?

pvlib/pvsystem.py
```diff
@@ -378,6 +378,71 @@ def get_iam(self, aoi, iam_model='physical'):
...
+            Supported models include ``'sapm'``, ``'pvsyst'``,
+            ``'faiman'``, ``'fuentes'``, and ``'noct_sam'``
+
+        **kwargs
+            Extra arguments passed to the model function.
+
+        Returns
+        -------
+        numeric or tuple of numeric
+            values in degrees C.
```
wholmgren, line 408:
```suggestion
            Values in degrees C.
```

pvlib/pvsystem.py
```diff
@@ -1466,6 +1439,90 @@ def get_iam(self, aoi, iam_model='physical'):
...
+            Supported models include ``'sapm'``, ``'pvsyst'``,
+            ``'faiman'``, ``'fuentes'``, and ``'noct_sam'``
+
+        **kwargs
+            Extra arguments passed to the model function.
+
+        Returns
+        -------
+        numeric
+            values in degrees C.
```
wholmgren, line 1468:
```suggestion
            Values in degrees C.
```

pvlib/pvsystem.py
```diff
@@ -1466,6 +1439,90 @@ def get_iam(self, aoi, iam_model='physical'):
...
+
+        Notes
+        -----
+        Some temperature models have requirements for the input types;
+        see the documentation of the underlying model function for details.
```
wholmgren, line 1473: This note is begging for a "See Also" section.

pvlib/pvsystem.py
```diff
@@ -1466,6 +1439,90 @@ def get_iam(self, aoi, iam_model='physical'):
...
+        """
+        def _build_tcell_args(keys):
+            # convenience wrapper to avoid passing args 2 and 3 every call
+            return _build_args(keys, self.temperature_model_parameters,
+                               'temperature_model_parameters')
```
wholmgren, line 1478: not opposed, but is there a reason to prefer this to `functools.partial`?

pvlib/pvsystem.py
```diff
@@ -1466,6 +1439,90 @@ def get_iam(self, aoi, iam_model='physical'):
...
+            func = temperature.noct_sam
+            required = _build_tcell_args(['noct', 'module_efficiency'])
+            optional = _build_kwargs(['transmittance_absorptance',
+                                      'array_height', 'mount_standoff'],
+                                     self.temperature_model_parameters)
+        else:
+            raise ValueError(f'{model} is not a valid cell temperature model')
+
+        # allow kwargs to override
+        optional.update(kwargs)
```
wholmgren, line 1521: Is there a compelling use case for kwargs here?

kwargs may be a necessary evil in `PVSystem.get_cell_temperature` due to `effective_irradiance`, but I'm -1 on allowing kwargs to arbitrarily update the parameters here. The object layer is already complicated and difficult to maintain. I also keep hitting bugs in my own code due to kwargs.

wholmgren: I also don't see that this is tested.

pvlib/tests/test_modelchain.py
```diff
@@ -642,13 +642,14 @@ def test_run_model_with_weather_sapm_temp(sapm_dc_snl_ac_system, location,
...
     mc.temperature_model = 'sapm'
-    m_sapm = mocker.spy(sapm_dc_snl_ac_system, 'sapm_celltemp')
+    m_sapm = mocker.spy(sapm_dc_snl_ac_system, 'get_cell_temperature')
     mc.run_model(weather)
     assert m_sapm.call_count == 1
     # assert_called_once_with cannot be used with series, so need to use
     # assert_series_equal on call_args
     assert_series_equal(m_sapm.call_args[0][1], weather['temp_air'])  # temp
     assert_series_equal(m_sapm.call_args[0][2], weather['wind_speed'])  # wind
+    assert m_sapm.call_args[1]['model'] == 'sapm'
```
wholmgren, line 652: No objection to this. An alternative is modifying the spy line to `mocker.spy(temperature, 'sapm_cell')` and dropping this line.

pvlib/pvsystem.py
```diff
@@ -378,6 +378,71 @@ def get_iam(self, aoi, iam_model='physical'):
...
+        passed as a tuple then the same value is used for each Array.
+        """
+        poa_global = self._validate_per_array(poa_global)
+        temp_air = self._validate_per_array(temp_air, system_wide=True)
+        wind_speed = self._validate_per_array(wind_speed, system_wide=True)
+
+        # additional model-specific (and array-specific) inputs
+        extra_inputs = [{}] * self.num_arrays
+
+        if 'effective_irradiance' in kwargs:
```
wholmgren, line 428: Is there a reason we cannot add `effective_irradiance=None` to the method signature and drop `kwargs`? I don't see one, but I feel like I might be forgetting something from a previous discussion.

wholmgren (commented):
Close. A few minor things and a larger question about if we can replace `**kwargs` with `effective_irradiance=None`.

pvlib/pvsystem.py
kanderso-nrel: Not sure why I thought this was a good idea.  Maybe I forgot that `temperature_model_parameters` existed.  I agree that kwargs is unnecessary and undesirable here, and it's now removed.

pvlib/pvsystem.py
kanderso-nrel: I think I was anticipating new models with new weather inputs, but better to cross that bridge if we come to it.  The code is much simpler now without this.

wholmgren (approved):
Thanks @kanderso-nrel. @cwhanse can you take another look, especially double checking the handling of `effective_irradiance`? Ok with me to merge if you approve.

pvlib/modelchain.py
cwhanse: > With this change, is ModelChain.temperature_model needed?

I think this question remains, but it should be explored in a separate issue.

cwhanse (approved):
It looks to me that `ModelChain.temperature_model` becomes obsolete with addition of these methods. I could be mistaken. Separate issue, in either case.

PR #1247, opened 2021-06-13
Add bhi, pressure, and wind_ to list of variables

AdamRJensen:
All the terms except bhi are in agreement with pv-terms.

The BHI (Beam Horizontal Irradiance) term was introduced in the variable map for the `get_cams`  and `read_cams` functions in #1175 

### DIFF
--- a/pvlib/data/variables_style_rules.csv
+++ b/pvlib/data/variables_style_rules.csv
@@ -5,6 +5,7 @@ longitude;longitude
 dni;direct normal irradiance
 dni_extra;direct normal irradiance at top of atmosphere (extraterrestrial)
 dhi;diffuse horizontal irradiance
+bhi;beam/direct horizontal irradiance
 ghi;global horizontal irradiance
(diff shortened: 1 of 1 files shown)


pvlib/data/variables_style_rules.csv
```diff
@@ -27,6 +28,9 @@ temp_module;temperature of the module
 temp_air;temperature of the air
 temp_dew;dewpoint temperature
 relative_humidity;relative humidity
+wind_speed;wind speed
+wind_direction;wind direction
```
kanderso-nrel, line 32: Should we document the angle convention like we do for `solar_azimuth`, e.g. `degrees East of North`?

cwhanse: I'm not opposed, but, this is primarily a list of terms and (where needed) definitions. It seems better to leave units, angle conventions and limits to the function docstrings.

wholmgren (approved):

PR #1251, opened 2021-07-02
Townsend snow

### LINKED ISSUE #1246: Implement monthly snow loss model
Interesting model for predicting monthly energy loss due to snow, from monthly quantities (snow total, plane-of-array insolation).

Photovoltaics and snow: An update from two winters of measurements in the SIERRA
June 2011Conference Record of the IEEE Photovoltaic Specialists Conference
[DOI: 10.1109/PVSC.2011.6186627](http://doi.org/10.1109/PVSC.2011.6186627)

Would be a nice addition to `pvlib.snow`

abhisheksparikh:
- [x] Closes #1246
 - [x] Tests added


pvlib/snow.py
```diff
@@ -185,3 +185,104 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    .. [1] Townsend, Tim & Powers, Loren. (2011). Photovoltaics and snow: An 
+            update from two winters of measurements in the SIERRA. Conference 
+            Record of the IEEE Photovoltaic Specialists Conference. 
+            003231-003236. 10.1109/PVSC.2011.6186627. 
+       Available at https://www.researchgate.net/publication/261042016_Photovoltaics_and_snow_An_update_from_two_winters_of_measurements_in_the_SIERRA
+
+    '''
+    return(np.where(N>0, 0.5 * S * (1 + 1/N), 0))
+
+def townsend_snow_loss_model(S, N, tilt, RH, T_air, POA, R, H, P=40):
```
cwhanse, line 219: Prefer `loss_townsend`, and would prefer more descriptive parameter names e.g., `snow_total` instead of `S`, `num_events` instead of `N`. It's a good idea to maintain traceability to the terms used in the reference but we can do that in the description of each parameter.

pvlib/snow.py
```diff
@@ -185,3 +185,104 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+        Number of snowfall events with snowfall > 1"
+
+    tilt : numeric
+        Array tilt in degrees
+
+    RH : numeric
+        Relative humidity in percentage
+
+    T_air : numeric
+        Ambient temperature in celcius
```
cwhanse, line 238:
```suggestion
    temp_air : numeric
        Ambient temperature [C].
```

Where these variables are already in pvlib we should use the same parameter name.

pvlib/snow.py
```diff
@@ -185,3 +185,104 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+
+    POA : numeric
+        Plane of array irradiance in kWh/m2/month
```
cwhanse, line 241: This is insolation not irradiance, correct?

pvlib/snow.py
```diff
@@ -185,3 +185,104 @@ def dc_loss_nrel(snow_coverage, num_strings):
        Available at https://www.nrel.gov/docs/fy18osti/67399.pdf
     '''
     return np.ceil(snow_coverage * num_strings) / num_strings
+
+def townsend_Se(S, N):
```
cwhanse, line 189: I would make this function private.

cwhanse (commented):
Thanks @abhisheksparikh this is a great start. Some comments for you to consider as you work through the formatting.

pvlib/snow.py
mikofski: Do we have a document with existing variable names?

kanderso-nrel: > Do we have a document with existing variable names?

Yes, although it would probably benefit from an update: https://pvlib-python.readthedocs.io/en/latest/variables_style_rules.html

abhisheksparikh:
Excuse me for the long delay! I will be working on resolving the issues of this PR in the coming couple of days.

pvlib/snow.py
```diff
@@ -185,3 +185,107 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    ----------
+    snow_total : numeric
+        Inches of snow received in the current month. Referred as S in the
+        paper
+
+    snow_events : numeric
+        Number of snowfall events with snowfall > 1". Referred as N in the
+        paper
+
+    tilt : numeric
```
kanderso-nrel, line 237: `surface_tilt`, again for consistency with the rest of the library

kanderso-nrel (commented):
Thanks @abhisheksparikh, looking good so far!  Can you please also add an entry to `docs/sphinx/source/whatsnew/v0.9.1.rst` under `Enhancements`, and yourself to the contributors list?

pvlib/snow.py
```diff
@@ -185,3 +185,107 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    relative_humidity : numeric
+        Relative humidity in percentage
+
+    temp_air : numeric
+        Ambient temperature [C]
+
+    poa_global : numeric
+        Plane of array insolation in kWh/m2/month
+
+    row_len : float
```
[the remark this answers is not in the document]
abhisheksparikh: Yep - will be changed

pvlib/snow.py
```diff
@@ -185,3 +185,107 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+        Row length in the slanted plane of array dimension in inches
+
+    H : float
+        Drop height from array edge to ground in inches
+
+    P : float
```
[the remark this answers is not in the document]
abhisheksparikh: For `H`, I think I will use `lower_edge_drop_height`. For `P`, I like `angle_of_repose`.

pvlib/snow.py
```diff
@@ -185,3 +185,107 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+        Average monthly DC capacity loss in percentage due to snow coverage
+
+    References
+    ----------
+    .. [1] Townsend, Tim & Powers, Loren. (2011). Photovoltaics and snow: An
+            update from two winters of measurements in the SIERRA. Conference
+            Record of the IEEE Photovoltaic Specialists Conference.
+            003231-003236. 10.1109/PVSC.2011.6186627.
+       Available at https://www.researchgate.net/publication/261042016_Photovoltaics_and_snow_An_update_from_two_winters_of_measurements_in_the_SIERRA
+    '''# noqa
```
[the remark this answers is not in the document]
abhisheksparikh: Good to know - using `# noqa: E501`

pvlib/snow.py
```diff
@@ -185,3 +185,107 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    snow_total_prev = np.roll(snow_total, 1)
+    snow_events_prev = np.roll(snow_events, 1)
+
+    Se = _townsend_Se(snow_total, snow_events)
+    Se_prev = _townsend_Se(snow_total_prev, snow_events_prev)
+
+    Se_weighted = 1/3 * Se_prev + 2/3 * Se
+    gamma = (row_len * Se_weighted * np.cos(np.deg2rad(tilt))) / \
+        (np.clip((H**2 - Se_weighted**2), a_min=0.01, a_max=None) / 2 /
+            np.tan(np.deg2rad(P)))
```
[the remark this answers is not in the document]
abhisheksparikh: Good to know - will switch to those ones 👍

pvlib/snow.py
```diff
@@ -185,3 +185,107 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    loss : numeric
+        Average monthly DC capacity loss in percentage due to snow coverage
+
+    References
+    ----------
+    .. [1] Townsend, Tim & Powers, Loren. (2011). Photovoltaics and snow: An
+            update from two winters of measurements in the SIERRA. Conference
+            Record of the IEEE Photovoltaic Specialists Conference.
+            003231-003236. 10.1109/PVSC.2011.6186627.
+       Available at https://www.researchgate.net/publication/261042016_Photovoltaics_and_snow_An_update_from_two_winters_of_measurements_in_the_SIERRA
```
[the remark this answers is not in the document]
abhisheksparikh: I will remove these whitespaces in my commit

pvlib/snow.py
```diff
@@ -185,3 +185,107 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    Se_weighted = 1/3 * Se_prev + 2/3 * Se
+    gamma = (row_len * Se_weighted * np.cos(np.deg2rad(tilt))) / \
+        (np.clip((H**2 - Se_weighted**2), a_min=0.01, a_max=None) / 2 /
+            np.tan(np.deg2rad(P)))
+
+    GIT = 1 - C2 * np.exp(-gamma)
+    loss = C1 * Se_weighted * (np.cos(np.deg2rad(tilt)))**2 * GIT * \
+        relative_humidity / (temp_air+273.15)**2 / poa_global**0.67
+
+    return (np.round(loss, 2))
```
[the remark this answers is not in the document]
abhisheksparikh: Will leave non-rounded losses.

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+
+    P : float
+        piled snow angle, assumed to stabilize at 40° , the midpoint of
+        25°-55° avalanching slope angles [deg]
+
+    Returns
+    -------
+    loss : numeric
+        Average monthly DC capacity loss due to snow coverage [%]
+
```
kanderso-nrel, line 264: Let's add a `Notes` section here pointing out that this model has not been validated for tracking arrays, but the reference suggests using the maximum rotation angle in place of `surface_tilt`.

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    poa_global : numeric
+        Plane of array insolation [kWh/m2/month]
+
+    slant_height : float
+        Row length in the slanted plane of array dimension [in]
+
+    lower_edge_drop_height : float
+        Drop height from array edge to ground [in]
+
+    P : float
```
kanderso-nrel, line 256: This `P` slipped through the cracks :)

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    Se_prev = _townsend_Se(snow_total_prev, snow_events_prev)
+
+    Se_weighted = 1/3 * Se_prev + 2/3 * Se
+    gamma = (slant_height * Se_weighted * cosd(surface_tilt)) / \
+        (np.clip((lower_edge_drop_height**2 - Se_weighted**2), a_min=0.01,
+         a_max=None) / 2 / tand(angle_of_repose))
+
+    GIT = 1 - C2 * np.exp(-gamma)
+    loss = (C1 * Se_weighted * (cosd(surface_tilt))**2 * GIT *
+            relative_humidity / (temp_air+273.15)**2 / poa_global**0.67) / 100
```
kanderso-nrel, line 290: Good catch that temperature is in Kelvin 👍

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+        paper [-]
+
+    surface_tilt : numeric
+        Array surface_tilt [deg]
+
+    relative_humidity : numeric
+        Relative humidity [%]
+
+    temp_air : numeric
+        Ambient temperature [°C]
```
kanderso-nrel, line 245: I think it's worth explicitly using the word "monthly" for all the monthly inputs, and "average/total/etc" as appropriate.  Temperature is monthly average, and I'm not seeing it in the reference but I guess RH is monthly average as well?

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    discussed in [1]_.
+
+    Parameters
+    ----------
+    snow_total : numeric
+        Inches of snow received in the current month. Referred as S in the
+        paper [in]
+
+    snow_events : numeric
+        Number of snowfall events with snowfall > 1". Referred as N in the
```
kanderso-nrel, line 235: Is this `> 1"` requirement in the reference?  I'm not seeing it.

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+        Ambient temperature [°C]
+
+    poa_global : numeric
+        Plane of array insolation [kWh/m2/month]
+
+    slant_height : float
+        Row length in the slanted plane of array dimension [in]
+
+    lower_edge_drop_height : float
+        Drop height from array edge to ground [in]
```
kanderso-nrel, line 254: To discuss: what units should `snow_total`, `slant_height`, and `lower_edge_drop_height` use?  I propose `cm`, `m`, `m` respectively, all for consistency with other pvlib functions.

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+
+    P : float
+        piled snow angle, assumed to stabilize at 40° , the midpoint of
+        25°-55° avalanching slope angles [deg]
+
+    Returns
+    -------
+    loss : numeric
+        Average monthly DC capacity loss due to snow coverage [%]
```
kanderso-nrel, line 263:
```suggestion
        Average monthly DC capacity loss fraction due to snow coverage
```

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    snow_total_prev = np.roll(snow_total, 1)
+    snow_events_prev = np.roll(snow_events, 1)
+
+    Se = _townsend_Se(snow_total, snow_events)
+    Se_prev = _townsend_Se(snow_total_prev, snow_events_prev)
+
+    Se_weighted = 1/3 * Se_prev + 2/3 * Se
+    gamma = (slant_height * Se_weighted * cosd(surface_tilt)) / \
+        (np.clip((lower_edge_drop_height**2 - Se_weighted**2), a_min=0.01,
+         a_max=None) / 2 / tand(angle_of_repose))
```
kanderso-nrel, line 286: I don't disagree with the spirit of having a lower bound of 0.01, but is there anything special about that value other than being small but nonzero?  May be worth a comment explaining it.

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+def loss_townsend(snow_total, snow_events, surface_tilt, relative_humidity,
+                  temp_air, poa_global, slant_height, lower_edge_drop_height,
+                  angle_of_repose=40):
+    '''
+    Calculates monthly snow loss based on a generalized monthly snow loss model
+    discussed in [1]_.
+
+    Parameters
+    ----------
+    snow_total : numeric
```
kanderso-nrel, line 230: Our current definition of `numeric` includes scalars (e.g. `float`), which I don't think makes sense here.  Would `array-like` make sense?  See https://pvlib-python.readthedocs.io/en/stable/contributing.html#documentation

pvlib/tests/test_snow.py
```diff
@@ -95,3 +95,32 @@ def test_dc_loss_nrel():
...
+
+def test__townsend_Se():
+    S = np.array([10, 10, 5, 1, 0, 0, 0, 0, 0, 0, 5, 10])
+    N = np.array([2, 2, 1, 0, 0, 0, 0, 0, 0, 0, 2, 3])
+    expected = np.array([7.5, 7.5, 5, 0, 0, 0, 0, 0, 0, 0, 3.75, 6.66666667])
+    actual = snow._townsend_Se(S, N)
+    np.testing.assert_allclose(expected, actual, rtol=1e-07)
+
+
+def test_loss_townsend():
```
kanderso-nrel, line 108: Question for @cwhanse: referring to https://github.com/pvlib/pvlib-python/pull/1393#issuecomment-1022336293, should we have a policy of always testing both array and Series for `array-like` parameters?

kanderso-nrel (commented):
More nitpicking.  I think after this it will be close!

pvlib/snow.py
abhisheksparikh: Nope - it isn't, hence removing it. I guess it is better to leave that up to the user (to define what a snow event is).

pvlib/snow.py
abhisheksparikh: I am prefer metric system over everything else :)

abhisheksparikh: I can change `slant_height` and `lower_edge_drop_height` inputs to `m`. However, I think, we should leave `snow_total` in `in` because that's what is available in most of the weather sources. Thoughts?

pvlib/snow.py
abhisheksparikh: I think `a_min` is set to a bound of 0.01 to prevent gamma from blowing up.

kanderso-nrel: Right, I was more wondering why specifically 0.01 is the lower limit (why not 0.1, 0.001, or any other small nonzero value?).  I guess I was more curious than anything else -- the specific lower bound, so long as it's <<1, doesn't seem to matter much.  Feel free to ignore this comment :)

mikofski: A common small value is eps or sometimes `EPS**(1/3)` but I don’t know the mathematics behind why. EPS is the smallest number the stupendous can represent on the runtime machine. See `np.finfo`
https://numpy.org/doc/stable/reference/generated/numpy.finfo.html

pvlib/tests/test_snow.py
cwhanse: > should we have a policy of always testing both array and Series for array-like parameters?

I think so

pvlib/snow.py
cwhanse: In this case, it seems the danger is having negative `gamma`, which upstream could result in a performance gain (from negative GIT in Eq. 3). To me, clipping `gamma` to be positive (not zero) is reasonable. It is not clear if `gamma` should have an upper bound, I haven't untangled the model equations enough to determine,

pvlib/snow.py
wholmgren: NOAA NCEI provides monthly climate normals in both metric and imperial units. 

I agree that sticking with metric units in pvlib is preferable. I think sticking with a single unit system within a single function is a must.

pvlib/tests/test_snow.py
wholmgren: I'd say it's good practice but importance and extent depends on the model and implementation. I'm usually more concerned that we test for compatibility with scalars in "float, array, series" situations because I think it's easier for regressions to slip in with scalars (e.g. by introducing masking).

wholmgren (commented):
This is looking great @abhisheksparikh! I made some suggestions for readability but I'm ok if the group prefers the terser code that more closely matches the reference.

pvlib/snow.py
wholmgren: My comment above is not correct. Sorry. Gamma will blow up as effective snow fall approaches the drop height. Beyond that, gamma is large and negative, eventually approaching 0 again. 

image

The ground interference term goes negative when gamma is smaller than ln(c2) ~= -0.67.

pvlib/snow.py
```diff
@@ -185,3 +185,108 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+       update from two winters of measurements in the SIERRA. Conference
+       Record of the IEEE Photovoltaic Specialists Conference.
+       003231-003236. :doi:`10.1109/PVSC.2011.6186627`
+       Available at https://www.researchgate.net/publication/261042016_Photovoltaics_and_snow_An_update_from_two_winters_of_measurements_in_the_SIERRA
+
+    '''  # noqa: E501
+    return(np.where(N > 0, 0.5 * S * (1 + 1/N), 0))
+
+
+def loss_townsend(snow_total, snow_events, surface_tilt, relative_humidity,
```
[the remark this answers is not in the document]
mikofski: It's a tough call, but I believe most folks already refer to this model as the "Townsend snow model"

kanderso-nrel:
Superseded by #1468

PR #1254, opened 2021-07-05
Add retrieval function for BSRN data

AdamRJensen:
 - [x] Tests added

Add a retrieval function for BSRN data `pvlib.iotools.get_bsrn`. Data is obtained from the BSRN's FTP server.


AdamRJensen:
@kanderso-nrel This is my initial idea for the `get_bsrn` function. 

The `read_bsrn` function needs to be updated to take a file buffer or a `parse_bsrn` function needs to be created.

wholmgren:
@AdamRJensen our [solarforecastarbiter-core adaptation](https://github.com/SolarArbiter/solarforecastarbiter-core/blob/master/solarforecastarbiter/io/fetch/bsrn.py) of the BSRN code that you contributed to pvlib includes a change to make a `parse_bsrn` function that only takes a file buffer. Maybe that would be helpful to copy back over here.

AdamRJensen:
@kanderso-nrel Any idea of why the Linux versions fail, but the others pass?

The log states that the test `test_get_bsrn_no_files` is failing on Linux due to "Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) was emitted." 

My knowledge of Linux is too limited to understand why it would fail on Linux and not the others.. any ideas?

kanderso-nrel:
> @kanderso-nrel Any idea of why the Linux versions fail, but the others pass?

I don't think it's actually a linux-specific issue, it's just that only the `conda_linux` versions are configured to run the `remote_data` tests.  I can recreate the CI failure on a local ubuntu VM.  The problem is that it actually does find a file for the bogus date -- I get back data for the month of `2000-06`.  I guess the issue is that files are named with the format `f"{station}%m%y.dat.gz"`, so year 1800 gets formatted as year 00 and interpreted as year 2000.  Changing the `start` and `end` from 1800 to 1899 gets the test to pass for me.

AdamRJensen:
@cwhanse Any opinion on adding **"gri" (ground-reflected irradiance)** to the pvlib list of variables? Better alternatives for the naming of *upward short-wave reflected irradiance*?

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+       Global Climate Observing System (GCOS) GCOS-174.
+       <https://bsrn.awi.de/fileadmin/user_upload/bsrn.awi.de/Publications/gcos-174.pdf>`_
+    """  # noqa: E501
+    # The FTP server uses lowercase station abbreviations
+    station = station.lower()
 
-    username and password are easily obtainable as described in the BSRN's
-    Data Release Guidelines [4]_.
+    # Generate list files to download based on start/end (SSSMMYY.dat.gz)
+    filenames = pd.date_range(start, end + pd.DateOffset(months=1), freq='1M')\
```
kanderso-nrel, line 156: When passing `start=end=2019-01-31`, data for both 2019-01 and 2019-02 are returned.  It's a little clunky, but what do you think about `pd.date_range(start.replace(day=1), end + pd.tseries.offsets.MonthEnd(n=0), freq='1M')`?

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+                    raise ftplib.Error(response)
+                # Save file locally if local_path is specified
+                if local_path is not None:
+                    # Create local file
+                    with open(os.path.join(local_path, filename), 'wb') as f:
+                        f.write(bio.getbuffer())  # Write local file
+                # Decompress/unzip and decode the binary file
+                text = gzip.decompress(bio.getvalue()).decode('latin1')
+                # Convert string to StringIO and parse data
+                dfi, metadata = parse_bsrn(io.StringIO(text), logical_records)
```
kanderso-nrel, line 186: There's a lot of shuffling data to/from buffers here (note that `gzip.decompress` wraps the input in a buffer as well).  I wonder if it might be cleaner to use `gzip.GzipFile` directly, maybe with an `io.TextIOWrapper`?  Maybe not, and probably this is a small fraction of total runtime anyway, so maybe not worth optimizing.

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+    metadata['version of data'] = int(date_line.split()[-1])
+    for line in fbuf:
+        if line[2:6] == '0004':  # stop once LR0004 has been reached
+            break
+        elif line == '':
+            raise ValueError('Mandatory record LR0004 not found.')
+    metadata['date when station description changed'] = fbuf.readline().strip()
+    metadata['surface type'] = int(fbuf.readline(3))
+    metadata['topography type'] = int(fbuf.readline())
+    metadata['address'] = fbuf.readline().strip().strip()
```
kanderso-nrel, line 256: More for my own education: what does the second `strip()` accomplish here?

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+    metadata['telephone no. of station'] = fbuf.readline(20).strip()
+    metadata['FAX no. of station'] = fbuf.readline().strip()
+    metadata['TCP/IP no. of station'] = fbuf.readline(15).strip()
+    metadata['e-mail address of station'] = fbuf.readline().strip()
+    metadata['latitude'] = float(fbuf.readline(8))
+    metadata['longitude'] = float(fbuf.readline(8))
```
kanderso-nrel, line 262: The BSRN's lat/lon convention isn't consistent with pvlib's.  Should we convert it?  As always I'm torn between making things convenient for pvlib users vs faithfully reporting whatever the provider gives us.

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+        else:
+            horizon += [int(i) for i in line.split()]
+    horizon = pd.Series(horizon[1::2], horizon[::2], name='horizon_elevation',
+                        dtype=int).drop(-1, errors='ignore').sort_index()
+    horizon.index.name = 'azimuth'
+    metadata['horizon'] = horizon
+
+    # Read file and store the starting line number and number of lines for
+    # each logical record (LR)
+    fbuf.seek(0)  # reset buffer to start of file
```
kanderso-nrel, line 281: What's the reason for `seek`ing back to the beginning here?  I think the data records will always be after `LR0004`, so is there any problem with just mapping the sections that appear after `LR0004`?

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+    lr_startrow = {}  # Dictionary of starting line number for each LR
+    lr_nrows = {}  # Dictionary of end line number for each LR
+    for num, line in enumerate(fbuf):
+        if line.startswith('*'):  # Find start of all logical records
+            if len(lr_startrow) >= 1:
+                lr_nrows[lr] = num - max(lr_startrow.values())-1  # noqa: F821
```
kanderso-nrel, line 287:
```suggestion
                lr_nrows[lr] = num - lr_startrow[lr]
```
maybe clearer?  might need a -1 or something too

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+    """
+    Parse a file-like buffer of a BSRN station-to-archive file.
 
     Parameters
     ----------
-    filename: str
-        A relative or absolute file path.
+    fbuf: file-like buffer
+        Buffer of a BSRN station-to-archive data file
+    logical_records: str or list, default: ['0100']
```
kanderso-nrel, line 220: Maybe I'm doing it wrong but I don't think `str` is actually allowed here (or the other functions)

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+            if len(lr_startrow) >= 1:
+                lr_nrows[lr] = num - max(lr_startrow.values())-1  # noqa: F821
+            lr = line[2:6]  # string of 4 digit LR number
+            lr_startrow[lr] = num
+    lr_nrows[lr] = num - lr_startrow[lr]
+
+    for lr in list(logical_records):
+        if lr not in ['0100', '0300', '0500']:
+            raise ValueError(f"Logical record {lr} not in "
+                             f"{['0100', '0300','0500']}.")
```
kanderso-nrel, line 295:
```suggestion
                             "['0100', '0300', '0500'].")
```

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+        List of the logical records (LR) to parse. Options are: '0100', '0300',
+        and '0500'.
 
     Returns
     -------
     data: DataFrame
         A DataFrame with the columns as described below. For more extensive
-        description of the variables, consult [2]_.
+        description of the variables, consult [2]_. An empty DataFrame is
+        returned if the specificed logical records were not found.
```
kanderso-nrel, line 384:
```suggestion
        returned if the specified logical records were not found.

```

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+    dhi†                     float   Mean diffuse horizontal irradiance [W/m^2]
+    lwd†                     float   Mean. downward long-wave radiation [W/m^2]
     temp_air                 float   Air temperature [°C]
     relative_humidity        float   Relative humidity [%]
     pressure                 float   Atmospheric pressure [hPa]
+    -----------------------  ------  ------------------------------------------
+    **Logical record 0300**
+    ---------------------------------------------------------------------------
+    gri†                     float   Mean ground-reflected irradiance [W/m^2]
+    lwu†                     float   Mean lowng-wave upwelling irradiance [W/m^2]
```
kanderso-nrel, line 408:
```suggestion
    lwu†                     float   Mean long-wave upwelling irradiance [W/m^2]
```

pvlib/iotools/bsrn.py
```diff
@@ -4,139 +4,449 @@
...
+    uvb_reflected†           float   Mean UV-B reflected irradiance [W/m^2]
     =======================  ======  ==========================================
 
+    † Marked variables have corresponding columns for the standard deviation
+    (_std), minimum (_min), and maximum (_max) based on the 60 samples made
+    for each minute.
+
+    Hint
+    ----
+    According to [2]_ ""All time labels in the station-to-archive files denote
```
kanderso-nrel, line 426:
```suggestion
    According to [2]_ "All time labels in the station-to-archive files denote
```

pvlib/tests/iotools/test_bsrn.py
```diff
@@ -2,25 +2,123 @@
...
+                        requires_bsrn_credentials)
+
+
+@pytest.fixture(scope="module")
+def bsrn_credentials():
+    """Supplies the BSRN FTP credentials for testing purposes.
 
-from pvlib.iotools import bsrn
-from ..conftest import DATA_DIR, assert_index_equal
+    Users should obtain there own credentials as described in the `read_bsrn`
```
kanderso-nrel, line 17:
```suggestion
    Users should obtain their own credentials as described in the `read_bsrn`
```

pvlib/tests/iotools/test_bsrn.py
```diff
@@ -2,25 +2,123 @@
...
+
+def test_read_bsrn_bad_logical_record():
+    # Test if ValueError is raised if an unsupported logical record is passed
+    with pytest.raises(ValueError, match='not in'):
+        read_bsrn(DATA_DIR / 'bsrn-lr0100-pay0616.dat',
+                  logical_records=['dummy'])
+
+
+def test_read_bsrn_logical_records_not_found():
+    # Test if an empty dataframe is returned if specified LRs are not present
```
kanderso-nrel, line 64: seems like this comment doesn't match the actual test

kanderso-nrel (commented):
This looks great @AdamRJensen!  Comments below, mostly very nitpicky...

pvlib/iotools/bsrn.py
AdamRJensen: Well, the second strip() is redundant. But strip() removes starting and trailing whitespaces, as well as "\r\n" and "\n"

pvlib/iotools/bsrn.py
AdamRJensen: If the seek is remove here, then we need to manually add the line number offset (where the metadata section ends) to `enumerate` using *start* as otherwise, the line numbering will be incorrect in the dictionary.

As the metadata section and the earlier sections have a variable number of lines this will probably be rather complex to keep track of.

pvlib/iotools/bsrn.py
AdamRJensen: what about just:
`pd.date_range(start.replace(day=1), end.replace(day=1) + pd.DateOffset(months=1), freq='1M')`

I believe you might not even need the `replace()` on start.

pvlib/iotools/bsrn.py
```diff
@@ -4,56 +4,367 @@
...
+                       'empty', 'empty', 'empty', 'empty',
+                       'uvb_global_mean', 'uvb_global_std', 'uvb_global_min',
+                       'uvb_global_max', 'uvb_diffuse_mean', 'uvb_diffuse_std',
+                       'uvb_diffuse_mean', 'uvb_diffuse_std',
+                       'uvb_diffuse_min', 'uvb_diffuse_max',
+                       'uvb_reflect_mean', 'uvb_reflect_std',
+                       'uvb_reflect_min', 'uvb_reflect_max']
+
+
+def get_bsrn(start, end, station, username, password, logical_records=['0100'],
```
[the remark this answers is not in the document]
AdamRJensen: I actually have it right now so it can either take a list or string. What do you think of that?
`    logical_records: str or list, default: ['0100']`

Therefore, the default could just be the string version: '0100'.

Otherwise, should I also change the input type in the doc string to tuple?

pvlib/iotools/bsrn.py
AdamRJensen: In the parse function I have this line, which allows it to be string:
`    for lr in list(logical_records):`

This does however also have to be added to the empty dataframe function.

First though, is this a desireable functionality or should I remove this?

kanderso-nrel: Note that `list('0100') == ['0', '1', '0', '0']`.  Does it work for you if you pass `logical_records='0100'`?  I get `ValueError: Logical record 0 not in ['0100', '0300', '0500'].`  I'd say just require it to be a list (or tuple etc).

pvlib/iotools/bsrn.py
AdamRJensen: Sorry for all the small typos...

pvlib/iotools/bsrn.py
kanderso-nrel: Works for me.  I wasn't sure about the `replace()` on start, I included it to be safe but I think you're right that it's unnecessary.

pvlib/tests/iotools/test_bsrn.py
AdamRJensen: What makes you think that?

In line 67 of the same test, the index is checked if it is empty:
`    assert_index_equal(pd.DataFrame().index, data.index)`

Perhaps that I check for certain columns to be present mislead you. That is only to check that the empty dataframe has the correct columns.

kanderso-nrel: > What makes you think that?

Careless reading.  I retract my objection :)

`assert data.empty` might be a little easier to parse than `assert_index_equal(pd.DataFrame().index, data.index)`

pvlib/iotools/bsrn.py
AdamRJensen: You are indeed correct! List or tuples it is :)

pvlib/iotools/bsrn.py
AdamRJensen: Nice tough. Certainly makes it much clearer. The -1 is needed fyi.

pvlib/iotools/bsrn.py
AdamRJensen: There certainly are pros and cons. Either way it would be nice if it was specified, perhaps one way of doing this is simply to have two entries for latitude and longitude.

E.g., `metadata['latitude']` would be the pvlib convention and `metadata['latitude_bsrn_convention']` would be the BSRN method.

pvlib/tests/iotools/test_bsrn.py
AdamRJensen: Hahah. I have changed it to `data.empty`, which probably would have avoided the confusion.

pvlib/iotools/bsrn.py
AdamRJensen: I took a lot of fiddling around to find this solution, as I was stumbling into errors for all the other solutions I investigated.

When I run `gzip.GzipFile(bio)`, I get the following error:
`TypeError: expected str, bytes or os.PathLike object, not _io.BytesIO`

kanderso-nrel: Right, it would use the `fileobj` parameter of `GzipFile`, I think this might work:

```python
bio.seek(0)
gzip_file = io.TextIOWrapper(gzip.GzipFile(fileobj=bio), encoding='utf-8')  # or latin1?
dfi, metadata = parse_bsrn(gzip_file, logical_records)
```

Again, not sure this is worth worrying about though.

AdamRJensen: My solution takes 83 ms, whereas yours take 20.6 µs. So certainly a bit of a performance gain :)

I'll change it to your solution

AdamRJensen:
Thanks for the review @kanderso-nrel! Do we need a second review on this?

kanderso-nrel:
> Do we need a second review on this?

My answer to that question is always yes :) https://en.wikipedia.org/wiki/Linus%27s_law

kanderso-nrel (approved):
LGTM, @pvlib/pvlib-maintainer anyone else want to review this?

cwhanse:
> LGTM, @pvlib/pvlib-maintainer anyone else want to review this?

I don't see how I will make time for it. I vote merge and improve later, if needed.

wholmgren (approved):
Looks good to me. I confirmed the parse function works with data I obtained from the NASA Langley site using requests.

The lines that are not covered are either expected due to the test configuration or are ok to skip for iotools.

Thanks @AdamRJensen! I expect this will be useful to many people.

Issue #1214, opened 2021-05-05
pvlib v0.9 release planning


cwhanse:
We've got a lot of new capability pending, and should move to formalize a release. @wholmgren @mikofski @kevinsa5 @wfvining  I am willing to manage v0.9 release. Here's the checklists. Please indicate what is needed to close a PR, or if you are willing to close any of the identified issues, or recommend deferring.

PRs to merge or defer: 

- [x] #1213 merged
- [x] #1211 
- [x] #1196 
- [x] #1191 
- [ ] #717 @mikofski I'm guessing the infinite sheds PR will be deferred?

Issues to close or defer:

- [x] #1188  will close with #1218 
- [x] #1185
- [x] #1184 PR #1231
- [x] #1148  PR #1224
- [x] #1126  PR #1233
- [x] #1118 PR #1222
- [x] #1115 PR #1223 
- [ ] #1113 
- [x] #1109 
- [ ] #1069 
- [ ] #964 suggest deferring
- [ ] #958 suggest deferring
- [x] #954 deferred 
- [x] #949 PR #1225 
- [ ] #825 recommend deferring
- [ ] #483 recommend deferring
- [ ] #436 maybe this can be closed, we've accomplished a substantial part of the scope
- [ ] #418 recommend deferring
- [ ] #329 perennial can to be kicked down the development road

wholmgren:
I'd say #1176 is the key remaining PR, which depends on several of the other PRs. @kanderso-nrel is behind most of those and appeared to be slowed down because he didn't get timely feedback from the rest of us (sorry Kevin!). I'm fine deferring anything that's not related to the object layer refactor.

kanderso-nrel:
> I'd say #1176 is the key remaining PR, which depends on several of the other PRs. @kanderso-nrel is behind most of those and appeared to be slowed down because he didn't get timely feedback from the rest of us (sorry Kevin!). I'm fine deferring anything that's not related to the object layer refactor.

Slowed down by my having to prioritize non-pvlib work lately as well.  #1211 is waiting for me to pick it up again, but #1196 is ready for others to take a closer look.  I need to remind myself of the details but I think #1176 will be close to ready once the other two are merged, that is unless I continue expanding its scope...

[...]

PR #1264, opened 2021-07-27
Add retrieval function for ERA5 reanalysis data

AdamRJensen:
 - [x] Tests added

**Description**
ERA5 is the reanalysis dataset provided by the ECMWF and provides hourly irradiance data for the entire world starting in 1970.

The proposed function retrieves ERA5 data from the Climate Data Store (CDS) using the `cdsapi`.


AdamRJensen:
In order to test ERA5, `CDSAPI_KEY` needs to be saved as an environment variable.

The API key can be obtained by [creating a CDS account](https://cds.climate.copernicus.eu/user/80797/edit). Note the tests do not need a local file with the key info as described [here](https://cds.climate.copernicus.eu/api-how-to).

@wholmgren Sorry to bother, but I think maybe you're the only one who has access to the pvlib email and email?

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+        Last day of the requested period
+    api_key: str, optional
+        Personal API key for the CDS
+    variables: list, default: ERA5_DEFAULT_VARIABLES
+        List of variables to retrieve (according to CDS naming convention)
+    dataset: str, default 'reanalysis-era5-single-levels'
+        Name of the dataset to retrieve the variables from. Can be either
+        'reanalysis-era5-single-levels' or 'reanalysis-era5-land'.
+    product_type: str, {'reanalysis', 'ensemble_members', 'ensemble_mean',
+                        'ensemble_spread'}, default: 'reanalysis'
```
kanderso-nrel, line 142: Sphinx's rendering here is not ideal (the second line becomes part of the text description rather than the list of allowed values) but I'm not sure off the top of my head how to fix it.  Maybe a backslash at the end of the first line?  Or maybe just make it one long line and use `# noqa: E501`?

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+    Notes
+    -----
+    The returned data includes the following fields by default:
+
+    ========================  ======  =========================================
+    Key, mapped key           Format  Description
+    ========================  ======  =========================================
+    *Mapped field names are returned when the map_variables argument is True*
+    ---------------------------------------------------------------------------
+    2tm, temp_air             float   Air temperature at 2 m above ground (K)
```
kanderso-nrel, line 167: Should we convert to C?  I forget if other iotools functions do that

kanderso-nrel: Also nitpick, some units in this table are in parentheses and some are in brackets

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+        'grid': grid,
+        'area': area}
+
+    # Retrieve path to the file
+    file_location = cds_client.retrieve(dataset, params)
+
+    # Load file into memory
+    with requests.get(file_location.location) as res:
+
+        # Save the file locally if local_path has been specified
```
kanderso-nrel, line 230:
```suggestion
        # Save the file locally if save_path has been specified
```

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+        'reanalysis-era5-single-levels' or 'reanalysis-era5-land'.
+    product_type: str, {'reanalysis', 'ensemble_members', 'ensemble_mean',
+                        'ensemble_spread'}, default: 'reanalysis'
+        ERA5 product type
+    grid: list or tuple, default: (0.25, 0.25)
+        User specified grid resolution
+    save_path: str or path-like, optional
+        Filename of where to save data. Should have ".nc" extension.
+    cds_client: CDS API client object, optional
+        CDS API client
```
kanderso-nrel, line 149: What's the use case for having this be a parameter?

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+        # Renaming of xarray datasets throws an error if keys are missing
+        ds = ds.rename_vars(
+            {k: v for k, v in ERA5_VARIABLE_MAP.items() if k in list(ds)})
+
+    if (output_format == 'dataframe') or (
+            (output_format is None) & (ds['latitude'].size == 1) &
+            (ds['longitude'].size == 1)):
+        if (ds['latitude'].size == 1) & (ds['longitude'].size == 1):
+            data = ds.to_dataframe().droplevel(['latitude', 'longitude'])
+        return data, metadata
```
kanderso-nrel, line 292: Not sure this is the logic we want -- `data` isn't necessarily defined here.  What do you think about raising an error if dataframe is requested but the sizes aren't equal to 1?

pvlib/tests/iotools/test_era5.py
```diff
@@ -0,0 +1,120 @@
...
+from pvlib.iotools import read_era5, get_era5
+from ..conftest import (DATA_DIR, RERUNS, RERUNS_DELAY, assert_index_equal,
+                        requires_cds_credentials, requires_xarray)
+
+
+@pytest.fixture(scope="module")
+def cds_api_key():
+    """Supplies pvlib-python's CDS API key.
+
+    Users should obtain their own credentials as described in the `get_bsrn`
```
kanderso-nrel, line 18: this has a copy/paste reference to BSRN

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+    import cdsapi
+except ImportError:
+    class cdsapi:
+        @staticmethod
+        def Client(*a, **kw):
+            raise ImportError(
+                'Retrieving ERA5 data requires cdsapi to be installed.')
+
+
+def extract_metadata_from_dataset(ds):
```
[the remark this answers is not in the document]
AdamRJensen: No, I don't see any reason for it not to be private.

pvlib/iotools/era5.py
AdamRJensen: I was initially against converting to C as it would get confusing if users imports other temperatures (e.g., dew point temperature) which then still will be in degrees C.

pvlib/iotools/era5.py
AdamRJensen: Probably none... I'd seen other wrappers do it. Maybe if you wanted to create your own API client and submit multiple jobs? I'm ok removing it.

pvlib/iotools/era5.py
AdamRJensen: It should only have been the droplevel that was in the if statement and not the to_dataframe(). It looks like this now:

    if (output_format == 'dataframe') or (
            (output_format is None) & (ds['latitude'].size == 1) &
            (ds['longitude'].size == 1)):
        data = ds.to_dataframe()
        if (ds['latitude'].size == 1) & (ds['longitude'].size == 1):
            data.droplevel(['latitude', 'longitude'])
        return data, metadata

You can still get a DataFrame with multiple locations (and multiindex) but then you shouldn't remove the latitude/longitude.

pvlib/iotools/era5.py
AdamRJensen: The entire docstring is already covered by the `# noqa: E501` at the end of the docstring, so making it one line is an easy solution.

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+    longitude: float or list
+        in decimal degrees, between -180 and 180, east is positive (ISO 19115).
+        If longitude is a list, it should have the format [W, E] and
+        longitudes within the range are selected according to the grid.
+    start: datetime like
+        First day of the requested period
+    end: datetime like
+        Last day of the requested period
+    api_key: str, optional
+        Personal API key for the CDS
```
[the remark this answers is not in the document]
AdamRJensen: Good point. Would this be a sufficient description?

    api_key: str, optional
        Personal API key for the CDS with the format "uid:key"

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def extract_metadata_from_dataset(ds):
+    metadata = {}
+    for v in list(ds.variables):
+        metadata[v] = {
+            'name': ds[v].name,
+            'long_name': ds[v].long_name}
+        if v.lower() != 'time':
+            metadata[v].update({'units': ds[v].units})
```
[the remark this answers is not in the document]
AdamRJensen: I guess it does exactly the same, but your suggestion is just cleaner?

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+"""
+# The functions only support single-level 2D data and not 3D / pressure-level
+# data. Also, monthly datasets and grib files are no supported.
+
+import requests
+
+try:
+    import xarray as xr
+except ImportError:
+    class xr:
```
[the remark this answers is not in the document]
AdamRJensen: I'm assuming this is just for reference and I can resolve this?

pvlib/iotools/era5.py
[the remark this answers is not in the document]
kanderso-nrel: I'm okay with that.  Maybe a dummy example would help, e.g. `'00000:aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee'`?  Up to you

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,294 @@
...
+    .. [5] `Climate Data Store (CDS) API Keywords
+       <https://confluence.ecmwf.int/display/CKB/Climate+Data+Store+%28CDS%29+API+Keywords>`_
+    .. [6] `Climate Data Storage user registration
+       <https://cds.climate.copernicus.eu/user/register>`_
+    """  # noqa: E501
+    if cds_client is None:
+        cds_client = cdsapi.Client(url=CDSAPI_URL, key=api_key)
+
+    # Area is selected by a box made by the four coordinates: [N, W, S, E]
+    if type(latitude) == list:
```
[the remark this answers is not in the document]
AdamRJensen: Looks like a nice solution to me. Consider it implemented!

pvlib/iotools/era5.py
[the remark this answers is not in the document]
AdamRJensen: I like it.

pvlib/iotools/era5.py
AdamRJensen: I've added a function that converts all variables with units of K to units C. The function is placed in `pvlib.tools` as it will also be used in #1274.

kanderso-nrel (commented):
Looking pretty good, a few minor notes below.  It's a shame that the CDS API is so slow to return data, but I guess it is what it is.  Ready for @pvlib/pvlib-maintainer to take a look if anyone has the time.

setup.py
```diff
@@ -54,7 +54,7 @@
 EXTRAS_REQUIRE = {
     'optional': ['cython', 'ephem', 'netcdf4', 'nrel-pysam', 'numba',
                  'pvfactors', 'siphon', 'statsmodels', 'tables',
-                 'cftime >= 1.1.1'],
+                 'cftime >= 1.1.1', 'xarray', 'dask', 'cdsapi'],
```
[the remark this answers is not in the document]
AdamRJensen: > Tangent: this list is getting pretty long and covers a lot of the package. Might be worth splitting it up by area like some other packages do. Not in this PR though

So for example have an `EXTRAS_IOTOOLS`?

Also, I've added "xarray, dask, and cdsapi are now optional requirements. (:pull:`1264`)" to the whatsnew file

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,276 @@
...
+       <https://github.com/ecmwf/cdsapi>`_
+    .. [5] `Climate Data Store (CDS) API Keywords
+       <https://confluence.ecmwf.int/display/CKB/Climate+Data+Store+%28CDS%29+API+Keywords>`_
+    .. [6] `Climate Data Storage user registration
+       <https://cds.climate.copernicus.eu/user/register>`_
+    """  # noqa: E501
+    if cdsapi is None:
+        raise ImportError('Retrieving ERA5 data requires cdsapi to be installed.')  # noqa: E501
+
+    cds_client = cdsapi.Client(url=CDSAPI_URL, key=api_key, verify=1)
```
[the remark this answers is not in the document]
AdamRJensen: You don't think this output is useful for the user?

I've set `quiet=True`.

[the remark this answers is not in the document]
kanderso-nrel: from https://pvlib-python.readthedocs.io/en/latest/contributing.html#code-style:

> Remove any logging calls and print statements that you added during development.

I guess it's not 100% relevant since you didn't add the logging yourself, but I think the point is that logging should be disabled by default

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,277 @@
...
+    params = {
+        'product_type': product_type,
+        'variable': variables,
+        'date': start.strftime('%Y-%m-%d') + '/' + end.strftime('%Y-%m-%d'),
+        'time': ERA5_HOURS,
+        'grid': grid,
+        'area': area,
+        'format': 'netcdf'}
+
+    # Retrieve remote path to the file
```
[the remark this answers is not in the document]
adriesse: Yes, it is better to avoid having the user download the same thing more than once.

There is a potential danger in this function: if an exception is raised while saving or reading/conversion, the in-memory data will be lost.

kanderso-nrel:
@wholmgren said in https://github.com/pvlib/pvlib-python/issues/1214#issuecomment-902819052 that we should think through how xarray and dask are used in pvlib, and now that 0.9.0 is out I guess we can do that.  This PR as it currently stands adds three new optional dependencies: `xarray`, `dask`, and `cdsapi`.  #1274 currently needs `pydap` as well.  I assume that what needs to be thought through is whether we are willing to add these new dependencies; if there are other questions to answer please point that out.  Here are my thoughts:

In the context of this PR, as I understand it (@AdamRJensen please correct any errors), I think `dask` is only needed for reading multiple files in a single `read_era5` call, so we might be able to drop `dask` without losing much functionality.  And if we do restrict to single files, I suspect we could drop `xarray` as well and replace it with a lower level netcdf library (can `h5py` read netcdf?) combined with something like `pvlib.forecast.ForecastModel._netcdf2pandas`.  

#1274 is less clear.  MERRA2 data retrieval is supposed to be done using OpenDAP instead of basic HTTP requests like most APIs.  My very limited understanding of OpenDAP suggests that it is advisable to use a dedicated library for it instead of just using `requests` like we normally do.  `xarray` combined with `pydap` is one way to access the data, but I am pretty reluctant to use `pydap`: the project seems abandoned and we could only get things working with an unreleased development version.  I think there are alternatives worth looking into but I don't know yet whether `xarray` will need to be involved.

AdamRJensen:
Closing this PR in favor #2573, as 2573 doesn't rely on the cdsapi package and uses the new time series endpoint.

PR #1283, opened 2021-08-16
Clean up how pvlib imports optional dependencies

kanderso-nrel:
- ~[ ] Closes #xxxx~
 - [ ] Tests added

I'd have made an issue for this but I thought a PR would make it easier to see the idea in action.

pvlib has multiple ways of handling optional imports.  One is to hide the imports inside the function that needs them, e.g. how `solarposition.py` handles the `ephem` dependency:
https://github.com/pvlib/pvlib-python/blob/f318c1c1527f69d9bf9aed6167ca1f6ce9e9d764/pvlib/solarposition.py#L633-L636

Another way used in `ecmwf_macc.py` and #1264, is to keep the import at the top of the file for clarity, but defer an import error until the missing module is actually called on to do something, like this:
https://github.com/pvlib/pvlib-python/blob/f318c1c1527f69d9bf9aed6167ca1f6ce9e9d764/pvlib/iotools/ecmwf_macc.py#L8-L15

So only when some bit of code actually tries to create a `netcdf4.Dataset` will the `ImportError` be raised.

Summarizing some of @AdamRJensen's thoughts here: the first is simple but adds a bit of clutter to the functions (`solarposition.py` imports `ephem` four separate times) and doesn't make it easy to see what external packages each pvlib module uses.  The second keeps imports at the top of the file but is a little opaque and gets clunky if more than one function is needed from a single import.

This PR extends the second approach, but with improvements:
- use `__getattr__` so there's no need to explicitly list out the module functions of interest.  For example the ERA5 PR wouldn't have to handle `open_dataset` and `open_mfdataset` separately like [this](https://github.com/AdamRJensen/pvlib-python/blob/792166697e9fa790bda91ea6f1066b0906a7bdec/pvlib/iotools/era5.py#L9-L21)
- less boilerplate in the modules because the machinery is defined in `pvlib.tools`, so each module doesn't need to define a new class for each import

What do people think?  Too "clever"?  Solving a problem that doesn't need solved?  There might be other places to use this too, but I only looked at `ecwmf_macc.py` and `solarposition.py` for now to illustrate its usage.

### DIFF
--- a/pvlib/iotools/ecmwf_macc.py
+++ b/pvlib/iotools/ecmwf_macc.py
@@ -4,24 +4,14 @@ Read data from ECMWF MACC Reanalysis.
 
 import threading
 import pandas as pd
+from pvlib.tools import _optional_import
 
-try:
-    import netCDF4
-except ImportError:
-    class netCDF4:
-        @staticmethod
-        def Dataset(*a, **kw):
-            raise ImportError(
-                'Reading ECMWF data requires netCDF4 to be installed.')
+netCDF4 = _optional_import('netCDF4', (
+    'Reading ECMWF data requires netCDF4 to be installed.'))
+ecmwfapi = _optional_import('ecmwfapi', (
+    'To download data from ECMWF requires the API client.\nSee https:/'
+    '/confluence.ecmwf.int/display/WEBAPI/Access+ECMWF+Public+Datasets'))
 
-try:
-    from ecmwfapi import ECMWFDataServer
-except ImportError:
-    def ECMWFDataServer(*a, **kw):
-        raise ImportError(
-            'To download data from ECMWF requires the API client.\nSee https:/'
-            '/confluence.ecmwf.int/display/WEBAPI/Access+ECMWF+Public+Datasets'
-        )
 
 #: map of ECMWF MACC parameter keynames and codes used in API
 PARAMS = {
@@ -164,7 +154,7 @@ def get_ecmwf_macc(filename, params, start, end, lookup_params=True,
     startdate = start.strftime('%Y-%m-%d')
     enddate = end.strftime('%Y-%m-%d')
     if not server:
-        server = ECMWFDataServer()
+        server = ecmwfapi.ECMWFDataServer()
     t = threading.Thread(target=target, daemon=True,
                          args=(server, startdate, enddate, params, filename))
     t.start()
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -24,7 +24,8 @@ import scipy.optimize as so
 import warnings
 
 from pvlib import atmosphere
-from pvlib.tools import datetime_to_djd, djd_to_datetime
+from pvlib.tools import datetime_to_djd, djd_to_datetime, _optional_import
+ephem = _optional_import('ephem', 'PyEphem must be installed')
 
 
 NS_PER_HR = 1.e9 * 3600.  # nanoseconds per hour
@@ -487,7 +488,6 @@ def _ephem_to_timezone(date, tzinfo):
 
 def _ephem_setup(latitude, longitude, altitude, pressure, temperature,
                  horizon):
-    import ephem
     # initialize a PyEphem observer
     obs = ephem.Observer()
     obs.lat = str(latitude)
@@ -543,11 +543,6 @@ def sun_rise_set_transit_ephem(times, latitude, longitude,
     --------
     pyephem
     """
-
-    try:
-        import ephem
-    except ImportError:
-        raise ImportError('PyEphem must be installed')
 
     # times must be localized
     if times.tz:
@@ -630,10 +625,6 @@ def pyephem(time, latitude, longitude, altitude=0, pressure=101325,
     """
 
     # Written by Will Holmgren (@wholmgren), University of Arizona, 2014
-    try:
-        import ephem
-    except ImportError:
-        raise ImportError('PyEphem must be installed')
 
     # if localized, convert to UTC. otherwise, assume UTC.
     try:
@@ -942,8 +933,6 @@ def pyephem_earthsun_distance(time):
     -------
     pd.Series. Earth-sun distance in AU.
     """
-
-    import ephem
 
     sun = ephem.Sun()
     earthsun = []
--- a/pvlib/tools.py
+++ b/pvlib/tools.py
@@ -6,6 +6,7 @@ import datetime as dt
 import numpy as np
 import pandas as pd
 import pytz
+import importlib
 
 
 def cosd(angle):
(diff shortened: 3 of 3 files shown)


wholmgren:
> Too "clever"? Solving a problem that doesn't need solved?

Sorry to be a downer, but I'd say yes and yes. I also understand that others may reasonably disagree.

> and doesn't make it easy to see what external packages each pvlib module uses.

This could be solved with comments near the external imports at the top of the file e.g. `# some functions import the optional dependency ephem`

AdamRJensen:
I agreed that it is quite clever/complex, but I like the fact that the 'cleverness' is contained in one function that most developers don't necessarily need to understand. The current solution requires each function to have some slightly less clever method (e.g, as with `get_ecmwf_macc`) which is less useful as it only covers the specific module used in the library (and not all instances).

As a newbie contributor to pvlib, I would have preferred to add `from pvlib.tools import _optional_import` and use `_optional_import` than the 13 lines I ended up writing in #1264 instead.

What I also quite like about the solution is that all module imports are kept at the top (which is where I go to look for dependencies) instead of numerous times throughout the function as in the case of `solarposition.py`.

Just a few thoughts from this side of the pond :)

mikofski:
Even tho I once thought it better for all imports at top, I agree with Will that the mocking of functions in `try: except` at module top is too clever and IMHO may have some downsides:
* I haven't seen a standard, or a pattern in other libraries. This means we'd be breaking new ground
* in the past I've had some bugs b/c of these, specifically in `singldiode.py`
* mocking imported modules may delay execution,  but IMO it's better to fail ASAP when a function with optional import dependency is called

An alternate solution which might be slightly less clever would be a decorator & still commenting at module top:

```python
from pvlib.tools import import_optional_dependency
# has optional dependency import

@import_optional_dependency("foobar")
def myfun():
    return


# in tools, untested psuedocode
import imp  # or importlib whatever
def import_optional_dependency(imports):
    def wrapper(f):
        def wrapt_fcn(*args, **kwargs):
            imp.import(imports)
                return f(*args, **kwargs)
        return wrapt_fcn
# end wrapper?
```

cwhanse:
I'm going to inject a related question: can we move `tables` off the optional list?  Its required to run the default clearsky model, and, it's a pain to diagnose for users when their python environment can't find it.

wholmgren:
Some research on what other projects do:

`pandas` has a private module [`_optional.py`](https://github.com/pandas-dev/pandas/blob/059c8bac51e47d6eaaa3e36d6a293a22312925e6/pandas/compat/_optional.py) that we could copy (with license attribution). Check out the git blame for all of the thought that has gone into it - that's part of why I want to keep it simple or at least avoid reinventing wheels. Note that pandas is fine with putting imports within functions e.g. [`lxml = import_optional_dependency("lxml.etree", errors="ignore")`](https://github.com/pandas-dev/pandas/blob/059c8bac51e47d6eaaa3e36d6a293a22312925e6/pandas/io/xml.py#L691). They also track installed state for some optional dependencies: [`from pandas.core.computation.check import NUMEXPR_INSTALLED`](https://github.com/pandas-dev/pandas/blob/8e844a2c4504de856b609c92523d384ec672ef0b/pandas/core/computation/eval.py#L45)

`siphon` puts `try/except ImportError` statements [within methods](https://github.com/Unidata/siphon/blob/baa8cacd024b244a34fb3a4cfd1fe144fc0c05b8/siphon/catalog.py#L721-L725).

So does [`dask`](https://github.com/dask/dask/blob/95fb60a31a87c6b94b01ed75ab6533fa04d51f19/dask/cache.py#L32-L37)

I think [this approach](https://stackoverflow.com/a/52826085/2802993) is also reasonable, while [this one](https://stackoverflow.com/a/27361558/2802993) is overkill.

> I agreed that it is quite clever/complex, but I like the fact that the 'cleverness' is contained in one function that most developers don't necessarily need to understand (like me).

Developer considerations are important, but multiple maintainers do need to understand it and I just don't have it in me at the moment.

kanderso-nrel:
Fine with me to KISS and just use `try/except ImportError` inside the function like that siphon example and our ephem code.

> What I also quite like about the solution is that all module imports are kept at the top (which is where I go to look for dependencies) 

I was going to say that this is not a compelling point to me personally, but maybe that's because I already have a decent sense for what packages are used in each pvlib module.  Or maybe I would just go to `setup.py` to see dependencies instead of reading module headers.  Anyway would this be addressed by including a comment next to the required imports like @wholmgren suggested?

AdamRJensen:
Sounds good to me. I was never super attached to the new strategy, but figured it was worth discussing.

AdamRJensen:
So for functions using a non-required package (e.g., xarray in `get_era5` #1264) would it look something like this immediately after the function docstring:

    try:
        import xarray as xr
    except ImportError:
        raise ImportError('Reading ERA5 data requires xarray to be installed.')

Also, is there a reason not to use the subclass `ModuleNotFoundError` or are we interested in catching all ImportErrors (the message "requires xarary to be installed" implies that it wasn't found and not that it couldn't be loaded)?

kanderso-nrel:
It's not clear to me when `ModuleNotFoundError` is more appropriate than `ImportError`, and googling didn't really help.  I looked at what some other packages use in this situation and found a mix:
- astropy uses [ImportError](https://github.com/astropy/astropy/blob/2aeda166006ca9f4be66b3f51c571928a3fc59a7/astropy/io/misc/pandas/connect.py#L61), [ModuleNotFoundError](https://github.com/astropy/astropy/blob/e183e398d8bdbbe5c7ccec7a793afed2ecdac4e1/astropy/io/fits/hdu/hdulist.py#L1322), plain [Exception](https://github.com/astropy/astropy/blob/4654cde65f233f14b2d809fcd1ccf6d97e4638ec/astropy/io/misc/hdf5.py#L86), and a [custom error](https://github.com/astropy/astropy/blob/734f8df777d8655cc4bf29ec742ae07c75fdb619/astropy/io/ascii/html.py#L86) they defined.  `ImportError` seems the most common
- siphon uses [ImportError](https://github.com/Unidata/siphon/blob/baa8cacd024b244a34fb3a4cfd1fe144fc0c05b8/siphon/catalog.py#L725) but I didn't find any `ModuleNotFoundError`
- xarray uses [ImportError](https://github.com/pydata/xarray/blob/41099669c63c7dbb51b112fdec09dfae07379fe7/xarray/plot/utils.py#L422) and [ValueError](https://github.com/pydata/xarray/blob/5ccb06951cecd59b890c1457e36ee3c2030a67aa/xarray/backends/api.py#L69) for some reason

In the absence of any better information I'd say raise `ImportError` since that's what the rest of pvlib already does.

mikofski:
from [zen of python](https://www.python.org/dev/peps/pep-0020/#id2):
>Simple is better than complex.

and
>In the face of ambiguity, refuse the temptation to guess.
>There should be one-- and preferably only one --obvious way to do it.

which I interpret to mean, try to observe the most-accepted and widely used methods because:
>Readability counts.
>Special cases aren't special enough to break the rules.

kanderso-nrel:
Thanks all for the discussion, I think we can close this now.

PR #1315, opened 2021-09-27
Added rounding of aa and b in _schumaker_qspline.py, see Issue #1311

### LINKED ISSUE #1311: Spline fitting failure because of rounding
Hello,

I'm doing an internship about SDM and DDM parameter extraction from experimental IV curves and I've begun to use pvlib (by the way, thank you for the great documentation, it was really helpful to understand the library !).
After some tinkering with my data, I got errors in numpy.linalg but I found out it boils down to the _schumaker_qspline function. #1235 
I don't know if this is really a bug, or if it is already known (the only issue that mentioned the qspline #1228 maybe briefly talks about it)
After some analysis, here is what I understand : 

**Describe the bug**

It all starts in the first else of the 5th step of algo 4.1 of the Schumaker publication, (lines 313-314 of pvlib.ivtools.utils.py) : 

a = s_i - \delta_i
b = s_{i+1} - \delta_i

the former difference can at some points lead to numerical residues : i.e. b_i = -1e-18 instead of 0.
It can lead to a_i * b_i being negative but really small and thus pass the next if of the algorithm.

But the computation of xi afterwards might lead to rounding of this really small value and in the end we can get xi_i = t_i

This will induce a division by zero in the definition of eta_i (line 354 of utils.py). No problem so far but this np.inf will become np.nan after other calculations. These np.nan will thwart the execution of the coming mean() and cumsum() functions. In the end, it is the least_squares fitting function that throws an error because in cannot converge when one of its arguments is an array of np.nan

**To Reproduce**
1. Load the following zip file : [data_to_reproduce.zip](https://github.com/pvlib/pvlib-python/files/7217832/data_to_reproduce.zip)
2. Extract the files
3. Run reproduce_bug.py in the same folder as the three .txt and the two .npz
4. See error : numpy.linalg.LinAlgError: SVD did not converge in Linear Least Squares*

So far the only indices where the division by zero occur are within the w indices. In the data provided, I sometimes found other indices where b was really small and a * b negative without leading to a division by zero because rounding did not occur (e.g. xi was 32.0000000000004 and uu 32). I've only checked for the first five curves but it is always the same : 

there are some indices where xi == uu. At those indices, w is True and v and z are False. The b values at those indices are really small (max 1e-14) and aa * b at those indices is negative (were it really should be equal to 0)

**Expected behavior**
IMHO, a test should be added after the computation of aa and b, and if they are close to zero, they should be rounded to ensure proper behaviour of the upcoming tests.
Let me know if you can reproduce this and what you think of it. I think the actual treatment of NaNs is good, replacing mean by nanmean will not do good in the long run.
I realized that smoothing the data could significantly reduce the number of errors but not all of them, so that is why I decided to open this issue.

**Versions:**
 - ``pvlib.__version__``: 0.9.0
 - ``numpy.__version__``: 1.20.1
 - python: 3.8

Antoine-0:
- [x] Closes #1311 
 - [x] I have read the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 I'm not sure how to do the following items : 
 - [x] Tests added
 - ~~[ ] Updates entries to [`docs/sphinx/source/api.rst`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes.~~


cwhanse:
I'll take a look at the failing test. I can add the text for the whatsnew document, and take care of assigning labels. Don't worry about the API item.

cwhanse:
I removed the data that was causing the test failure. Those data were made up to check consistency between Matlab and python implementations of `pvlib.ivtools.sde._fit_sandia_cocontent`. The test still has 2 of the original 3 made-up IV curves, which pass the test criteria. So I am not concerned about dropping part of the test data, the test still meets its intent. The removed data that are removed would still produce values for the 5 IV curve parameters, but they would not be close to what is produced by the Matlab code (e.g. 16% relative difference in one parameter). 

I also split the `EPS` parameter into `EPS_slope` to determine when a slope is 0, and `EPS_val` to determine when a number is zero. The `EPS_slope` value is around 10^-6, the `EPS_val` value is single digit precision.
 
Why did the removed data fail the test? The single-curve fitting function is rather sensitive to its input data. Rounding small values to 0 changes the spline fit, which affects the cocontent integral, which affects a regression, etc. This sensitivity is why the diode model fitting in `pvlib.ivtools.sdm.fit_pvsyst_sandia` (and similar) retains only the shunt resistance from the single IV curve fits; the other four parameters aren't reliably estimated by the cocontent method.

cwhanse:
any objections to merging this?

Antoine-0:
no, this is all fine for me, thanks again for running the tests !

cwhanse:
Thanks @Antoine-0

Antoine-0:
I merely noticed an issue, thanks to you for updating the tests and the great work you put into pvlib !

PR #1319, opened 2021-10-02
Check longitude and latitude are numbers (float or integer)

ndclt:
- [x] Closes #xxxx: **I didn't find any issue linked to the modification** 
 - [x] Tests added
 - [x] Updates entries to [`docs/sphinx/source/api.rst`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes. (not relevant for this MR)
 - [ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``). **I didn't find a file for the unreleased version where I can add my modification**

# Context

In [this Stackoverflow question](https://stackoverflow.com/questions/69418240/ufunc-add-did-not-contain-a-loop-with-signature-matching-types-dtypeu32/69419181#69419181), the OP creates a `Location` object with strings without errors and this leads to a Numpy error when using the function `get_solarposition`. Numpy couldn't add a string with two float array.

# Goal
This PR aims at raising an error when something different from a number is given to the `Location` init function.

# Implementation choice
I hesitate between the solution proposed in the PR and something like this:
```python
# for longitude
try:
   self.longitude = float(longitude)
except ValueError:
   raise TypeError('Invalid longitude specification')
```
I prefer the PR version because it doesn't hide the conversion (which is done with the `float` use) but the PR version won't work with a size-1 numpy array whereas the `float` one will (but I don't know if it's a common use case).

### DIFF
--- a/pvlib/location.py
+++ b/pvlib/location.py
@@ -57,6 +57,8 @@ class Location:
 
         if not isinstance(latitude, (float, int)):
             raise TypeError("latitude should be a float")
+        if not isinstance(longitude, (float, int)):
+            raise TypeError("longitude should be a float")
         self.latitude = latitude
         self.longitude = longitude
 
--- a/pvlib/location.py
+++ b/pvlib/location.py
@@ -56,9 +56,9 @@ class Location:
     def __init__(self, latitude, longitude, tz='UTC', altitude=0, name=None):
 
         if not isinstance(latitude, (float, int)):
-            raise TypeError("Invalid latitude specification")
+            raise TypeError('Invalid latitude specification')
         if not isinstance(longitude, (float, int)):
-            raise TypeError("Invalid longitude specification")
+            raise TypeError('Invalid longitude specification')
         self.latitude = latitude
         self.longitude = longitude
 
--- a/pvlib/location.py
+++ b/pvlib/location.py
@@ -56,9 +56,9 @@ class Location:
     def __init__(self, latitude, longitude, tz='UTC', altitude=0, name=None):
 
         if not isinstance(latitude, (float, int)):
-            raise TypeError("latitude should be a float")
+            raise TypeError("Invalid latitude specification")
         if not isinstance(longitude, (float, int)):
-            raise TypeError("longitude should be a float")
+            raise TypeError("Invalid longitude specification")
         self.latitude = latitude
         self.longitude = longitude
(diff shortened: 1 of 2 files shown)


wholmgren:
> Hi @ndclt, thanks for the PR, and answering the stackoverflow question :)

+1

I'm -1 on type validation but +1 on [type annotations](https://docs.python.org/3/library/typing.html) for the `Location` class. Type annotations are growing in python and I'd rather rely on that flexible tooling than building our own inflexible validators.

wholmgren:
> I also give a try to the monkeytype package in this branch. Would you like me to create a PR with these modifications?

I previously wrote type hints for `Location.__init__` (and other objects) but never got to the point of creating a PR. The relevant commit for Location is https://github.com/wholmgren/pvlib-python/commit/5f1b7ad53c530b631f779df0b6f31f01f474e7b5. If I recall correctly, making mypy happy required a bit more than the monkeytype package suggests.

Some in the pvlib community are not enthusiastic about type hints, so to get started I would not annotate the other methods. Simply annotating the object constructors/attributes should catch a fair number of problems.

> other than the convenience of raising hints in an IDE.

To me they are most helpful when the IDE is setup to flag/underline conflicts as I write/review code. Most of the pvlib function layer is straightforward enough, but the object layer could benefit from hints. I don't recall the details, but the `ModelChainResult` annotations helped identify a few bugs during that big refactor. Annotations are relatively helpful to people that write applications (like we did for Solar Performance Insight) or production scripts but less helpful (or worse) for people that write research code. pvlib serves both communities so it's hard to know the right answer. I'm not aware of any good tooling to combine type hints with documentation, so there's a bit of duplication involved.

In any case, it seems there's not an appetite for type checking these inputs. I'd be happy for help with experimenting with type annotations in the `Location`/`PVSystem`/`Mount`/`ModelChain` `__init__` methods and setting up mypy in the CI, but I'm not confident the efforts will ultimately be accepted.

PR #717, opened 2019-05-10
ENH: infinite sheds

mikofski:
- ~~[ ] Closes issue #xxxx~~
 - [x] I am familiar with the [contributing guidelines](http://pvlib-python.readthedocs.io/en/latest/contributing.html).
 - [x] Fully tested. Added and/or modified tests to ensure correct behavior for all reasonable inputs. Tests (usually) must pass on the TravisCI and Appveyor testing services.
 - [x] Updates entries to `docs/sphinx/source/api.rst` for API changes.
 - [x] Adds description and name entries in the appropriate `docs/sphinx/source/whatsnew` file for all changes.
 - [X] Code quality and style is sufficient. Passes LGTM and SticklerCI checks.
 - [X] New code is fully documented. Includes sphinx/numpydoc compliant docstrings and comments in the code where necessary.
 - [X] Pull request is nearly complete and ready for detailed review.

Implements "infinite sheds" model for irradiance on long, regular bifacial array (fixed or single-axis tracked) on horizontal ground.

PR #1354, opened 2021-12-10
add mlfm

### LINKED ISSUE #925: Adding the Loss Factors Model and Mechanistic Performance Models to PVLIB
I'll be adding the Loss Factors Models and Mechanistic Performance Models to PVLIB, . 

See http://www.steveransome.com/PUBS/1906_PVSC46_Chicago_Ransome.pdf for more details on what they do

I hope it should be ready by the conference in Salt Lake in May

I've uploaded a couple of graphs and some sample code and Initially I'm looking for feedback and comments as I will need to enhance the present naming convention

With the LFM I have been using prefix and suffix letters to describe the type of measurement and any corrections. For example this Python 3 code shows the naming I have been using and does normalised LFM and temperature correction.
I have just shown nIsc and nVoc for clarity but will upload code for all variables soon plus graphs and MPM fits.

Once I have this in a correct form I can add more of features

mSi0188_nVoc_U_5) mSi0188_nVoc_U_C4 ~  102 1%; -0 33%; 14 64%; -2 27%
mSi0188_PRdc_lfm_gi_3
[MLFM_SRCL_200302.py.txt](https://github.com/pvlib/pvlib-python/files/4278331/MLFM_SRCL_200302.py.txt)

Any feedback is welcome
Steve

steve-ransome:
- [x] Closes #925 
 - [x] Tests added


cwhanse:
@wholmgren @kanderso-nrel would appreciate some advice here. As currently organized, #1354 adds two python modules to pvlib-python: `mlfm.py` contains the parameter estimation and model functions, and `mlfm_graphs.py` contains code to produce figures which are important to the model fitting and model evaluation processes. `mlfm_graphs.py` imports matplotlib which is not part of pvlib's requirements. @steve-ransome that's why the tests are failing.

What makes sense?
- add matplotlib to requirements. I feel like we'd regret this but I can't articulate why.
- move the plotting functions to the tutorial code. Users could get the code but to using it would involve copying and pasting etc. Not the most convenient.
- exclude `mlfm_graphs.py` from testing.

I'm not enamored with both `pvlib.mlfm` and `pvlib.mlfm_graphs` but I don't have a better idea atm.

kanderso-nrel:
I've not yet done more than a quick skim of this code, so take these as general comments.

> exclude mlfm_graphs.py from testing.

RdTools has a plotting module.  We decided that it would be too brittle to test those functions beyond (1) does it run without error and (2) does it return a `Figure`.  Examples of more complete testing strategies that seem too brittle: inspect the `Figure` object attributes; save a PNG and compare pixels with a reference image.

> add matplotlib to requirements. I feel like we'd regret this but I can't articulate why.

FWIW requiring matplotlib has not really been a technical burden in other projects I work on.  Including plotting functions in general might be a pain because everyone has their own favorite plotting package; for example, RdTools uses matplotlib for some functions and plotly for others.  

> I'm not enamored with both pvlib.mlfm and pvlib.mlfm_graphs

Could have `pvlib.mlfm` be a folder with an `__init__.py` and separate submodules, the same way #717 is (re)structuring `pvlib.bifacial`.

One other note: I wonder if `mlfm` would work better as a separate repo, i.e. `pvlib/mlfm` instead of `pvlib.mlfm`.  If looking at plots is important and there isn't much (any?) integration with the rest of pvlib-python, maybe it's more of an application than a component of a modeling chain?

cwhanse:
> I wonder if mlfm would work better as a separate repo

Personally, I think it would, but that is very unlikely to happen. I'm aware of at least one commercial pvlib user who is also using the mlfm (as private code). We discussed pros and cons of including it in pvlib [here](https://github.com/pvlib/pvlib-python/issues/925#issuecomment-951116697)

steve-ransome:
Thanks for the comments.

The MLFM can do many tasks including 
- fitting and analysing IV curves (comparing with desoto etc)
- apportioning loss mechanisms and limits
- derives low light and temperature coefficients more easily than desoto or SAPM.
- fitting indoor or outdoor measurement data (including IEC 61853) and can be compared with desoto and SAPM
- predicting PR, power, voltage or current vs. irradiance and temperature for energy yield predictions (as sapm)
- derive degradation rates and causes versus irradiance and temperature (e.g. shunt resistance fall, worse at low light than high light) and can be compared with other models. 

To keep this code short I haven't shown any integration with the rest of pvlib here but have been using it with pvlib weather functions, angles of incidence, tmy 2 power calculations and inverters since at least 2015 and have published a few papers with Sandia including http://www.steveransome.com/PUBS/502_42pvsc_NewOrleans_150610t12.pdf and https://pvpmc.sandia.gov/download/7879/

Although the MLFM uses graphics to highlight its findings (I used matplotlib because all the tutorials did) it does not need to, all of the graphs come from calculations stored in pandas dataframes.

My feeling is it should be integrated with pvlib and I will do whatever I can to reduce any problems with graphics.

cwhanse:
@steve-ransome I will push some changes to implement @wholmgren's suggestions.

steve-ransome (commented):
Thanks

cwhanse:
@steve-ransome sorry I set this aside for a while. Next step: we need to write tests for each function that is being added. For numerical functions, can you code a set of inputs with known output and put them in a new file `tests/test_mlfm.py`?  We can re-use these data to test the plot functions.

steve-ransome:
Thanks Cliff. I looked through some of the existing tests to get an idea what was needed, there were several different styles/things included.

This file should run and check three functions with one date_time. If anyone has comments I will follow them then put in the repository.

Should I write one for mlfm_fit as this will need many date_times (probably 20 would do).

That might be enough to check the graph functions - if you wish me to do that

[test_mlfm_220111t17.py.txt](https://github.com/pvlib/pvlib-python/files/7848604/test_mlfm_220111t17.py.txt)
.

cwhanse:
@steve-ransome if you can develop a test for `mlfm_fit` I'll go ahead and organize that python script into `test_mlfm`. I'm still unsure how to test the plotting functions; perhaps we only check that the functions return a figure and leave it at that.

steve-ransome:
Here's an update with a test for mlfm_fit using the matrix data from mlfm.ipynb selecting option 2.
I think although we can get calculations with tolerance_calc atol=0.0000001 we might only get fits to be tolerance_fit atol=0.001 or so as the Excel and Python RMSE only differ 

[test_mlfm_220114t17.py.txt](https://github.com/pvlib/pvlib-python/files/7872325/test_mlfm_220114t17.py.txt)
in the 5th significant figure 0.0027961 vs. 0.0027962 

*If* we wanted to test then graphics then the mlfm.ipynb with file option 2 gives these graphs, we could check that they plot and look ok.

mlfm

steve-ransome:
I'm thinking of submitting a paper on this work for WCPEC8 . It would be a little about the mlfm and code but mostly highlighting PVLIB and the methods and procedures I had to go through to get it up to standard and included, bearing in mind that I have no programming qualifications and little experience in collaborative work so hadn't used git hub etc.
Any thoughts? I'd obviously provide all the acknowledgements  etc. for the conference and give links etc to help anyone else contribute their work.

steve-ransome:
I have now written an abstract paper called "ADDING THE MLFM TO PVPMC/PVLIB" and intend to submit it to WCPEC8 this evening European time 4th Feb (the deadline). It will have lots of links to PVPMC. If anyone wishes to make any comments before I submit it, please do. If it's accepted I would obviously interact more and could add some extra authors.

steve-ransome:
"ADDING THE MLFM TO PVPMC/PVLIB" was accepted as a poster at WCPEC8. It would be great to get some feedback for this, I can put the abstract up here if you wish. All contributions will be acknowledged or I could add people as joint authors.

cwhanse:
@kanderso-nrel what is the status of pvlib-python\docs\tutorials?  I'm not seeing them on the new documentation pages; are they retired, or am I missing the link?

kanderso-nrel:
Other than a brief mention on the [matlab comparison page](https://pvlib-python.readthedocs.io/en/stable/user_guide/comparison_pvlib_matlab.html#documentation) I don't think the notebooks in `docs/tutorials` have ever been integrated with the readthedocs sphinx pages.  I usually forget those notebooks exist and if I had to guess I'd speculate that most users don't know about them.  I don't think we've officially retired them, but they do tend to collect dust and get out of date.  Maybe we should just merge their content into the readthedocs User Guide pages and get rid of the notebooks.

For this PR I wouldn't object to tossing `mlfm.ipynb` onto the `docs/tutorials` pile and sorting out what to do with the notebooks later.  But I doubt many people will find it in there in the meantime.

adriesse:
> But I doubt many people will find it in there in the meantime.

A paper referring and linking to it should help.

cwhanse:
@steve-ransome I have the three tutorial notebooks cleaned up and running.

There are some parts of `mlfm` functions that are not hit by the tests. Could you take a look at providing data for test cases that hit `mlfm_norm_to_stack` lines 271-294. Here, we need a normalized data set that has `'v_mp'` and `'i_mp'` instead of `'v_ff', 'r_oc', 'i_ff', 'r_sc'`. The same data will improve test coverage for `plot_mlfm_stack`.

steve-ransome:
Thanks Cliff, I am looking at this now to try to find the easiest way. 

If we have 6 mlfm variables then the 4 variables derive straight from them, i_sc and v_oc are unchanged and i_mp =i_ff * r_sc and v_mp = v_ff * r_oc 

If we check the 6 (i_sc, r_sc, i_ff, v_ff, r_oc, v_oc) and it works, so must the 4 (i_sc, r_sc * i_ff, v_ff * r_oc, v_oc).

Do you think we still need to create and test the 4 numbers separately?

cwhanse:
> Do you think we still need to create and test the 4 numbers separately?

I do, so that the tests touch all lines of code which do computations.  Sounds like we can make a 4 variable test from the existing 6 variable test data?

steve-ransome:
Exactly. We don't need new measurement data of a 4 variable set, we can convert 6 --> 4 and use that  
as i_mp =i_ff * r_sc and v_mp = v_ff * r_oc.

pvlib/mlfm.py
```diff
@@ -0,0 +1,803 @@
...
+    # put mlfm values in a stack from pr_dc (bottom) to 1/ff_ref (top)
+    dstack['pr_dc'] = + dnorm['pr_dc']  # initialise
+    dstack['i_sc'] = -(dnorm['i_sc'] - 1) * corr
+    dstack['i_mp'] = -(dnorm['i_mp'] - 1) * corr - gap/2
+    dstack['i_v'] = gap
+    dstack['v_mp'] = -(dnorm['v_mp'] - 1) * corr - gap/2
+    dstack['v_oc'] = -(dnorm['v_oc'] - 1) * corr
+
+    dstack['temp_module_corr'] = (
+        - (dnorm['v_oc'] - dnorm['v_oc_temp_corr']) * corr)
```
cwhanse, line 292: @steve-ransome can you confirm that these statements are all correct? When I made a new test from the 6-element data, replacing `v_ff` and `r_oc` by `v_mp = v_ff * r_oc` and `i_ff` and `r_sc` by `i_mp = i_ff * r_sc`, the stacked `i_sc` wasn't very close to the value computed from the 6-element test. I had to make a new fixture `stack4` and just copied output from this code block.

cwhanse:
@pvlib/pvlib-maintainer I think this is ready for a look. Coverage is a bit shy because the tests don't hit all of the try/except blocks.

pvlib/mlfm.py
```diff
@@ -0,0 +1,803 @@
...
+        c_3 * np.log10(dmeas['poa_global_kwm2']) + \
+        c_4 * dmeas['poa_global_kwm2'] + c_6 / dmeas['poa_global_kwm2']
+    if 'wind_speed' in dmeas.columns:
+        mlfm_out += c_5 * dmeas['wind_speed']
+    return mlfm_out
+
+
+def mlfm_fit(data, var_to_fit):
+    '''
+    Fit MLFM to data.
```
kanderso-nrel, line 359: Newbies like me might benefit from a mention of what variables are appropriate to fit with this function.  `pr_dc`, others?

kandersolar (commented):
A couple general questions/comments (sorry if these have already been discussed elsewhere):
- Was the decision to use bundled DataFrame and dict inputs an intentional departure from the more typical (for pvlib) separate Series and float inputs?
- Would it make more sense to have the plotting functions be defined in notebooks/gallery examples instead of pvlib itself?  I feel like there's enough going on in the plotting functions that people will want to modify/customize the code to suit their needs.

A few things I noticed while reading the notebooks:
- The tutorials rely on external CSV data files which aren't currently included.  Should they be added to this PR?
- Cell 4: seems like the `while` loop is unnecessary
- Cell 16: `plot_mlfm_scatter` returns a figure with two overlapping legends
- Underscores/asterisks are being interpreted as markdown instead of literal characters:
  + `1) Select MLFM measurement`
  + `8) Normalise MLFM values...`
  + `14) Convert multiplicative to subtractive...`
  + `17) Fit mechanistic model...`

pvlib/mlfm.py
```diff
@@ -0,0 +1,803 @@
...
+
+    fig, ax1 = plt.subplots()
+
+    ax1.set_title(title)
+
+    ax1.set_ylabel('Normalised values')
+    ax1.axhline(y=1, c='grey', linewidth=3)  # show 100% line
+    ax1.set_ylim(0.8, 1.1)  # optional normalised y scale
+
+    ax1.set_xlabel('Plane of array irradiance [W/m$^2$]')
```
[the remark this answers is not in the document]
steve-ransome: Yes, thanks

pvlib/mlfm.py
steve-ransome: Thanks Cliff. The norm data is multiplicative so should work with multiplication as I wrote,
The stack data is subtractive (to get the stack plot to show each loss term separately)

norm_v_mp = norm_v_ff * norm_r_oc and norm_i_mp = norm_i_ff * norm_r_sc

stack_v_mp = stack_v_ff + stack_r_oc and stack_i_mp = stack_i_ff + stack_r_sc

Does that work?

pvlib/mlfm.py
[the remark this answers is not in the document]
steve-ransome: Thanks. The original python code came from things I had written and was using with clients where most data handling has been done with data frames, there's no intentional departure from pvlib but I was looking at code such  as 
def sapm(effective_irradiance, temp_cell, module):
which uses dicts and dataframes. 

@cwhanse has been modifying my code to fit better with pvlib

The stacked loss plotting code is quite complicated and I suggest having it somewhere that the users can test unless they want to alter it. It's much more complex than just having simple xy plots.

The csv files were sent to @cwhanse, they should be in the tutorials soon I hope

cell 4 : OK

cell 16 : Thanks, will correct it

underscores / asterisks : thanks will correct

pvlib/mlfm.py
cwhanse: Not exactly the same values. So I think we need to create new, known output for the 4-variable case. Because I did that using the code referenced above, if you confirm that code is correct, then we'll trust the values in the new fixture. 

```
stack6['v_ff'] + stack6['r_oc']
Out[87]: 
0    0.217098
dtype: float64

stack4['v_mp']
Out[88]: 
0    0.214483
Name: v_mp, dtype: float64
```

steve-ransome (commented):
ok

pvlib/mlfm.py
steve-ransome: Thanks Cliff, I checked it and your figures are correct. 

While line 1 is true
1) norm_v_mp = norm_v_ff * norm_r_oc , norm_i_mp = norm_i_ff * norm_r_sc
Line 2 is only approximate (to about 0.2%)
2) stack_v_mp ~ stack_v_ff + stack_r_oc , stack_i_mp ~ stack_i_ff + stack_r_sc

We are taking accurate multiplicative losses and the putting them onto a linear graph so there may be some small distortion but it is small. 

It's the overall relative losses (e.g. r_oc vs. r_sc) and any changes with time that matter. The sum of all the stack losses and the product of all multiplicative losses are the same from stack_6 and stack_4 as that's the way they drop from 1/ff to pr_dc and the stack graphs are very useful.

Therefore we need to check the stack_4 figures as you have done.

steve-ransome: Cliff, as discussed at the PVPMC the multiplicative LFM losses are all correct (the product of them takes us from the initial 1/ff to pr_dc) but in translating them to subtractive losses where the sum of them takes us from 1/ff to pr_dc correctly, but there's a tiny bit of non-linearity meaning the lfm4 and lfm6 figures won't be exactly the same, they'll be close and as the stack loss is just a visual interpretation of losses should be good enough to identify losses and degradation as in previous publications. If we want to get them exactly the same (not necessary but nice) then we could do stack plots of 1-loss on a log  y scale which shows it well with the thickness of the bar indicating the relative loss but we then might have to rethink possible negative losses such as isc gain due to spectral effects and the log graph would need to go from only 0.5 to 1.5 or so. What's the next stage on this as isn't the other code working OK?

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+        c_1 + c_2 (T_m - 25) + c_3 \log10(G_{POA}) + c_4 G_{POA}
+        + c_5 WS + c_6 / G_{POA}
+
+    where :math:`G_{POA}` is global plane-of-array (POA) irradiance in kW/m2,
+    :math:`T_m` is module temperature in C and :math:`WS` is wind speed in
+    m/s.
+
+    Parameters
+    ----------
+    dmeas : DataFrame
```
adriesse, line 314: Suggest making three separate parameters for the three input columns as is usually done in pvlib.

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+    return mlfm_out
+
+
+def mlfm_fit(data, var_to_fit):
+    '''
+    Fit MLFM to data.
+
+    Parameters
+    ----------
+    data : DataFrame
```
adriesse, line 364: Suggest making multiple separate parameters for the input columns as is usually done in pvlib.

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+    # meaningless but a non-zero value may have been returned.
+    if c5_zero:
+        coeff[4] = 0.
+
+    # get error of mlfm coefficients as sqrt of covariance
+    perr = np.sqrt(np.diag(pcov))
+    coeff_err = list(perr)
+
+    # save fit and error to dataframe
+    pred = mlfm_6(data, coeff[0], coeff[1], coeff[2], coeff[3], coeff[4],
```
adriesse, line 435: consider using *coeff

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+
+    # plot stack in order bottom to top,
+    # allowing self_ref and temp_module corrections
+    ax1.stackplot(xdata, *tuple(ydata), labels=labels, colors=color_map)
+
+    ax1.axhline(y=1/fill_factor, c='grey', lw=3)  # show initial 1/FF
+    ax1.axhline(y=1, c='grey', lw=3)  # show 100% line
+    ax1.set_ylabel('stacked mlfm losses')
+
+    # find number of x date values
```
adriesse, line 708: Fancy date labels could be a separate utility function reusable elsewhere and by others.

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+    except KeyError:
+        pass
+
+    ax2.legend(bbox_to_anchor=(bbox, 0.3), loc='upper left', borderaxespad=0.)
+    ax1.set_xticklabels(xax2, rotation=90)
+
+    return fig
+
+
+REFS = """
```
adriesse, line 765: Where does REFS get used?

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+            except IndexError:
+                xax2[x_count] = xdata[x_count]
+            except TypeError:  # xdata can't be subscripted
+                xax2[x_count] = xdata[0]
+
+        x_count += 1
+
+    ax1.set_xticklabels(xax2)
+    ax1.set_ylim(0.6, 1/fill_factor + 0.1)  # optional normalised y scale
+    plt.legend(bbox_to_anchor=(bbox, 1), loc='upper left', borderaxespad=0.)
```
adriesse, line 740: There seems to be a mix of pyplot and non-pyplot usage.  It might be preferable to stick with one approach.

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
+'''
+This ``mlfm code`` module contains functions to analyse and predict
+performance of PV modules using the mechanistic performance (MPM) and
+loss factors models (LFM). The module also contains functions to display
+performance of PV modules using the mechanistic performance (MPM) and
+loss factors models (LFM)
+
+Authors : Steve Ransome (SRCL) and Juergen Sutterlueti (Gantner Instruments)
+Thanks to Cliff Hansen (Sandia National Laboratories)
```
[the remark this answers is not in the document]
steve-ransome: Thanks for the helpful comments Anton. This is the first code I have written for PVLIB so some things come as I am accustomed to doing for clients and my own use, obviously some things need to be changed to fit in with the rest of the code. I would have thought a big chunk of code that's already used by other teams could have an acknowledgement, but if it's not desired then we could take it out. All the references show where the code came from and a link to NREL data. Maybe we should add CFV too for their matrix stuff. Will go with whatever you all agree,

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+#
+T_STC = 25.0  # STC temperature [C]      temperature_ref
+T_HTC = 75.0  # HTC temperature [C]
+G_STC = 1000.0   # STC irradiance  [W/m^2]
+G_LIC = 0.2   # LIC irradiance  [kW/m^2]
+
+
+#  Define standardised MLFM graph colours as a dict ``clr``
+
+clr = {
```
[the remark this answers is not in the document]
steve-ransome: Good point. It is effectively a global constant to get uniform colours. I think @cwhanse got me to do it this way but I am happy with CAPITALS for a constant

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+}
+
+
+def mlfm_meas_to_norm(dmeas, ref):
+    '''
+    Convert measured power, current and voltage to normalized values.
+
+    Parameters
+    ----------
+    dmeas : DataFrame
```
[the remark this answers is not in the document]
steve-ransome: Thanks. Yes it should be specified.  Outdoor measurements it's date and time which can be selected to say show only the first year of data. Indoors it's usually just the measurement number (e.g. 1..28 for a matrix) which can be ignored. Will look at this.

pvlib/mlfm.py
steve-ransome: Thanks, I had that comment before but hadn't gone with it as I had been using a weather+measurement dataframe which I am used to. However if it fits in better with PVLIB with separate G,T and W then we can do it.

pvlib/mlfm.py
steve-ransome: Thanks

pvlib/mlfm.py
steve-ransome: Am very happy for that to happen as making date axes legible with lots of data isn't easy. Will look at ways to do it but am open to suggestions.

pvlib/mlfm.py
steve-ransome: OK would need guidance as to what you suggest

steve-ransome:
I'll be presenting a poster "3BV3 WCPEC #8 ADDING THE MLFM TO PVPMC/PVLIB" at WCPEC8.
Basically it summarises the flow chart, graphs and some code. 

If any PVLIB guru is around at the session it would be great if they could see me there if there are pvlib questions I can't answer.

The poster has a section (cut and pasted below) on how to contribute code with lots of links. I paraphrased this from the latest pvlib page https://pvlib-python.readthedocs.io/en/stable/contributing.html#documentation

Please let me know if you have any comments, additions or corrections. I can send a draft poster now or can forward it the final version before I get it printed.

Is there an official QR code to use? I've added the "Powered by PVLIB" logo. If anyone wants to leave some stickers at the poster leave them with me and I can hand them out.

Thanks in advance. 

"
**How to contribute to PVLIB (add a QR code?)**

INFORMATION :
Stack Overflow : 	[http](http://stackoverflow.com/questions/tagged/pvlib)[://stackoverflow.com/questions/tagged/pvlib](http://stackoverflow.com/questions/tagged/pvlib)
Google groups : 	[https](https://groups.google.com/forum/%23!forum/pvlib-python)[://groups.google.com/forum/#!forum/pvlib-python](https://groups.google.com/forum/%23!forum/pvlib-python)
GitHub issues : 	[https](https://github.com/pvlib/pvlib-python/issues)[://github.com/pvlib/pvlib-python/](https://github.com/pvlib/pvlib-python/issues)[issues](https://github.com/pvlib/pvlib-python/issues)
Pull requests : 	https://github.com/pvlib/pvlib-python/pulls
Jupyter Notebook tutorials :	https://github.com/pvlib/pvlib-python/tree/master/docs/tutorials

Follow PEP8 : 	https://www.python.org/dev/peps/pep-0008/. (Max line length 79 chars)
Add your project : 	https://github.com/pvlib/pvlib-python/wiki/Projects-and-publications-that-use-pvlib-python.
Variables and Symbols : 	[https://pvlib-python.readthedocs.io/en/stable/user_guide/variables_style_rules.html#variables-style-rules](https://pvlib-python.readthedocs.io/en/stable/user_guide/variables_style_rules.html)
Documentation style : 	[https://pvlib-python.readthedocs.io/en/stable/contributing.html#documentation](https://pvlib-python.readthedocs.io/en/stable/contributing.html) 

Coding : 
PVSystem and Location classes provide convenience wrappers around the core pvlib functions. 
Remove logging calls and print statements.
Include documentation and Comprehensive unit tests 
Functions must return the desired output for all inputs 	(see https://github.com/pvlib/pvlib-python/issues/394). 

Code of conduct : 	https://github.com/pvlib/pvlib-python/blob/master/CODE_OF_CONDUCT.md"

AdamRJensen:
> I'll be presenting a poster "3BV3 WCPEC #8 ADDING THE MLFM TO PVPMC/PVLIB" at WCPEC8. Basically it summarises the flow chart, graphs and some code.
> If any PVLIB guru is around at the session it would be great if they could see me there if there are pvlib questions I can't answer.

I will unfortunately not be at WCPEC as I'm attending EuroSun which occurs simultaneously - maybe @adriesse will be?

> Is there an official QR code to use? I've added the "Powered by PVLIB" logo. If anyone wants to leave some stickers at the poster leave them with me and I can hand them out.

Sadly, all of the stickers reside with Silvana and me - I suppose I should have done more to spread them out...

steve-ransome:
@adriesse is due to chair a session with me at wcpec8 so I hope he'll be there. You could always mail some to me or him. Let me know if you want my address in the UK.

adriesse:
I'll be there!  Adam, could you bring some stickers to Freiburg the week before?

AdamRJensen:
@adriesse I sadly will only attend our IEA PVPS Task 16 virtually this time, as I will be returning from my stay at NREL the same day.

steve-ransome:
I presented my poster at WCPEC8 today in Milano.

There was a lot of interest, not only in the MLFM but also using pvlib and also several people said they'd like to add their code.

This is the poster that I will submit (without the "to be presented" box). Any comments gratefully received. 
image

cwhanse:
Thank you, Steve. You and I will likely need to get on another Zoom call to address @adriesse comments.

steve-ransome:
Here's a "powered by PVLIB" sticker in the wild. 
image

steve-ransome:
Thanks @cwhanse . All @adriesse comments can be easily added. Most of the other changes made so far are OK but I think a couple of things are complicating procedures so will look at better ways of doing things. At the moment we are allowing for 4-LFM parameter and 6-LFM parameter fits, but we can go from 2 to 12 in the future (adding spectrum, aoi, soiling, mismatch, rollover etc,) so I think it would be best to have a number identifying the total parameters and using that as easier than looking for which parameters are defined. Will do some code when I am back in the UK next week then circulate it and will be ok for a zoom.

steve-ransome:
This is the MLFM paper as submitted (without the header)
[2209_MLFM_WCPEC8_Ransome_3BV.3.14_pvlib.pdf](https://github.com/pvlib/pvlib-python/files/9678797/2209_MLFM_WCPEC8_Ransome_3BV.3.14_pvlib.pdf)

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+    mlfm_6
+    '''
+
+    # drop missing data
+    data = data.dropna()
+
+    c5_zero = 'wind_speed' not in data.columns
+    # if wind_speed is not present, add it
+    if c5_zero:
+        data['wind_speed'] = 0.
```
markcampanelli, line 404: Technically, c5 is _unidentifiable_ if the data do not contain windspeed information. As such, the algorithm may not converge robustly and may report convergence problems when you nonetheless keep c5 in the model. (Your bounds on c5 may temper the convergence issues to some degree.)

Unfortunately, `scipy.optimize.curve_fit` does not allow you to specify equal upper and lower bounds for c5 in this case, which here is presumably zero _by convention_. I can suggest a couple ways to work around this, but first I suggest examining the convergence info from `scipy.optimize.curve_fit` in a case where there is no windspeed info. (Do this by passing `full_output=True` and examining the  `mesg` and `ierint` entries.)

pvlib/mlfm.py
markcampanelli: I think that you can pass `full_output=True` to `scipy.optimize.curve_fit`, then get the `pred` variable pre-calculated as `infodict["fvec"]` from the additional output `infodict` from `scipy.optimize.curve_fit`.

`full_output=True` also provides two more outputs: `mesg` and `ierint`, which you should inspect for convergence information, possibly warning the user and/or passing back to the caller.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html

pvlib/tests/conftest.py
```diff
@@ -150,6 +150,17 @@ def has_numba():
 requires_siphon = pytest.mark.skipif(not has_siphon,
                                      reason='requires siphon')
 
+try:
+    import matplotlib.pyplot as plt  # noqa: F401
```
markcampanelli, line 154: Why is this import made here?

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+    Returns
+    -------
+    dnorm : DataFrame
+        Normalised values.
+
+        * `'pr_dc'` is `'p_mp'` normalised (divided) by reference `'p_mp'` \
+            and by `'poa_global'` in kW/m^2.
+        * `'pr_dc_temp_corr'` is `'pr_dc'` adjusted to 25C.
+        * Columns `'i_sc'`, `'i_mp'`, `'v_oc'`, `'v_mp'`, `'v_oc_temp_corr'`,
+          `'r_sc'`, `'r_oc'`, `'i_ff'`, `'v_ff'` are returned when the
```
markcampanelli, line 106: The naming convention here strikes me as inconsistent. These are also all normalized values for the DC circuit, but they don't follow a naming convention like `pr_dc` and `pr_dc_temp_corr`.

markcampanelli (changes requested):
@steve-ransome Thanks for this submission.

Here is my review of your core algorithm, esp. the normalization and fitting. I have not yet run any examples or looked at the error stacking, plotting functions, or tests.

The main issue I see is potentially more careful handling of the fits in the special case where no windspeed info is provided. I think this should be investigated a bit further and would be happy to discuss.

pvlib/mlfm.py
markcampanelli: Not sure if this helps: https://matplotlib.org/stable/gallery/text_labels_and_annotations/date.html

pvlib/mlfm.py
steve-ransome: It doesn't, that's what  @cwhanse changed it to, Cliff?

pvlib/__init__.py
```diff
@@ -12,6 +12,7 @@
     ivtools,
     location,
     modelchain,
+    mlfm,
```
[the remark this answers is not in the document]
steve-ransome: Good spot

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+
+#  DEFINE REFERENCE MEASUREMENT CONDITIONS
+# or use existing definitions in pvlib
+
+# NAME  value  comment         unit     PV_LIB name
+#
+T_STC = 25.0  # STC temperature [C]      temperature_ref
+T_HTC = 75.0  # HTC temperature [C]
+G_STC = 1000.0   # STC irradiance  [W/m^2]
+G_LIC = 0.2   # LIC irradiance  [kW/m^2]
```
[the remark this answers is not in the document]
steve-ransome: They should be the same units, one was changed without the other. G_LIC hasn't been used here but will be used in later code changes as the MLFM can create useful performance values at other conditions such as LIC, HTC, NOCT etc. for validation.

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+        * `'p_mp'` - Power at maximum power point at Standard Test Condition
+          (STC). [W]
+        * `'gamma_pdc'` - Temperature coefficient of power at STC. [1/C]
+
+        May include:
+
+        * `'i_sc'` - Current at short circuit at STC. Required if `'i_sc'` is
+          present in ``dmeas``. [A]
+        * `'v_oc'` - Voltage at open circuit at STC. Required if `'v_oc'` is
+          present in ``dmeas``. [A]
```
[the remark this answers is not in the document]
steve-ransome: Thanks, that's been corrected

pvlib/mlfm.py
cwhanse: I don't recall why or when that was changed to a constant string.

pvlib/mlfm.py
steve-ransome: pr is usually defined as the performance ratio ac as PR = Yf/Yr. 
There isn't a single naming convention for the equivalent at dc conditions Ya/Yr. 
I've seen both performance_factor, module_factor (which I don't like because it can be used for modules, strings, dc_combiners and inverter_inputs) so had used pr_ac and pr_dc to differentiate the two.

markcampanelli: Thanks for explaining the Performance Ratio connection that I totally missed (doh!). This might be worth a comment, unless I missed it.

My pattern-matching brain would prefer to just use `p_mp` and `p_mp_temp_corr` because (1) this is consistent with the remaining `i_sc`, `i_mp`, etc., and (2) the indication of "normalization" is already in the dataframe name `dnorm`.

Alternatively, one could apply some sort of consistent normalization pre/suffix to each variable in the dataframe.

steve-ransome: Thanks Mark. 
The problem I see is that on the dc side there are values for pr and p_mp (W or normalised), on the ac side there are also pr and p_mp (W or normalised) and these will differ by the factor inv_eff.
That's why I've used _dc and _ac suffixes to differentiate between values that would have the same name.

I've also used meas_ or norm_ as prefixes or dataframe names.

We could use p_dc and p_ac  but unless we have a dc equivalent of pr (ac) then I think I would need a pr_dc suffix.

I always use p_mp = i_mp * v_mp as at perfect mppt and p_dc = i_dc * v_dc or p_ac = i_ac * v_ac to cope with imperfect mpp tracking or clipping.

pvlib/mlfm.py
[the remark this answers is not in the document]
markcampanelli: For completeness and explicitness, you could include `G_HTC`, and `T_LIC`, as well as the `G_LTC` and `T_LTC` (and perhaps `G_NOCT`) conditions defined in IEC 61853-1.

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+        * `'beta_v_oc'` - Temperature coefficient of open circuit voltage at
+          STC. Required if `'v_oc'` is present in ``dmeas``. [1/C]
+
+    Returns
+    -------
+    dnorm : DataFrame
+        Normalised values.
+
+        * `'pr_dc'` is `'p_mp'` normalised (divided) by reference `'p_mp'` \
+            and by `'poa_global'` in kW/m^2.
```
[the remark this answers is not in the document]
markcampanelli: Aha: I now see that here is where you might want to include the words "direct-current performance ratio" for the slow folks like me :).

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+        / (dmeas['poa_global'] / G_STC)
+
+    # temperature corrected
+    dnorm['pr_dc_temp_corr'] = (
+        dnorm['pr_dc'] *
+        (1 - ref['gamma_pdc']*(dmeas['temp_module'] - T_STC)))
+
+    if 'i_sc' in dmeas.columns:
+        dnorm['i_sc'] = dmeas['i_sc'] / (dmeas['poa_global'] / G_STC) \
+            / ref['i_sc']
```
markcampanelli, line 128: Nit: This code would read a bit quicker if it followed the same arrangement as for `dnorm['pr_dc']`, i.e.,

```python
dnorm['i_sc'] = dmeas['i_sc']  / ref['i_sc'] / \
    (dmeas['poa_global'] / G_STC)
```

pvlib/mlfm.py
```diff
@@ -0,0 +1,804 @@
...
+    c_2 : float
+        Temperature coefficient in model. [1/C]
+    c_3 : float
+        Coefficient for low light log irradiance drop.
+    c_4 : float
+        Coefficient for high light linear irradiance drop.
+    c_5 : float, default 0
+        Coefficient for wind speed dependence.
+    c_6 : float, default 0
+        Coefficient for dependence on inverse irradiance.
```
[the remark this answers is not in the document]
markcampanelli: Looking closer, I think it could be that G_{POA} in the math equation is actually the POA irradiance normalized by `G_STC`, and does NOT actually have units kW/m^2, which means coefficients `c4` and `c6` are indeed unitless. However, I think `c_5` should have units of s/m.

markcampanelli:
Because there are several mflm-related tutorial notebooks and supporting files, we may want to add a general `docs/tutorials/mflm/` directory (with a `data` sub-directory) to hold everything.

pvlib/mlfm.py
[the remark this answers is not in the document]
steve-ransome: Maybe an easier way to look at this is to use the dimensionless value "suns" where [G=1000W/m^2 / G_STC] is 1sun.
This way all the coefficients can be written as "dimensioned"
c_1 is constant,
c_2 is 1/K,
c_3 is /sun,
c_4 is /sun,
c_5 is 1/(ms^-1) as you said
c_6 is *sun,

The values are meaningful in that the
c_3 is the low light "v_oc and r_sh drop" at G=0.1sun (if we use log10 then log10(0.1)=-1)

c_4 is the high light rseries drop at 1sun due to I^2.Rseries

The graph shows the LFM's c_3 and c_4 coefficients quantifiably change the drops at 0.1sun and 1.0sun respectively

image

steve-ransome:
> 
We already do that.

markcampanelli:
> We already do that. , there are directories for reference data, meas_gtw measurements+weather, figs and outputs

I may have not been clear. I was thinking in terms of how this structure might better grow and coexist with other multi-file tutorials.

mflm alternative directory structure

pvlib/mlfm.py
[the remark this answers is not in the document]
markcampanelli: I personally do often think in terms of suns, but I don't think we need to add "suns" units here.

Rather, in the code's documentation I'm seeing some conflation between an irradiance in units of kW/m^2 and a unit-less normalized irradiance that's been divided by G_STC = 1000 W/m^2. Those are significantly different concepts, even if numerically they end up being the same.

markcampanelli:
@steve-ransome In the example notebooks, don't use Windows-specific path separators, which don't work on Mac/Linux. For example, instead of
`ref_file_name = (root_dir + '\\mlfm_data\\ref\\' + 'mlfm_reference_modules.csv')`,
use
ref_file_name = os.path.join(root_dir, 'mlfm_data', 'ref', 'mlfm_reference_modules.csv')

pvlib/mlfm.py
markcampanelli: I made comparison `curve_fit`s using datasets n05667_Y13_R1k6_fClear_041.csv and x19074001_iec61853_041.csv, both of which have no `wind_speed` info. For the comparison, I removed the `c5` coefficient altogether from `mlfm_6` (instead using a new `mlfm_5` function for `func`). The remaining coefficients appear to be stable between the two approaches without convergence errors in either (although the convergence criterion may change from `gtol` to `ftol`). I did not check the results of of the fit beyond this. It appears that it is ok to use `mlfm_6` in all cases, but I would still recommend checking for convergence.

steve-ransome:
> 

Thanks, have updated mlfm to use that

markcampanelli:
@steve-ransome Would it be helpful at this point for me to review all the comments that I initiated and resolve/follow up with them?

steve-ransome:
Thanks @markcampanelli  that would be very useful. I think I've got as far as I can for the moment.

pvlib/mlfm.py
```diff
@@ -0,0 +1,803 @@
...
+def mlfm_meas_to_norm(dmeas, ref):
+    '''
+    Convert measured power, current and voltage to normalized values.
+
+    Parameters
+    ----------
+    dmeas : DataFrame
+        Measurements. Must include columns:
+
+        * `'poa_global_kwm2'` global plane of array irradiance [kW/m^2]
```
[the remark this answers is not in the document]
markcampanelli: I pointed out some suspected conflation in the code between using kW/m^2 units and using the (unitless) normalized irradiance G/(1000 W/m^2), where G is in W/m^2. I think it is important to avoid conflating this normalization with a units change.

pvlib/tests/test_mlfm.py
```diff
@@ -0,0 +1,262 @@
+
+import numpy as np
+import pandas as pd
+from pvlib import mlfm
+from numpy.testing import assert_allclose  # assert_almost_equal,
```
markcampanelli, line 5: Remove commented out import.

markcampanelli: Also, imports are typically alphabetized and broken out into standard library packages/modules, external packages/modules, and internal packages/modules.

pvlib/tests/conftest.py
markcampanelli: I'm 99% sure that the `import matplotlib.pyplot as plt` is not needed.

markcampanelli:
@steve-ransome I walked through the comment history, and there seems to be some unresolved ones from all the reviewers (including me). Not sure if the maintainers are open to arranging a conference call to walk through all of these?

Personally I have few outstanding small issues, whereas my main issue is that the coefficient-solver information is not propagated back to the user, including a potential convergence failure.

In particular, I think the `whatsnew` info is still missing. This depends on the intended release version, and I wonder if the MLFM addition warrants a "new feature" version bump to v0.10.0, which would likely queue it behind the pending v0.9.4 release.

adriesse:
> Personally I have few outstanding small issues, whereas my main issue is that the coefficient-solver information is not propagated back to the user, including a potential convergence failure.

This could be a more general question about how we would like fitting functions in pvlib to behave.  ```curve_fit``` will raise a warning or exception if things go terribly wrong.  Is that suitable and/or sufficient?  Admittedly the user will have to have a higher level of expertise to figure out what might be the problem.  If this is not good enough, then maybe we shouldn't be using ```curve_fit``` but rather some lower level function like ```least_squares``` or ```minimize```.

pvlib/mlfm.py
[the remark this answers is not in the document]
steve-ransome: > # NAME  value     # comment          unit       PV_LIB name
> 
> T_STC = 25.0      # STC temperature  [C]        temperature_ref
> G_STC = 1000.0    # STC irradiance   [W/m^2]
> 
> # not all yet used below , added here for completeness
> T_LIC = 25.0      # LIC temperature  [C]
> G_LIC = 200.0     # LIC irradiance   [W/m^2]
> 
> T_HTC = 75.0      # HTC temperature  [C]
> G_HTC = 1000.0    # HTC irradiance   [W/m^2]
> 
> T_PTC = 55.0      # HTC temperature  [C]
> G_PTC = 1000.0    # HTC irradiance   [W/m^2]
> 
> G_LTC = 500.0    # HTC irradiance   [W/m^2]
> T_LTC = 15.0      # LTC temperature  [C]
> 
> G_NOCT = 800      # NOCT irradiance  [W/m^2]
> T_NOCT = 45       # NOCT temperature [C]

pvlib/mlfm.py
steve-ransome: >     # calculate normalised loss coefficients
>     if 'i_sc' in dmeas.columns:
>         dnorm['i_sc'] = (dmeas['i_sc'] / ref['i_sc']
>                          / (dmeas['poa_global'] / G_STC))

pvlib/mlfm.py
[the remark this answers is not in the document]
steve-ransome: Note that some of these such as T_NOCT depend on the module type so I have added an example value of 45.
T_PTC is also a ballpark figure as it is taken at T_amb=20 and has to assume a noct. 
T_ptc ~ 20 + 1000/800 * (Noct -20) ~ 50C

https://sinovoltaics.com/learning-center/quality/ptc-pv-usa-test-conditions/

kandersolar:
This PR is in a sorry state.  Looks like some bad merges got in here somehow.  If this ever goes forward, I think it would be easiest to just start a fresh PR, so I'm closing this one.

FWIW I also think this feature is complicated enough that it would need a more experienced pvlib contributor to champion it.

cwhanse:
Been a while since I tried to help this one. My memory is that a primary (perhaps fatal) deficiency is a lack of clear references.

adriesse:
My impression was that the scope was too large and ill-defined.  Break it up and start with a small chunk would be my advice for any future PR on this.

PR #1378, opened 2022-01-15
Implement initial storage support

### LINKED ISSUE #1333: Add initial storage support
Abstract
---

The [storage support proposal for NumFOCUS Small Development Grants was recently accepted](https://groups.google.com/g/pvlib-python/c/ZuCHKhDwyso).

This issue aims to track progress of the development and discussions related to [the proposal](https://08916725177196330806.googlegroups.com/attach/558aad2b63ae0/pvlib%20-%20NumFOCUS%203rd%20round%20of%20Small%20Development%20Grants%20for%202021.pdf?part=0.1&view=1&vt=ANaJVrGQ_jQRA_73S2Tpc-QtowJ_LNuHdznLLKc6exF-tWz8jlmUZ3TYbPrE0lrzo73LnUkHfkAWCpUQyWYsYWFihUkJS-zhEQMJ21fZFyp3Q6g8aa0YDtk).

Goal
---

The goal of this project is to make a first step in supporting storage within pvlib. To do so, it
aims to complete at least:

- One use-case (PV for self-consumption)
- One configuration (i.e.: behind-the-meter)
- One technology (i.e.: lithium-ion batteries)
- One connection (i.e.: DC)
- One model (i.e.: NREL)
- One dispatch (i.e.: store all excess non-self-consumed energy)

Phases
---

1. Use case refinement phase
  a. Reading and understanding the state-of-the-art battery models and use cases
  b. Talking to stakeholders/experts to understand the most needed use case (PV installers, HW manufacturers and other professionals)
  c. Adjustment of the scope, first use case and detailed planning for the implementation phase
2. Implementation of the first prototype
  a. Public discussion with pvlib maintainers for an initial candidate API
  b. Implementation of the first functions and tests
  c. Validation of the first use case with well-known data
3. Integration
  a. Adjustments to the API for a ready-to-be-integrated version
  b. Project documentation update (API, storage, example code)
  c. Fix any issues or comments in the pull-request dis

Timeline
---

-  October 2021: Notification (no phase or work load associated)
- November 2021: Use case refinement phase
- December 2021 - January 2022: Implementation phase
- February-March 2022: Integration phase

Peque:
- [x] Closes #1333
 - [x] Tests added


williamhobbs:
In the documentation [here](https://pvlib-python--1378.org.readthedocs.build/en/1378/user_guide/storage.html#conventions), it notes that "Timestamps are associated to the beginning of the interval". My understanding is that the default convention for most of pvlib is that timestamps are instantaneous. If interval-averaged data with beginning of interval timestamps will be the convention for storage (which make sense to me), it might be helpful to clarify that this is different from other parts of pvlib.

mikofski:
@Peque would you consider showing your work on storage at the SciPy conference?
https://www.scipy2022.scipy.org/

Peque:
@williamhobbs Thanks for the comment! I will make sure it is more clearly explained. :blush: 

@mikofski Do you know if traveling is required for that conference?

mikofski:
> Do you know if traveling is required for that conference

moved this to discussions: https://github.com/pvlib/pvlib-python/discussions/1383

Peque:
@cwhanse @wholmgren If you could have a quick look at this initial proposal, I would love to know about your opinion! :blush: 

The idea is to implement other models/dispatching algorithms but, before moving forward to more complicated things, I would like to make sure this is aligned with pvlib. :rocket:

cwhanse:
@Peque this is a good start. What would you think about `pvlib.storage.battery` (battery efficiency/state-of-charge models), with `pvlib.storage.dispatch` (functions that take in a Series of PV power and return some kind of optimal dispatch)? Musing if it makes sense to group storage-related code, but I don't have much sense for the breadth of functions yet.

About power vs. energy, I think we can make a simple assumption that power is assumed to be constant in each time interval, which makes conversion to energy easy. I'm OK with a left-labeled time convention, and the positive means discharge convention.

I have two hesitations. Although the PR doesn't have financial calculations, I sense we may be headed that way if we start coding dispatch optimization. I'm pretty sure that we don't want to start building financial calculations into pvlib. That capability is quite mature in SAM, and I for one am not enthusiastic about redundant capability, nor to learn all that stuff.

I am not sure about `pvlib.flow` (`pvlib.powerflow` is better). I think we'll want to set some clear boundaries for this module to manage expectations for this module. For example, I don't think we want to include interfaces for power flow software such as Open DSS, that's a lot to maintain.

Peque:
@cwhanse Thanks for your feedback! I'll update the PR with your suggestions before moving forward and adding more code.

One question: should the "flow" module assume power flow or energy flow? Right now it is implemented as power flow (and SAM seems to be using that convention). That feels a bit weird to me (as opposed to energy flow), but if it is the most common convention, I'll leave it as-is.

I agree that financial calculations should not be the scope of pvlib. These would include things like payback, discount rate, IRR, replacement costs... However, I think price series could be supported as input of dispatch strategies. That means pvlib could support some kind of prize-based strategies. It could be as simple as "dispatch only when price is above X" or it could even be "dispatch only when price is above an optimal and automatically converged value". If the latter is something out-of-scope for pvlib (considering the optimization could be based on already-existing dependencies like SciPy), then I could maybe think about the interface (and add an example) on how to implement these types of dispatch strategies on your own. Any thoughts on this?

PS: not sure if the discussion should happen here or in the related issue (https://github.com/pvlib/pvlib-python/issues/1333). If the latter, then maybe we can copy these last 2 comments there. :blush:

janinefreeman:
We've gone with power flow rather than energy flow in SAM primarily because it is independent of time, which significantly simplifies calculations if you want to be able to model different timestep lengths.

Peque:
@cwhanse Where you able to take a look at the latest changes in this MR? (see https://github.com/pvlib/pvlib-python/issues/1333#issuecomment-1264761572) :blush:

mikofski:
Hi @Peque sorry for the delay - I have been meaning to go through your PR and respond to your comments. Without having reviewed it yet, I can't say much. Although I was a little concerned that you said you rewrote some of the inverter modeling. I was understand the need to reverse some of the models, but I was hoping that we could reuse code in pvlib if possible. I was also very much in favor of an agnostic rules or dispatch based model, rather than a self optimizing one that integrates an objective function as part of the package. My reasoning is that

1. the rules or dispatch based storage models have been documented in the literature, which is important in pvlib
2. optimization can be a secondary function that wraps the rules based approach, IMO this is another important philosophy of pvlib, is that it's a _toolbox_ not a polished solution. So we provide an implementation of the documented function, and then others, like your company, can take that function and wrap it in their own optimization routines.
3. Wrapping a rules or dispatch based storage method in optimization routine (like those from scipy.optimize) allows more flexibility in the criteria - for example some may wish to optimize the charging rate for full on shifting, vs. others may want to optimize the dispatch schedule to leverage TOD pricing. 

Given a rules or dispatch based storage function such as the one in SAM, then using scipy.optimize should be straight forward by defining an objective function, tolerances, boundaries, and starting guesses. EG:

```python

from pvlib.storage import pvbess
from scipy.optimize import minimize

DISPATCH_SCHEDULE = pd.DataFrame(index=pd.daterange(start, end), [True, True, True, ..., False, False, False, ...])
PVSYSTEM = pvlib.pvsystem(location=pvlib.location(lat, lon, elev, tz), inverter, module, ...)
TOD = pd.DataFrame(index=pd.daterange(start, end), [0.14, 0.14, 0.14, ..., 0.07, 0.07, 0.07, ...])

def lcoe(dispatch, tod, pvsystem, max_cap, charge_rate, discharge_rate, aux_loss, ...):
    energy = pvbess(pvsystem, dispatch, max_cap, charge_rate, discharge_rate, axu_loss, ...)
    return energy * tod  # revenue


def objective(dispatch, tod, pvsystem, max_cap, charge_rate, discharge_rate, aux_loss, ...):
    return -lcoe*lcoe

best_dispatch = minimize(objective, DISPATCH_SCHEDULE, args...)
```

Peque:
@mikofski Thanks for sharing your thoughts Mark! :blush: 

The current implementation always leaves the dispatch optimization to the user (i.e.: a dispatch series needs to be provided to run the simulation). Only, just like with AC-connected batteries, some constraints are enforced to make sure the energy flow is physically possible. These constraints may alter the original dispatch series.

Note that the new inverter model for DC-connected batteries is not defined in `inverter.py`, since it is meant to be an example for people wanting to try this non-official model. I reused as much code as possible, but it was the only way I found without having to modify code in `inverter.py` (which I think should be out-of-scope for this MR). The implementation is based on the current Sandia-multi inverter, and uses the same efficiency calculations. The battery is treated as a separate MPP, only working at a fixed voltage. The output of the function changes as well, since it makes a lot of sense to return a DataFrame instead of a Series, indicating the energy flow better and adding the clipping losses as well, which are affected by the use of a DC-connected battery.

mikofski:
Thanks @Peque for your patient explanations! The progress on this sounds great! How close to being complete? Will there be some presentation or write up to summarize the work to the community and NumFOCUS?

Peque:
@mikofski Thanks to you for your feedback an interest. :blush: 

I would consider this to be ready as-is since it fulfills the initial goals set for the project and seems to align well with the already-discussed expectations of keeping away the optimization algorithms and avoid touching the current inverter models (which would require further discussion).

As for NumFOCUS, I asked them and the only requirement is a very simple form (the [SDG Report Back Form](https://numfocus.typeform.com/to/ZC5YEN?typeform-source=numfocus.org)) that I am supposed to fill. You can [have a look at the answers I plan to submit](https://docs.google.com/document/d/1r3NMCeAVS6uMgxvRWxB1UXtv8UxoRX3K99W7NLrddYQ/edit#). Feel free to comment, specially in the last question, which I feel could be better answered by you (the pvlib maintainers).

As for further documentation, I think the best summary is included in the docs now. I do have some more information regarding the initial market research and meetings/discussions we had with the NREL team (regarding models and code reuse). I could share them if you want, but this are just notes. Also, some do not really apply. For example, after those meeting we decided to start with AC-connected batteries only and SAM's holistic model. The final result includes DC-connected batteries too and uses SAM's BatteryStateful, which adds more flexibility to support different period/frequency time series.

If there is interest, I could clean-up those notes to share them so that there is a document to understand "why", before the how/what (that is better described in the docs).

Also if/when this MR gets integrated, I could prepare a write up to summarize the work to the community (just in case there are further modifications before this gets integrated). :blush:

adriesse:
Could you 

> Updates entries to [docs/sphinx/source/api.rst](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes.

so that we can preview the doc strings?

AdamRJensen:
Closing this PR as this work is no longer being pursued.

Peque:
@AdamRJensen You mean that this is no longer pursued by pvlib? If there is something missing/pending on my side, please let me know. 😊

AdamRJensen:
> @AdamRJensen You mean that this is no longer pursued by pvlib? If there is something missing/pending on my side, please let me know. 😊

@Peque It was more in regard to there having been no action in +2 years. In regard to whether pvlib is interested in pursuing this, then I will make sure to bring it up for the next Core Team meeting (next meeting is in about two months).

*Update: The Core Team decided that pvlib will not support modeling energy storage as it is sufficiently far from pvlib's core capabilities and would be better handled by other packages. Also, no maintainers are actively engaged in storage modeling.*

adriesse:
> Closing this PR as this work is no longer being pursued.

It would be more polite to ask before closing.

PR #1391, opened 2022-01-25
Add Prilliman et al transience model to pvlib.temperature

### LINKED ISSUE #1081: implement transient moving-average thermal model
**New Feature**
Implement the [transient moving-average model proposed by Matt Prillman](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9095219)

a new function `pvlib.temperature.prillman(times, ...)`

see also #1080 

**Additional context**
see https://github.com/NREL/ssc/issues/261

kanderso-nrel:
- [x] Closes #1081
 - [x] Tests added

This PR adds a cleaned-up version of `prilliman_v5` from this notebook: https://gist.github.com/kanderso-nrel/1d6da384d7af8afc24c230f1f144eb57

Also: I noticed that one of the links in the PR template is out of date so I updated that here too.

### DIFF
--- a/pvlib/temperature.py
+++ b/pvlib/temperature.py
@@ -904,7 +904,7 @@ def prilliman(temp_cell, wind_speed, unit_mass=11.1, coefficients=None):
 
     wind_speed = wind_speed.values
     P = a[0] + a[1]*wind_speed + a[2]*unit_mass + a[3]*wind_speed*unit_mass
-    timedeltas = np.arange(window, 0, -1) * (time_step*60)  # s to min
+    timedeltas = np.arange(window, 0, -1) * (time_step*60)  # min to s
     weights = np.exp(-P[:, np.newaxis] * timedeltas)
 
     # set weights corresponding to the prefix values to zero; otherwise the
--- a/pvlib/temperature.py
+++ b/pvlib/temperature.py
@@ -869,6 +869,10 @@ def prilliman(temp_cell, wind_speed, unit_mass=11.1, coefficients=None):
     value. At the beginning of the series where a full 20 minute window is not
     possible, partial windows are used instead.
 
+    Output ``temp_cell[k]`` is NaN when input ``wind_speed[k]`` is NaN, or
+    when no non-NaN data are in the input temperature for the 20 minute window
+    preceding index ``k``.
+
     References
     ----------
     .. [1] M. Prilliman, J. S. Stein, D. Riley and G. Tamizhmani,
--- a/pvlib/temperature.py
+++ b/pvlib/temperature.py
@@ -7,6 +7,8 @@ import numpy as np
 import pandas as pd
 from pvlib.tools import sind
 from pvlib._deprecation import warn_deprecated
+import scipy
+
 
 TEMPERATURE_MODEL_PARAMETERS = {
     'sapm': {
@@ -821,3 +823,98 @@ def noct_sam(poa_global, temp_air, wind_speed, noct, module_efficiency,
     heat_loss = 1 - module_efficiency / tau_alpha
     wind_loss = 9.5 / (5.7 + 3.8 * wind_adj)
     return temp_air + cell_temp_init * heat_loss * wind_loss
+
+
+def prilliman(temp_cell, wind_speed, unit_mass=11.1, coefficients=None):
+    """
+    Smooth out short-term model transience using the Prilliman model [1]_.
+
+    The Prilliman et al. model applies an exponential moving average to
+    the output of a steady-state cell temperature model to account for a
+    module's thermal inertia and smooth out the cell temperature's response
+    to changing weather conditions.
+
+    .. warning::
+        This implementation requires the time series inputs to be regularly
+        sampled in time.  Data with irregular time steps should be resampled
+        prior to using this function.
+
+    Parameters
+    ----------
+    temp_cell : pandas Series
+        Cell temperature modeled with steady-state assumptions [C]
+
+    wind_speed : pandas Series
+        Wind speed, adjusted to correspond to array height [m/s]
+
+    unit_mass : float, default 11.1
+        Total mass of module divided by its one-sided surface area [kg/m^2]
+
+    coefficients : 4-element list-like, optional
+        Values for coefficients a_0–a_3 from [1]_
+
+    Returns
+    -------
+    temp_cell : pandas Series
+        Smoothed version of the input cell temperature [C]
+
+    Notes
+    -----
+    This smoothing model was developed and validated using the SAPM
+    model for the steady-state input.
+
+    References
+    ----------
+    .. [1] M. Prilliman, J. S. Stein, D. Riley and G. Tamizhmani,
+       "Transient Weighted Moving-Average Model of Photovoltaic Module
+       Back-Surface Temperature," IEEE Journal of Photovoltaics, 2020.
+       :doi:`10.1109/JPHOTOV.2020.2992351`
+    """
+
+    # TODO: check inputs to ensure regular spacing?
+
+    wind_speed = wind_speed.values
+    time_step = (temp_cell.index[1] - temp_cell.index[0]).total_seconds()
+    if time_step >= 1200:
+        # too coarsely sampled for smoothing to be relevant
+        return temp_cell
+        
+    window = int(1200 / time_step)
+
+    # prefix with NaNs so that the rolling window is "full",
+    # even for the first actual value:
+    prefix = np.full(window, np.nan)
+    temp_cell_prefixed = np.append(prefix, temp_cell.values)
+
+    # get one row per 20-minute window
+    H = scipy.linalg.hankel(np.arange(window),
+                            np.arange(window - 1, len(temp_cell_prefixed)))
+    subsets = temp_cell_prefixed[H].T
+
+    # calculate weights for the values in each window
+    if coefficients is None:
+        a = coefficients
+    else:
+        # values from [1], Table II
+        a = [0.0046, 0.00046, -0.00023, -1.6e-5]
+
+    P = a[0] + a[1]*wind_speed + a[2]*unit_mass + a[3]*wind_speed*unit_mass
+    timedeltas = np.arange(window, 0, -1) * time_step
+    weights = np.exp(-P[:, np.newaxis] * timedeltas)
+
+    # set weights corresponding to the prefix values to zero; otherwise the
+    # denominator of the weighted average below would be wrong
+    mask_idx = np.triu_indices(window)
+    np.fliplr(weights)[mask_idx] = 0
+
+    # change the first row of weights from zero to nan -- this is a
+    # trick to prevent div by zero warning on the next line
+    weights[0, :] = np.nan
+
+    # finally, take the weighted average of each window
+    numerator = np.nansum(subsets[:-1] * weights, axis=1)
+    denominator = np.sum(weights, axis=1)
+    smoothed = numerator / denominator
+    smoothed[0] = temp_cell.values[0]
+    smoothed = pd.Series(smoothed, index=temp_cell.index)
+    return smoothed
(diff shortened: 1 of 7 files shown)


kanderso-nrel:
Ready for review.  

@mjprilliman comments are welcome :)

cwhanse (commented):
Haven't looked at the test details yet

pvlib/temperature.py
```diff
@@ -900,7 +904,7 @@ def prilliman(temp_cell, wind_speed, unit_mass=11.1, coefficients=None):
 
     wind_speed = wind_speed.values
     P = a[0] + a[1]*wind_speed + a[2]*unit_mass + a[3]*wind_speed*unit_mass
-    timedeltas = np.arange(window, 0, -1) * time_step
+    timedeltas = np.arange(window, 0, -1) * (time_step*60)  # s to min
```
adriesse, line 907: If `timestep` is in seconds, then you would have to divide by 60 to get minutes...

kanderso-nrel:
```suggestion
    timedeltas = np.arange(window, 0, -1) * (time_step*60)  # min to s
```

wholmgren:
Great to see this. I'll review soon.

Reproducing something like Figure 5 from the reference would be a great addition to the example library.

prilliman figure 5 winter

pvlib/temperature.py
```diff
@@ -821,3 +825,101 @@ def noct_sam(poa_global, temp_air, wind_speed, noct, module_efficiency,
...
+        a = coefficients
+    else:
+        # values from [1], Table II
+        a = [0.0046, 0.00046, -0.00023, -1.6e-5]
+
+    wind_speed = wind_speed.values
+    P = a[0] + a[1]*wind_speed + a[2]*unit_mass + a[3]*wind_speed*unit_mass
+    timedeltas = np.arange(window, 0, -1) * (time_step*60)  # min to s
+    weights = np.exp(-P[:, np.newaxis] * timedeltas)
+
```
wholmgren, line 909: The array manipulation below is very clever, but will fly over most people's heads (well, at least I had to look up the function examples). I don't think the code needs to change, and I don't have a great idea for how to make it more understandable, and it requires understanding the hankel matrix too, nor do I think it's your job to teach numpy here, but any further comments below might help. So yeah, tough spot, sorry.

pvlib/temperature.py
```diff
@@ -821,3 +825,101 @@ def noct_sam(poa_global, temp_air, wind_speed, noct, module_efficiency,
...
+    window = min(window,          # time series > 20 minutes total
+                 len(temp_cell))  # time series < 20 minutes total
+
+    # prefix with NaNs so that the rolling window is "full",
+    # even for the first actual value:
+    prefix = np.full(window, np.nan)
+    temp_cell_prefixed = np.append(prefix, temp_cell.values)
+
+    # get one row per 20-minute window
```
wholmgren, line 893: I like this comment in `clearsky.py` "# generate matrix of integers for creating windows with indexing"

pvlib/temperature.py
```diff
@@ -821,3 +825,101 @@ def noct_sam(poa_global, temp_air, wind_speed, noct, module_efficiency,
...
+    mask_idx = np.triu_indices(window)
+    np.fliplr(weights)[mask_idx] = 0
+
+    # change the first row of weights from zero to nan -- this is a
+    # trick to prevent div by zero warning when dividing by summed weights
+    weights[0, :] = np.nan
+
+    # finally, take the weighted average of each window
+    numerator = np.nansum(subsets[:-1] * weights, axis=1)
+    denominator = np.sum(weights, axis=1)
```
wholmgren, line 921: why nansum above but sum here?

wholmgren (commented):
Looks good! All minor (hopefully) comments below.

mjprilliman (commented):
I'm not familiar enough with pvlib to give a full review but the function implementation looks good to me. Thanks @kanderso-nrel !

pvlib/temperature.py
```diff
@@ -821,3 +825,101 @@ def noct_sam(poa_global, temp_air, wind_speed, noct, module_efficiency,
...
+
+    if time_step >= 20:
+        warnings.warn("temperature.prilliman only applies smoothing when "
+                      "the sampling interval is shorter than 20 minutes "
+                      f"(input sampling interval: {time_step} minutes)")
+        # too coarsely sampled for smoothing to be relevant
+        return temp_cell
+
+    window = min(window,          # time series > 20 minutes total
+                 len(temp_cell))  # time series < 20 minutes total
```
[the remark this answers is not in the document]
kanderso-nrel: Indeed I took that behavior from @mjprilliman's example code in #1081, but I think using whatever is available is also consistent with Equation 11 in the reference.  There is some relevant discussion in this comment and the ones following: https://github.com/pvlib/pvlib-python/issues/1081#issuecomment-1019593000

Either way I am +1 to adding a docstring note about the partial windows at the beginning :+1:

pvlib/temperature.py
```diff
@@ -821,3 +825,101 @@ def noct_sam(poa_global, temp_air, wind_speed, noct, module_efficiency,
     heat_loss = 1 - module_efficiency / tau_alpha
     wind_loss = 9.5 / (5.7 + 3.8 * wind_adj)
     return temp_air + cell_temp_init * heat_loss * wind_loss
+
+
+def prilliman(temp_cell, wind_speed, unit_mass=11.1, coefficients=None):
+    """
+    Smooth short-term cell temperature transients using the Prilliman model.
+
+    The Prilliman et al. model [1]_ applies a weighted moving average to
```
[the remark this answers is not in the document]
cwhanse: I'm -1 on using the term "exponentially weighted" because that implies that the weight are of the form exp(k * delta_t) where k is a constant. In this model, k=k(t) is a function of t (P in Eq. 11 changes at each time step).

pvlib/tests/test_temperature.py
```diff
@@ -293,3 +295,59 @@ def test_noct_sam_options():
...
+    wind_speed = pd.Series([0, 1, 2])
+    msg = re.escape("temperature.prilliman only applies smoothing when the "
+                    "sampling interval is shorter than 20 minutes (input "
+                    "sampling interval: 30.0 minutes)")
+    with pytest.warns(UserWarning, match=msg):
+        actual = temperature.prilliman(cell_temperature, wind_speed)
+    assert_series_equal(cell_temperature, actual)
+
+
+def test_prilliman_nans():
```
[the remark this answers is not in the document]
kanderso-nrel: My idea was that each calculation operates on a single value of wind speed and many values of temperature, so a nan wind speed is a deal breaker but a nan temperature is ok so long as there are some non-nan temperatures in the window.  The reference indicates that gaps in the data are acceptable, which I think is functionally equivalent to ignoring nans.  I'm open to switching to nan output if any of the input temperature is nan, but to me that seems less consistent with the reference.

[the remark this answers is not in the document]
wholmgren: Thanks, now I agree with you. An `Examples` section of the docstring might help prevent user confusion, but I don't want to require that for this PR.

kanderso-nrel:
Thanks @wholmgren for the comments, was just lazy of me to not document that numpy magic.  I came up with it and still had to remind myself how it worked!

wholmgren (approved):
Any further comments or should we merge?

cwhanse:
I'd like to take one more look.

pvlib/temperature.py
```diff
@@ -821,3 +825,148 @@ def noct_sam(poa_global, temp_air, wind_speed, noct, module_efficiency,
...
+        Smoothed version of the input cell temperature [C]
+
+    Notes
+    -----
+    This smoothing model was developed and validated using the SAPM
+    cell temperature model for the steady-state input.
+
+    At the beginning of the series where a full 20 minute window is not
+    possible, "partial" windows including whatever values are available
+    is used instead.
```
cwhanse, line 870:
```suggestion
    Smoothing is done using the 20 minute window behind each temperature
    value. At the beginning of the series where a full 20 minute window is not
    possible, partial windows are used instead.
```

cwhanse (approved):
You win some unnamed award for excellent in-line comments

kanderso-nrel:
Actually I think there may be a flaw in this implementation regarding treatment of nans.  Please do not merge yet!

kanderso-nrel:
Ok, fixed.  The problem was that NaNs in the input temperature were correctly ignored in the numerator (`np.nansum(subsets * weights, axis=1)`), but the corresponding weights were not set to zero in the denominator, so in effect it was treating NaN temperature as zero temperature instead of ignoring it.  The fix was to replace the `triu` incantation, which properly ignored NaNs for the "prefix", with a simple mask that ignores all NaNs, not just the prefix.  So the code is simpler now too.

I don't think this should be controversial but it's probably worth another pair of eyes.  Thanks!

pvlib/temperature.py
```diff
@@ -821,3 +825,149 @@ def noct_sam(poa_global, temp_air, wind_speed, noct, module_efficiency,
...
+    Notes
+    -----
+    This smoothing model was developed and validated using the SAPM
+    cell temperature model for the steady-state input.
+
+    Smoothing is done using the 20 minute window behind each temperature
+    value. At the beginning of the series where a full 20 minute window is not
+    possible, partial windows are used instead.
+
+    References
```
cwhanse, line 872: One more suggestion: explain in Notes how `nan` in inputs corresponds to `nan` in output. It makes sense (to me) that output at `k` is `nan` if `windspeed[k]` is `nan`. As a user, I may be surprised non-`nan` at position `k` in output when input `temp_cell[k]` is `nan`. But that also makes sense, because the non-`nan` value can be produced when window of `temp_cell` prior to `k` has data.

```
Output ``temp_cell[k]`` is NaN when input ``wind_speed[k]`` is NaN, or when no non-NaN data are
in the input temperature for the 20 minute window preceding index ``k``.
```

adriesse:
> You win some unnamed award for excellent in-line comments

I had noticed these also, with appreciation.

cwhanse:
If not objections, will merge tomorrow.

pvlib/temperature.py
```diff
@@ -821,3 +825,155 @@ def noct_sam(poa_global, temp_air, wind_speed, noct, module_efficiency,
...
+    # change the first row of weights from zero to nan -- this is a
+    # trick to prevent div by zero warning when dividing by summed weights
+    weights[0, :] = np.nan
+
+    # finally, take the weighted average of each window:
+    # use np.nansum for numerator to ignore nans in input temperature, but
+    # np.sum for denominator to propagate nans in input wind speed.
+    numerator = np.nansum(subsets * weights, axis=1)
+    denominator = np.sum(weights, axis=1)
+    smoothed = numerator / denominator
```
wholmgren, line 976: @kanderso-nrel this line emitted a warning despite the comment on lines 967-969 that suggested otherwise. Is this a version-specific issue? My versions are below...

```
INSTALLED VERSIONS
------------------
commit           : 66e3805b8cabe977f40c05259cc3fcf7ead5687d
python           : 3.9.7.final.0
python-bits      : 64
OS               : Windows
OS-release       : 10
Version          : 10.0.18363
machine          : AMD64
processor        : Intel64 Family 6 Model 140 Stepping 1, GenuineIntel
byteorder        : little
LC_ALL           : None
LANG             : None
LOCALE           : English_United States.1252

pandas           : 1.3.5
numpy            : 1.20.3
```

kanderso-nrel: Hmm, the only way I can reproduce this is if there is a stretch of NaN in the input lasting longer than 20 minutes -- is that true for your case?  That comment is referring to a div by zero warning cause by a quirk of this implementation (no weights for the very first value), not a div by zero warning caused by nans in the input.

wholmgren: Ah, thanks. Yes, I had 38 minutes of NaN temperatures. I filled only the middle point with a valid number and the warning disappeared.

PR #1394, opened 2022-01-26
gallery example for bifacial modeling with pvfactors

spaneja:
- ~[ ] Closes #xxxx~
 - ~[ ] Tests added~

The gallery does not currently have any examples for bifacial modeling. After looking into the issue, it appears that it is somewhat clunky to perform bifacial modeling with the modelchain. As such, I have created two examples of bifacial modeling, with the pvfactors approach, and would like the community to weigh in. One example uses the modelchain, while another is procedural. I will wait for feedback on which approach is more suited for the gallery.

At this time, much of the documentation inside the examples is not complete or is copy/pasted from another example. I will clean and polish this when a decision is made on modeling approach.


kanderso-nrel:
Thanks @spaneja!  I agree it makes sense to decide what we want the examples to do before getting into the nitty gritty about code formatting and such.  

First though, the docs build is failing because ReadTheDocs doesn't currently install pvfactors before trying to run the examples.  Can you update this PR and add a `'pvfactors'` entry to this list in `setup.py`?  https://github.com/pvlib/pvlib-python/blob/master/setup.py#L59-L62

cwhanse:
Thanks @spaneja. I would vote to move forward.

Generally, I think the ModelChain approach is about as clean as pvlib v0.9.0 can support. This is an important example to add, which may also help us see how to deliver bifacial capabilities through ModelChain methods.

The procedural approach looks OK to me also, with the exception of the list structure for `irrad`; maybe that's a `pvfactors` issue but references like `irrad[0]` are rather opaque.

Ping me when you want to start working on commenting, etc.

spaneja:
Thanks @cwhanse and @kanderso-nrel. I have gone ahead and cleaned up and edited the submissions here. Please take a look for your approval and let me know of any feedback that you have.

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,96 @@
...
+axis_tilt = 0
+axis_azimuth = 180
+gcr = 0.35
+max_angle = 60
+pvrow_height = 3
+pvrow_width = 4
+albedo = 0.2
+
+# load temperature parameters and module/inverter specifications
+temp_model_parameters = PARAMS['sapm']['open_rack_glass_polymer']
```
kanderso-nrel, line 36:
```suggestion
temp_model_parameters = PARAMS['sapm']['open_rack_glass_glass']
```

The glass-glass coeffs are probably more appropriate for bifacial mods, though I'll defer to the Sandia folks on that

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,96 @@
...
+sandia_modules = pvsystem.retrieve_sam('SandiaMod')
```
kanderso-nrel, line 37: I think examples should favor the CEC module database over the Sandia module database.  CEC is much more up to date and more likely to have what users are after.  Though maybe I'll defer to the Sandia folks again on that ;)

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,96 @@
...
+bifaciality = 0.75
+irrad['effective_irradiance'] = (
+    irrad['total_abs_front'] + (irrad['total_abs_back'] * bifaciality)
+)
+
+# create modelchain object for bifacial system and run bifacial simulation
+mc_bifi = modelchain.ModelChain(system, site_location)
+mc_bifi.run_model_from_effective_irradiance(irrad)
+
+# plot results of both monofacial and bifacial
```
kanderso-nrel, line 94: There's no monofacial in this example :)

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,96 @@
+"""
+Bifacial Modeling - modelchain
```
kanderso-nrel, line 2: I think this example would benefit from having the code split into two smaller sections instead of one monolith of code.  The first code block would be what is interesting about this example (getting to the `effective_irradiance` variable) and the second would be more or less pvlib as usual (`array`, `system`, `mc_bifi`).

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,96 @@
...
+                                      pvrow_height,
+                                      pvrow_width,
+                                      albedo)
+
+# turn into pandas DataFrame
+irrad = pd.concat(irrad, axis=1)
+
+# define bifaciality coefficient (specific to module type being used)
+# create bifacial effective irradiance using aoi-corrected timeseries values
+bifaciality = 0.75
```
kanderso-nrel, line 85: Might make sense to put this up top with the other system specs

docs/examples/bifacial/plot_bifi_model_pvwatts.py
```diff
@@ -0,0 +1,107 @@
+"""
+Bifacial Modeling - procedural
+==============================
+
+Example of bifacial modeling using procedural method
+"""
+
+# %%
+# This example shows how to complete a bifacial modeling example using the
+# :py:class:`pvlib.modelchain.ModelChain` with the
```
kanderso-nrel, line 10: No ModelChain in this example :)

docs/examples/bifacial/plot_bifi_model_pvwatts.py
```diff
@@ -0,0 +1,107 @@
...
+
+# create location object and get clearsky data
+site_location = location.Location(lat, lon, tz=tz, name='Greensboro, NC')
+cs = site_location.get_clearsky(times)
+
+# get solar position data
+solar_position = solarposition.get_solarposition(cs.index,
+                                                 lat,
+                                                 lon
+                                                 )
```
kanderso-nrel, line 36: `site_location.get_solarposition(times)`?

docs/examples/bifacial/plot_bifi_model_pvwatts.py
```diff
@@ -0,0 +1,107 @@
...
+orientation = tracking.singleaxis(solar_position['apparent_zenith'],
+                                  solar_position['azimuth'],
+                                  max_angle=max_phi,
+                                  backtrack=True,
+                                  gcr=gcr
+                                  )
+
+# set axis_azimuth, albedo, pvrow width and height, and use
+# the pvfactors  engine for both front and rear-side absorbed irradiance
+axis_azmuth = 180
```
kanderso-nrel, line 51:
```suggestion
axis_azimuth = 180
```

docs/examples/bifacial/plot_bifi_model_pvwatts.py
```diff
@@ -0,0 +1,107 @@
...
+                                      gcr,
+                                      pvrow_height,
+                                      pvrow_width,
+                                      albedo
+                                      )
+
+# using bifaciality factor and pvfactors results, create effective
+# irradiance data
+bifaciality = 0.75
+effective_irrad_bifi = irrad[2] + (irrad[3] * bifaciality)
```
kanderso-nrel, line 72: I agree with @cwhanse that `irrad[i]` is opaque.  Can we turn it into a dataframe like the ModelChain example does?

docs/examples/bifacial/plot_bifi_model_pvwatts.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+# get cell temperature using the Faiman model
+temp_cell = temperature.faiman(irrad[0], 25, 1)
```
kanderso-nrel, line 76:
```suggestion
temp_cell = temperature.faiman(irrad[0], temp_air=25, wind_speed=1)
```

Best to avoid unexplained magic numbers

kanderso-nrel: Also, question: should cell temperature be calculated using total incident irradiance or only front incident irradiance?  The ModelChain example doesn't know the difference and is using total.

If total, I suppose cell temperature should be recalculated for the monofacial comparison farther down.

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,96 @@
...
+==============================
+
+Example of bifacial modeling using the modelchain
+"""
+
+# %%
+# This example shows how to complete a bifacial modeling example using the
+# :py:class:`pvlib.modelchain.ModelChain` with the
+# :py:func:`pvlib.bifacial.pvfactors_timeseries` function to transpose
+# GHI data to both front and rear Plane of Array (POA) irradiance.
```
kanderso-nrel, line 12: Probably a good idea to explain why we have to mix procedural and ModelChain code here.  Maybe something like this?

    # Unfortunately ``ModelChain`` does not yet support bifacial simulation
    # directly so we have to do the bifacial irradiance simulation ourselves.
    # Once the combined front + rear irradiance is known, we can pass that
    # to ``ModelChain`` and proceed as usual.
    #
    # Future versions of pvlib may make it easier to do bifacial modeling
    # with ``ModelChain``.

kanderso-nrel (commented):
Thanks @spaneja, here are some notes.

docs/examples/bifacial/plot_bifi_model_mc.py
cwhanse: 100% agree with @kanderso-nrel here

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,109 @@
...
+                       module_parameters=cec_module,
+                       temperature_model_parameters=temp_model_parameters)
+
+# create system object
+system = pvsystem.PVSystem(arrays=[array],
+                           inverter_parameters=cec_inverter)
+
+# create modelchain object for bifacial system and run bifacial simulation
+mc_bifi = modelchain.ModelChain(system, site_location, aoi_model='no_loss')
+mc_bifi.run_model_from_effective_irradiance(irrad)
```
kanderso-nrel, line 105: It's a small inelegance that specifying `aoi_model` is necessary for a ModelChain that only uses `run_model_from_effective_irradiance`.  I don't see a way around it, so I don't think there's anything to be done in this PR, just thinking out loud about the ModelChain design.

kanderso-nrel:
Referring to #1398, we should think about the choice of `n_pvrows` in these two examples.  I think it would be a good idea to explicitly specify `n_pvrows=something` (so that at least the choice is explicit in the code instead of implicit via the default value) and leave a comment briefly explaining why the user may want to choose a different value.

docs/examples/bifacial/plot_bifi_model_mc.py
cwhanse: Could default to `None` but then we'd need to check for a valid model in other `run_model` methods.

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,108 @@
...
+array = pvsystem.Array(mount=sat_mount,
+                       module_parameters=cec_module,
+                       temperature_model_parameters=temp_model_parameters)
+
+# create system object
+system = pvsystem.PVSystem(arrays=[array],
+                           inverter_parameters=cec_inverter)
+
+# create modelchain object for bifacial system and run bifacial simulation
+mc_bifi = modelchain.ModelChain(system, site_location, aoi_model='no_loss')
```
cwhanse, line 103: Add a comment explaining why `aoi_model='no_loss'` is needed, perhaps

```
# ModelChain requires the parameter aoi_loss to have a value. pvfactors
# applies surface reflection models in the calculation of front and back
# irradiance, so assign aoi_model='no_loss' to avoid double counting
# reflections.

File: docs/examples/bifacial/plot_bifi_model_pvwatts.py [hunk 14]
```diff
@@ -0,0 +1,111 @@
...
+max_phi = 60
+orientation = tracking.singleaxis(solar_position['apparent_zenith'],
+                                  solar_position['azimuth'],
+                                  max_angle=max_phi,
+                                  backtrack=True,
+                                  gcr=gcr
+                                  )
+
+# set axis_azimuth, albedo, pvrow width and height, and use
+# the pvfactors  engine for both front and rear-side absorbed irradiance
```
cwhanse, line 46:
```suggestion
# the pvfactors engine for both front and rear-side absorbed irradiance
```

docs/examples/bifacial/plot_bifi_model_pvwatts.py
```diff
@@ -0,0 +1,111 @@
...
+# irradiance data
+bifaciality = 0.75
+effective_irrad_bifi = irrad['total_abs_front'] + (irrad['total_abs_back']
+                                                   * bifaciality)
+
+
+# get cell temperature using the Faiman model
+# it is worth noting that this example uses total frontside incident
+# irradiance, however, for bifacial modeling, total irradiance (inclusive of
+# front and rear side) may be needed.
```
cwhanse, line 78: The wording of this comment undercuts the example a bit. To avoid the question, it may be better to input `effective_irrad_bifi` rather than `irrad['total_inc_front']`. AFAIK, there is no cell temperature model specific to bifacial modules. [This paper](https://ieeexplore.ieee.org/document/8366045) found that bifacial and monofacial module temperatures are predicted equally well by one of the available cell temperature models, so it's not a stretch to assume that Faiman applies also to bifacial models.

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,108 @@
...
+                                            max_angle=max_angle,
+                                            backtrack=True,
+                                            gcr=gcr)
+
+# created for use in pvfactors timeseries
+orientation = sat_mount.get_orientation(solar_position['apparent_zenith'],
+                                        solar_position['azimuth'])
+
+# get rear and front side irradiance from pvfactors transposition engine
+irrad = bifacial.pvfactors_timeseries(solar_position['azimuth'],
```
cwhanse, line 68: Is it worth adding some kind of warning suppression here? On execution, ShapelyDeprecationWarning and RuntimeWarning are raised, by shapely (pfactors dependency) and pvfactors, respectively. It clutters up the example a bit.

docs/examples/bifacial/plot_bifi_model_pvwatts.py
```diff
@@ -0,0 +1,111 @@
...
+                                  gcr=gcr
+                                  )
+
+# set axis_azimuth, albedo, pvrow width and height, and use
+# the pvfactors  engine for both front and rear-side absorbed irradiance
+axis_azimuth = 180
+pvrow_height = 3
+pvrow_width = 4
+albedo = 0.2
+irrad = bifacial.pvfactors_timeseries(solar_position['azimuth'],
```
cwhanse, line 51: Same comment about suppressing warnings, although I do not understand why only the RuntimeWarnings are raised here.

kanderso-nrel: I think this is because the Shapely warnings happen when python imports pvfactors, not because of any calculations.  That, plus the fact that all gallery examples run in the same python process (different namespaces but shared `sys.modules`), means that pvfactors is only actually imported once during the docs build, so any warning emitted during the import will only appear in whatever example happens to import it first.  

+1 to suppressing the Shapely warnings, and I wouldn't object to suppressing the runtime warnings as well.

campanelli-sunpower:
PSA: I have a "pre-release" of pvfactors v1.5.2 up at https://github.com/SunPower/pvfactors/releases/tag/v1.5.2. All that this left is to name it and for me to get cred's to upload to PyPI, see https://github.com/SunPower/pvfactors/issues/137.

docs/examples/bifacial/plot_bifi_model_mc.py
```diff
@@ -0,0 +1,120 @@
+"""
+Bifacial Modeling - modelchain
+==============================
+
+Example of bifacial modeling using the modelchain
```
cwhanse, line 5:
```suggestion
Example of bifacial modeling using pvfactors and ModelChain
```

docs/examples/bifacial/plot_bifi_model_mc.py
cwhanse: Converted to #1411

docs/examples/bifacial/plot_bifi_model_pvwatts.py
```diff
@@ -0,0 +1,115 @@
+"""
+Bifacial Modeling - procedural
+==============================
+
+Example of bifacial modeling using procedural method
```
cwhanse, line 5:
```suggestion
Example of bifacial modeling using pvfactors and procedural method
```

docs/sphinx/source/whatsnew/v0.9.1.rst
```diff
@@ -53,7 +53,10 @@ Documentation
 ~~~~~~~~~~~~~
 * Fix documentation return error in :py:meth:`pvlib.forecast.ForecastModel.cloud_cover_to_transmittance_linear`
   (:issue:`1367`, :pull:`1370`)
-
+* Add gallery example illustrating bifacial simulation using :py:class:`pvlib.modelchain.ModelChain`
+  and the :py:func:`pvlib.bifacial.pvfactors_timeseries` function (:pull:`1394`)
+* Add gallery example illustrating bifacial simulation using procedural functions
+  and the :py:func:`pvlib.bifacial.pvfactors_timeseries` function (:pull:`1394`)
```
cwhanse, line 59:
```suggestion
* Add gallery example illustrating bifacial simulation using the
  :py:func:`pvlib.bifacial.pvfactors_timeseries` function and either
   :py:class:`pvlib.modelchain.ModelChain` or procedural functions.
   (:pull:`1394`)
```

cwhanse (approved):
Thanks @spaneja

docs/sphinx/source/whatsnew/v0.9.1.rst
```diff
@@ -53,6 +53,10 @@ Documentation
 ~~~~~~~~~~~~~
 * Fix documentation return error in :py:meth:`pvlib.forecast.ForecastModel.cloud_cover_to_transmittance_linear`
   (:issue:`1367`, :pull:`1370`)
+* Add gallery example illustrating bifacial simulation using the
+  :py:func:`pvlib.bifacial.pvfactors_timeseries` function and either
+   :py:class:`pvlib.modelchain.ModelChain` or procedural functions.
+   (:pull:`1394`)
```
kanderso-nrel, line 59:
```suggestion
  :py:class:`pvlib.modelchain.ModelChain` or procedural functions.
  (:pull:`1394`)
```

sphinx rendering issue caused by indentation mismatch

kanderso-nrel:
Thanks @spaneja!

PR #1408, opened 2022-02-17
handle nan in tools.golden_sect_DataFrame

cwhanse:
- ~~[ ] Closes #xxxx~~
 - [x] Tests added
 - ~~[ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).~~

Supplements #1089. See [discussion](https://github.com/pvlib/pvlib-python/pull/1089/files/8b55eed569e41c258219d1b565e8d296516d542f#r809145242).


pvlib/tools.py
```diff
@@ -345,15 +349,18 @@ def _golden_sect_DataFrame(params, lower, upper, func, atol=1e-8):
...
         # but just to be safe
         iterations += 1
 
     if iterations > iterlimit:
-        raise Exception("iterations exceeded maximum")  # pragma: no cover
+        raise Exception("Iterations exceeded maximum. Check that func",
+                        " is not NaN in (lower, upper)")
 
-    return func(df, 'V1'), df['V1']
+    func_result = func(df, 'V1')
```
kanderso-nrel, line 361: This PR gets #1394 working again for me locally, so +1 for that.  However I think this will still fail with a KeyError if either of `upper` or `lower` is all nans: in that case `iterlimit` will still be nan, so the while loop will not execute at all, causing `df['V1']` to not exist when this line executes.  I think the all-nan case is worth supporting:  a ModelChain simulation of only nighttime is not particularly useful, but shouldn't throw an error IMHO.

kanderso-nrel (approved):
Not an expert about the context of this function but LGTM.  Stickler is complaining still.

mikofski:
I guess my preference long term would be to replace this with the [`SciPy Cython Optimize Zeros API`](https://docs.scipy.org/doc/scipy/reference/optimize.cython_optimize.html) `brentq()` function. Not having to maintain a math library seems preferable to me.

PR #1416, opened 2022-03-02
Improve get_cams documentation

soma2000-lang:
- [x] Closes #1412
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [x] Tests added
 - [x] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/reference) for API changes.

PR #1418, opened 2022-03-02
Update v0.9.1.rst #1416

soma2000-lang:
@cwhanse  Added myself to the contributors section as you sugested
fixes #1416


cwhanse:
@soma2000-lang could you add this change to #1416?

wholmgren:
I think with new contributors we should be forgiving about this kind of workflow mistake... easy enough for us to hit merge twice or for us to make the change in the original PR. Of course @soma2000-lang is welcome to make the change in #1416 too.

cwhanse:
We'll merge #1416 and this together. Thanks @soma2000-lang !

cwhanse:
Thanks @soma2000-lang !

PR #1450, opened 2022-04-26
Updated plot_ghi_transposition.py

RoyCoding8:
- [ ] Closes #xxxx
 - [ ] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [ ] Tests added
 - [ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/reference) for API changes.
 - [ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [ ] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [ ] Pull request is nearly complete and ready for detailed review.
 - [ ] Maintainer: Appropriate GitHub Labels and Milestone are assigned to the Pull Request and linked Issue.

PR #1451, opened 2022-04-26
Updated plot_ghi_transposition.py

RoyCoding8:
- [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/docs/examples/irradiance-transposition/plot_ghi_transposition.py
+++ b/docs/examples/irradiance-transposition/plot_ghi_transposition.py
@@ -19,7 +19,7 @@ from matplotlib import pyplot as plt
 
 # For this example, we will be using Golden, Colorado
 lat, lon = 39.755, -105.221
-tz='MST'
+tz = 'MST'
 
 # Create location object to store lat, lon, timezone
 site = location.Location(lat, lon, tz=tz)
--- a/docs/examples/irradiance-transposition/plot_ghi_transposition.py
+++ b/docs/examples/irradiance-transposition/plot_ghi_transposition.py
@@ -19,9 +19,10 @@ from matplotlib import pyplot as plt
 
 # For this example, we will be using Golden, Colorado
 lat, lon = 39.755, -105.221
+tz='MST'
 
 # Create location object to store lat, lon, timezone
-site = location.Location(lat, lon, tz='MST')
+site = location.Location(lat, lon, tz=tz)
 
 
 # Calculate clear-sky GHI and transpose to plane of array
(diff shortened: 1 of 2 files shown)


RoyCoding8 (commented):
Changed PR #1450 to #1451

docs/examples/irradiance-transposition/plot_ghi_transposition.py
```diff
@@ -18,11 +18,10 @@
 from matplotlib import pyplot as plt
 
 # For this example, we will be using Golden, Colorado
-tz = 'MST'
 lat, lon = 39.755, -105.221
 
 # Create location object to store lat, lon, timezone
-site = location.Location(lat, lon, tz=tz)
+site = location.Location(lat, lon, tz='MST')
```
mikofski, line 25: I actually liked it the way it was before when `tz` was a variable, defined together with lat and lon where it's easy to comment. IMHO moving it here makes the code more terse, less verbose, and less accessible. In examples especially, I think it's better to lean towards over-explaining and writing as clearly and verbosely as possible.

docs/sphinx/source/whatsnew/v0.9.1.rst
```diff
@@ -88,3 +89,4 @@ Contributors
 * Will Hobbs (:ghuser:`williamhobbs`)
 * Uday Guntupalli (:ghuser:`uguntupalli`)
 * Shashwata Roy (:ghuser:`RoyCoding8`)
+* Shashwata Roy (:ghuser: `RoyCoding8`)
```
mikofski, line 92: I don't think this second entry is needed, since it just repeats the line immediately before it. Also I don't think the extra space is necessary.

docs/examples/irradiance-transposition/plot_ghi_transposition.py
kanderso-nrel: I agree, I like the pattern of having configurations in a group at the top of the script so that people don't have to hunt through the code to make sure they didn't miss something if they want to model a different location.  

Also if we get rid of the variable for `tz`, why not do the same for `lat` and `lon`?

docs/sphinx/source/whatsnew/v0.9.1.rst
kanderso-nrel: It should be moved to v0.9.2.rst anyway since we've already released v0.9.1 :)

mikofski:
Hi @RoyCoding8 , I applaud your enthusiasm, but I don't feel like this PR adds value to the example. One way to judge how much value new changes add is to open an issue _before_ starting work on a PR. The issue will draw attention to the problem and encourage a discussion which involves the folks in the community who are most affected. There's a place in the PR template to refer to the issue that the PR closes. I think this process is the best way to contribute to pvlib. Thanks for your consideration.

docs/sphinx/source/whatsnew/v0.9.1.rst
RoyCoding8: Sure, I will move and remove the duplicate entry.

kanderso-nrel:
@RoyCoding8 yes, I think we don't plan to merge this so it makes sense to close this PR.  Thanks anyway!

PR #1454, opened 2022-04-27
CI asv check

### LINKED ISSUE #1446: A basic CI asv check would be useful
Currently there's no automated check to ensure that changes to the benchmarks are valid.  For example in #1443 and #1445 I think currently the best way to be sure that the new benchmarks are valid (i.e. they won't error when being run in the nightly job) is to checkout the PR locally and try it out manually.  It would be nice if this was done automatically somehow. 

An automated benchmark check would also prevent us from forgetting to update the benchmarks when we make breaking changes to pvlib itself. 

A new github actions workflow that builds the asv environments and executes the benchmarks at least once to ensure validity.  Note that I'm not suggesting that we actually use the timing results for anything: the goal is to verify that the benchmarks execute without error, not to detect performance regressions.  The latter will still be the nightly VM's responsibility. 

Running the benchmarks in earnest for PRs would also solve this, but that is still a complicated problem that I don't want to take on at this point.  I think this small step in that direction makes more sense for now. 

**Additional context**
`asv run --quick` seems to do what I want ([ref](https://asv.readthedocs.io/en/stable/commands.html#asv-run)):

> Do a "quick" run, where each benchmark function is run only once. This is useful to find basic errors in the benchmark functions faster. The results are unlikely to be useful, and thus are not saved.

`--strict` is probably also useful here, although see https://github.com/airspeed-velocity/asv/issues/1199

roger-lcc:
- [x] Closes #1446

A basic CI ASV CHECK, which is aimed to verify that the benchmarks execute without error.


.github/workflows/asv_check.yml
```diff
@@ -0,0 +1,39 @@
...
+          pip install ephem==3.7.6.0
+          pip install numba==0.40.0
+          pip install asv==0.4.2
+
+      - name: Run asv benchmarks
+        run: |
+          cd benchmarks
+          asv machine --yes
+          asv run --quick --dry-run --show-stderr --python=same\
+          | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
```
kanderso-nrel, line 35: I'm not sure the backslash works as a line continuation here.  I'd just put it all on one line instead of trying to figure out whatever weird YAML syntax is needed to make a multi-line command work.

.github/workflows/asv_check.yml
```diff
@@ -0,0 +1,39 @@
+name: CI ASV CHECK
```
kanderso-nrel, line 1:
```suggestion
name: asv
```

.github/workflows/asv_check.yml
```diff
@@ -0,0 +1,34 @@
...
+          python-version: '3.9.7'
+
+      - name: Install asv
+        run: pip install asv==0.4.2
+
+      - name: Run asv benchmarks
+        run: |
+          cd benchmarks
+          asv machine --yes
+          asv run --quick --dry-run --show-stderr | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
```
kanderso-nrel, line 30: Ah, it's talking about `master` because that's the default if you don't specify a specific commit or commit range.  Try using `HEAD^!` here (`asv run HEAD^! --quick ...`), which I think will tell asv to just run the currently checked-out commit.

roger-lcc: Yes, I think it's the key. I was always thinking about the `asv.conf` before this. 
Thanks for your valuable advice!

docs/sphinx/source/whatsnew/v0.9.2.rst
```diff
@@ -24,8 +24,10 @@ Benchmarking
 ~~~~~~~~~~~~~
 * Updated version of numba in asv.conf from 0.36.1 to 0.40.0 to solve numba/numpy conflict. (:issue:`1439`, :pull:`1440`)
 Requirements
+* Added a basic CI asv check(:issue:`1446`, :pull:`1454`)
 ~~~~~~~~~~~~
```
kanderso-nrel, line 28:
```suggestion
* Added a basic CI asv check (:issue:`1446`, :pull:`1454`)

Requirements
~~~~~~~~~~~~
```

kanderso-nrel (approved):
Minor formatting issue in the what'snew file, but otherwise I think I'm good with this.  I'll leave this PR open for a bit in case anyone else wants to take look before merging. 

Thanks again @roger-lcc!

.github/workflows/asv_check.yml
```diff
@@ -0,0 +1,34 @@
+name: asv
+
+# CI ASV CHECK is aimed to verify that the benchmarks execute without error.
+on: [pull_request, push]
```
wholmgren, line 4: @kanderso-nrel do we really need this on push? I was surprised to push a branch to my fork and then get a notification that the test was done.

kanderso-nrel: I thought it would be nice to run it on pushes to master. I didn't consider the effect on forks.  Fine with me to change it to something like this:
```yaml
on:
  push:
    branches:
    - master
  pull_request:
```

PR #1468, opened 2022-06-07
Townsend snow (building on #1251)

### LINKED ISSUE #1246: Implement monthly snow loss model

reepoi:
This builds off of pull request #1251 to add the changes suggested in the conversation there and wrap it up.
 - [x] Closes #1246
 - [x] Tests added


docs/sphinx/source/whatsnew/v0.9.2.rst
```diff
@@ -18,6 +18,8 @@ Bug fixes
   where passing localized timezones with large UTC offsets could return
   rise/set/transit times for the wrong day in recent versions of ``ephem``
   (:issue:`1449`, :pull:`1448`)
+* Added Townsend-Powers monthly snow loss model: :py:func:`pvlib.snow.loss_townsend`
+  (:issue:`1246`, :pull:`1251`, :pull:`1468`)
```
kanderso-nrel, line 22: Let's move this from `Bug fixes` to `Enhancements` :)

kanderso-nrel (commented):
Thanks @reepoi this is looking pretty good!  A few comments below.

pvlib/snow.py
```diff
@@ -185,3 +185,130 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    This model has not been validated for tracking arrays; however, for tracking
+    arrays [1]_ suggests using the maximum rotation angle in place of surface_tilt.
+
+    References
+    ----------
+    .. [1] Townsend, Tim & Powers, Loren. (2011). Photovoltaics and snow: An
+       update from two winters of measurements in the SIERRA. Conference
+       Record of the IEEE Photovoltaic Specialists Conference.
+       003231-003236. 10.1109/PVSC.2011.6186627.
+       Available at https://www.researchgate.net/publication/261042016_Photovoltaics_and_snow_An_update_from_two_winters_of_measurements_in_the_SIERRA
```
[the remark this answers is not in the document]
reepoi: Ok, I removed the links and the noqa tags under the docstrings.

docs/sphinx/source/whatsnew/v0.9.2.rst
reepoi: Ah, right, thank you.

pvlib/snow.py
```diff
@@ -185,3 +185,130 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+def loss_townsend(snow_total, snow_events, surface_tilt, relative_humidity,
+                  temp_air, poa_global, slant_height, lower_edge_drop_height,
+                  angle_of_repose=40):
+    '''
+    Calculates monthly snow loss based on the Townsend monthly snow loss model [1]_.
+
+    Parameters
+    ----------
+    snow_total : array-like
+        Inches of snow received each month. Referred to as S in the paper [in]
```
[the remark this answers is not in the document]
kanderso-nrel: @reepoi note that I think changing to metric requires more than just editing the docstring -- I think some conversions will be needed in the code as well, since the equations and coefficients in the reference assume non-metric inputs.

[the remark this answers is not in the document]
reepoi: Yes, the code does need to be changed, and here is a description of what I did. Also, I want to bring attention to the variable `loss_percentage` ([this line](https://github.com/pvlib/pvlib-python/blob/bbe5bc33f8e6fff0d0e405484a7bd70863cc6c45/pvlib/snow.py#L305)) and how it is affected by the unit conversion. I'll describe it more next.
Code changes for converting to metric:
- Method `_townsend_effective_snow`: This method's code was not changed. Here is the line calculates effective snow: `effective_snow = 0.5 * snow_total * (1 + 1 / snow_events_no_zeros)`. Let `X = 0.5 * (1 + 1 / snow_events_no_zeros)`, which is unitless. `effective_snow = X * snow_total` and `effective_snow` has the same units as `snow_total`.
- Method `loss_townsend`: This method's code was changed. The variable `gamma` is unitless in the paper, so it should remain unitless after changing to metric. Here is the formula for `gamma`:
  ```slant_height (in) * effective_snow_weighted (in) / (((drop_height_clipped (in))^2 - (effective_snow_weighted (in))^2) * 2 * tand(angle_of_repose)```
The (in)^2 cancel. Next change the inputs to metric. `effective_snow_weighted` is now in centimeters. Let `effective_snow_weighted_m` be `effective_snow_weighted` in meters. The formula for `gamma` becomes
```slant_height (m) * effective_snow_weighted_m (m) / (((drop_height_clipped (m))^2 - (effective_snow_weighted (m))^2) * 2 * tand(angle_of_repose)```
The (m)^2 cancel.
I compared the values of `gamma` calculated using the unit test data ([from this test](https://github.com/pvlib/pvlib-python/blob/bbe5bc33f8e6fff0d0e405484a7bd70863cc6c45/pvlib/tests/test_snow.py#L110)) in imperial units and in metric units, and they were the same. Note the data of the test I linked is in metric.
`ground_interference_term` is also unitless.

Next, changing `loss_percentage` to metric affects the result of `loss_townsend` because it is not unitless. The unit of the returned result `Loss` in the Townsend-Powers paper is `inches/(Kelvin)^2/(kWh/m2)^0.67`. From the unit conversions to metric (and changing `poa_global` from kWh/m2 to Wh/m2), `loss_percentage` has the unit `centimeters/(Kelvin)^2/(Wh/m2)^0.67`. Does either of these units make sense as a capacity loss fraction?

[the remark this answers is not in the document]
Comment (cwhanse) [reply pvlib/snow.py]: Eq.3 gives the loss percentage if I am reading this paper correctly. I would assume that the constant C1 has units that, when multiplied by the other factors, give a product with unit kWhr_DC (loss) / kWhr_DC (no snow) * 100%. C1 is the slope of the fitted line in Fig. 7, so the unit is the y-axis unit (kWhr (loss) / kWhr (no snow) * 100) divided by the bizarre unit on the x-axis.

Bottom line: I would convert `poa_global` from its input unit of Whr /m2 to kWh / m2 inside the function, and not change C1's value.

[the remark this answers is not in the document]
Comment (reepoi) [reply pvlib/snow.py]: Ah, I see. Since weighted effective snow (inches) is also in the unit of Figure 7's x-axis, we should also convert `effective_snow_weighted` from centimeters to inches in the `loss_percentage` calculation.

File: pvlib/snow.py [hunk 4]
```diff
@@ -185,3 +185,133 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    ground_interference_term = 1 - C2 * np.exp(-gamma)
+    temp_air_kelvin = temp_air + 273.15
+    effective_snow_weighted_in = effective_snow_weighted / 2.54
+    poa_global_kWh = poa_global / 1000
+    loss_percentage = (
+        C1
+        * effective_snow_weighted_in
+        * cosd(surface_tilt)**2
+        * ground_interference_term
+        * relative_humidity
```
kanderso-nrel, line 312: More units... do we need to divide humidity by 100 to convert input units (percent) to the units of Eqn 3 (dimensionless fraction)?

reepoi: Oh yes, I think you're right. I've pushed a commit with that change. I've also adjusted the unit test from the original author so that it passes, but I don't know where the original author got the unit test's data. I think next it would be nice to add a unit test with data that reproduces one of the results in Table 2 of the paper.

pvlib/snow.py
```diff
@@ -185,3 +185,134 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    Calculates effective snow using the total snowfall received each month and
+    the number of snowfall events each month.
+
+    Parameters
+    ----------
+    snow_total : array-like
+        Snow received each month. Referred to as S in [1]_ [cm]
+
+    snow_events : array-like
+        Number of snowfall events each month. Referred to as N in [1]_ [-]
```
cwhanse, line 201:
```suggestion
        Number of snowfall events each month. Referred to as N in [1]_. [-]
```

pvlib/snow.py
```diff
@@ -185,3 +185,134 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+
+def _townsend_effective_snow(snow_total, snow_events):
+    '''
+    Calculates effective snow using the total snowfall received each month and
+    the number of snowfall events each month.
+
+    Parameters
+    ----------
+    snow_total : array-like
+        Snow received each month. Referred to as S in [1]_ [cm]
```
cwhanse, line 198:
```suggestion
        Snow received each month. Referred to as S in [1]_. [cm]
```

pvlib/snow.py
```diff
@@ -185,3 +185,134 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    -------
+    effective_snowfall : array-like
+        Effective snowfall as defined in the Townsend model [cm]
+
+    References
+    ----------
+    .. [1] Townsend, Tim & Powers, Loren. (2011). Photovoltaics and snow: An
+       update from two winters of measurements in the SIERRA. Conference
+       Record of the IEEE Photovoltaic Specialists Conference.
+       003231-003236. :doi:`10.1109/PVSC.2011.6186627`
```
cwhanse, line 213:
```suggestion
       update from two winters of measurements in the SIERRA. 37th IEEE
       Photovoltaic Specialists Conference, Seattle, WA, USA.
       :doi:`10.1109/PVSC.2011.6186627`
```

pvlib/snow.py
```diff
@@ -185,3 +185,134 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    Notes
+    -----
+    This model has not been validated for tracking arrays; however, for
+    tracking arrays [1]_ suggests using the maximum rotation angle in place
+    of surface_tilt.
+
+    References
+    ----------
+    .. [1] Townsend, Tim & Powers, Loren. (2011). Photovoltaics and snow: An
+       update from two winters of measurements in the SIERRA. Conference
```
cwhanse, line 271: Copy reference from above function

pvlib/snow.py
```diff
@@ -185,3 +185,134 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+                  temp_air, poa_global, slant_height, lower_edge_drop_height,
+                  angle_of_repose=40):
+    '''
+    Calculates monthly snow loss based on the Townsend monthly snow loss
+    model [1]_.
+
+    Parameters
+    ----------
+    snow_total : array-like
+        Snow received each month. Referred to as S in [1]_ [cm]
```
cwhanse, line 230: Periods at the end of each description.

pvlib/snow.py
```diff
@@ -185,3 +185,134 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+        Monthly average ambient temperature [C]
+
+    poa_global : array-like
+        Monthly plane of array insolation [Wh/m2]
+
+    slant_height : float
+        Row length in the slanted plane of array dimension [m]
+
+    lower_edge_drop_height : float
+        Drop height from array edge to ground [m]
```
cwhanse, line 251: What about `lower_edge_height` instead? Defined as "Distance from array lower edge to the ground. [m]"

pvlib/snow.py
```diff
@@ -185,3 +185,134 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+
+
+def loss_townsend(snow_total, snow_events, surface_tilt, relative_humidity,
+                  temp_air, poa_global, slant_height, lower_edge_drop_height,
+                  angle_of_repose=40):
+    '''
+    Calculates monthly snow loss based on the Townsend monthly snow loss
+    model [1]_.
+
+    Parameters
```
cwhanse, line 227:
```suggestion
    Model is applicable to arrays on fixed mounting.

    Parameters
```

cwhanse (approved):
Editorial comments, thanks @reepoi

pvlib/snow.py
```diff
@@ -185,3 +185,134 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+    Returns
+    -------
+    loss : array-like
+        Monthly average DC capacity loss fraction due to snow coverage
+
+    Notes
+    -----
+    This model has not been validated for tracking arrays; however, for
+    tracking arrays [1]_ suggests using the maximum rotation angle in place
+    of surface_tilt.
```
cwhanse, line 266:
```suggestion
    of ``surface_tilt``.
```

pvlib/tests/test_snow.py
```diff
@@ -95,3 +95,36 @@ def test_dc_loss_nrel():
...
+                                  80, 80])
+    temp_air = np.array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
+    poa_global = np.array([350000, 350000, 350000, 350000, 350000, 350000,
+                           350000, 350000, 350000, 350000, 350000, 350000])
+    angle_of_repose = 40
+    slant_height = 2.54
+    lower_edge_height = 0.254
+    expected = np.array([0.0007696253, 0.0007992262, 0.0006216201,
+                         0.0001715392, 0, 0, 0, 0, 0, 0, 0.0002643821,
+                         0.0006068194])
```
kanderso-nrel, line 125: These expected losses seem a couple orders of magnitude too small for the inputs, so I suspect this implementation is not quite correct yet.  I'm not immediately seeing any remaining input unit issues, but I do notice that the y-axis in Fig 7 is fractional (not percentage) but has the same slope/C1 value, so maybe Eqn 3 is mislabeled in the reference and actually calculates fractional loss, not percentage loss?  If that's true, these values ought to be 100x what they are now, which would be a lot more reasonable.

pvlib/snow.py
```diff
@@ -185,3 +185,134 @@ def dc_loss_nrel(snow_coverage, num_strings):
...
+        C1
+        * effective_snow_weighted_in
+        * cosd(surface_tilt)**2
+        * ground_interference_term
+        * relative_humidity_fraction
+        / temp_air_kelvin**2
+        / poa_global_kWh**0.67
+    )
+
+    return loss_percentage / 100
```
kanderso-nrel, line 318: We might want to constrain the loss to [0, 1] just to be safe

kanderso-nrel (commented):
One or two things below that should be addressed before merging.  

A comment more about the model than this PR: the geometry of the ground interference term seem strange to me in that it's only accounting for half of what I would expect from a snow pile.  I expected the pile to be symmetrical (angle of repose on both sides), but I can only get the math to match the reference if one side of the pile is vertical.  Is the idea that the forward momentum of the shedded snow allows it to populate only the "forward" half of the pile?  To be explicit using the labels in my diagram below, I expected the dissipation area to comprise `A_1`, `A_2`, and their reflected counterparts on the other side of the "drop height" line, but it seems that Townsend & Powers don't count the reflected counterparts. 

image

cwhanse:
A communication with the model author T. Townsend: "the pile of snow was modeled as a triangle atop a rectangle, with yes, a vertical wall on the "north" side of the pile. This assumption, while an engineering approximation, is based on observing the shape of shedded snow piles, piles that are shaped by the slight forward (i.e., southward) velocity of the sliding/falling snow and ice."

pvlib/tests/test_snow.py
reepoi: loss_townsend now takes its Eqn 3 calculation as a fractional loss, and no longer divides it by 100. I also added a comment to explain why we think Eqn 3 calculates a fractional loss.

cwhanse (approved):
LGTM!

kanderso-nrel (approved):
Thanks @abhisheksparikh for getting this started and @reepoi for finishing it!  Great to see this merged.

PR #1490, opened 2022-07-11
Improve nonuniform timestamps error message for prilliman and detect_clearsky

kanderso-nrel:
- [x] Closes #1476
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - ~[ ] Tests added~
 - ~[ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/reference) for API changes.~
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - ~[ ] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.~
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

@williamhobbs, does this change adequately address #1476?

PR #1493, opened 2022-07-13
Update CI - Micromamba

### LINKED ISSUE #1472: Look into using mamba instead of conda for CI jobs
One step in our pytest CI jobs is setting up a conda environment.  Setting up that environment can be a large fraction of total job execution time.  Here's an [example job run](https://github.com/pvlib/pvlib-python/runs/6886771043?check_suite_focus=true) where 5 out of the total 8 minutes were spent setting up the conda environment.  It would be nice to speed this up somehow; faster jobs means less time spent twiddling thumbs waiting for tests to pass and are a more responsible use of the free CI services GitHub provides to us.

[Mamba](https://mamba.readthedocs.io/en/latest/) is supposedly a faster drop-in replacement for conda, so using mamba instead of conda in our CI jobs might speed them up without much hassle.

As suggested by @wholmgren in https://github.com/pvlib/pvlib-python/pull/1306#issuecomment-1152660707, the [micromamba github action](https://github.com/mamba-org/provision-with-micromamba) might be one way to do this.  This method pandas uses seems worth exploring too: https://github.com/pandas-dev/pandas/blob/3b6015f24f088597846c757b1e9f85e6e379ece7/.github/actions/setup-conda/action.yml#L35

Cached environments may be another option.  In general I am wary of persisting anything between CI sessions for fear of introducing some difficult-to-debug state issue, but if it can be done robustly then it may be worthwhile here.

**Additional context**
Starting with the pytest jobs probably makes the most sense, but it might be possible to use mamba for the asv job as well.  But see https://github.com/airspeed-velocity/asv/pull/970

jules-ch:
- [x] Closes #1472 
 - [x] Tests added
 ~~- [ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/reference) for API changes.~~
~~- [ ] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.~~

Update workflow to replace conda with micromamba.

- Add step to install pvlib without installing deps
  - It is required to test the proper version of pvlib (pvlib is a dependency of pvfactors which install last version of pvlib)
- Update Checkout step to limit depth
- Add step to get tags to have version inference when installing pvlib

@kanderso-nrel I only added cache-downloads which is purged everyday. It does not cache the whole environment.
We could do that to speed things up even more, cached environment is cached for one day also.

See https://github.com/mamba-org/provision-with-micromamba

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -8,7 +8,6 @@ on:
 
 jobs:
   test:
-
     strategy:
       fail-fast: false  # don't cancel other matrix jobs when one fails
       matrix:
@@ -31,16 +30,23 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v3
-
-      - name: Set up conda environment
-        if: matrix.environment-type == 'conda'
-        uses: conda-incubator/setup-miniconda@v2
+      # We check out only a limited depth and then pull tags to save time
+      - name: Checkout source
+        uses: actions/checkout@v3
         with:
-          activate-environment: test_env
+          fetch-depth: 100
+
+      - name: Get tags
+        run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+
+      - name: Install Conda environment with Micromamba
+        if: matrix.environment-type == 'conda'
+        uses: mamba-org/provision-with-micromamba@v12
+        with:
           environment-file: ${{ env.REQUIREMENTS }}
-          python-version: ${{ matrix.python-version }}
-          auto-activate-base: false
+          cache-downloads: true
+          extra-specs: |
+            python=${{ matrix.python-version }}
         env:
           # build requirement filename.  First replacement is for the python
           # version, second is to add "-min" if needed
@@ -49,13 +55,18 @@ jobs:
       - name: List installed package versions (conda)
         if: matrix.environment-type == 'conda'
         shell: bash -l {0}  # necessary for conda env to be active
-        run: conda list
+        run: micromamba list
 
       - name: Install bare Python ${{ matrix.python-version }}${{ matrix.suffix }}
         if: matrix.environment-type == 'bare'
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
+
+      - name: Install pvlib
+        if: matrix.environment-type == 'conda'
+        shell: bash -l {0}
+        run: python -m pip install --no-deps .
 
       - name: Set up bare environment
         if: matrix.environment-type == 'bare'
--- a/docs/sphinx/source/whatsnew/v0.9.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.2.rst
@@ -26,6 +26,7 @@ Bug fixes
 Testing
 ~~~~~~~
 * Switched CI testing provider from Azure to GitHub Actions (:pull:`1306`)
+* Speed up CI setup using micromamba instead of conda (:pull:`1493`)
 
 
 Documentation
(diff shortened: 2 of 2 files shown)


.github/workflows/pytest.yml
```diff
@@ -31,16 +30,23 @@ jobs:
...
+      - name: Install Conda environment with Micromamba
         if: matrix.environment-type == 'conda'
-        uses: conda-incubator/setup-miniconda@v2
+        uses: mamba-org/provision-with-micromamba@v12
         with:
-          activate-environment: test_env
           environment-file: ${{ env.REQUIREMENTS }}
-          python-version: ${{ matrix.python-version }}
-          auto-activate-base: false
+          cache-downloads: true
```
kanderso-nrel, line 47: A bit of interesting info here regarding caching in case anyone is interested: https://github.com/mamba-org/provision-with-micromamba/pull/38#discussion_r808837618

.github/workflows/pytest.yml
```diff
@@ -31,16 +30,23 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v3
+      # We check out only a limited depth and then pull tags to save time
+      - name: Checkout source
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 100
```
kanderso-nrel, line 37: Why 100 here?  I think the default is just 1, so I'd have thought that this would tend to be slower, not faster.  Although either way this step is pretty fast relative to the pytest step, so maybe not worth optimizing anyway?

kanderso-nrel (commented):
Thanks @jules-ch!  One comment below, otherwise I think this LGTM.  Time taken to set up the Windows 3.6 environment dropped from ~8 minutes in #1490 to ~2 minutes here, which is a very nice improvement.  Here's a less precise comparison of total job runtimes between this PR and #1490, just using the "Successful in 1m" text in the checks summary:

|                                 | #1490   | #1493   |
|:--------------------------------|:--------|:--------|
| macos-10.15, 3.6, bare          | 1m      | 1m      |
| macos-10.15, 3.7, bare          | 2m      | 1m      |
| macos-10.15, 3.8, bare          | 2m      | 1m      |
| macos-10.15, 3.9, bare          | 2m      | 2m      |
| ubuntu-latest, 3.6, bare        | 1m      | 1m      |
| ubuntu-latest, 3.6, conda       | 4m      | 2m      |
| ubuntu-latest, 3.6, conda, -min | 1m      | 1m      |
| ubuntu-latest, 3.7, bare        | 1m      | 1m      |
| ubuntu-latest, 3.7, conda       | 3m      | 4m      |
| ubuntu-latest, 3.8, bare        | 54s     | 54s     |
| ubuntu-latest, 3.8, conda       | 3m      | 2m      |
| ubuntu-latest, 3.9, bare        | 50s     | 1m      |
| ubuntu-latest, 3.9, conda       | 4m      | 2m      |
| windows-latest, 3.6, conda      | 10m     | 5m      |
| windows-latest, 3.7, conda      | 9m      | 5m      |
| windows-latest, 3.8, conda      | 8m      | 5m      |
| windows-latest, 3.9, conda      | 6m      | 6m      |

.github/workflows/pytest.yml
jules-ch: I think we need to fetch some commits to get the proper version using versionner with versionner.py when using `pip install --no-deps .` in further steps.

Example in this workflow :

`Successfully installed pvlib-0.9.1+17.g3fcda76`

Without this you cannot infer version 0.9.1 with 17 commit ahead

100 is purely arbitrary

.github/workflows/pytest.yml
jules-ch: Tell me if you want to enable cache-env, stay with cache-downloads or removing cache alltogether

kanderso-nrel (approved):
I think I'm okay with this.  Faster, plus it fixes at least one issue with versioning that probably would have popped up eventually.  Anyone else want to take a look?

PR #1494, opened 2022-07-19
Allow read_tmy3 to handle 00:00 timestamps

AdamRJensen:
- ~~[ ] Closes #xxxx~~
 - [x] Tests added

As mentioned by @williamhobbs in https://github.com/pvlib/pvlib-python/discussions/1310#discussioncomment-2996349, SolarAnywhere TMY3 files specify midnight as 00:00 instead of 24:00 (which is the NREL TMY3 convention). Currently, the pvlib `read_tmy3` function incorrectly treats 00:00 and 24:00 equally by renaming both time stamps to 00:00 and shifting the time stamp one day ahead. This behavior is correct for the 24:00 timestamps but incorrectly shifts 00:00 instances one day ahead.

This PR proposes a minor change to handle both 00:00 and 24:00 timestamps correctly. Additionally, I've made trivial documentation updates to more closely follow the numpydoc style.


pvlib/iotools/tmy.py
```diff
@@ -181,8 +176,9 @@ def read_tmy3(filename, coerce_year=None, recolumn=True):
     data_ymd = pd.to_datetime(data['Date (MM/DD/YYYY)'], format='%m/%d/%Y')
     # shift the time column so that midnite is 00:00 instead of 24:00
     shifted_hour = data['Time (HH:MM)'].str[:2].astype(int) % 24
-    # shift the dates at midnite so they correspond to the next day
-    data_ymd[shifted_hour == 0] += datetime.timedelta(days=1)
+    # shift the dates at midnight (24:00) so they correspond to the next day.
+    # If midnight is specified as 00:00 do not shift date.
+    data_ymd[data['Time (HH:MM)'].str[:2] == '24'] += datetime.timedelta(days=1)  # noqa: E501
```
AdamRJensen, line 181: These lines contain the changes to the function itself. The remaining changes are only related to the formatting of the documentation.

AdamRJensen:
@KWagnerCPR could you provide a TMY3 file from SolarAnywhere that can be added to pvlib for testing purposes?

KWagnerCPR:
@AdamRJensen I've attached the full historical time series in TMY3 format for one of our Public sites here. 
[SolarAnywhereData_TMY3.zip](https://github.com/pvlib/pvlib-python/files/9162508/SolarAnywhereData_TMY3.zip)

AdamRJensen:
@pvlib/pvlib-maintainer Ready for review :)

Does anyone know why it says "TMYData" 70 times in the documentation? I'd like to remove all those instances..
image

Also, I think the following line in the documentation might be incorrect: *"NOTE, the index is currently timezone unaware, and times are set to local standard time (daylight savings is not included)"*. The index seems to have the timezone offset included - I'd appreciate a second pair of eyes on that though.

kanderso-nrel:
> Does anyone know why it says "TMYData" 70 times in the documentation?

A long time ago in a commit far, far away, that's what the function's return value was called:

https://github.com/pvlib/pvlib-python/blob/fef3d518b6ebe719f05bb51beeb7dfa05b45b23c/pvlib/tmy.py#L42-L45

Inherited from the MATLAB code, presumably: https://github.com/sandialabs/MATLAB_PV_LIB/blob/master/pvl_readtmy3.m

> Also, I think the following line in the documentation might be incorrect: "NOTE, the index is currently timezone unaware, and times are set to local standard time (daylight savings is not included)". The index seems to have the timezone offset included - I'd appreciate a second pair of eyes on that though.

This pair of eyes agrees with yours; the index includes a fixed UTC offset.  The part about not including DST should be kept though.

pvlib/tests/iotools/test_tmy.py
```diff
@@ -2,13 +2,15 @@
...
 from pvlib.iotools import tmy
 from ..conftest import DATA_DIR
+import pytest
 
 # test the API works
 from pvlib.iotools import read_tmy3
 
 TMY3_TESTFILE = DATA_DIR / '703165TY.csv'
 TMY2_TESTFILE = DATA_DIR / '12839.tm2'
 TMY3_FEB_LEAPYEAR = DATA_DIR / '723170TYA.CSV'
```
cwhanse, line 12: I would add a comment here identifying the source of these files as the NSRDB, because the NSRDB TMY3 format is different from SolarAnywhere's.

cwhanse (commented):
I think we need to add a Note to `read_tmy3` to the effect that the function can read TMY3 files from either the NRSDB or from SolarAnywhere. The document references are specific to the NSRDB. SolarAnywhere TMY3 is described here: https://www.solaranywhere.com/support/historical-data/file-formats/

wholmgren:
I tried to remove the `TMYData.` at some point, ran into some issue with rst table formatting, and decided it wasn't worth the effort. Maybe things have changed or you're more patient/efficient than I.

pvlib/tests/iotools/test_tmy.py
AdamRJensen: I've added comments to the tests and an admonition in the documentation:
image

pvlib/iotools/tmy.py
```diff
@@ -62,80 +59,85 @@ def read_tmy3(filename, coerce_year=None, recolumn=True):
...
-    TMYData.LprecipSource               See [1]_, Table 1-5, 8760x1 cell array of strings
-    TMYData.LprecipUncertainty          See [1]_, Table 1-6
-    TMYData.PresWth                     Present weather code, see [2]_.
-    TMYData.PresWthSource               Present weather code source, see [2]_.
-    TMYData.PresWthUncertainty          Present weather code uncertainty, see [2]_.
-    =============================       ======================================================================================================================================================
+    =====================       ======================================================================================================================================================
+    field                       description
+    =====================       ======================================================================================================================================================
+    Index                       A pandas datetime index. NOTE, the index is timezone aware, and times are set to local time (daylight savings is not included)
```
kanderso-nrel, line 65: I would keep this as "local standard time".  "local time" alone implies DST *is* included, for most places in the US anyway.

kanderso-nrel (commented):
I just tagged this PR with `remote-data` so that the iotools tests run.  I get the same encoding failure locally, FYI.

In hindsight, requiring the `remote-data` tag to be present before running *any* iotools tests, not just the actual `--remote-data` tests, is probably not actually what we want.  Might be as simple as just removing the `--ignore` options in the workflow command.

AdamRJensen:
> I just tagged this PR with `remote-data` so that the iotools tests run. I get the same encoding failure locally, FYI.

Huh! I was very surprised to see test failures. I could not replicate them locally, so I imagine it's a windows/linux thing. My best guess is that the ® (registered) symbol in the test file is not parseable on Linux (at least not by default). For now, I have just removed it from the file. I think the best thing would be for CPR to remove it generally from their files and perhaps replace it with (R) or the phrase "Registered, U.S. Patent and Trademark Office". Alternatively, we could look into specifying some specific encoding settings when opening the file 🤷  @kanderso-nrel thoughts?

> In hindsight, requiring the `remote-data` tag to be present before running _any_ iotools tests, not just the actual `--remote-data` tests, is probably not actually what we want. Might be as simple as just removing the `--ignore` options in the workflow command.

I'm not quite sure what you're talking about here..

kanderso-nrel:
> For now, I have just removed it from the file.

I think technically that would be in violation of the CPR license file included in the zip archive posted earlier:

> Licensee shall not remove, alter, cover, or disguise any acknowledgements, copyright notice, trademark, or other proprietary rights notice placed by Clean Power Research on the Data or any portion thereof.

I'm not optimistic that CPR would get rid of that character or change their file encoding just for us (maybe I'm wrong!), but I'm not sure what the best alternative is.  FWIW specifying `encoding='iso-8859-01'` (determined using the [file](https://en.wikipedia.org/wiki/File_(command)) command) gets it working for me locally, but I'm wary of hardcoding that in a general purpose function `read_tmy3`.  Maybe expose `encoding` as an optional parameter and suggest a value for SolarAnywhere files?  

> I'm not quite sure what you're talking about here..

Sorry, was just thinking out loud that we should probably follow up #1306 with some tweaks.  No changes needed here.

pvlib/iotools/tmy.py
```diff
@@ -23,15 +20,16 @@ def read_tmy3(filename, coerce_year=None, recolumn=True):
...
-
     coerce_year : None or int, default None
         If supplied, the year of the index will be set to `coerce_year`, except
         for the last index value which will be set to the *next* year so that
         the index increases monotonically.
-
     recolumn : bool, default True
         If ``True``, apply standard names to TMY3 columns. Typically this
         results in stripping the units from the column name.
+    encoding, str, optional
```
kanderso-nrel, line 30:
```suggestion
    encoding : str, optional
```

kanderso-nrel (approved):

wholmgren:
`read_csv` also has an `encoding_errors` kwarg since pandas 1.3. I've never used it but might help here. We could also try/except with default encoding and then `iso-8859-01`. Or read the first N bytes of the file to see if it matches the SolarAnywhere format to determine the encoding.

Changing the SolarAnywhere encoding would break existing code, so I wouldn't recommend that unless they bumped their API to 4.0.

AdamRJensen:
> `read_csv` also has an `encoding_errors` kwarg since pandas 1.3. I've never used it but might help here. We could also try/except with default encoding and then `iso-8859-01`. Or read the first N bytes of the file to see if it matches the SolarAnywhere format to determine the encoding.

The `read_tmy3` function utilizes the built-in [`open`](https://docs.python.org/3/library/functions.html#open) function, which has an argument called `errors`. Options for the errors argument are `ignore` and `replace`, neither of which I think are great options. I have implemented try/except which I think is the best option (also better than adding an `encoding` parameter).

pvlib/iotools/tmy.py
```diff
@@ -150,24 +150,28 @@ def read_tmy3(filename, coerce_year=None, recolumn=True):
...
-        # header is actually the second line, but tell pandas to look for
-        # header information on the 1st line (0 indexing) because we've already
-        # advanced past the true first line with the readline call above.
-        data = pd.read_csv(csvdata, header=0)
+    try:
+        fbuf = open(str(filename), 'r')
+    # SolarAnywhere files contain non-UTF8 characters and require
+    # encoding='iso-8859-1' on Linux in order to be parsed
+    except UnicodeDecodeError:
+        fbuf = open(str(filename), 'r', encoding='iso-8859-1')
```
kanderso-nrel, line 166: I believe it was the `readline()` call (not the `open()` call) that errors if the encoding is not correct.  Also if we're not using the context manager form of `open()` then best practice is to include an explicit `fbuf.close()`, preferably inside a `finally` clause, to ensure that open file handles aren't left lying around.

P.S. you currently have to re-tag the PR w/ `remote-data` to trigger those tests again.  That is something else to improve about the remote-data tests.  Sorry this PR is the guinea pig for the `remote-data` part of #1306!

AdamRJensen: @kanderso-nrel Sadly you are correct. The try/except method gets a bit messier but I guess that's just how it is.

I think the remote-test failures are unrelated (`get_surfrad`). Also, I'm puzzled why four tests were canceled..

pvlib/iotools/tmy.py
```diff
@@ -150,24 +150,30 @@ def read_tmy3(filename, coerce_year=None, recolumn=True):
...
+        firstline = fbuf.readline()  # first line contains the metadata
+    finally:
         # use pandas to read the csv file buffer
         # header is actually the second line, but tell pandas to look for
         # header information on the 1st line (0 indexing) because we've already
         # advanced past the true first line with the readline call above.
-        data = pd.read_csv(csvdata, header=0)
+        data = pd.read_csv(fbuf, header=0)
+        # close file
+        fbuf.close()
```
kanderso-nrel, line 176: I'm probably just being picky at this point, but if the `pd.read_csv` call fails for some reason then there's still risk of not closing the file.  How about this?  Better to use `with open(...):` than keep track of handle state manually. 

```python
def _parse(fbuf):
    firstline = fbuf.readline()
    data = pd.read_csv(fbuf, header=0)
    return firstline, data

try:
    with open(str(filename), 'r') as fbuf:
        firstline, data = _parse(fbuf)
except UnicodeDecodeError:
    with open(str(filename), 'r', encoding='iso-8859-1') as fbuf:
        firstline, data = _parse(fbuf)
```

pvlib/iotools/tmy.py
```diff
@@ -150,24 +150,30 @@ def read_tmy3(filename, coerce_year=None, recolumn=True):
...
     head = ['USAF', 'Name', 'State', 'TZ', 'latitude', 'longitude', 'altitude']
-
-    with open(str(filename), 'r') as csvdata:
-        # read in file metadata, advance buffer to second line
-        firstline = csvdata.readline()
+    try:
+        fbuf = open(str(filename), 'r')
+        firstline = fbuf.readline()  # first line contains the metadata
+    # SolarAnywhere files contain non-UTF8 characters and require
+    # encoding='iso-8859-1' on Linux in order to be parsed
```
wholmgren, line 165: I don't think this is particular to linux. I haven't tested this particular PR or file, but from other work I know that I need to specify the encoding when using solar anywhere files on my windows machine.

pvlib/iotools/tmy.py
AdamRJensen: Consider it done.

pvlib/iotools/tmy.py
AdamRJensen: The Linux part has been removed.

wholmgren (approved):
Thanks @AdamRJensen!

PR #1483, opened 2022-06-26
setup.py: use setuptools.find_namespace_packages()

kanderso-nrel:
- [x] Closes #1474
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - ~[ ] Tests added~
 - ~[ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/reference) for API changes.~
 - ~[ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).~
 - ~[ ] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.~
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

According to people who keep better track than I do of the evolution of python packaging over time, any folder is now considered a python package, even if it doesn't have any python-related files (see https://github.com/pypa/setuptools/issues/3340#issuecomment-1146676266).  This prompted some deprecations in `setuptools`, leading to the warnings in #1474.  According to https://github.com/pypa/setuptools/issues/3340#issuecomment-1146678086 and similar posts elsewhere, using `setuptools.find_namespace_packages()` is a recommended fix.  

In our case `find_namespace_packages()` includes a bunch of other stuff we don't actually want, but `find_packages()` seems closer to the mark.  Making that change seems to get rid of the warnings without changing which files get included in the `sdist`:

```bash
# use `cut` to remove the first component of filepaths, which we don't care about here
$ tar tf pvlib-0.9.1+12.gc8b04a56.tar.gz | cut -d / -f 2- > old.txt  # current master
$ tar tf pvlib-0.9.1+18.g071cc215.tar.gz  | cut -d / -f 2- > new.txt  # current HEAD on this branch
$ diff old.txt new.txt  # no output (except unrelated changes, e.g. new pvfactors gallery example file), so list of filenames did not change
```

[...]

PR #1495, opened 2022-07-20
Use pep517/518 build system for modern build isolation

jules-ch:
Following #1486

- Remove pvlib/_version.py (versioneer)
- Use importlib.metadata to retrieve version from PKG-INFO
- Add importlib-metadata backport for python < 3.8
- Create pyproject.toml with setuptools build backend
- Update workflow to use pypa/build module to build sdist & wheel

editable installation still works with the new setuptools 63

 ~- [ ] Closes #xxxx~
 ~- [ ] Tests added~
 ~- [ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/reference) for API changes.~
 ~- [ ] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.~


pyproject.toml
```diff
@@ -0,0 +1,25 @@
+[build-system]
+requires = ["setuptools>=45", "wheel", "setuptools_scm>=6.2"]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools_scm]
```
kanderso-nrel, line 5: The version string for non-tagged commits using this method references the next pvlib version instead of the previous.  For example this PR branch currently references 0.9.2:
- `master`: `pvlib-0.9.1+18.g3f397ed4`
- this PR: `pvlib-0.9.2.dev16+g2c4fbe3f`

The number of commits since the last tag is `.devN` instead of `+N`.  "dirty" builds are also a bit different: `pvlib-0.9.2.dev16+g2c4fbe3f.d20220721` vs `pvlib-0.9.1+18.g3f397ed4.dirty`.

Not sure I see either of these as a problem, just pointing it out.

kanderso-nrel (commented):
Thanks for these excellent PRs @jules-ch!  Can you please also remove the `versioneer.py` and `_version.py` references in `.lgtm.yml`, `.stickler.yml`, and `MANIFEST.in`?

Also, any idea why a bunch of top-level files are being included in the dist bundles?  Compare the build logs for [master](https://github.com/pvlib/pvlib-python/runs/7415055298?check_suite_focus=true#step:6:72) versus [this PR](https://github.com/pvlib/pvlib-python/runs/7437689503?check_suite_focus=true#step:6:107).

jules-ch:
About top files included in dist, I think setuptools_scm by default include files checked in VCS.
I'll have to check.

About the version name, by default it generates a dev version with number as number of commit. This explains why it displays next version with dev.

Then local part, after `+` you get commit hash + date if dirty. This is customizable if we want to have the exact same as versioneer.

kanderso-nrel:
> I think setuptools_scm by default include files checked in VCS.

I think you're right, and we should adapt our MANIFEST.in accordingly: https://github.com/pypa/setuptools_scm#file-finders-hook-makes-most-of-manifestin-unnecessary

It's fine by me to change the version string format, especially if the new format is what the PyPA suggests/prefers.  I only brought it up in case the other maintainers care more about non-tagged version strings than I do :)

jules-ch:
I've removed include in `MANIFEST.in` since files tracked by VCS are by default included.
Also I've removed reference to versioneer files from .lgtm & .stickler files.

jules-ch:
@kanderso-nrel can you check building SPA C extension is still working ? Should not be a problem, setuptools build backend is till using setup.py under the hood.

kanderso-nrel:
> @kanderso-nrel can you check building SPA C extension is still working ? Should not be a problem, setuptools build backend is still using setup.py under the hood.

After building the `.so` file:

- `$ pip install .` seems to do the right thing: it produces and installs a platform-specific wheel (`pvlib-0.9.2.dev16+g3bcd63e-cp39-cp39-linux_x86_64.whl`) that includes the `.so` file, and the C function executes as expected from python.
- `$ python -m build --wheel` also produces a platform-specific wheel containing the `.so`.  Seems like it would work, although I did not actually install it to be sure.
- `$ python -m build` produces a platform agnostic wheel (`pvlib-0.9.2.dev16+g3bcd63e-py3-none-any.whl`) which doesn't contain the `.so`. 

I'm not sure why `python -m build` behaves differently from `python -m build --wheel` in that regard, but I also think it doesn't really matter since our distribution files don't include the SPA C extension.  Building it locally and installing with `pip install .` still works, and I think that's all we care about (I hope another maintainer will confirm this). 

---

Unrelated to the SPA C extension, I think we should add a bunch of excludes to the MANIFEST so that `docs`, `benchmarks`, `ci` etc don't get bundled.  But I wouldn't object to skipping that for now and adding it to #1483.

kanderso-nrel (approved):
I think I'm okay with these changes, assuming we follow up elsewhere on the MANIFEST excludes.  Would definitely be good for another maintainer to take a look though.

wholmgren:
I thought [editable installs do not work](https://github.com/pvlib/pvlib-python/discussions/1486#discussioncomment-3085581) with this pattern. That would destroy my workflow.

No objection to the change in version string format. I'm guessing we'll want to make alpha tags when we anticipate that the next release will be a 0.x.0. (I'd like to see more pre-releases anyways.)

Are the changes to MANIFEST consistent with the discussion in https://github.com/pvlib/pvlib-python/pull/580? I am a -1 on removing anything from the distribution at this time, so let's take up that discussion somewhere else.

kanderso-nrel:
> I thought https://github.com/pvlib/pvlib-python/discussions/1486#discussioncomment-3085581 with this pattern. That would destroy my workflow.

I *think* editable installs are only a problem if you ditch `setup.py` altogether (https://github.com/pypa/setuptools/issues/2816#issuecomment-1195451458), which is not the case here.  `pip install -e .` seemed to do the right thing when I tried it out just now, but it would be good if you could try it on your end too since I don't know how to thoroughly vet that it actually works.

> Are the changes to MANIFEST consistent with the discussion in https://github.com/pvlib/pvlib-python/pull/580? I am a -1 on removing anything from the distribution at this time, so let's take up that discussion somewhere else.

Sorry, maybe my phrasing was confusing.  As I understand it, this PR doesn't remove anything from the distribution.  It does *add* things to the distribution because of the switch to `setuptools-scm`, which includes everything tracked by git ([ref](https://github.com/pypa/setuptools_scm#file-finders-hook-makes-most-of-manifestin-unnecessary)), and this PR doesn't `prune` them back out again.  

I thought it would be easier to prune them back out in another PR (e.g. #1483, which is already dealing with what gets included in the dist) so that we don't ask too much of @jules-ch :)

wholmgren:
Seems to work. I did the following

1. check out branch
2. activate conda env with existing pvlib dev installation
3. import pvlib. `pvlib.__version__` returns 0.7.2 (that's what was checked out when I last ran `pip install -e .`)
4. confirm that despite version report above, I actually have all the features on this branch (like Array, which was created long after 0.7.2)
5. exit ipython
6. `pip install -e .` ``Successfully uninstalled pvlib-0.7.2`` ``Successfully installed pvlib-0.9.2.dev16+g3bcd63e``
7. `import pvlib; pvlib.__version__` `'0.9.2.dev16+g3bcd63e'`
8. define a new function in irradiance.py. use `importlib.reload` to reload module. new function works

I'm good with merging. I'd like to give @mikofski an opportunity to review too (or decline to review) since he's been fighting python packaging longer than any of us.

wholmgren (approved):

wholmgren:
@jules-ch can you resolve the merge conflicts?

jules-ch:
I've rebased, should be good now @wholmgren

jules-ch:
FYI CI is failing because I'm ignoring coverage on version.py & it is reducing code coverage which is making CI fails.

wholmgren:
thanks @jules-ch! let's see how it goes...

wholmgren:
One thing that I failed to test above... make a change, then check the version *without reinstalling*. It used to be the case that versioneer would immediately report `.dirty` or, if a commit was made, `+X.abcdefg`. Now it seems that I must rerun `pip install -e .` to get the new version. Is that correct? That seems to track the discussion in https://github.com/matplotlib/matplotlib/pull/18971. And mpl now uses a [function in `__init__.py`](https://github.com/matplotlib/matplotlib/blob/276436aba3f52ab3158efb89212fe31ed45806ed/lib/matplotlib/__init__.py#L165-L182) as a workaround. Thoughts on a follow up PR for that?

jules-ch:
Shoud not be too difficult

kanderso-nrel:
Ok, makes sense that switching from versioneer (run-time version determination) to setuptools_scm (install-time determination) would cause this issue. After some searching, it seems like people either just live with outdated version strings for editable installs or use some variant of the matplotlib approach of dynamically computing the version if the code detects that it's being imported from a git repo.  I don't love it but it seems there's not a better option, or at least I haven't found one!

PR #1497, opened 2022-07-22
Add functions to read and retrieve SolarAnywhere irradiance data

AdamRJensen:
- [x] Closes #1310
 - [x] Tests added

**Description**
This PR adds functions for reading and retrieving SolarAnywhere data as discussed in #1310. SolarAnywhere is a commercial dataset, although it provides irradiance data for a few locations for free.

**Minor comments**
The SolarAnywhere API is asynchronous, hence two requests are made for each data retrieval.

The `get_solaranywhere` function is currently not designed to handle annual and monthly data, as this seems less relevant to pvlib users.


AdamRJensen:
@KWagnerCPR It would be great if you could test the functions and give some feedback on the default values.

docs/sphinx/source/whatsnew/v0.9.2.rst
```diff
@@ -11,6 +11,9 @@ Enhancements
 * Add :py:func:`pvlib.tracking.calc_surface_orientation` for calculating
   single-axis tracker ``surface_tilt`` and ``surface_azimuth`` from
   rotation angles. (:issue:`1471`, :pull:`1480`)
+* Add :py:func:`pvlib.iotools.read_solaranywhere` and
+  :py:func:`pvlib.iotools.get_solaranywhere` for reading and retrieving
+  SolarAnywhere solar irradiance data. (:pull:`1497`, :discussion:`1310`)
```
kanderso-nrel, line 16: This will need a new [extlinks](https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html) entry for `discussion` here: 

https://github.com/pvlib/pvlib-python/blob/3f397ed44075b01f4ba535750356859195a51b2d/docs/sphinx/source/conf.py#L295-L301

AdamRJensen: Perhaps it is more appropriate to convert the discussion to an issue?

KWagnerCPR:
> @KWagnerCPR It would be great if you could test the functions and give some feedback on the default values. 😄

@AdamRJensen Yes, I'll have a few people on our team test it out and get back to you. Which default values are you referring to?

AdamRJensen:
@KWagnerCPR In regards to default values I'm thinking of these:

    def get_solaranywhere(latitude, longitude, api_key, start=None, end=None,
                          time_resolution=60, spatial_resolution=0.1,
                          true_dynamics=False, source='SolarAnywhereLatest',
                          variables=DEFAULT_VARIABLES, missing_data='Omit',
                          url=URL, map_variables=True, max_response_time=300)

Also, I have chosen not to parse any annual and monthly data, as that seems to be outside the scope of pvlib and would just unnecessarily complicate the function.

Furthermore, the `get_solaranywhere` function currently does not support retrieval of "probability of exceedance" time series, as the public license does not allow me to retrieve those.

KWagnerCPR:
@AdamRJensen 

In terms of default values, I would recommend the following: 
time_resolution=60, 
spatial_resolution=0.01,
true_dynamics=False, 
source='SolarAnywhereLatest',
missing_data='FillAverage',
                    
With SolarAnywhere Public, users get access to 1 km (0.01 x 0.01 degree) spatial resolution data, so I think it makes sense to set the default as 0.01 if our focus is on time series requests rather than TGY requests. Also, using the "FillAverage" default for missing data will ensure periods of missing satellite imagery are filled with long term averages for those times, which is likely preferable. 

You should be able to request PXX data as a part of the SolarAnywhere public license. The issue might have been that you had the spatial resolution set to 0.1 when you requested it. You also have to include "ProbabilityOfExceedance": 90, in the request. 

I think it makes sense not to request the monthly and annual totals at this time.

AdamRJensen:
@KWagnerCPR I have updated the default values accordingly. Thanks for the feedback! :smile:

Also, I figured out how to work with PXX data, so I've added a keyword called `probability_of_exceedance`.

AdamRJensen:
@pvlib/pvlib-maintainer I am thinking of developing primarily mock tests for the `get_solaranywhere` function as the SolarAnywhere data retrieval is slow (minutes). However, I would like to include at least one real request of historical data (just two days of data, which would make the time reasonable). Thus, unless I hear any objections I will go ahead and register the pvlib email for a free SolarAnywhere account and add a GitHub Secret called `SOLARANYWHERE_API_KEY`. If there is a desire to avoid live tests at all, let me know. FYI I imagine writing mock tests for this will be a real pain.

KWagnerCPR:
@AdamRJensen We've tested these functions and they are working well!

One thing I noticed is that when I request time series data (WeatherDataSource = SolarAnywhereLatest) without specifying a start or end time, I don't get the error message you specify in the script: "'When requesting non-TMY data, specifying `start` and 'end` is required.'". Instead I get: AttributeError: 'NoneType' object has no attribute 'tz'. Not a huge deal. Just wanted to point it out. 

I think these will be really helpful for our customers as well as new SolarAnywhere users. I'd love to include links to these in our developers documentation. Do you know if these will get documented in the official pvlib documentation?: https://pvlib-python.readthedocs.io/en/stable/index.html

AdamRJensen:
> One thing I noticed is that when I request time series data (WeatherDataSource = SolarAnywhereLatest) without specifying a start or end time, I don't get the error message you specify in the script: "'When requesting non-TMY data, specifying `start` and 'end` is required.'". Instead I get: AttributeError: 'NoneType' object has no attribute 'tz'. Not a huge deal. Just wanted to point it out.

@KWagnerCPR Thank you for catching this error. I had forgotten to raise the error, hence nothing was happening - it was been corrected.

> I think these will be really helpful for our customers as well as new SolarAnywhere users. I'd love to include links to these in our developers documentation. Do you know if these will get documented in the official pvlib documentation?: https://pvlib-python.readthedocs.io/en/stable/index.html

Once the functions get merged and we make a new release of pvlib, the documentation will be available publicly. We're releasing version 0.9.2 within a few weeks, but I can't promise that this pull request will be merged by then.

KWagnerCPR:
> > One thing I noticed is that when I request time series data (WeatherDataSource = SolarAnywhereLatest) without specifying a start or end time, I don't get the error message you specify in the script: "'When requesting non-TMY data, specifying `start` and 'end` is required.'". Instead I get: AttributeError: 'NoneType' object has no attribute 'tz'. Not a huge deal. Just wanted to point it out.
> 
> @KWagnerCPR Thank you for catching this error. I had forgotten to raise the error, hence nothing was happening - it was been corrected.
> 
> > I think these will be really helpful for our customers as well as new SolarAnywhere users. I'd love to include links to these in our developers documentation. Do you know if these will get documented in the official pvlib documentation?: https://pvlib-python.readthedocs.io/en/stable/index.html
> 
> Once the functions get merged and we make a new release of pvlib, the documentation will be available publicly. We're releasing version 0.9.2 within a few weeks, but I can't promise that this pull request will be merged by then.

@AdamRJensen Thank you! I'll stay tuned.

AdamRJensen:
When downloading data from the SolarAnywhere website, missing data may either be filled with average data or represented as missing using one of three options:
image

Missing values are by default correctly converted to Numpy nan. Similarly, I belive that the 'NaN' option would also be correctly converted to Numpy nan since we're using pd.read_csv (though I haven't tested it). However, -999 will have to be converted to numpy nan by the `read_solaranywhere` function.

@KWagnerCPR Do you have a file or a know station/time period where there is missing data that I could use for testing?

AdamRJensen:
The remote test failures are caused by this line:

    solaranywhere_api_key = os.environ["SOLARANYWHERE_API_KEY"]

which makes me assume that the API key added to the GitHub Secrets by @kanderso-nrel isn't named what I think they are...

kanderso-nrel:
I think it is just the annoying but secure behavior of `pull-request-target` not respecting `yml` changes in the PR -- your update to include the SA key in the environment isn't taking effect because the action is defined by the `.github/workflows/pytest-remote-data.yml` file on `master`, not this PR branch.  I guess the cleanest workaround is to submit a separate PR to change `pytest-remote-data.yml`, merge that, then merge master into this PR?

KWagnerCPR:
@cwhanse I see, thank you for sharing that link! Looks like this item was slated for the 0.10.2 release on September 6, but was then deferred. Do you have an update on when the next release may take place?

kandersolar:
@KWagnerCPR I've added this to the v0.10.3 milestone, which is scheduled to be released mid-December.  But (as seen in the previous milestones for this PR), that's just an aspiration, not a guarantee.  Getting this PR merged depends on reviewer availability, which is in rather short supply at the moment.

KWagnerCPR:
@kandersolar Understood. Thank you for this background info!

docs/sphinx/source/whatsnew/v0.9.2.rst
```diff
@@ -14,6 +14,9 @@ Enhancements
 * Add :py:func:`pvlib.tracking.calc_surface_orientation` for calculating
   single-axis tracker ``surface_tilt`` and ``surface_azimuth`` from
   rotation angles. (:issue:`1471`, :pull:`1480`)
+* Add :py:func:`pvlib.iotools.read_solaranywhere` and
+  :py:func:`pvlib.iotools.get_solaranywhere` for reading and retrieving
+  SolarAnywhere solar irradiance data. (:pull:`1497`)
```
kandersolar, line 19: Move to 0.10.3

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,285 @@
...
+        Time resolution in minutes. For TMY data, time resolution has to be 60
+        min. (hourly).
+    spatial_resolution: {0.1, 0.01, 0.005}, default: 0.01
+        Spatial resolution in degrees.
+    true_dynamics: bool, default: False
+        Whether to apply SolarAnywhere TrueDynamics statistical processing.
+        Only available for the 5-min time resolution.
+    probability_of_exceedance: int, optional
+        Probability of exceedance in the range of 1 to 99. Only relevant when
+        requesting probability of exceedance (POE) time series.
```
kandersolar, line 88:
```suggestion
        requesting probability of exceedance (POE) time series. [%]
```

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,285 @@
...
+        Variables to retrieve (described in [4]_).  Available variables depend
+        on whether historical or TMY data is requested.
+    missing_data: {'Omit', 'FillAverage'}, default: 'FillAverage'
+        Method for treating missing data.
+    url: str, default: :const:`pvlib.iotools.solaranywhere.URL`
+        Base url of SolarAnywhere API.
+    map_variables: bool, default: True
+        When true, renames columns of the DataFrame to pvlib variable names
+        where applicable. See variable :const:`VARIABLE_MAP`.
+    max_response_time: float, default: 300
```
kandersolar, line 99: Some other iotools functions have a `timeout` parameter that serves a similar purpose (although the mechanical details are a bit different under the hood).  Worth considering using the same parameter name here for consistency.

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,285 @@
...
+            break
+        elif (time.time()-start_time) > max_response_time:
+            raise TimeoutError('Time exceeded the `max_response_time`.')
+        time.sleep(5)  # Sleep for 5 seconds before each data retrieval attempt
+
+    # Extract time series data
+    data = pd.DataFrame(results_json['WeatherDataResults'][0]['WeatherDataPeriods']['WeatherDataPeriods'])  # noqa: E501
+    # Set index and convert to UTC time
+    data.index = pd.to_datetime(data['ObservationTime'])
+    data.index = data.index.tz_convert('UTC')
```
kandersolar, line 204: Other iotools functions don't convert to UTC; why do it here?

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,285 @@
...
+    See Also
+    --------
+    pvlib.iotools.get_solaranywhere
+
+    References
+    ----------
+    .. [1] `SolarAnywhere historical data file formats
+       <https://www.solaranywhere.com/support/historical-data/file-formats/>`_
+    """
+    with open(str(filename), 'r', encoding='iso-8859-1') as fbuf:
```
kandersolar, line 250: Mild preference to expose `encoding` as an optional parameter

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,285 @@
...
+    Read a SolarAnywhere formatted file into a pandas DataFrame.
+
+    The SolarAnywhere file format and variables are described in [1]_. Note,
+    the SolarAnywhere file format resembles the TMY3 file format but contains
+    additional variables and meatadata.
+
+    Parameters
+    ----------
+    fbuf: file-like object
+        File-like object containing data to read.
```
kandersolar, line 229: Should be `filename`

pvlib/tests/iotools/test_solaranywhere.py
```diff
@@ -0,0 +1,230 @@
...
+    pd.testing.assert_index_equal(data.index, time_series_index)
+    # Test one column
+    pd.testing.assert_series_equal(data['temp_air'], timeseries_temp_air)
+
+
+@requires_solaranywhere_credentials
+@pytest.mark.remote_data
+@pytest.mark.flaky(reruns=RERUNS, reruns_delay=RERUNS_DELAY)
+def test_get_solaranywhere_max_response_time(solaranywhere_api_key):
+    # Test if ValueError is raised if start/end is missing for non-TMY request
```
kandersolar, line 202: This comment doesn't apply to this test

pvlib/tests/iotools/test_solaranywhere.py
```diff
@@ -0,0 +1,230 @@
...
+            start=pd.Timestamp('2020-01-01 00:00:00+0000'),
+            end=pd.Timestamp('2020-01-05 12:00:00+0000'))
+
+
+# Second live test
+# True dynamics is False
+# Different variables
+
+# Mock tets:
+# TGY
```
kandersolar, line 230: Pointing out these comments so they don't get forgotten :)

kandersolar (commented):
A few comments.  Generally looks good to me!

pvlib/iotools/solaranywhere.py
AdamRJensen: Has now been implemented

pvlib/iotools/solaranywhere.py
AdamRJensen: I think you are correct good sir!

pvlib/iotools/solaranywhere.py
AdamRJensen: Also for the ``get_solaranywhere`` function?

AdamRJensen:
@kandersolar Could you help with the flake8 issue - I can't really tell what it is?

pvlib/iotools/solaranywhere.py
kandersolar: I don't think `get_solaranywhere` has to worry about encoding problems; network requests shouldn't be subject to users unwittingly re-encoding files.

cwhanse:
> @kandersolar Could you help with the flake8 issue - I can't really tell what it is?

Do you get the error if the csv files are read with pandas?

kandersolar:
> @kandersolar Could you help with the flake8 issue - I can't really tell what it is?

I think the problem is this:
1) To use flake8's `--diff` option, we are piping `git diff` into flake8 via stdin rather than letting flake8 discover and read files itself
2) That means flake8 has to identify .py files by parsing the `git diff` text provided on stdin, which means it wants to decode the stdin text to UTF-8 first
3) The CSV data file has that troublesome trademark character that cannot be decoded as UTF-8.  This character is included in the output of `git diff`
4) Thus flake8 breaks while trying to decode the data passed to its stdin

What to do about it?  Modify the `git diff` to only show output for `.py` files like this, maybe?

```bash
git diff upstream/main -- "*.py" | flake8 --exclude pvlib/version.py --ignore E201,E241,E226,W503,W504 --max-line-length 79 --diff
```

AdamRJensen:
> I think the problem is this:
> 
> 1. To use flake8's `--diff` option, we are piping `git diff` into flake8 via stdin rather than letting flake8 discover and read files itself
> 2. That means flake8 has to identify .py files by parsing the `git diff` text provided on stdin, which means it wants to decode the stdin text to UTF-8 first
> 3. The CSV data file has that troublesome trademark character that cannot be decoded as UTF-8.  This character is included in the output of `git diff`
> 4. Thus flake8 breaks while trying to decode the data passed to its stdin
> 
> What to do about it? Modify the `git diff` to only show output for `.py` files like this, maybe?
> 
> ```shell
> git diff upstream/main -- "*.py" | flake8 --exclude pvlib/version.py --ignore E201,E241,E226,W503,W504 --max-line-length 79 --diff
> ```

This solved the issue. Thanks!

.github/workflows/flake8.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           git remote add upstream https://github.com/pvlib/pvlib-python.git
           git fetch upstream $GITHUB_BASE_REF
       - name: Run Flake8 linter
-        run: git diff upstream/$GITHUB_BASE_REF HEAD | flake8
+        run: git diff upstream/main -- "*.py" | flake8
```
kandersolar, line 23:
```suggestion
        run: git diff upstream/$GITHUB_BASE_REF HEAD -- "*.py" | flake8
```

We should probably keep this part, just in case we ever start sending PRs to branches other than `main`.  Sorry I didn't think to do that in the original suggestion!

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,292 @@
...
+        In decimal degrees, north is positive (ISO 19115).
+    longitude: float
+        In decimal degrees, east is positive (ISO 19115).
+    api_key: str
+        SolarAnywhere API key.
+    start: datetime like, optional
+        First timestamp of the requested period. If a timezone is not
+        specified, UTC is assumed. Not applicable for TMY data.
+    end: datetime like, optional
+        Last timtestamp of the requested period. If a timezone is not
```
kandersolar, line 70:
```suggestion
        Last timestamp of the requested period. If a timezone is not
```

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,292 @@
...
+    # is separated by a slash, and the key-value pairs are separated by a
+    # colon. E.g., 'Data Version: 3.4 / Type: Typical Year / ...'
+    for i in ','.join(firstline).replace('"', '').split('/'):
+        if ':' in i:
+            k, v = i.split(':')
+            meta[k.strip()] = v.strip()
+
+    # Set index to UTC
+    data.index = pd.to_datetime(data['ObservationTime(GMT)'],
+                                format='%m/%d/%Y %H:%M', utc=True)
```
kandersolar, line 284:
```suggestion
    data.index = pd.to_datetime(data['ObservationTime(LST)'],
                                format='%m/%d/%Y %H:%M')
```
Maybe better to use the local standard time column?

pvlib/tests/conftest.py
```diff
@@ -95,6 +95,19 @@ def assert_frame_equal(left, right, **kwargs):
     not has_bsrn_credentials, reason='requires bsrn credentials')
 
 
+try:
+    # Attempt to load SolarAnywhere API key used for testing
+    # pvlib.iotools.get_solaranywhere
+    solaranywhere_api_key = os.environ["SOLARANYWHERE_API_KEY"]
+    has_solaranywhere_credentials = True
+except KeyError:
+    has_solaranywhere_credentials = False
```
kandersolar, line 104: I don't understand why codecov is claiming that line 102 isn't covered.  The CI logs show that the tests are running, so I don't see why this line wouldn't be getting hit.  No action needed; just recording my confusion...

pvlib/iotools/solaranywhere.py
kandersolar: If so, I think the tests will need to be adjusted

kandersolar (approved):
A few minor comments, but this looks good to me!  Note that I did not actually run any code.

pvlib/iotools/solaranywhere.py
AdamRJensen: I think you are right that LST is preferred. My main point for saying this is that it's much easier for a user to convert to UTC than to convert to LST,

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,297 @@
...
+        Variables to retrieve (described in [4]_), must include
+        'ObservationTime'. Available variables depend on whether historical or
+        TMY data is requested.
+    missing_data: {'Omit', 'FillAverage'}, default: 'FillAverage'
+        Method for treating missing data.
+    url: str, default: :const:`pvlib.iotools.solaranywhere.URL`
+        Base url of SolarAnywhere API.
+    map_variables: bool, default: True
+        When true, renames columns of the DataFrame to pvlib variable names
+        where applicable. See variable :const:`VARIABLE_MAP`.
```
cwhanse, line 99:
```suggestion
        where applicable. See :const:`VARIABLE_MAP`.
```

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,297 @@
...
+    return data, meta
+
+
+def read_solaranywhere(filename, map_variables=True, encoding='iso-8859-1'):
+    """
+    Read a SolarAnywhere formatted file into a pandas DataFrame.
+
+    The SolarAnywhere file format and variables are described in [1]_. Note,
+    the SolarAnywhere file format resembles the TMY3 file format but contains
+    additional variables and meatadata.
```
cwhanse, line 228:
```suggestion
    additional variables and metadata.
```

Too bad we can't coin this term for substantial metadata, as opposed to thin data.

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,297 @@
...
+
+    Parameters
+    ----------
+    filename: str
+        Filename
+    map_variables: bool, default: True
+        When true, renames columns of the DataFrame to pvlib variable names
+        where applicable. See variable :const:`VARIABLE_MAP`.
```
cwhanse, line 236:
```suggestion
        where applicable. See :const:`VARIABLE_MAP`.
```

pvlib/tests/iotools/test_solaranywhere.py
```diff
@@ -0,0 +1,287 @@
...
+TESTFILE_HIGH_RESOLUTION = DATA_DIR / 'Burlington, United States SolarAnywhere Time Series 20210101 to 20210103 Lat_44_4675 Lon_-73_2075 SA format.csv'  # noqa: E501
+# TGY test file (v3.6) containing GHI/DHI and temperature.
+# Note, the test file only contains the first three days.
+TESTFILE_TMY = DATA_DIR / 'Burlington, United States SolarAnywhere Typical GHI Year Lat_44_465 Lon_-73_205 SA format.csv'  # noqa: E501
+
+
+@pytest.fixture(scope="module")
+def solaranywhere_api_key():
+    """Supplies the pvlib's SolarAnywhere API key for testing purposes.
+    Users can freely registre for an API key."""
```
cwhanse, line 18:
```suggestion
    Users can freely register for an API key."""
```

cwhanse (approved):
One comment about requirements `start` and `end`, rest are spelling-level nits.

docs/sphinx/source/whatsnew/v0.9.2.rst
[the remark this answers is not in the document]
AdamRJensen: There's no change there so I can't remove the file... I think it's fine leaving it. @kandersolar leave a thumb if you have an opinion

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,297 @@
...
+            "WeatherDataSource": source,
+            "MissingDataHandling": missing_data,
+        }
+    }
+
+    if true_dynamics:
+        payload['Options']['ApplyTrueDynamics'] = True
+
+    if probability_of_exceedance is not None:
+        if type(probability_of_exceedance) != int:
```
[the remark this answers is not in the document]
AdamRJensen: I switched to `isinstance`

pvlib/iotools/solaranywhere.py
```diff
@@ -0,0 +1,297 @@
...
+            raise ValueError('`probability_of_exceedance` must be an integer')
+        payload['Options']['ProbabilityOfExceedance'] = \
+            probability_of_exceedance
+
+    # Add start/end time if requesting non-TMY data
+    if (start is not None) or (end is not None):
```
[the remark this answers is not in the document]
AdamRJensen: If a user specified either `start` or `end` we would want them to get a useful feedback. Such as `None` cannot be converted to timestamp which is what would happen now. If you use `and` then the error message would first happen much later and be more difficult to interpret.

kandersolar:
Ok, going ahead with the merge.  Thanks @AdamRJensen!

KWagnerCPR:
We greatly appreciate the work on this!

PR #1511, opened 2022-08-01
Make leap_day=True default for PSM3 (deprecation)

### LINKED ISSUE #1481: Set leap_day=True as default in pvlib.iotools.get_psm3()
Several functions require equal timesteps with no gaps (e.g., `pvlib.temperature.prilliman` and `detect_clearsky`). Having weather data that excludes leap day will cause an error. Specifically, using `pvlib.iotools.get_psm3` for a leap year with the default `leap_day=True` would be a common reason for encountering this error.

As suggested by @adriesse here https://github.com/pvlib/pvlib-python/issues/1476#issuecomment-1163130675, perhaps the default should be changed to `leap_day=True`?

**Additional context**
Related to https://github.com/pvlib/pvlib-python/issues/1476. 

On the topic of "is this a good idea", some comments from issue 1476:
> "I wonder if there are any compelling reasons to keep it as False, other than the fact that other tools (e.g., NREL SAM) can't handle leap day." 

(by me)

> "I suspect `leap_day=False` is rooted in matching the PSM3 API's decision to set it False by default." 

(by @kanderso-nrel)

AdamRJensen:
- [x] Closes #1481 
 - [x] Tests added
 - ~~[ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).~~

It has been decided to change the default value for the `leap_day` argument in the `get_psm3` function from False to True (see #1481). This PR adds a deprecation warning stating that the default will be changed to `leap_day=True` in pvlib 0.11.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.9.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.2.rst
@@ -24,6 +24,10 @@ Bug fixes
   where passing localized timezones with large UTC offsets could return
   rise/set/transit times for the wrong day in recent versions of ``ephem``
   (:issue:`1449`, :pull:`1448`)
+* :py:func:`pvlib.iotools.get_psm3` now raises a deprecation warning if
+  the `leap_day` parameter is not specified. Starting in pvlib 0.11.0
+  `leap_day` will be default to True instead of False.
+  (:issue:`1481`, :pull:`1511`)
 
 
 Testing
--- a/docs/sphinx/source/whatsnew/v0.9.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.2.rst
@@ -24,9 +24,9 @@ Bug fixes
   where passing localized timezones with large UTC offsets could return
   rise/set/transit times for the wrong day in recent versions of ``ephem``
   (:issue:`1449`, :pull:`1448`)
-* :py:func:`pvlib.iotools.get_psm3` now raises a deprecation warning if
-  the `leap_day` parameter is not specified. Starting in pvlib 0.11.0
-  `leap_day` will be default to True instead of False.
+* :py:func:`pvlib.iotools.get_psm3` now raises a deprecation warning for
+  single-year requests where the `leap_day` parameter is not specified.
+  Starting in pvlib 0.11.0 `leap_day` will default to True instead of False.
   (:issue:`1481`, :pull:`1511`)
 
 
--- a/docs/sphinx/source/whatsnew/v0.9.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.2.rst
@@ -24,8 +24,8 @@ Bug fixes
   where passing localized timezones with large UTC offsets could return
   rise/set/transit times for the wrong day in recent versions of ``ephem``
   (:issue:`1449`, :pull:`1448`)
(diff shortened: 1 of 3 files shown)


pvlib/iotools/psm3.py
```diff
@@ -165,6 +165,14 @@ def get_psm3(latitude, longitude, api_key, email, names='tmy', interval=60,
     attributes = [amap.get(a, a) for a in attributes]
     attributes = list(set(attributes))  # remove duplicate values
 
+    if leap_day is None:
```
kanderso-nrel, line 168: Probably we should not warn if the request is for a TMY/TGY/TDY dataset, right?  No point in warning for cases where the parameter can't have an effect.

kanderso-nrel (commented):
Needs a whatsnew entry too

pvlib/iotools/psm3.py
AdamRJensen: Indeed, great point! I have modified it to be the following:

        if (leap_day is None) & (not names.startswith('t')):

pvlib/iotools/psm3.py
```diff
@@ -165,6 +165,14 @@ def get_psm3(latitude, longitude, api_key, email, names='tmy', interval=60,
     attributes = [amap.get(a, a) for a in attributes]
     attributes = list(set(attributes))  # remove duplicate values
 
+    if (leap_day is None) & (not names.startswith('t')):
```
kanderso-nrel, line 168: Nitpick: `and` and `&` have identical semantics when applied to booleans as far as I know (except short-circuiting), but stylistically I prefer to reserve `&` for situations where it actually behaves differently (e.g. element-wise array logic or bit fiddling).  But I don't think PEP8 opines on this point, and from how often I see people using `&` for scalar booleans I guess a lot of people would disagree.

kanderso-nrel (approved):
OK with me!

pvlib/iotools/psm3.py
AdamRJensen: Sure, why not :grinning:

AdamRJensen:
@pvlib/pvlib-maintainer This PR is ready for a second review if anyone wants to take a look

docs/sphinx/source/whatsnew/v0.9.2.rst
```diff
@@ -24,6 +24,10 @@ Bug fixes
   where passing localized timezones with large UTC offsets could return
   rise/set/transit times for the wrong day in recent versions of ``ephem``
   (:issue:`1449`, :pull:`1448`)
+* :py:func:`pvlib.iotools.get_psm3` now raises a deprecation warning for
+  single-year requests where the `leap_day` parameter is not specified.
```
cwhanse, line 28:
```suggestion
* :py:func:`pvlib.iotools.get_psm3` now raises a deprecation warning if
  the `leap_day` parameter is not specified in a single-year request.
```

cwhanse (approved):

PR #1524, opened 2022-08-13
Spectral mismatch calculation code

### LINKED ISSUE #1523: Spectral mismatch calculation
Currently the only function in pvlib that uses spectral irradiance is `spectrl2`.  

This feature will provide a function to calculate the spectral mismatch for a given spectral response and spectral irradiance in relation to the reference spectrum.  Its intended application is to estimate module performance gains or losses due spectral irradiance in the field differing from the reference spectrum used in the lab for module power rating.

**Additional context**
Efforts are ongoing to assemble and publish both spectral irradiance and spectral response data, such as:
https://datahub.duramat.org/project/about/spectral-irradiance-data-and-resources
The new feature will make it easier to make use of those spectral resources.

adriesse:
- [x] Closes #1523 23 
 - [x] Tests added


adriesse:
Empty module: ```mismatch```.  I anticipate some more broadly usable functions could find a home here over time, so a more generic name might be considered.

adriesse:
So we will probably have some interesting discussion on the type and shape of the outputs.  While working on this I noticed that spectrl2 has wavelength as the first dimension.  When converted to a dataframe this make wavelength the row index.  I would prefer to have wavelength as the last dimension, which I find works more naturally.

cwhanse:
What is the envisioned scope of `pvlib/spectrum`? Among other purposes, it seems like a natural place to collect the functions for spectral mismatch models that are currently in `pvsystem` and `atmosphere`. And add some code to help use `pvlib` with SMARTS or other such tools.

adriesse:
Yes, I think it would make some sense to move those two functions.  At the moment I can't imagine ```spectrum``` growing a lot more.

adriesse:
I've been thinking about numpy friendliness here, and I am tending toward making these functions pandas only.  The wavelengths values are an integral part of the data so it makes a lot of sense to me to carry them around as row or column index.

adriesse:
Ok, I think it works now.  :)

adriesse:
I think this is ready for review.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    Calculate the spectral mismatch under a given measured spectrum with
+    respect to a reference spectrum.
+
+    Parameters
+    ----------
+    sr: pandas.Series
+        The spectral response of one (photovoltaic) device.
+        The index of the Series must contain wavelength values in nm.
+    e_sun: pandas.DataFrame or pandase.Series
```
cwhanse, line 116:
```suggestion
    e_sun: pandas.DataFrame or pandas.Series
```

cwhanse (commented):
Some comments while I'm working on wrapping my brain around this

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,164 @@
...
+import pvlib
+import numpy as np
+import pandas as pd
+from scipy.interpolate import interp1d
+import os
+
+
+def get_sample_sr(wavelength=None):
+    '''
+    Generate a generic smooth c-si spectral response for tests and experiments.
```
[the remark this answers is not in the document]
adriesse: Spectral response data should generally be in data files I think.  (Lots coming.)  This is really just something for people who have nothing.

kanderso-nrel:
>  Not sure whether I cause all the automated test failure--I hope not.

At least some of the failures are because `xlrd` is needed to parse the excel file but it's not installed as a pvlib dependency.  Maybe better to convert the file to CSV?  Doesn't seem like a big loss to drop the `Chart1` sheet.

adriesse:
Should I try attempt to change [docs/sphinx/source/reference](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/reference), or leave that to someone more familiar with this?

kanderso-nrel:
Certainly not opposed to housing mismatch functions in `pvlib.spectrum.mismatch`, but should an alias be set up in `pvlib/spectrum/__init__.py` so that the functions are also accessible as e.g. `pvlib.spectrum.calc_spectral_mismatch` (in addition to `pvlib.spectrum.mismatch.calc_spectral_mismatch`)?  We did that for `spectrl2`.

I wonder if `spectrl2` ought to have been pandas-only as well.  

> Should I try attempt to change [docs/sphinx/source/reference](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/reference)

Please do!  I think adding a couple entries at the bottom of `docs/sphinx/source/reference/effects_on_pv_system_output.rst` would do.

adriesse:
Ok, I made those adjustments.

pvlib/spectrum/mismatch.py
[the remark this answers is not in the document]
cwhanse: I still vote for adding a kwarg now, rather than later.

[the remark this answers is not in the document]
adriesse: Are you thinking of a NotImplemented exception for values other thane "csi" this time around?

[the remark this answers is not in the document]
cwhanse: Something like that, yes

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+    Parameters
+    ----------
+    sr: pandas.Series
+        The relative spectral response of one (photovoltaic) test device.
+        The index of the Series must contain wavelength values in nm. [-]
+    e_sun: pandas.DataFrame or pandas.Series
+        One or more measured solar irradiance spectra in a pandas.DataFrame
+        having wavelength in nm as column index.  A single spectrum may be
+        be given as a pandas.Series having wavelength in nm as index.
+        [W/m^2/nm]
```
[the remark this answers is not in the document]
adriesse: Spectral data files often have a column for each wavelength.  It's not so much about where the wavelength are, but rather about where the timestamps are.

adriesse:
Can someone suggest how to remove docs/sphinx/source/whatsnew/v0.9.2.rst from my PR?

kanderso-nrel:
I think (hope?) the 0.9.2 whatsnew changes will go away by resolving the merge conflict (`Resolve conflicts` button here on GH, or can be done locally).  When resolving the conflict I would keep the `master` version in each case.

adriesse:
I feel a bit silly here, but I don't know where to click to actually resolve the conflicts it shows.

kanderso-nrel:
Oh!  This guide might be useful: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-on-github

tl;dr: resolving the conflicts is done mostly via keyboard rather than mouse.  In an editable text box, it presents you with the two versions (this PR's version and master's version) of each conflict, demarcated by `===` and `>>>` lines.  Pick the version you want to keep, delete the other, delete the demarcation lines, and commit the changes.  Doing it locally is basically the same except you use your own text editor instead of github's in-browser editor.

adriesse:
Thanks, I had no idea that box was editable!!

markcampanelli:
@adriesse I have just started to take a look at this. I hit some snags trying to pull the branch, but I think that's on my end. Hopefully I'll have that sorted soon.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+    Parameters
+    ----------
+    wavelength: 1-D sequence of numeric, optional
+        The wavelengths at which the spectrum should be interpolated.
+        By default the 2002 wavelengths of the standard are returned. [nm]
+
+    Returns
+    -------
+    am15g: pandas.Series
+        The AM1.5g standard spectrum indexed by wavelength in nm. [W/m^2/nm]
```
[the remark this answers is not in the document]
adriesse: Sure.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+
+    Notes
+    -----
+    This function uses linear interpolation.  If the reference spectrum is too
+    coarsely interpolated, its integral may deviate from the standard value
```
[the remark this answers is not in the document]
adriesse: Sure.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,164 @@
...
+        The wavelengths at which the sample sr should be interpolated. [nm]
+        By default the wavelengths are from 280 to 1200 in 5 nm intervals.
+
+    Notes
+    -----
+    This sr is based on measurements taken on a reference cell.
+    The measured data points have been adjusted by PV Performance Labs so that
+    standard cubic spline interpolation produces a curve without oscillations.
+    '''
+    SR_DATA = np.array([[ 290, 0.00],
```
AdamRJensen, line 43: Maybe name "SR_SAMPLE_DATA"

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,164 @@
...
+    Notes
+    -----
+    This function uses linear interpolation.  If the reference spectrum is too
+    coarsely interpolated, its integral may deviate from 1000.37 W/m2.
+
+    More information about reference spectra is found here:
+    https://www.nrel.gov/grid/solar-resource/spectra-am1.5.html
+
+    The original file containing the reference spectra can be found here:
+    https://www.nrel.gov/grid/solar-resource/assets/data/astmg173.xls
```
AdamRJensen:
```suggestion
    More information about reference spectra can be found in [1]_.
    The original file containing the reference spectra can be found in [2]_.

    References
    ----------
    .. [1] `Reference Air Mass 1.5 Spectra
       <https://www.nrel.gov/grid/solar-resource/spectra-am1.5.html>`_
    .. [2] `ASTM G-173-03 spreadsheet
       <https://www.nrel.gov/grid/solar-resource/assets/data/astmg173.xls>`_
```

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+    The measured data points have been adjusted by PV Performance Labs so that
+    standard cubic spline interpolation produces a curve without oscillations
+    as shown in [1]_.
+
+    References
+    ----------
+    .. [1] Driesse, Anton, and Stein, Joshua. "Global Normal Spectral
+       Irradiance in Albuquerque: a One-Year Open Dataset for PV Research".
+       United States 2020. https://doi.org/10.2172/1814068.
+       https://www.osti.gov/servlets/purl/1814068.
```
AdamRJensen, line 39: The links resolve to the same URL, so maybe just keep the doi link. Also, then you could format the URL as a hyperlink.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+    References
+    ----------
+    .. [1] ASTM "E973-16 Standard Test Method for Determination of the
+       Spectral Mismatch Parameter Between a Photovoltaic Device and a
+       Photovoltaic Reference Cell"
+    .. [2] ASTM "E973-10 Standard Test Method for Determination of the
+       Spectral Mismatch Parameter Between a Photovoltaic Device and a
+       Photovoltaic Reference Cell"
+    .. [3] IEC 60904-7 "Computation of the spectral mismatch correction
+       for measurements of photovoltaic devices"
```
AdamRJensen, line 186: Could provide hyperlinks for these references. The ASTM links look fairly stable, e.g., https://www.astm.org/e0973-16r20.html

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+
+    sr.index.name = 'wavelength'
+    sr.name = 'sr'
+
+    return sr
+
+
+def get_am15g(wavelength=None):
+    '''
+    Read the ASTM G173-03 AM1.5 global tilted spectrum, optionally interpolated
```
markcampanelli, line 78: According to a strict adherence to the G173-03 standard, I think the word "global" here (and elsewhere) should be "hemispherical". I recognize that "global" is nevertheless very common parlance, so perhaps just make a note of the synonymous usage here with "hemispherical".

From G173-03---

> 3.3.4.1 Discussion—For the special condition of a horizon-
tal plane the hemispherical solar irradiance is properly termed
global solar irradiance, E G . Incorrectly, global tilted, or total
global irradiance is often used to indicate hemispherical
irradiance for a tilted plane. In case of a sun-tracking receiver,
this hemispherical irradiance is commonly called global nor-
mal irradiance. The adjective global should refer only to
hemispherical solar radiation on a horizontal, not a tilted,
surface.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+
+        am15g = pd.Series(data=interpolator(wavelength), index=wavelength)
+
+    am15g.index.name = 'wavelength'
+    am15g.name = 'am15g'
+
+    return am15g
+
+
+def calc_spectral_mismatch(sr, e_sun, e_ref=None):
```
markcampanelli, line 133: I think the name of this function should somehow reflect that it is specific to the case of a broadband reference device.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+    achieve alternate behavior the spectral response can be transformed
+    before calling this function.
+
+    The standards describing mismatch calculations focus on indoor laboratory
+    applications, but are applicable to outdoor performance as well.
+    The most recent version of ASTM E973 [1]_ is somewhat more difficult to
+    read because it includes adjustments for temperature dependency of the
+    spectral response, which led to a formulation using quantum efficiency
+    (QE). IEC 60904-7 is clearer and also discusses the use of a
+    broadband reference device.
```
markcampanelli, line 175: The temperature dependence of the SR would typically be relevant for PV devices deployed outside. NREL found this effect to be significant in their primary reference cell calibrations that are performed outdoors against broadband cavity radiometers. For other applications, perhaps the effect could be considered negligible.

markcampanelli (changes requested):
Here is my initial review base on reading the code, but not yet running it. Next, I will try to run some computation comparisons with the M-calculator in PVfit.

One question: Where would the spectral mismatch result here be "inserted" into the `pvlib-python` performance modeling workflow, and could you point me to an example of this?

pvlib/spectrum/mismatch.py
adriesse: However, none of those will be downloadable for free.

pvlib/data/astm_g173_am15g.csv
```diff
@@ -0,0 +1,2003 @@
+wavelength,am15g
```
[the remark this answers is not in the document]
adriesse: Sure

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,164 @@
...
+"""
+
+import pvlib
+import numpy as np
+import pandas as pd
+from scipy.interpolate import interp1d
+import os
+
+
+def get_sample_sr(wavelength=None):
```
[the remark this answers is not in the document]
AdamRJensen: Agreed `get_example_spectral_response` would be good.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+                                kind='linear',
+                                bounds_error=False,
+                                fill_value=0.0,
+                                copy=False,
+                                assume_sorted=True)
+
+        am15g = pd.Series(data=interpolator(wavelength), index=wavelength)
+
+    am15g.index.name = 'wavelength'
+    am15g.name = 'am15g'
```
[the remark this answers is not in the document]
AdamRJensen: I don't recall seeing any other column names in pvlib that have units fyi.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+    ----------
+    .. [1] ASTM "G173-03 Standard Tables for Reference Solar Spectral
+        Irradiances: Direct Normal and Hemispherical on 37° Tilted Surface."
+    '''
+    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Aug. 2022
+
+    pvlib_path = pvlib.__path__[0]
+    filepath = os.path.join(pvlib_path, 'data', 'astm_g173_am15g.csv')
+
+    am15g = pd.read_csv(filepath, index_col=0, squeeze=True)
```
markcampanelli, line 115: This line issues a deprecation warning:
`pvlib-python/pvlib/spectrum/mismatch.py:115: FutureWarning: The squeeze argument has been deprecated and will be removed in a future version. Append .squeeze("columns") to the call to squeeze.`

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,213 @@
...
+    sr_sun = np.interp(e_sun.T.index, sr.index, sr, left=0.0, right=0.0)
+    sr_ref = np.interp(e_ref.T.index, sr.index, sr, left=0.0, right=0.0)
+
+    # a helper function to make usable fraction calculations more readable
+    def integrate(e):
+        return np.trapz(e, x=e.T.index, axis=-1)
+
+    # calculate usable fractions
+    uf_sun = integrate(e_sun * sr_sun) / integrate(e_sun)
+    uf_ref = integrate(e_ref * sr_ref) / integrate(e_ref)
```
markcampanelli, line 205: Concerning the validity of these two computations, consider these words of caution from IEC 60904-7 (Ed. 4, 2009)---

> In Formula (6) the two integrals in the first fraction have to be taken over the entire
wavelength range of sensitivity of the thermopile detector. This in general poses a problem,
as the sensitivity range of such detectors is larger than the range readily measurable with
spectroradiometers.

I'm trying to put together a clear example where this leads to error in the computation, as well as some idea of the magnitude. (IIRC, this is also where the energy beyond 4 microns in the standard spectrum matters, which is specified differently between ASTM G173-03 and 60904-3 (2008).)

pvlib/spectrum/mismatch.py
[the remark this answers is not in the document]
adriesse: Rather not

pvlib/spectrum/mismatch.py
adriesse: Sure

pvlib/spectrum/mismatch.py
adriesse: I don't think it's necessarily negligible, but the information needed for the calculation is even more unavailable than SR data at 25C.

pvlib/spectrum/mismatch.py
adriesse: If only our present global squeeze could be deprecated and removed in a future version!

pvlib/spectrum/mismatch.py
adriesse: The words of caution are certainly applicable, but I think an error in the result can only arise from a deficiency in the inputs.

pvlib/spectrum/mismatch.py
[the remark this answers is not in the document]
AdamRJensen: @adriesse When suggestions are made such as the one by Cliff you can click the "Commit suggestion" button, thereby avoiding having to make the change yourself. And for trivial stuff it probably makes sense to click the "resolve button" after the change has been made to make the PR conversation make manageable to read through. Just a hint in case you were unaware :)

pvlib/spectrum/mismatch.py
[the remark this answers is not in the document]
adriesse: I have no problems making a name and parameter change to pass the review but I think it would be inappropriate to put additional technology options into this function in the future.  I believe SR data should be in data files in general.

pvlib/spectrum/mismatch.py
adriesse: I think it's fine as an internal name in the function.

pvlib/spectrum/mismatch.py
adriesse: How about: ```calc_spectral_mismatch_to_broadband()```

pvlib/spectrum/mismatch.py
adriesse: It would be good to set a general guideline for when to include hyperlinks.

adriesse:
Ok, I've resolved a bunch of stuff.  There were some that perhaps would have been more appropriately marked as resolved by the originators, but I don't know whether that is possible or common practice.

pvlib/spectrum/mismatch.py
markcampanelli: I strongly suggest that this algorithm be updated to allow more precise calculations when the total broadband "sun" irradiance is known. This covers, for example, the case for primary reference cell calibrations at NREL where a cavity radiometer provides the total irradiance alongside a spectral radiometer with a smaller response window. (This is where the discrepancy was previously discovered as a significant source of error. Admittedly, this level of error may not be important for other types of workflows.)

Here is a suggestion on how to address this using two changes:

1) In the function `get_am15g`, scale this spectrum from ASTM G173-03 so that it matches the broadband spectrum in IEC  60904-3 (2008), in which the tail irradiance is fully specified and the full curve (to lambda=infinity) integrates to precisely 1000 W/m^2. Namely, change this line in `get_am15g`
```python
am15g = pd.read_csv(filepath, index_col=0, squeeze=True)
```
to
```python
    am15g_raw = pd.read_csv(filepath, index_col=0, squeeze=True)
    am15g_raw_integral_to_4_micron = np.trapz(am15g_raw, x=am15g_raw.index, axis=-1)  # W/m^2
    iec_integral_to_4_micron = 997.47  # W/m^2, from IEC 60904-3 (2008)
    am15g = iec_integral_to_4_micron / am15g_raw_integral_to_4_micron * am15g_raw
```

2) Pass optional total irradiance parameters to the function `calc_spectral_mismatch`, namely
```python
def calc_spectral_mismatch(sr, e_sun, e_sun_tot=None, e_ref=None, e_ref_tot=1000):
```
and then add the following logic to `calc_spectral_mismatch` before the usable fraction calculations:
```python
    if e_sun_tot is None:
        # Use integrated spectral irradiance, typically an approximation
        # that misses "tail" of distribution.
        e_sun_tot = integrate(e_sun)

    if e_ref_tot is None:
        # Use integrated spectral irradiance, typically an approximation
        # that misses "tail" of distribution.
        e_ref_tot = integrate(e_ref)

    # calculate usable fractions
    uf_sun = integrate(e_sun * sr_sun) / e_sun_tot
    uf_ref = integrate(e_ref * sr_ref) / e_ref_tot
```

I have an example that shows an error on the order of 0.25% in the spectral correction factor from ignoring the integral of the tail. This well approximates the degree of error that I recall discovering at NREL. I am happy to share this out separately but am omitting here for brevity.

I have also reached out to Carl Osterwald about the scaling and "tail" discrepancies between ASTM G173-03 and IEC 60904-3 (2008), which seems to persist.

Finally, I still have some questions about the numerical effect of partitioning on the integrations for coarse SR functions, but I would like to settle the above first before further investigation.

markcampanelli:
@adriesse I added a new comment under the resolved comment for the usable fraction calculations that may not show up as new for you.

pvlib/spectrum/mismatch.py
markcampanelli: One change to the above: I recommend that no default value be provided for `e_sun_tot` in `calc_spectral_mismatch`, so that the user must explicitly provide a `None` value in order to use the approximation via the integral of the spectral irradiance over a partial interval, i.e.,
```python
def calc_spectral_mismatch(sr, e_sun, e_sun_tot, e_ref=None, e_ref_tot=1000):
```

pvlib/spectrum/mismatch.py
adriesse: Actually, I am going to update my suggestion to: ```calc_field_spectral_mismatch()```

pvlib/spectrum/mismatch.py
adriesse: I understand what you're talking about and had multiple exchanges with Carl on the subject some years ago.  If the sun produced something close to the reference spectrum we would expect about 2.5 or 2.9 W/m^2 beyond 4000 nm.

The main focus in this work is on modules and systems in the field, where someone is lucky if they have a spectro-radiometer that goes to 1700 nm.  What is beyond the measurement range is essentially unknown, therefore, by default this function also truncates the reference spectrum to 1700 nm (in this case).  This is not the true mismatch to full broadband, but the portion of the mismatch that is calculable based on the available measurements.  The tail beyond 4000 nm does not make any difference here.

Actually, an advanced user can already achieve what you propose simply by providing a suitable reference spectrum of their own.  They can take the far IR portion from E1125 and append it, for example, and scale it as they like, so this does not require changes in the code. 

But if you do that, you also have to extend the measured sun spectra from 1100 or 1700 to 10000 nm (or infinity) or at least estimate its integral that far out.  NREL measures to 2400 nm and makes this estimate under optimal clear sky conditions, which is far easier than for all-sky field applications.

I think the functions work correctly for the intended scope of this feature, which should perhaps have been more clearly defined.

pvlib/spectrum/mismatch.py
cwhanse: I would prefer `calc_spectral_mismatch_field` to make space for `_lab`, someday.

docs/sphinx/source/reference/effects_on_pv_system_output.rst
```diff
@@ -48,3 +48,6 @@ Spectrum
    :toctree: generated/
 
    spectrum.spectrl2
+   spectrum.get_sample_sr
```
cwhanse, line 51:
```suggestion
   spectrum.get_example_spectral_response
```

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,219 @@
...
+    Parameters
+    ----------
+    wavelength: 1-D sequence of numeric, optional
+        Wavelengths at which the spectral response should be interpolated.
+        By default the wavelengths are from 280 to 1200 in 5 nm intervals. [nm]
+
+    Returns
+    -------
+    sr: pandas.Series
+        The relative spectral response indexed by wavelength in nm. [-]
```
cwhanse, line 25:
```suggestion
        The relative spectral response indexed by ``wavelength`` in nm. [-]
```

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def get_example_spectral_response(wavelength=None):
+    '''
+    Generate a generic smooth spectral response (SR) for tests and experiments.
+
+    Parameters
+    ----------
+    wavelength: 1-D sequence of numeric, optional
+        Wavelengths at which the spectral response should be interpolated.
```
cwhanse, line 19:
```suggestion
        Wavelengths at which the spectral response is returned.
```

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,219 @@
...
+    -------
+    sr: pandas.Series
+        The relative spectral response indexed by wavelength in nm. [-]
+
+    Notes
+    -----
+    This spectral response is based on measurements taken on a c-Si cell.
+    The measured data points have been adjusted by PV Performance Labs so that
+    standard cubic spline interpolation produces a curve without oscillations
+    as shown in [1]_, which makes it suitable for experimenting with spectral
```
cwhanse, line 32:
```suggestion
    The measured data points are interpolated using a cubic spline that produces
    a curve without oscillations as shown in [1]_, which makes it suitable for
    experimenting with spectral
```

The phrasing "adjusted by" might invite questions here which may be better left to the reference, unless we add to this note and explain what the adjustment was.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,219 @@
...
+    to the specified wavelength(s).
+
+    Global (tilted) irradiance includes direct and diffuse irradiance from sky
+    and ground reflections, and is more formally called hemispherical
+    irradiance (on a tilted surface).
+
+    Parameters
+    ----------
+    wavelength: 1-D sequence of numeric, optional
+        The wavelengths at which the spectrum should be interpolated.
```
cwhanse, line 88:
```suggestion
        The wavelengths at which the spectrum is interpolated.
```

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,219 @@
...
+        By default the 2002 wavelengths of the standard are returned. [nm]
+
+    Returns
+    -------
+    am15g: pandas.Series
+        The AM1.5g standard spectrum indexed by wavelength in nm. [(W/m^2)/nm]
```
cwhanse, line 94:
```suggestion
        The AM1.5g standard spectrum indexed by ``wavelength`` in nm. [(W/m^2)/nm]
```

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,219 @@
...
+    Returns
+    -------
+    smm: pandas.Series or float if a single measured spectrum is provided. [-]
+
+    Notes
+    -----
+    If the default reference spectrum is used it is linearly interpolated
+    to the wavelengths of the measured spectrum.  To achieve alternate
+    behavior e_ref can be transformed before calling this function and
+    provided as an argument.
```
cwhanse, line 168:
```suggestion
    If the default reference spectrum ``e_ref`` is used, it is linearly interpolated
    to the wavelengths of the measured spectrum ``e_sun``. 
```

What is meant by "To achieve alternate behavior e_ref can be transformed before calling this function and provided as an argument." If `e_ref` is default or input, it is interpolated to the frequencies of `sr`, so it's not clear what the alternate behavior is.

cwhanse (approved):
Thanks @adriesse

pvlib/spectrum/mismatch.py
markcampanelli: Without supplying total irradiances separately, I don't agree that I could use the code as is to confidently compute a primary reference cell calibration using the IEC 60904-3 (2008) reference spectrum and a cavity radiometer for the measured broadband irradiance. I think users should be fairly warned that this usage is out of scope.

I will run some calculations for a spectro-radiometer cutoff at 1700nm. I anticipate that the error in the spectral correction could go significantly higher than 0.25% in this case.

adriesse: I should add that adding extra parameters for integrated values is not a bad idea in itself.  They're just not needed for the field application and would create more opportunities for mistakes by non-expert users.

markcampanelli: I do appreciate this point about usability. I think I come from the viewpoint that the algorithm should ask by default for the inputs to get the theoretically correct computation and then make the user "opt in" to any approximations. That said, I see that this approach could lead to more confusion, especially if the approximation is typically (1) good and/or (2) the only possible solution given the available measurements.

In any case, here I think we have the outstanding question of when the truncation of two of the four integrals in the "field" spectral correction creates "cancelling errors". Hopefully I'll soon have some estimates to share on this.

markcampanelli:
Using the example data for the spectral radiance of the sun and the spectral responsivity of a Si reference device, the error in the SMM when truncating both integrals appears to small, on the order of 0.1%. One does not, however, want to only truncate the incomplete sun integral.
```
e_ref totals to 1.7 um, 4 um, and infinity (W/m^2): 942.88 997.47 1000.0
e_sun totals to 1.7 um, 4 um, and infinity (W/m^2): 668.028852782929 708.4855034837434 710.2825182549283
no_trunc smm: 0.9913868433999364
1.7 um single truncation smm: 1.054093308636961
4 um single truncation smm: 0.9939014139772988
1.7 um double truncation smm: 0.9902678450884946
4 um double truncation smm: 0.9923965546943567
1.7 um single truncation error in smm: 6.325125822930477 %
4 um single truncation error in smm: 0.25364171353523535 %
1.7 um double truncation error in smm: -0.11287201548935144 %
4 um double truncation error in smm: 0.10184836536235586 %
```

Here is the code that produced these results:
```python
import numpy as np
from pvlib import spectrum
from pvlib.tests import test_spectrum

e_ref = spectrum.get_am15g()  # 0.28 to 4 micron, scaled to "match" IEC 60904-3 Ed. 2 (2008)
e_ref_integral_to_1p7_micron = 942.88  # W/m^2, from IEC 60904-3 Ed. 2 (2008)
e_ref_integral_to_4_micron = 997.47  # W/m^2, from IEC 60904-3 Ed. 2 (2008)
e_ref_integral_to_infinity = 1000.  # W/m^2, from IEC 60904-3 Ed. 2 (2008)

# Get example solar spectrum from test fixture.
_, e_sun = test_spectrum.spectrl2_data.__pytest_wrapped__.obj()  # 0.3 to 4 micron
e_sun = e_sun.set_index('wavelength')
e_sun = e_sun.transpose().loc['specglo']

idx_to_1p7_micron = e_sun.T.index <= 1700
e_sun_integral_to_1p7_micron = np.trapz(e_sun[idx_to_1p7_micron], x=e_sun.T.index[idx_to_1p7_micron], axis=-1)
e_sun_integral_to_4_micron = np.trapz(e_sun, x=e_sun.T.index, axis=-1)  # 708.4855034837434 W/m^2
scale_factor_estimate = e_ref_integral_to_infinity / e_ref_integral_to_4_micron
e_sun_integral_to_infinity = scale_factor_estimate * e_sun_integral_to_4_micron

sr = spectrum.get_example_spectral_response()

no_trunc = spectrum.calc_spectral_mismatch_field(sr, e_sun, e_sun_integral_to_infinity, e_ref=e_ref, e_ref_tot=e_ref_integral_to_infinity)
single_trunc_1p7_micron = spectrum.calc_spectral_mismatch_field(sr, e_sun[idx_to_1p7_micron], None, e_ref=e_ref, e_ref_tot=e_ref_integral_to_infinity)
single_trunc_4_micron_ = spectrum.calc_spectral_mismatch_field(sr, e_sun, None, e_ref=e_ref, e_ref_tot=e_ref_integral_to_infinity)
double_trunc_1p7_micron = spectrum.calc_spectral_mismatch_field(sr, e_sun[idx_to_1p7_micron], None)
double_trunc_4_micron = spectrum.calc_spectral_mismatch_field(sr, e_sun, None)

print("e_ref totals to 1.7 um, 4 um, and infinity (W/m^2):", e_ref_integral_to_1p7_micron, e_ref_integral_to_4_micron, e_ref_integral_to_infinity)
print("e_sun totals to 1.7 um, 4 um, and infinity (W/m^2):", e_sun_integral_to_1p7_micron, e_sun_integral_to_4_micron, e_sun_integral_to_infinity)
print("no_trunc smm:", no_trunc)
print("1.7 um single truncation smm:", single_trunc_1p7_micron)
print("4 um single truncation smm:", single_trunc_4_micron_)
print("1.7 um double truncation smm:", double_trunc_1p7_micron)
print("4 um double truncation smm:", double_trunc_4_micron)
print("1.7 um single truncation error in smm:", 100*(single_trunc_1p7_micron / no_trunc - 1), "%")
print("4 um single truncation error in smm:", 100*(single_trunc_4_micron_ / no_trunc - 1), "%")
print("1.7 um double truncation error in smm:", 100*(double_trunc_1p7_micron / no_trunc - 1), "%")
print("4 um double truncation error in smm:", 100*(double_trunc_4_micron / no_trunc - 1), "%")
```

I had to change the source code in this PR in order to calculate the "no_trunc" result (doc strings omitted):
```python
def get_am15g(wavelength=None):
    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Aug. 2022
    # Changed by Mark Campanelli for testing purposes.

    pvlib_path = pvlib.__path__[0]
    filepath = os.path.join(pvlib_path, 'data', 'astm_g173_am15g.csv')

    am15g = pd.read_csv(filepath, index_col=0).squeeze()  # 0.28 to 4 microns
    am15g_integral_to_4_micron = np.trapz(am15g, x=am15g.index, axis=-1)  # W/m^2
    iec_integral_to_4_micron = 997.47  # W/m^2, from IEC 60904-3 Ed. 2 (2008)
    am15g_scaled_to_iec = iec_integral_to_4_micron / am15g_integral_to_4_micron * am15g

    if wavelength is not None:
        interpolator = interp1d(am15g_scaled_to_iec.index, am15g_scaled_to_iec,
                                kind='linear',
                                bounds_error=False,
                                fill_value=0.0,
                                copy=False,
                                assume_sorted=True)

        am15g_scaled_to_iec = pd.Series(data=interpolator(wavelength), index=wavelength)

    am15g_scaled_to_iec.index.name = 'wavelength'
    am15g_scaled_to_iec.name = 'am15g'

    return am15g_scaled_to_iec


def calc_spectral_mismatch_field(sr, e_sun, e_sun_tot, e_ref=None, e_ref_tot=None):
    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Aug. 2022.
    # Changed by Mark Campanelli for testing purposes.

    # get the reference spectrum at wavelengths matching the measured spectra
    if e_ref is None:
        e_ref = get_am15g(wavelength=e_sun.T.index)

    # a helper function to make usable fraction calculations more readable
    def integrate(e):
        return np.trapz(e, x=e.T.index, axis=-1)

    if e_sun_tot is None:
        # Use integrated spectral irradiance, typically an approximation
        # that misses "tail" of distribution.
        e_sun_tot = integrate(e_sun)

    if e_ref_tot is None:
        # Use integrated spectral irradiance, typically an approximation
        # that misses "tail" of distribution.
        e_ref_tot = integrate(e_ref)

    # interpolate the sr at the wavelengths of the spectra
    # reference spectrum wavelengths may differ if e_ref is from caller
    sr_sun = np.interp(e_sun.T.index, sr.index, sr, left=0.0, right=0.0)
    sr_ref = np.interp(e_ref.T.index, sr.index, sr, left=0.0, right=0.0)

    # calculate usable fractions
    uf_sun = integrate(e_sun * sr_sun) / e_sun_tot
    uf_ref = integrate(e_ref * sr_ref) / e_ref_tot

    # mismatch is the ratio or quotient of the usable fractions
    smm = uf_sun / uf_ref

    if isinstance(e_sun, pd.DataFrame):
        smm = pd.Series(smm, index=e_sun.index)

    return smm
```

I am guessing that double truncation is not a bad strategy in many (most?) cases, but I will point out that an alternate strategy (ref. Carl Osterwald) is to estimate the total sun spectrum by integrating the tail of the reference spectrum after scaling it to "match" some portion of the sun spectrum. I suppose we would have to do a bigger study to fairly compare the two approaches.

I have some additional thoughts now about how we could (optionally) include the total irradiance value(s) but still try to keep the function interface from being too confusing. (This idea is different than the code changes I used above.) That said, I would also be ok if the double truncation approximation was more clearly stated in the docstring and perhaps a note saying that the code is not appropriate for computations such as primary reference cell calibrations such as ASTM E1125. (It turns out that ASTM E1125 is where the differences between ASTM G173-03 and IEC 60904-3 is called out and handled.)

adriesse:
Thanks very much, @markcampanelli for checking this out so thoroughly and confirming at the very least that I'm not terribly mistaken!  :) The double truncation is more or less imposed when one of the inputs is pre-truncated in field measurements.   One additional easy test is to image the sun spectrum being equal to the reference spectrum, so you  want to get a factor of 1.0.  If the spectroradiometer only measures to 1700 nm, then the only way to get 1.0 is to truncate the reference spectrum to 1700 as well.

I've added a paragraph to the docstring to explain the truncation.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,232 @@
...
+    ----------
+    .. [1] ASTM "E973-16 Standard Test Method for Determination of the
+       Spectral Mismatch Parameter Between a Photovoltaic Device and a
+       Photovoltaic Reference Cell" :doi:`10.1520/E0973-16R20`
+    .. [2] ASTM "E973-10 Standard Test Method for Determination of the
+       Spectral Mismatch Parameter Between a Photovoltaic Device and a
+       Photovoltaic Reference Cell" :doi:`10.1520/E0973-10`
+    .. [3] IEC 60904-7 "Computation of the spectral mismatch correction
+       for measurements of photovoltaic devices"
+    """
```
markcampanelli, line 206: What about IEC 61853-3 (ed. 1.0), specifically, equation (6) in section 7.3?

This is where the implementation here does not cover that formula precisely.

adriesse: Could you elaborate a little bit on where you see the problem?

adriesse: In any case, the equation shows the pre-integrated values of the reference and sun/field spectra in a similar manner to your proposed function signature. Do the pre-integrated values need to be different from the values that would be obtained from an actual integration?

I have some comments about the problems with this calculation of the standard in the docstring of my implementation in [pvpltools](https://github.com/adriesse/pvpltools-python/blob/master/pvpltools/iec61853.py#L127).

adriesse:
@nicorie would you be interested in having a look?

pvlib/spectrum/mismatch.py
kanderso-nrel: Another idea: keep the function general (not specifically relative to broadband reference) add an optional `sr_ref` parameter that defaults to 1.0?

kanderso-nrel (approved):
Very nice PR @adriesse!  One minor change and one idea to consider below.

AdamRJensen (approved):
Minor comments that can be considered, but looks good :)

pvlib/spectrum/mismatch.py
markcampanelli: @adriesse Sorry that I have not been able to look at this closely again this week.

> Do the pre-integrated values need to be different from the values that would be obtained from an actual integration?

This is perhaps the key question, assuming that the IEC 61853-3 use case is an important one for this code addition. (I have some issues with this formula, but it is what it is.) At this point I am assuming that this use case is important, and it requires some consideration in addition to the "primary reference cell calibration with a cavity radiometer" use case that I previously examined here. I intend to look closer at this this weekend.

There is also one last use case, which I am personally interested in. This is for PVfit's "irradiance" calculation of `F=Isc/Isc0`. The spectral-mismatch approach here provides an alternative to the approach that I presented at the recent PVPMC workshop. While I would like to see this code cover this case to my satisfaction, please know that I am not trying to hold up this PR based on this criterion.

docs/sphinx/source/whatsnew/v0.9.3.rst
```diff
@@ -12,6 +12,9 @@ Enhancements
 * Low resolution altitude lookup map
   :py:func:`~pvlib.location.lookup_altitude`
   (:issue:`1516`, :pull:`1518`)
+* New module to calculate spectral mismatch
+  :py:func:`~pvlib.spectrum.mismatch.calc_spectral_mismatch_field`
+  (:issue:`1523`, :pull:`1524`)
```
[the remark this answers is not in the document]
adriesse: Will do.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,232 @@
...
+
+    Parameters
+    ----------
+    wavelength: 1-D sequence of numeric, optional
+        Wavelengths at which spectral response values are generated.
+        By default ``wavelength`` is from 280 to 1200 in 5 nm intervals. [nm]
+
+    Returns
+    -------
+    sr: pandas.Series
```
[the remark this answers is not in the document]
adriesse: See below (or above).

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,232 @@
...
+                            kind='cubic',
+                            bounds_error=False,
+                            fill_value=0.0,
+                            copy=False,
+                            assume_sorted=True)
+
+    sr = pd.Series(data=interpolator(wavelength), index=wavelength)
+
+    sr.index.name = 'wavelength'
+    sr.name = 'sr'
```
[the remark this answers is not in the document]
adriesse: I disagree, but will go with the majority opinion.  I believe anyone who uses this code will quickly learn what sr means and will appreciate not having to type it out in full.

pvlib/spectrum/mismatch.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def get_am15g(wavelength=None):
+    '''
+    Read the ASTM G173-03 AM1.5 global tilted spectrum, optionally interpolated
+    to the specified wavelength(s).
+
+    Global (tilted) irradiance includes direct and diffuse irradiance from sky
+    and ground reflections, and is more formally called hemispherical
+    irradiance (on a tilted surface).
```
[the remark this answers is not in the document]
adriesse: Sure.

pvlib/spectrum/mismatch.py
adriesse: I might prefer to see future functions for specific use cases such as calibration or energy rating.  A single generic function could be fewer lines of code, but may not be as inviting for users who are not as broadly informed on the subject.

I have been involved in many discussions about IEC 61853-3.  The next version will look different, probably skipping the mismatch or spectral correction factor in this form altogether.  Anyway, it would be hard to cover the current version in this function due to the use of banded solar spectral irradiance data.

pvlib/spectrum/mismatch.py
adriesse: Its a very reasonable suggestion, but perhaps it could be done in a future PR with some more debate about the range of interesting options and different ways of implementing them?

pvlib/spectrum/mismatch.py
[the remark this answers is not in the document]
markcampanelli: I see units in the source data file now, which is plenty. Thanks.

pvlib/spectrum/mismatch.py
markcampanelli: I derived equation (6) in section 7.3 in IEC 61853-3 (ed. 1.0) from first principles. Recall that this is used in IEC 61853-1 is to compute the equivalent irradiance under the standard spectrum that would generate the same short-circuit current in the device under the given operating condition, denote here as `E_equiv`. One issue I found is that technically the spectral response of the device should depend on temperature. A second issue to note is that the upper limit of integration on the integrals, 'lambda_e", is fixed at 4 microns, which corresponds to the assumed spectral-response cutoff of total POA irradiance as measured by a pyranometer and (presumably) reported in the profiles in IEC 61853-4. In the study below, a cutoff change to 1.7 microns requires scaling the total POA irradiance accordingly.

I then computed the SSM and `E_equiv` that would be computed by code under review vs. an implementation that covers equation (6) in IEC 61853-3. The error appears small for the 4 micron cutoff, but becomes very significant for the 1.7 micron cutoff. What is quite interesting, and what I did not expect, is that the SMM changes significantly based on the wavelength cutoff, and this change is necessary to compute a consistent `E_equiv`! Of course, these results are contingent upon 3rd-party review :).

This is the modified source code to run the comparison. It can be used to compute both algorithms.
```python
def get_am15g(wavelength=None):
    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Aug. 2022
    # Modified by Mark Campanelli for testing purposes.

    pvlib_path = pvlib.__path__[0]
    filepath = os.path.join(pvlib_path, 'data', 'astm_g173_am15g.csv')

    am15g = pd.read_csv(filepath, index_col=0).squeeze()  # 0.28 to 4 microns
    am15g_integral_to_4_micron = np.trapz(am15g, x=am15g.index, axis=-1)  # W/m^2
    iec_integral_to_4_micron = 997.47  # W/m^2, from IEC 60904-3 Ed. 2 (2008)
    am15g_scaled_to_iec = iec_integral_to_4_micron / am15g_integral_to_4_micron * am15g

    if wavelength is not None:
        interpolator = interp1d(am15g.index, am15g,
                                kind='linear',
                                bounds_error=False,
                                fill_value=0.0,
                                copy=False,
                                assume_sorted=True)

        am15g = pd.Series(data=interpolator(wavelength), index=wavelength)

        interpolator = interp1d(am15g_scaled_to_iec.index, am15g_scaled_to_iec,
                                kind='linear',
                                bounds_error=False,
                                fill_value=0.0,
                                copy=False,
                                assume_sorted=True)

        am15g_scaled_to_iec = pd.Series(data=interpolator(wavelength), index=wavelength)

    am15g.index.name = 'wavelength'
    am15g.name = 'am15g'

    am15g_scaled_to_iec.index.name = 'wavelength'
    am15g_scaled_to_iec.name = 'am15g'

    return am15g, am15g_scaled_to_iec


def calc_spectral_mismatch_field(sr, e_sun, e_sun_tot, e_ref=None, e_ref_tot=None):
    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Aug. 2022
    # Modified by Mark Campanelli for testing purposes.

    # get the reference spectrum at wavelengths matching the measured spectra
    if e_ref is None:
        e_ref, _ = get_am15g(wavelength=e_sun.T.index)

    # a helper function to make usable fraction calculations more readable
    def integrate(e):
        return np.trapz(e, x=e.T.index, axis=-1)

    if e_sun_tot is None:
        # Use integrated spectral irradiance, typically an approximation
        # that misses "tail" of distribution.
        e_sun_tot = integrate(e_sun)

    if e_ref_tot is None:
        # Use integrated spectral irradiance, typically an approximation
        # that misses "tail" of distribution.
        e_ref_tot = integrate(e_ref)

    # interpolate the sr at the wavelengths of the spectra
    # reference spectrum wavelengths may differ if e_ref is from caller
    sr_sun = np.interp(e_sun.T.index, sr.index, sr, left=0.0, right=0.0)
    sr_ref = np.interp(e_ref.T.index, sr.index, sr, left=0.0, right=0.0)

    # calculate usable fractions
    uf_sun = integrate(e_sun * sr_sun) / e_sun_tot
    uf_ref = integrate(e_ref * sr_ref) / e_ref_tot

    # mismatch is the ratio or quotient of the usable fractions
    smm = uf_sun / uf_ref

    if isinstance(e_sun, pd.DataFrame):
        smm = pd.Series(smm, index=e_sun.index)

    return smm
```

This is the Jupyter notebook code that runs the comparison:
```python
import numpy as np
from pvlib import spectrum
from pvlib.tests import test_spectrum

# Standard spectra from 0.28 to 4 micron, e_ref_iec is am15g (from ASTM G173-03) scaled to "match" IEC 60904-3 Ed. 2 (2008).
e_ref_am15g_astm, e_ref_am15g_iec = spectrum.get_am15g()

# IEC 60904-3 Ed. 2 (2008), cumulatives are provided in standard.
e_ref_am15g_iec_idx_to_1p7_micron = e_ref_am15g_iec.T.index <= 1700
e_ref_am15g_iec_integral_to_1p7_micron = 942.88  # W/m^2
e_ref_am15g_iec_integral_to_4_micron = 997.47  # W/m^2
e_ref_am15g_iec_integral_to_infinity = 1000.  # W/m^2

# ASTM G173-03, calculate cumulatives not provided in standard.
e_ref_am15g_astm_idx_to_1p7_micron = e_ref_am15g_astm.T.index <= 1700
e_ref_am15g_astm_integral_to_1p7_micron = np.trapz(e_ref_am15g_astm[e_ref_am15g_astm_idx_to_1p7_micron], x=e_ref_am15g_astm.index[e_ref_am15g_astm_idx_to_1p7_micron], axis=-1)  # W/m^2
e_ref_am15g_astm_integral_to_4_micron = np.trapz(e_ref_am15g_astm, x=e_ref_am15g_astm.index, axis=-1)  # W/m^2
e_ref_am15g_astm_integral_to_infinity = None  # W/m^2, undefined in standard

# Get example solar spectrum from test fixture.
_, e_sun = test_spectrum.spectrl2_data.__pytest_wrapped__.obj()  # 0.3 to 4 micron
e_sun = e_sun.set_index('wavelength')
e_sun = e_sun.transpose().loc['specglo']
e_sun_idx_to_1p7_micron = e_sun.T.index <= 1700
e_sun_integral_to_1p7_micron = np.trapz(e_sun[e_sun_idx_to_1p7_micron], x=e_sun.T.index[e_sun_idx_to_1p7_micron], axis=-1)
e_sun_integral_to_4_micron = np.trapz(e_sun, x=e_sun.T.index, axis=-1)  # 708.4855034837434 W/m^2
scale_factor_estimate = e_ref_am15g_iec_integral_to_infinity / e_ref_am15g_iec_integral_to_4_micron
e_sun_integral_to_infinity = scale_factor_estimate * e_sun_integral_to_4_micron

# Relative spectral response for Si, 0.29 to 1.19 micron.
sr = spectrum.get_example_spectral_response()

# SMM computations IEC 61853-1 imply reference temperature in all situations, which is technically incorrect.

# Calculate SMM using formula in IEC 61853-1, using IEC 60904-3 Ed. 2 (2008) spectrum.
# Different cutoff wavelengths used for solar total irradiance integral according to the spectraradiometer cutoff.
iec_smm_inf = spectrum.calc_spectral_mismatch_field(
    sr, e_sun, e_sun_integral_to_infinity, e_ref=e_ref_am15g_iec, e_ref_tot=e_ref_am15g_iec_integral_to_infinity)
iec_smm_4_micron = spectrum.calc_spectral_mismatch_field(
    sr, e_sun, e_sun_integral_to_4_micron, e_ref=e_ref_am15g_iec, e_ref_tot=e_ref_am15g_iec_integral_to_infinity)
iec_smm_1p7_micron = spectrum.calc_spectral_mismatch_field(
    sr, e_sun[e_sun_idx_to_1p7_micron], e_sun_integral_to_1p7_micron, e_ref=e_ref_am15g_iec, e_ref_tot=e_ref_am15g_iec_integral_to_infinity)

# Calculate SMM using proposed formula double truncation of integrals with ASTM G173-03 spectrum.
astm_smm_4_micron = spectrum.calc_spectral_mismatch_field(sr, e_sun, None)
astm_smm_1p7_micron = spectrum.calc_spectral_mismatch_field(sr, e_sun[e_sun_idx_to_1p7_micron], None)

print("IEC am15g totals to 1.7 um, 4 um, and infinity (W/m^2):", e_ref_am15g_iec_integral_to_1p7_micron, e_ref_am15g_iec_integral_to_4_micron, e_ref_am15g_iec_integral_to_infinity)
print("ASTM am15g totals to 1.7 um, 4 um, and infinity (W/m^2):", e_ref_am15g_astm_integral_to_1p7_micron, e_ref_am15g_astm_integral_to_4_micron, e_ref_am15g_astm_integral_to_infinity)
print("e_sun totals to 1.7 um, 4 um, and infinity (W/m^2):", e_sun_integral_to_1p7_micron, e_sun_integral_to_4_micron, e_sun_integral_to_infinity)

print("IEC smm to infinity:", iec_smm_inf)
print("IEC smm to 4 um:", iec_smm_4_micron)
print("IEC smm to 1.7 um:", iec_smm_1p7_micron)
print("IEC am15g-equivalent E to infinity:", iec_smm_inf*e_sun_integral_to_infinity)
print("IEC am15g-equivalent E to 4 um:", iec_smm_4_micron*e_sun_integral_to_4_micron)
print("IEC am15g-equivalent E to 1.7 um:", iec_smm_1p7_micron*e_sun_integral_to_1p7_micron)

print("ASTM smm to 4 um:", astm_smm_4_micron)
print("ASTM smm to 1.7 um:", astm_smm_1p7_micron)
print("ASTM am15g-equivalent E to 4 um:", astm_smm_4_micron*e_sun_integral_to_4_micron)
print("ASTM am15g-equivalent E to 1.7 um:", astm_smm_1p7_micron*e_sun_integral_to_1p7_micron)
```

Here is the output for the comparison. Note the consistency of `E_equiv` in the "IEC version" of the computation.
```terminal
IEC am15g totals to 1.7 um, 4 um, and infinity (W/m^2): 942.88 997.47 1000.0
ASTM am15g totals to 1.7 um, 4 um, and infinity (W/m^2): 945.6188921247738 1000.3706555734423 None
e_sun totals to 1.7 um, 4 um, and infinity (W/m^2): 668.028852782929 708.4855034837434 710.2825182549283
IEC smm to infinity: 0.9913868433999364
IEC smm to 4 um: 0.9939014139772988
IEC smm to 1.7 um: 1.054093308636961
IEC am15g-equivalent E to infinity: 704.1647436949111
IEC am15g-equivalent E to 4 um: 704.1647436949111
IEC am15g-equivalent E to 1.7 um: 704.164743694911
ASTM smm to 4 um: 0.9923965546943565
ASTM smm to 1.7 um: 0.9902678450884945
ASTM am15g-equivalent E to 4 um: 703.0985727081635
ASTM am15g-equivalent E to 1.7 um: 661.5274925022902
```

adriesse: Thanks again for your detailed analysis, Mark.  I agree that there is potential for small differences with the tails, and large differences due to the cut-off at 1700 nm.  In this field application the cutoff at 1700 nm is only made if there is no information available about the spectrum beyond.  In that situation there are not a lot of choices available, and cutting off the reference spectrum at the same wavelength appears to be a reasonable solution.

docs/sphinx/source/whatsnew/v0.9.3.rst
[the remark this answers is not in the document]
adriesse: Regarding your other comment on offering `sr_ref`: it's a reasonable suggestion.  Perhaps for a more general companion function combined with other advanced options?  Comparing two devices can also be easily and intuitively done by running this function twice and taking the quotient.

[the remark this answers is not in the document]
kanderso-nrel: My thought was that, if more general functionality will be added eventually and one general function is better than two partially overlapping functions, we might want to avoid locking ourselves into the `calc_spectral_mismatch_field` name.  But clearly others are more familiar than I am with the typical workflow here, so I defer to your judgement on what makes the most sense.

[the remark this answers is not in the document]
adriesse: The name could be changed, if it that helps.  There is a quite range of possibilities for making a more general function.  I wouldn't claim that this makes the most sense but it gets us one step further in the general direction.

adriesse:
To all: if you feel that this PR is premature, don't feel like you have to merge it.  It's not a lot of code and I can easily just pack it into an example notebook.

kanderso-nrel:
Well this PR does have three approving reviews, which is more than most pvlib PRs :) I haven't been following the discussion with @markcampanelli closely, but I have the sense that at this point whatever limitations there are in this implementation are noted in the docstring, and of course we can always make additional improvements later if we want.  Anyone please correct me if I'm mistaken there.  Otherwise let's plan to merge this tomorrow?

markcampanelli:
@adriesse Would it be possible to include an example in this PR of how this code is used for a specific use case or cases?

If I were to try to use it for an IEC 61853-3 like calculation with a 1.7 micron cutoff, then I could get tripped up about which total irradiance to multiply the SMM by to get the equivalent irradiance under the standard spectrum used in the 61853-1 matrix (i.e., using the double truncation approximation).

I am curious if `pvlib` workflows supports any use cases. (Not that the answer to this has to be positive, the code can be useful regardless.)

adriesse:
@markcampanelli An example will accompany the forthcoming sr collection and spectroradiometer measurements on Duramat.

mikofski:
There are conflicts with the main branch, maybe what’s new? That must be addressed before we can merge

adriesse:
I think this needs coordination with the other PR(s), doesn't it?

cwhanse:
The conflict was created when #1468 was merged. I can resolve if you'd like the help.

adriesse:
Sure @cwhanse, it seems easiest to do this at the time of merging.

cwhanse:
@markcampanelli are you OK with this being merged and opening issues for improvements?

markcampanelli:
> @markcampanelli are you OK with this being merged and opening issues for improvements?

Yes.

Note: I am running some computation comparisons with PVfit, and if I find anything to reconcile, then I will make a followup issue.

kanderso-nrel:
Thanks @adriesse for this addition (which I myself plan to use) and @markcampanelli for the effort in ensuring pvlib continues to be a reliable and high-quality tool :)

adriesse:
Thanks to all participants here as well!!  There's definitely room for further work on  mismatch calculation methods, both for pvlib and for scientific publications.

markcampanelli:
> Note: I am running some computation comparisons with PVfit, and if I find anything to reconcile, then I will make a followup issue.

For the example SR and reference and solar spectra provided herein, I did not find any significant numerical differences between this implementation and PVfit's `pvfit.measurement.spectral_correction.M`. Very small differences (<0.02% in SMM) appear to be explained by choices of function interpolation, integral truncation, and numerical integration technique.

adriesse:
Thanks a lot for checking that @markcampanelli !

adriesse:
> There's definitely room for further work on mismatch calculation methods, both for pvlib and for scientific publications.

I didn't use a strong enough word there: there a ***need*** for further work.

Issue #1371, opened 2022-01-07
Add iotools function to write SAM-formatted CSV files


williamhobbs:
**Is your feature request related to a problem? Please describe.**
Manual formatting is required before writing weather data into a CSV file that NREL SAM can read. 

**Describe the solution you'd like**
An iotools function, e.g., `pvlib.iotools.write_sam()`, that writes data to a CSV file in SAM's format in a standardized way would be nice.

**Describe alternatives you've considered**
Manually re-formatting data, adding relevant headers, and writing a CSV. An example of code I've used is below:

```
## Write SAM Weather file - SURFRAD Goodwin Creek 2019
# note: start with dataframes for data and metadata created with pvlib.iotools.read_surfrad, e.g., `(surfrad_data, surfrad_metadata) = pvlib.iotools.read_surfrad(...)`

sam_data = surfrad_data

# Make sure file starts at midnight local time. Here I use roll (circular shift) as a quick fix
sam_data = sam_data.reindex(index=np.roll(surfrad_data.index,-6*60))    

# metadata
source='SURFRAD Goodwin Creek 2019'
latitude=surfrad_metadata.loc['latitude',0]
longitude=surfrad_metadata.loc['longitude',0]
elevation=surfrad_metadata.loc['elevation',0]
timezone_offset = -6 # standard time in hours ahead of GMT, e.g., -7 for Denver

filename = source + '.csv'

# make a header        
header = '\n'.join(
        [ 'Source,Latitude,Longitude,Time Zone,Elevation',
        source + ',' + str(latitude) + ',' + str(longitude) + ',' + str(timezone_offset) + ',' + str(elevation),
        'year,month,day,hour,minute,ghi,dni,dhi,temperature,wspd',
        ]) + '\n'

# write 1st part of header
with open(filename, 'w') as fp:
    fp.write(header)

# write rest of header with SAM-friendly names and write weather data
columns = ['year','month','day','hour','minute','ghi','dni','dhi','temp_air','wind_speed']
sam_data.to_csv(filename, columns = columns, mode='a', index=False, header=False)
```

**Additional context**
There is an open issue in https://github.com/NREL/SAM/, https://github.com/NREL/SAM/issues/866, to update SAM to allow weather variable names used by pvlib. That would simplify things. 

There may need to be additional consideration given to weather datasets that:
- include less than one year, which currently will not work in SAM
- include a full year, but don't start January 1 at midnight. Automatically "rolling" could fix this
[...]

PR #1556, opened 2022-09-22
sam iotools #1371

shirubana:
adds a sam.py, and pytest

 - [ X ]  I think Closes #xxxx
 - [ X ] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [  X] Tests added
 - [ X] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [ X] Pull request is nearly complete and ready for detailed review.

### DIFF
--- a/pvlib/iotools/sam.py
+++ b/pvlib/iotools/sam.py
@@ -1,0 +1,164 @@
+"""Functions for reading and writing SAM data files."""
+
+import pandas as pd
+
+def saveSAM_WeatherFile(data, metadata, savefile='SAM_WeatherFile.csv', standardSAM = True, includeminute=False):
+    """
+    Saves a dataframe with weather data from pvlib format on SAM-friendly format.
+
+    Parameters
+    -----------
+    data : pandas.DataFrame
+        timeseries data in PVLib format. Should be TZ converted (not UTC). Ideally it is one sequential year data; if not suggested to use standardSAM = False. 
+    metdata : dictionary
+        Dictionary with at least 'latitude', 'longitude', 'elevation', 'source', and 'TZ' for timezone.
+    savefile : str
+        Name of file to save output as. 
+    standardSAM : boolean
+        This checks the dataframe to avoid having a leap day, then averages it to SAM style (closed to the right), 
+        and fills the years so it starst on YEAR/1/1 0:0 and ends on YEAR/12/31 23:00.
+    includeminute ; Bool
+        For hourly data, if SAM input does not have Minutes, it calculates the sun position 30 minutes 
+        prior to the hour (i.e. 12 timestamp means sun position at 11:30)
+        If minutes are included, it will calculate the sun position at the time of the timestamp (12:00 at 12:00)
+        Set to true if resolution of data is sub-hourly.
+
+    Returns
+    -------
+    Nothing, it just writes the file.
+    
+    """
+
+    def _is_leap_and_29Feb(s):
+        ''' Creates a mask to help remove Leap Years. Obtained from: 
+            https://stackoverflow.com/questions/34966422/remove-leap-year-day-from-pandas-dataframe/34966636
+        '''
+        return (s.index.year % 4 == 0) & \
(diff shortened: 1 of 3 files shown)


cwhanse:
@shirubana when sticker turns green we'll approve and run the tests.

kanderso-nrel:
@williamhobbs, since you opened #1371, would you be interested in taking a look at this PR from a user's perspective?

pvlib/iotools/sam.py
```diff
@@ -0,0 +1,178 @@
...
+        savedata['Minute'] = data.index.minute
+
+    windspeed = list(data.wind_speed)
+    temp_amb = list(data.temp_air)
+    savedata['Wspd'] = windspeed
+    savedata['Tdry'] = temp_amb
+
+    if 'dni' in data:
+        dni = list(data.dni)
+        savedata['DHI'] = dni
```
adriesse, line 113: I think you could just write savedata['X'] = data.X.values if this is to avoid index alignment.

shirubana: tx, pushed this change

williamhobbs:
> @williamhobbs, since you opened #1371, would you be interested in taking a look at this PR from a user's perspective?

Sure! I just ran a test with some data I previously grabbed with `pvlib.iotools.read_surfrad`. Here are a few comments:

- The `metadata` dictionary returned by `pvlib.iotools.read_surfrad` stores the time zone as `"tz": "UTC"`. Two issues come up from that:
  -  `sam.py` is looking for "TZ" instead of "tz". Not sure if this matters, since "tz" doesn't seem to be standardized across `iotools`, but it might help a bit to be consistent.
  - NREL SAM needs the time zone as a number that is UTC offset, e.g., 0 for UTC or -6 for MDT. It would be nice if `sam.py` could return an error or warning when passing a time zone string, or even better, convert time zones that might get created by other functions in `iotools` into the number fomat SAM is looking for.
- When filling in the year so that it starts Jan-1 at 12 AM and ends with the last interval of the calendar year, I typically "roll" the data if it starts mid-year (e.g., if I have data from March 2018 through Feb 2019, move Jan and Feb 2019 to the beginning of the file). In my test case, I pulled data for a full calendar year in UTC (starting at 2019-01-01 00:00 UTC and ending 2019-12-31 23:59 UCT), but after running through `sam.saveSAM_WeatherFile()`, I ended up with a two-year weather file since 2019-01-01 00:00 UCT is 2018-12-31 18:00 CST and it filled in zeros for the rest of 2018. I'm not sure the best way to handle this, but I think the current approach could be confusing. 
- Naming the function `write_sam()` or similar might better match existing naming conventions, but I'm also not familiar with pvlib function name guidelines. 
- `sam.saveSAM_WeatherFile()` requires a `source` value in the metadata file, but that isn't produced by `read_surfrad` and some of the other `iotools` functions. Would it makes sense to have this function default to something, e.g., "pvlib export", if the user doesn't provide an input value?

shirubana:
@williamhobbs  do you have an example of your rolling method? 

I addressed everything else except that and seems veyr important to do so, and also the timezone string to number function, is a nice add on in general but I probably won't get around to that specifically for this PR.

shirubana:
@williamhobbs  I added a rolling method. Whenever you get an opporutnity can you re-try your surfrad data? 

Thanks!

williamhobbs:
@shirubana The rolling seems to work based on a quick test. 

I'm getting the time zone in the output file as "UTC" instead of "0". I think users would expect the time zone in the metadata returned from read_surfrad to work, or to get an error/warning. 

I've also just realized that it would be helpful to be able to split out hourly averaging from the other standardSAM behaviors (skipping leap day and making the file start with Jan 1 12:00 AM), since SAM does not require hour intervals but does require the other items.  I would suggest either:
1. removing the resampling (`_averageSAMStyle`) or 
2. creating an additional input that is either a boolean for hourly averaging or something that lets you pass an averaging interval in minutes.

PR #1562, opened 2022-09-30
Implement iam.schlick, iam.schlick_diffuse

### LINKED ISSUE #1564: Add iam function based on Schlick 1994

Equation 15 in Schlick 1994 gives an efficient approximation equation for surface reflectance as a function of incidence angle that could be used as the basis for an *iam* function.

Equation 4 in Xie 2022 gives an analytical integration of the Schlick equation over the portion of an isotropic sky that is visible from a tilted surface.

Equation 6 in Xie 2022 gives an analytical integration of the Schlick equation over the portion of an isotropic ground that is visible from a tilted surface.

A new function called ```schlick``` in the ```iam``` module.  The equation would be:

```
iam = 1 - (1 - cos(aoi)) ** 5
```
This equation has no parameters for fitting to iam measurements or to represent different types of PV module glass.

A second new function called ```schlick_diffuse``` in the ```iam``` module (similar to the function ```martin_ruiz_diffuse```) which implements Equations 4 and 6 from Xie 2022 (excluding the empirical weighting factor *w*, or in other words, with *w=1*).

**References**

Schlick, C. An inexpensive BRDF model for physically-based rendering. Computer graphics forum 13 (1994).
Xie, Y. et al. The 'Fresnel Equations' for Diffuse radiation on Inclined photovoltaic Surfaces (FEDIS). J.RSER (2022)

kanderso-nrel:
- [x] Closes #1564
 - [x] Tests added

Yu Xie from NREL presented a new IAM model at the recent PVPMC meeting ([slides](https://pvpmc.sandia.gov/download/8412/)) and wanted to contribute it to pvlib.  This one is a little unusual in that it calculates the IAM factor as a relative quantity, taking the indices of refraction of both the PV surface and the accompanying pyranometer into account.


pvlib/iam.py
```diff
@@ -748,3 +748,108 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    iam : dict
+        IAM values for each type of irradiance:
+
+            * 'direct': radiation from the solar disc
+            * 'sky': radiation from the sky dome (zenith <= 90)
+            * 'ground': radiation reflected from the ground (zenith >= 90)
+
+    Notes
+    -----
+    This implementation corrects a typo in the reference regarding the sign
```
cwhanse, line 793:
```suggestion
    This implementation corrects a typo in [1]_ regarding the sign
```

pvlib/iam.py
```diff
@@ -748,3 +748,108 @@ def marion_integrate(function, surface_tilt, region, num=None):
         Fd = pd.Series(Fd, surface_tilt.index)
 
     return Fd
+
+
+def fedis(aoi, surface_tilt, n=1.5, n_ref=1.4585):
+    """
+    Determine the incidence angle modifiers (iam) for direct, diffuse sky,
```
cwhanse, line 755:
```suggestion
    Determine the incidence angle modifiers (IAM) for direct, diffuse sky,
```

pvlib/iam.py
```diff
@@ -748,3 +748,108 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    n_ref : float, default 1.4585
+        Refractive index of the pyranometer cover. The default value
+        was used for a fused silica dome over a CMP22 in [1]_.
+
+    Returns
+    -------
+    iam : dict
+        IAM values for each type of irradiance:
+
+            * 'direct': radiation from the solar disc
```
cwhanse, line 787: Minor preference here for 'irradiance' instead of 'radiation', only for consistency of terms.

pvlib/iam.py
```diff
@@ -748,3 +748,108 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+            * 'sky': radiation from the sky dome (zenith <= 90)
+            * 'ground': radiation reflected from the ground (zenith >= 90)
```
cwhanse, line 789:
```suggestion
            * 'sky': diffuse radiation from the sky dome
            * 'ground': radiation reflected from the ground
```

pvlib/iam.py
```diff
@@ -748,3 +748,108 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    ----------
+    aoi : numeric
+        Angle of incidence. [degrees]
+
+    surface_tilt : numeric
+        Surface tilt angle measured from horizontal (e.g. surface facing
+        up = 0, surface facing horizon = 90). [degrees]
+
+    n : float, default 1.5
+        Refractive index of the PV cover.  The default value of 1.5
```
cwhanse, line 775:
```suggestion
        Refractive index of the PV front surface material.  The default value of 1.5
```

kanderso-nrel:
I suppose we could move the diffuse calculations to a separate function like was done with martin_ruiz.  It's a little strange to return scalar sky & ground IAM but time series direct IAM as the current code does for a fixed tilt simulation.

pvlib/iam.py
```diff
@@ -748,3 +748,108 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    ----------
+    .. [1] Xie, Y., M. Sengupta, A. Habte, A. Andreas, "The 'Fresnel Equations'
+       for Diffuse radiation on Inclined photovoltaic Surfaces (FEDIS)",
+       Renewable and Sustainable Energy Reviews, vol. 161, 112362. June 2022.
+       :doi:`10.1016/j.rser.2022.112362`
+    """
+
+    # avoid undefined results for horizontal or upside-down surfaces
+    zeroang = 1e-06
+    surface_tilt = np.where(surface_tilt == 0, zeroang, surface_tilt)
```
cwhanse, line 806: The need for this is a caution flag, to me. It addresses the divide by zero possiblity in the "/ (1 - cos(surface_tilt))" portion of the equation for the ground-reflected IAM equation.  I am not familiar with the reference. Is it possible that the integrand becomes singular as surface_tilt -> 0? If so, maybe there's an alternate way to show that the ground-reflected IAM converges to 0.

pvlib/tests/test_iam.py
```diff
@@ -322,3 +322,51 @@ def test_marion_integrate_invalid():
...
     with pytest.raises(ValueError):
         _iam.marion_integrate(_iam.ashrae, 0, 'bad', 180)
+
+
+def test_fedis_defaults():
+    # expected values generated with code from model authors
+    surface_tilt = np.array([0, 30, 60, 90])
+    aoi = np.array([0, 30, 60, 90])
+
+    expected = {
```
cwhanse, line 332: I would add a comment here stating the source of these expected results.

pvlib/iam.py
kanderso-nrel: I settled on this approach (lifted from `martin_ruiz_diffuse`) because it also prevents the obnoxious `RuntimeWarning: divide by zero encountered in true_divide` warnings from numpy.  I'm certainly happy to change to a better alternative if there is one.

cwhanse: Here's an alternative: for `surface_tilt` < minimum, set `cuk = 20 w / 21` (limiting case of Eq. A8a in [1]), and set `cug = 0` (there should be no ground-reflected irradiance on a horizontal, upward-facing surface.

pvlib/iam.py
```diff
@@ -748,3 +748,108 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+
+def fedis(aoi, surface_tilt, n=1.5, n_ref=1.4585):
+    """
+    Determine the incidence angle modifiers (IAM) for direct, diffuse sky,
+    and ground-reflected radiation using the FEDIS transmittance model.
+
+    The "Fresnel Equations" for Diffuse radiation on Inclined photovoltaic
+    Surfaces (FEDIS) [1]_ is an analytical solution of diffuse transmission
+    based on the rigorous integration of an alternate form of the
+    Fresnel equations. The approach leads to a simple yet accurate
```
adriesse, line 761: The claim made here and in the source publication is misleading.  I suggest that the so-called "alternate form of the Fresnel" equations" be implemented in pvlib before finalizing the docstring for Fedis.

kanderso-nrel: Is the issue that the phrase "alternate form" implies equivalence when in fact the math relies on an approximation?  And what would the function signature for that implementation look like (inputs & outputs)?

kanderso-nrel: >  And what would the function signature for that implementation look like (inputs & outputs)?

Nevermind, just saw #1564

kanderso-nrel:
Offline discussion with @cwhanse and the model's author determined that the parameter this PR calls `n_ref` may have some niche uses but should be set to the same value as `n` for standard PV modeling workflows (unlike as assumed in the reference).  That is now the default behavior. 

I also added `pvlib.iam.schlick` in this PR as suggested in #1564.

pvlib/iam.py
```diff
@@ -748,3 +749,177 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+        up = 0, surface facing horizon = 90). [degrees]
+
+    n : float, default 1.5
+        Refractive index of the PV front surface material.  The default value
+        of 1.5 was used for an IMT reference cell in [1]_. [unitless]
+
+    n_ref : float, optional
+        Reference refractive index. In [1]_ this was set to 1.4585 for
+        was used for a fused silica dome over a CMP22, but in conventional
+        PV applications it is appropriate to set this to the same value as
```
cwhanse, line 832:
```suggestion
        a fused silica dome over a CMP22, but in conventional
        PV applications it is appropriate to set n_ref to the same value as
```

pvlib/iam.py
```diff
@@ -748,3 +749,177 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+        ``n`` (the default behavior).
+
+    Returns
+    -------
+    iam : dict
+        IAM values for each type of irradiance:
+
+            * 'direct': irradiance from the solar disc
+            * 'sky': irradiance radiation from the sky dome
```
cwhanse, line 841:
```suggestion
            * 'sky': diffuse irradiance from the sky dome
```

cwhanse (approved):
I think there are some minor efficiency improvements that could be done in `iam.fedis` at the cost of less transparency when comparing to the reference. E.g., the default case `n_ref=n` means that `term1` is identically 1. On the whole I am preferring transparency.

adriesse:
Should/does fedis direct be equal to schlick?  If not, why not?

kanderso-nrel:
> Should/does fedis direct be equal to schlick? If not, why not?

Close-ish but not identical.  FEDIS is using the true Fresnel equations for the direct component so it cannot match exactly.  The Schlick approximation only comes in for the diffuse components: in the absence of an analytical integration of the real Fresnel equations, FEDIS's diffuse components are instead an integration of the Shlick approximation.  It's the alternative to Marion's integration method -- whereas Marion's is an approximate integration of an exact integrand, FEDIS is an exact integration of an approximate integrand.

adriesse:
> Close-ish but not identical. FEDIS is using the true Fresnel equations for the direct component 

I see, but we already have that in iam.physical, right?

kanderso-nrel:
> I see, but we already have that in iam.physical, right?

Yes, with two minor differences: FEDIS does not consider extinction (i.e. it assumes `K=0` in `iam.physical`), and FEDIS has the option to normalize the IAM profile with respect to another device with a different index of refraction.  The latter is what prevents `iam.physical` from being a superset of FEDIS's direct IAM calculation.  A separate index of refraction for the normal incidence normalization coefficient is probably only useful in unusual modeling scenarios, but nevertheless it is part of the model so I include it here.

For context, I have the sense that the FEDIS paper to some extent views its diffuse IAM equations as the primary scientific contribution with the direct component being included for completeness.

adriesse:
> FEDIS has the option to normalize the IAM profile with respect to another device with a different index of refraction.

Normally IAM is a property or characteristic of one device, so we should try to avoid confusion.  If a ratio is needed for some unusual modeling scenarios, this can be calculated with any pair of existing IAM models and parameters, I would think.

kanderso-nrel:
The behavior as currently implemented is to default to equal refractive index so that everything is for a single device.  I'm  open to removing the second refractive index altogether so long as we include sufficient justification in the docstring, but this would be a substantial deviation from the reference which, for better or worse, places a lot of focus on the multiple refractive index aspect.

adriesse:
> > Should/does fedis direct be equal to schlick? If not, why not?
> 
> Close-ish but not identical. FEDIS is using the true Fresnel equations for the direct component so it cannot match exactly. The Schlick approximation only comes in for the diffuse components: in the absence of an analytical integration of the real Fresnel equations, FEDIS's diffuse components are instead an integration of the Shlick approximation. It's the alternative to Marion's integration method -- whereas Marion's is an approximate integration of an exact integrand, FEDIS is an exact integration of an approximate integrand.

This is a good way to put the distinction.  Based on my most recent re-read, FEDIS is an exact integration of an approximate integrand ***plus*** a scaling/correction factor ***w*** to make up for the fact that the integrating the approximate integrand (Schlick) doesn't actually produce the desired result.

cwhanse:
@kanderso-nrel that bottom panel is an eyebrow-raiser. I had not seen that result before. Although ground-reflected irradiance onto the front surface is a very minor part of plane-of-array, that figure suggests it should be further reduced by roughly a factor of 2, for many installations.

adriesse:
Nicely done.  Let's see what other comments appear here over the coming days...

On a related note, I think it would be great if the original author could submit two examples to pvlib-python: one example illustrating an important use case for ```n_ref``` ; and one to replicate the validation comparisons that are presented in the publication.

docs/sphinx/source/whatsnew/v0.9.4.rst
```diff
@@ -10,6 +10,11 @@ Deprecations
 Enhancements
 ~~~~~~~~~~~~
 * Multiple code style issues fixed that were reported by LGTM analysis. (:issue:`1275`, :pull:`1559`)
+* Added two direct IAM models :py:func:`pvlib.iam.schlick` and :py:func:`pvlib.iam.fedis`,
+  both of which are also available in :py:func:`~pvlib.iam.marion_diffuse` (:pull:`1562`, :issue:`1564`)
```
cwhanse, line 14:
```suggestion
  both of which can be used with :py:func:`~pvlib.iam.marion_diffuse` (:pull:`1562`, :issue:`1564`)
```

pvlib/iam.py
```diff
@@ -748,3 +750,265 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+        iam = iam.item()
+    elif isinstance(aoi, pd.Series):
+        iam = pd.Series(iam, aoi.index)
+
+    return iam
+
+
+def schlick_diffuse(surface_tilt):
+    """
+    Determine the incidence angle modifiers (iam) for diffuse sky and
```
cwhanse, line 811: Is there a reason not to capitalize "IAM"?

pvlib/iam.py
```diff
@@ -748,3 +750,265 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    return cuk, cug
+
+
+def fedis(aoi, n=1.5, n_ref=None):
+    """
+    Determine the incidence angle modifier (IAM) for direct irradiance
+    using the FEDIS transmittance model.
+
+    Note that FEDIS [1]_ calculates direct IAM using the Fresnel equations
+    without extinction, thus in the default case of ``n_ref=n``, this is
```
cwhanse, line 883:
```suggestion
    without extinction, thus in the default case of ``n_ref=None``, this is
```

pvlib/iam.py
```diff
@@ -748,3 +750,265 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+
+    n : float, default 1.5
+        Refractive index of the PV front surface material.  The default value
+        of 1.5 was used for an IMT reference cell in [1]_. [unitless]
+
+    n_ref : float, optional
+        Reference refractive index. In [1]_ this was set to 1.4585 for
+        a fused silica dome over a CMP22, but in conventional
+        PV applications it is appropriate to set n_ref to the same value as
+        ``n`` (the default behavior).
```
cwhanse, line 899:
```suggestion
        Reference refractive index. If None (default), set equal to ``n``.
        In [1]_ ``n_ref`` was set to 1.4585 for
        a fused silica dome over a CMP22, but in conventional
        PV applications it is appropriate to set ``n_ref=n`` (the default behavior).
```

pvlib/iam.py
```diff
@@ -748,3 +750,265 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+
+    In PV contexts, the Schlick approximation has been used as an analytically
+    integrable alternative to the Fresnel equations for estimating IAM
+    for diffuse irradiance [2]_.
+
+    Parameters
+    ----------
+    aoi : numeric
+        The angle of incidence (AOI) between the module normal vector and the
+        sun-beam vector in degrees. Angles of nan will result in nan.
```
cwhanse, line 775:
```suggestion
        The angle of incidence (AOI) between the module normal vector and the
        sun-beam vector. Angles of nan will result in nan. [degrees]
```

pvlib/iam.py
```diff
@@ -748,3 +750,265 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    surface_tilt : numeric
+        Surface tilt angle measured from horizontal (e.g. surface facing
+        up = 0, surface facing horizon = 90). [degrees]
+
+    n : float, default 1.5
+        Refractive index of the PV front surface material.  The default value
+        of 1.5 was used for an IMT reference cell in [1]_. [unitless]
+
+    n_ref : float, optional
+        Reference refractive index. In [1]_ this was set to 1.4585 for
```
cwhanse, line 967: Copy from `fedis`

cwhanse (approved):
Some docstring suggestions.

pvlib/iam.py
kanderso-nrel: I blame @adriesse for this :)

adriesse: Guilty.  :(  Copied the line from ```martin-ruiz-diffuse()```, in which I first made the error.

pvlib/iam.py
```diff
@@ -748,3 +750,267 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    ----------
+    .. [1] Xie, Y., M. Sengupta, A. Habte, A. Andreas, "The 'Fresnel Equations'
+       for Diffuse radiation on Inclined photovoltaic Surfaces (FEDIS)",
+       Renewable and Sustainable Energy Reviews, vol. 161, 112362. June 2022.
+       :doi:`10.1016/j.rser.2022.112362`
+
+    See Also
+    --------
+    pvlib.iam.physical
+    """
```
adriesse, line 917:
```
def fedis(aoi, n=1.5, n_ref=None):

    iam_physical = physical(aoi, n, K=0, L=0)

    if n_ref is None:
        return iam_physical

    else:
        # reflectance for normal incidence with refractive index n:
        rd0 = ((n - 1) / (n + 1)) ** 2

        # reflectance for normal incidence with refractive index n_ref:
        r0 =  ((n_ref - 1) / (n_ref + 1)) ** 2

        # weighting function
        w = (1 - rd0) / (1 - r0)
        
        return w * iam_physical
```

kanderso-nrel: I wondered if something like this could be done but did not bother to look into it.  Thanks!

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    n : float, default 1.5
+        Refractive index of the PV front surface material.  The default value
+        of 1.5 was used for an IMT reference cell in [1]_. [unitless]
+
+    n_ref : float, optional
+        Reference refractive index. If None (default), set equal to ``n``.
+        In [1]_ ``n_ref`` was set to 1.4585 for
+        a fused silica dome over a CMP22, but in conventional
+        PV applications it is appropriate to set ``n_ref=n`` (the default
+        behavior).
```
cwhanse, line 900:
```suggestion
        behavior). [unitless]
```

cwhanse, line 900:
```suggestion
        behavior). [unitless]
```

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
         Fd = pd.Series(Fd, surface_tilt.index)
 
     return Fd
+
+
+def schlick(aoi):
+    """
+    Determine incidence angle modifier (IAM) using the Schlick approximation
+    to the Fresnel equations.
```
cwhanse, line 758:
```suggestion
    Determine incidence angle modifier (IAM) for direct irradiance using the
    Schlick approximation to the Fresnel equations.
```

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    n : float, default 1.5
+        Refractive index of the PV front surface material.  The default value
+        of 1.5 was used for an IMT reference cell in [1]_. [unitless]
+
+    n_ref : float, optional
+        Reference refractive index. If None (default), set equal to ``n``.
+        In [1]_ ``n_ref`` was set to 1.4585 for
+        a fused silica dome over a CMP22, but in conventional
+        PV applications it is appropriate to set ``n_ref=n`` (the default
+        behavior).
```
cwhanse, line 960:
```suggestion
        behavior). [unitless]
```

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    Parameters
+    ----------
+    aoi : numeric
+        The angle of incidence (AOI) between the module normal vector and the
+        sun-beam vector. Angles of nan will result in nan. [degrees]
+
+    Returns
+    -------
+    iam : numeric
+        The incident angle modifier
```
cwhanse, line 780:
```suggestion
        The incident angle modifier.
```

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    Parameters
+    ----------
+    surface_tilt : numeric
+        Surface tilt angle measured from horizontal (e.g. surface facing
+        up = 0, surface facing horizon = 90). [degrees]
+
+    Returns
+    -------
+    iam_sky : numeric
+        The incident angle modifier for sky diffuse
```
cwhanse, line 828:
```suggestion
        The incident angle modifier for sky diffuse.
```

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+
+    iam_ground : numeric
+        The incident angle modifier for ground-reflected diffuse
```
cwhanse, line 831:
```suggestion
        The incident angle modifier for ground-reflected diffuse.
```

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+        Reference refractive index. If None (default), set equal to ``n``.
+        In [1]_ ``n_ref`` was set to 1.4585 for
+        a fused silica dome over a CMP22, but in conventional
+        PV applications it is appropriate to set ``n_ref=n`` (the default
+        behavior).
+
+    Returns
+    -------
+    iam : numeric
+        The incident angle modifier
```
cwhanse, line 905:
```suggestion
        The incident angle modifier.
```

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+        Reference refractive index. If None (default), set equal to ``n``.
+        In [1]_ ``n_ref`` was set to 1.4585 for
+        a fused silica dome over a CMP22, but in conventional
+        PV applications it is appropriate to set ``n_ref=n`` (the default
+        behavior).
+
+    Returns
+    -------
+    iam_sky : numeric
+        The incident angle modifier for sky diffuse
```
cwhanse, line 965:
```suggestion
        The incident angle modifier for sky diffuse.
```

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+
+    iam_ground : numeric
+        The incident angle modifier for ground-reflected diffuse
```
cwhanse, line 968:
```suggestion
        The incident angle modifier for ground-reflected diffuse.
```

cwhanse (approved):
A little docstring tidying. OK with me to merge as is.

adriesse:
Perhaps we should invite a additional reviewer on this one?

pvlib/iam.py
```diff
@@ -748,3 +750,255 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+
+    # relative transmittance of sky diffuse radiation by PV cover:
+    cosB = cosd(surface_tilt)
+    sinB = sind(surface_tilt)
+    cuk = (2 / (np.pi * (1 + cosB))) * (
+        (30/7)*np.pi - (160/21)*np.radians(surface_tilt) - (10/3)*np.pi*cosB
+        + (160/21)*cosB*sinB - (5/3)*np.pi*cosB*sinB**2 + (20/7)*cosB*sinB**3
+        - (5/16)*np.pi*cosB*sinB**4 + (16/105)*cosB*sinB**5
+    )  # Eq 4 in [2]
+
```
adriesse, line 859: Just out of curiosity, doesn't stickler force spaces around math operators?

pvlib/iam.py
adriesse, line 900: The explanation of n_ref is not really clear enough.  I suggest two concrete examples, to illustrate the need for sometimes having them equal and sometimes not.  (Unfortunately I don't understand the author's thinking on this.)

pvlib/iam.py
adriesse: As above.

adriesse:
It turns out that ```term1```  in ```fedis_diffuse()``` is exactly the same as the weighting function w in ```fedis()```, just in simplified form.  I suggest making the code the same in both functions and referring to this ratio/weight/multiplier as ```w0```.  (Or the calculation could even be put in a separate helper function.)

For clarity, I would suggest using the long version because it shows what's actually going on, and putting the short version in a comment with reference to the paper.

pvlib/iam.py
kanderso-nrel: I agree this description is not very useful.  What do you think of this revision?

```
    n_ref : float, optional
        Refractive index for which IAM=1.0 at normal incidence.
        If None (default), set equal to ``n``. See Notes for
        possible use cases. [unitless]
```

And the following in `Notes`:
```
        In most PV applications it is appropriate to set ``n_ref=n``
        (the default behavior), but a custom ``n_ref`` may be useful when
        using irradiance data from a pyranometer whose calibration does not
        account for reflection off its glass dome.  It may also be used as
        the original index of refraction if the module surface's refractive
        index is changing over time (for example a degraded AR coating).
```

pvlib/iam.py
kanderso-nrel: Looks like our `.stickler.yml` configuration ignores `E226` (the flake8 code for missing whitespace around operators)

kanderso-nrel:
> For clarity, I would suggest using the long version because it shows what's actually going on, and putting the short version in a comment with reference to the paper.

I have kept the equations as-is so they continue being directly tied to the reference, but the latest commit at least leaves a comment noting the equivalence for any interested readers.  Hopefully that achieves the same goal?

> Perhaps we should invite a additional reviewer on this one?

I wouldn't mind input from additional reviewers here, but the three of us (@adriesse, @cwhanse, myself) reaching consensus would be good enough for me.

pvlib/iam.py
adriesse: > may be useful when
>         using irradiance data from a pyranometer whose calibration does not
>         account for reflection off its glass dome.

In other words: never.

adriesse: I think the author needs to provide this information.

cwhanse: Pending author input, I think the revised description is clear.

kanderso-nrel: Author input has been incorporated with a minor addition besides the revision proposed above.  Any other thoughts here @adriesse?

adriesse: I think the two use cases are still inadequately described.  This is why I suggested getting an example, which would be concrete; and another reviewer, who might evaluate this from a less biased perspective.

pvlib/iam.py
```diff
@@ -748,3 +750,266 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+
+def fedis_diffuse(surface_tilt, n=1.5, n_ref=None):
+    """
+    Determine the incidence angle modifiers (IAM) for diffuse sky and
+    and ground-reflected radiation using the FEDIS transmittance model.
+
+    The "Fresnel Equations" for Diffuse radiation on Inclined photovoltaic
+    Surfaces (FEDIS) [1]_ is the result of analytical integration
+    the Schlick approximation [2]_ to the Fresnel equations.
```
adriesse, line 952:
```suggestion
This model scales the :py:func:`schlick_diffuse` output using a 
polynomial to approximate the influence of refractive index n.  
An additional scaling factor is applied if ``n_ref`` is not equal to ``n``.

```

pvlib/iam.py
```diff
@@ -748,3 +750,266 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+    """
+    if n_ref is None:
+        n_ref = n
+
+    cuk, cug = schlick_diffuse(surface_tilt)
+
+    # weighting function
+    # note that this expression for term1 is algebraically equivalent to
+    # the "transmittance_ratio" in fedis()
+    term1 = n*(n_ref+1)**2 / (n_ref*(n+1)**2)
```
adriesse, line 1006:
```suggestion
    # note that the following line is algebraically equivalent to
    # sequence of calculations for the "normal_transmittance_ratio" found in fedis()
    normal_transmittance_ratio = n*(n_ref+1)**2 / (n_ref*(n+1)**2)
```

pvlib/iam.py
```diff
@@ -748,3 +750,266 @@ def marion_integrate(function, surface_tilt, region, num=None):
...
+        # reflectance for normal incidence with refractive index n:
+        rd0 = ((n - 1) / (n + 1)) ** 2
+
+        # reflectance for normal incidence with refractive index n_ref:
+        r0 = ((n_ref - 1) / (n_ref + 1)) ** 2
+
+        # correction coefficient for different indices of refraction
+        transmittance_ratio = (1 - rd0) / (1 - r0)
+
+        return transmittance_ratio * iam_physical
```
adriesse, line 942:
```suggestion
        normal_transmittance_ratio = (1 - rd0) / (1 - r0)

        return normal_transmittance_ratio * iam_physical
```

kanderso-nrel:
Following offline discussion, we have decided to move forward here with only the two schlick functions.  I'll wait a few days to merge this to leave opportunity to review for anyone who wants to.

adriesse (approved):

kanderso-nrel:
Thanks again for the thoughtful and productive discussion @adriesse and @cwhanse

adriesse:
No problem.  I learned a few things in the process!

PR #1573, opened 2022-10-16
Test pvsystem.singlediode against precisely generated IV curves

### LINKED ISSUE #411: Test singlediode() implementations using a "gold" function and data set

reepoi:
- [x] Closes #411
 - [x] Tests added

Generated IV curves are added to `pvlib/data` to test the precision of `pvsystem.singlediode`, `pvsystem.v_from_i`, and `pvsystem.i_from_v`. The parameters of the single diode equation that generated each IV curve are in CSV files, and the IV curve data are in JSON files following [this JSON schema](https://github.com/cwhanse/ivcurves/blob/main/ivcurves/ivcurve_jsonschema.json). `i_x` and `i_xx` are added in addition to the fields of the JSON schema since they are also calculated by `pvsystem.singlediode`.
The generated IV curve data are calculated using the functions in the [ivcurves repository](https://github.com/cwhanse/ivcurves). These functions use the arbitrary precision math library [mpmath](mpmath.org). The data may be reproduced by checking out the code of [this commit of ivcurves](https://github.com/reepoi/ivcurves/tree/19c0269884fbb5f35c385b61923175bfbc961a13) and running the command
```
python3 ivcurves/precise.py --save-json test_sets
```
The data are in the files `test_sets/case1.json` and `test_sets/case2.json`.


pvlib/pvsystem.py
```diff
@@ -2873,10 +2873,11 @@ def singlediode(photocurrent, saturation_current, resistance_series,
 
         # calculate the IV curve if requested using bishop88
         if ivcurve_pnts:
-            vd = v_oc * (
+            ivcurve_v = v_oc * (
                 (11.0 - np.logspace(np.log10(11.0), 0.0, ivcurve_pnts)) / 10.0
             )
-            ivcurve_i, ivcurve_v, _ = _singlediode.bishop88(vd, *args)
+            ivcurve_i = _singlediode.bishop88_i_from_v(ivcurve_v, *args,
```
cwhanse, line 2879: These changes should be reverted. `singlediode.bishop88` returns the current and voltage vectors. The returned voltage vector is not equal to the input `vd` which is the diode voltage.

Arguably, the returned voltage could be constructed in the way proposed here (compute voltage, not diode voltage, and find the corresponding currents). But that should be discussed in a separate issue.

reepoi: I reverted these changes and created issue #1578 for this. The test cases for the `brentq` and `newton` methods check the accuracy of the currents returned at the voltages of `vd`, so they will not pass. Since these unit tests depend on what voltage values are returned by `pvsystem.singlediode`, I think they should be removed for now and added when issue #1578 is resolved.

reepoi:
I updated the description of this pull request. The test cases added now test:
- `pvsystem.singlediode` with `ivcurve_pnts=None` and all options for `method`: checks all the returned values.
- `pvsystem.singlediode` with `ivcurve_pnts=100` and `method='lambertw': checks only the returned current and voltage values.
- `pvsystem.v_from_i` and `pvsystem.i_from_v` with all options for `method`: checks the returned voltage or current values.

The tests for `pvsystem.v_from_i` and `pvsystem.i_from_v` indirectly test the precision of `pvsystem.singlediode` with methods `brentq` and `newton` when `ivcurves_pnts=100` since it they all eventually call `pvsystem.bishop88`.

cwhanse:
Ready for review. @markcampanelli would appreciate your input if this closes #411

pvlib/data/precise_iv_curves_parameter_sets1.csv
```diff
@@ -0,0 +1,33 @@
+Index,photocurrent,saturation_current,resistance_series,resistance_shunt,n,cells_in_series
+1,1.0,5e-10,0.1,300,1.01,72
```
markcampanelli, line 2: Because this is a Python repo, you might consider indexing from 0, not 1.

pvlib/data/precise_iv_curves1.json
```diff
@@ -0,0 +1,10251 @@
...
+  "Manufacturer": "",
+  "Model": "",
+  "Serial Number": "",
+  "Module ID": "",
+  "Description": "",
+  "Material": "",
+  "cells_in_series": 72,
+  "IV Curves": [
+    {
+      "Index": 1,
```
markcampanelli, line 11: Because this is a Python repo, you might consider indexing from 0, not 1.

pvlib/tests/test_singlediode.py
```diff
@@ -69,6 +71,124 @@ def test_newton_fs_495(method, cec_module_fs_495):
...
+    joined[is_array] = joined[is_array].applymap(
+        lambda a: np.asarray(a, dtype=np.float64)
+    )
+    is_number = joined.columns.difference(
+        is_array + ['Index', 'Irradiance', 'Sweep direction', 'Datetime']
+    )
+    joined[is_number] = joined[is_number].applymap(np.float64)
+
+    joined['Boltzman'] = 1.380649e-23
+    joined['Electron Charge'] = 1.60217663e-19
```
markcampanelli, line 115: For better traceability, I would think that the choice of values for these constants should documented as part of the files containing the parameters used to precisely compute the curves.

For comparison, `scipy==1.8.1` gives:

scipy.constants.value("Boltzmann constant") = 1.380649e-23 J/K
scipy.constants.value("elementary charge") = 1.602176634e-19 C

pvlib/tests/test_singlediode.py
```diff
@@ -69,6 +71,123 @@ def test_newton_fs_495(method, cec_module_fs_495):
...
+    x, pc = precise_iv_curves
+    outs = pvsystem.singlediode(method=method, **x)
+
+    assert np.allclose(pc['i_sc'], outs['i_sc'], atol=1e-10, rtol=0)
+    assert np.allclose(pc['v_oc'], outs['v_oc'], atol=1e-10, rtol=0)
+    assert np.allclose(pc['i_mp'], outs['i_mp'], atol=5e-8, rtol=0)
+    assert np.allclose(pc['v_mp'], outs['v_mp'], atol=8e-7, rtol=0)
+    assert np.allclose(pc['p_mp'], outs['p_mp'], atol=1e-10, rtol=0)
+    assert np.allclose(pc['i_x'], outs['i_x'], atol=1e-10, rtol=0)
+    assert np.allclose(pc['i_xx'], outs['i_xx'], atol=8e-8, rtol=0)
```
markcampanelli, line 158: I'm super curious how these tolerances might be tightened for 'lambertw', 'brentq', vs. 'newton'.

markcampanelli (changes requested):
Thanks for this submission. Generally looks good.

The big thing I would like to see is some better documentation in this repo (mostly for posterity) about how the precise I-V data points were generated elsewhere. For example, how did you choose the precision cutoff using `mpmath` with respect to the precision used in the `pvsystem.singlediode` calculations? The choice of constants in the upstream calculation could also have a better "paper trail".

pvlib/tests/test_singlediode.py
markcampanelli: Also see https://github.com/pvlib/pvlib-python/issues/483

markcampanelli:
> Ready for review. @markcampanelli would appreciate your input if this closes #411

I think this addresses https://github.com/pvlib/pvlib-python/issues/411, so that can be closed once this merges.

I have closed https://github.com/pvlib/pvlib-python/pull/426, which this substitutes.

pvlib/tests/test_singlediode.py
reepoi: Ok, I updated these constants to match The SI Brochure (https://www.bipm.org/en/publications/si-brochure) and cited that document in the docstring. In that docstring, I also added the precision cutoff of mpmath and more detail about the precision of the IV curve data. I avoided mentioning [cwhanse/ivcurves](https://github.com/cwhanse/ivcurves), but maybe we should reference it. There is much more documentation and code that would help someone understand how the IV curve data was generated.

Also, if we use `scipy.constants`, pvlib may need to update its minimum SciPy version requirement because Boltzmann's constant changes depending on the SciPy version. The value for SciPy 1.9.3 is `1.602176634e-19 C`, which matches the value hard-coded here; however, the value is `1.6021766208e-19 C` for SciPy 1.2.0. It changes in SciPy 1.4.0, and pvlib's minimum version of SciPy is 1.2.0.

cwhanse: @pvlib/pvlib-maintainer I think we should advance scipy version and use the 2019 values, which are defined to be _exact_ rather than with uncertainty.

pvlib/tests/test_singlediode.py
cwhanse: And I'm curious why there are test failures only on ubuntu and python <=3.9 for the 'lambertw' case. I suspect something buried fairly deep behind `scipy.special.lambertw` which we won't solve.

reepoi: For [this failed test](https://github.com/pvlib/pvlib-python/actions/runs/3879773557/jobs/6617268775), the 30th `i_xx` value differs by `1e-6`. All of the tests related to this pull request pass with `atol=1e-6` for the `i_xx` value.

cwhanse: @reepoi I think there's a mistake in the precise curve library somewhere.

cwhanse:
> The current test failures are only on linux, only occur for numpy>=1.22.0, and have to do with precision checks, so I suspect this is due to the increased SIMD functionality [added in numpy 1.22.0](https://numpy.org/doc/stable/release/1.22.0-notes.html#vectorize-umath-module-using-avx-512).
> 
> In the past we have relaxed test tolerances to accommodate the slight numerical variation, but in this specific case that seems rather self-defeating :) 

My perspective is that the test tolerance is a measure of the pvlib functions' accuracy. Six digits (worst case) is at least useful for energy simulations and may also help provide context for other use cases. At least it establishes a benchmark from which we can measure improvements.

kanderso-nrel:
@cwhanse a lot has happened in the https://github.com/cwhanse/ivcurves repository in the last couple months.  Are any corresponding updates required in this PR?

cwhanse:
We may need to replace the json files

reepoi:
I regenerated the JSON files using the latest version of the https://github.com/cwhanse/ivcurves repository and there is no difference between them and the JSON files in this PR.

kanderso-nrel (approved):
A few minor docs suggestions but otherwise LGTM.  Thanks @reepoi!

cwhanse (approved):

kanderso-nrel:
Thanks @reepoi for the contribution and @markcampanelli and @cwhanse for the reviews!

Issue #1594, opened 2022-11-17
Create new temperature model(s) incorporating a simple radiative loss term.


adriesse:
**Is your feature request related to a problem? Please describe.**
The most frequently used operating temperature models do not account properly for radiative losses to the sky.

**Describe the solution you'd like**
Create enhanced versions of the existing models (at least one) incorporating a simple but effective radiative loss term as described in: Driesse, A. et al (2022) "Improving Common PV Module Temperature Models by Incorporating Radiative Losses to the Sky". SAND2022-11604. [10.2172/1884890](http://dx.doi.org/10.2172/1884890)

**Describe alternatives you've considered**
Watching TV.

cwhanse:
The point in the technical report is that temperature models fit measurements better when a radiative loss term is included. Is that correct?

The generic coefficients for e.g. Faiman, SAPM would not be the right values to use in a model including radiative loss terms, I think. Is there enough data at hand to come up with default coefficients for the new temperature model, or to convert from the generic coefficients?

adriesse:
> The point in the technical report is that temperature models fit measurements better when a radiative loss term is included. Is that correct?
> 
Yes.

> The generic coefficients for e.g. Faiman, SAPM would not be the right values to use in a model including radiative loss terms, I think. Is there enough data at hand to come up with default coefficients for the new temperature model, or to convert from the generic coefficients?

You're quite right, those would need to be changed to something (smaller).  I'm working on a data set and demo as well, and will reflect on what values to suggest.  In general I must say that I am trending toward having fewer default values filled in to encourage users to think more about what they're doing.  This might be a situation where no defaults would be appropriate in the function, but with sample values in the Notes.

adriesse:
> @adriesse what kind of TV you been watching?

As long as this issue is moving forward: None!

kanderso-nrel:
An idea, perhaps a bad one: as an alternative to creating `*_rad` variants of existing models, is it possible and practical to implement the radiative loss term as a standalone generic secondary function kind of like `temperature.prilliman`?

adriesse:
Not a bad thought.  I did actually consider some other approaches, like preprocessing/reducing poa_global or using a factory function to make the rad models on demand.  Or something cool with a class again.  But I didn't actually try them out.  

For parameter fitting it's certainly much easier to have a new function like this one, and I don't think I'll do more than one in the near future anyway.

adriesse:
[...]

PR #1083, opened 2020-10-16
Hayes temp model

stephenjkaplan:
- [x] Closes #1080 
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [x] Tests added **Note from Stephen: there are some placeholder unit tests to be updated after model is validated**
 - [x] Updates entries to [`docs/sphinx/source/api.rst`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes.
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``). **Note from Stephen: waiting on feature to be added to a particular release**
 - [x] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [ ] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels and Milestone are assigned to the Pull Request and linked Issue.

PR #1595, opened 2022-11-17
Implement enhanced Faiman temperature model

adriesse:
- [x] Motivated by #1594
 - [x] Tests added

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.9.4.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.4.rst
@@ -59,5 +59,5 @@ Contributors
 * Kevin Anderson (:ghuser:`kanderso-nrel`)
 * Karel De Brabandere (:ghuser:`kdebrab`)
 * Naman Priyadarshi (:ghuser:`Naman-Priyadarshi`)
-* Anton Driesse (:ghuser:`adriesse`)
+* Adam R. Jensen (:ghuser:`AdamRJensen`)
 * Echedey Luis (:ghuser:`echedey-ls`)
--- a/docs/sphinx/source/whatsnew/v0.9.4.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.4.rst
@@ -59,6 +59,4 @@ Contributors
 * Karel De Brabandere (:ghuser:`kdebrab`)
 * Naman Priyadarshi (:ghuser:`Naman-Priyadarshi`)
 * Anton Driesse (:ghuser:`adriesse`)
-* Cliff Hansen (:ghuser:`cwhanse`)
-* Kevin Anderson (:ghuser:`kanderso-nrel`)
 * Echedey Luis (:ghuser:`echedey-ls`)
--- a/docs/sphinx/source/whatsnew/v0.9.4.rst
+++ b/docs/sphinx/source/whatsnew/v0.9.4.rst
@@ -13,9 +13,12 @@ Enhancements
 * Added a function to calculate one of GHI, DHI, and DNI from values of the other two.
   :py:func:`~pvlib.irradiance.complete_irradiance`
   (:issue:`1565`, :pull:`1567`)
-* Add optional ``return_components`` parameter to :py:func:`pvlib.irradiance.haydavies` to return
+* Added optional ``return_components`` parameter to :py:func:`pvlib.irradiance.haydavies` to return
   individual diffuse irradiance components (:issue:`1553`, :pull:`1568`)
-
+* Added a module temperature model that accounts for radiative losses to the sky
+  in a simplified way, using the Faiman model as an example.
+  :py:func:`~pvlib.temperature.faiman_rad`
+  (:issue:`1594`, :pull:`1595`)
 
 Bug fixes
 ~~~~~~~~~
@@ -35,7 +38,6 @@ Benchmarking
 ~~~~~~~~~~~~~
 * Removed ``time_tracker_singleaxis`` function from tracking.py (:issue:`1508`, :pull:`1535`)
 
-
 Requirements
 ~~~~~~~~~~~~
 
@@ -48,3 +50,6 @@ Contributors
 * Marcus Boumans (:ghuser:`bowie2211`)
 * Karel De Brabandere (:ghuser:`kdebrab`)
 * Naman Priyadarshi (:ghuser:`Naman-Priyadarshi`)
+* Anton Driesse (:ghuser:`adriesse`)
+* Cliff Hansen (:ghuser:`cwhanse`)
+* Kevin Anderson (:ghuser:`kanderso-nrel`)
--- a/pvlib/temperature.py
+++ b/pvlib/temperature.py
@@ -457,8 +457,8 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
     return temp_air + temp_difference
 
 
-def faiman_rad(poa_global, temp_air, wind_speed=1.0, ir_down=0.0,
-               u0=25.0, u1=6.84, emissivity=0.88, sky_view=1.0):
+def faiman_rad(poa_global, temp_air, wind_speed=1.0, ir_down=None,
+               u0=25.0, u1=6.84, sky_view=1.0, emissivity=0.88):
     r'''
     Calculate cell or module temperature using the Faiman model augmented
     with a radiative loss term.
@@ -533,7 +533,7 @@ def faiman_rad(poa_global, temp_air, wind_speed=1.0, ir_down=0.0,
 
     # The following lines may seem odd since the values are probably scalar,
     # but they serve an indirect and easy way of allowing lists and
-    # tuples for the other function arguments.
+    # tuples for the other function arguments, and providing Series output.
     u0 = np.asanyarray(u0)
     u1 = np.asanyarray(u1)
     emissivity = np.asanyarray(emissivity)
@@ -541,8 +541,11 @@ def faiman_rad(poa_global, temp_air, wind_speed=1.0, ir_down=0.0,
     t_zero = np.array(-273.15)
     kstefbolz = np.array(5.670367e-8)
 
-    ir_up = kstefbolz * ((temp_air - t_zero)**4)
-    qrad_sky = emissivity * sky_view * (ir_up - ir_down)
+    if ir_down is None:
+        qrad_sky = np.array(0.0)
+    else:
+        ir_up = kstefbolz * ((temp_air - t_zero)**4)
+        qrad_sky = emissivity * sky_view * (ir_up - ir_down)
 
     heat_input = poa_global - qrad_sky
     total_loss_factor = u0 + u1 * wind_speed
--- a/pvlib/temperature.py
+++ b/pvlib/temperature.py
@@ -467,8 +467,7 @@ def faiman_rad(poa_global, temp_air, wind_speed=1.0, ir_down=None,
     adopted in the IEC 61853 standards [2]_ and [3]_.  The radiative loss
     term was developed in [4]_.
 
-    Usage of this model in the IEC 61853 standard does not distinguish
-    between cell and module temperature.
+    The model can be used to represent cell or module temperature.
 
     Parameters
     ----------
@@ -497,12 +496,16 @@ def faiman_rad(poa_global, temp_air, wind_speed=1.0, ir_down=None,
         determined by Faiman for 7 silicon modules.
         :math:`\left[ \frac{\text{W}/\text{m}^2}{\text{C}\ \left( \text{m/s} \right)} \right]`
 
-    emissivity : numeric, default 0.88
-        Infrared emissivity of the module surface facing the sky. [-]
-
     sky_view : numeric, default 1.0
         Effective view factor limiting the radiative exchange between the
-        module and the sky. [-]
+        module and the sky. For a tilted array the expressions
+        (1 + 3*cos(tilt)) / 4 can be used as a first estimate for sky_view
+        as discussed in [4]_. The default value is for a horizontal module. [-]
+
+    emissivity : numeric, default 0.88
+        Infrared emissivity of the module surface facing the sky. The default
+        value represents the middle of a range of values found in the
+        literature. [-]
 
     Returns
     -------
@@ -512,6 +515,9 @@ def faiman_rad(poa_global, temp_air, wind_speed=1.0, ir_down=None,
     -----
     All arguments may be scalars or vectors. If multiple arguments
     are vectors they must be the same length.
+
+    If the input `ir_down` is not supplied the model output is the same as the
+    original Faiman model.
 
     References
     ----------
--- a/pvlib/temperature.py
+++ b/pvlib/temperature.py
@@ -419,7 +419,7 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
     u1 : numeric, default 6.84
         Combined heat loss factor influenced by wind. The default value is one
         determined by Faiman for 7 silicon modules.
-        :math:`\left[ \frac{\text{W}/\text{m}^2}{\text{C}\ \left( \text{m/s} \right)} \right]`
+        :math:`\left[ \frac{\text{W}/\text{m}^2}{\text{C}\ \left( \text{m/s} \right)} \right]`  # noQA: E501
 
     Returns
     -------
@@ -494,18 +494,19 @@ def faiman_rad(poa_global, temp_air, wind_speed=1.0, ir_down=None,
     u1 : numeric, default 6.84
         Combined heat loss factor influenced by wind. The default value is one
         determined by Faiman for 7 silicon modules.
-        :math:`\left[ \frac{\text{W}/\text{m}^2}{\text{C}\ \left( \text{m/s} \right)} \right]`
+        :math:`\left[ \frac{\text{W}/\text{m}^2}{\text{C}\ \left( \text{m/s} \right)} \right]`  # noQA: E501
 
     sky_view : numeric, default 1.0
         Effective view factor limiting the radiative exchange between the
         module and the sky. For a tilted array the expressions
         (1 + 3*cos(tilt)) / 4 can be used as a first estimate for sky_view
-        as discussed in [4]_. The default value is for a horizontal module. [-]
+        as discussed in [4]_. The default value is for a horizontal module.
+        [unitless]
 
     emissivity : numeric, default 0.88
         Infrared emissivity of the module surface facing the sky. The default
         value represents the middle of a range of values found in the
-        literature. [-]
+        literature. [unitless]
 
     Returns
     -------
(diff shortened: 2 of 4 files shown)


pvlib/temperature.py
```diff
@@ -457,6 +457,108 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Nov., 2022
+
+    # The following lines may seem odd since the values are probably scalar,
+    # but they serve an indirect and easy way of allowing lists and
+    # tuples for the other function arguments, and providing Series output.
+    u0 = np.asanyarray(u0)
+    u1 = np.asanyarray(u1)
+    emissivity = np.asanyarray(emissivity)
+
+    t_zero = np.array(-273.15)
```
cwhanse, line 547: Should this constant be 298.15 to correspond with 25C (temperature at STC)?

pvlib/temperature.py
```diff
@@ -457,6 +457,108 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    Parameters
+    ----------
+    poa_global : numeric
+        Total incident irradiance [W/m^2].
+
+    temp_air : numeric
+        Ambient dry bulb temperature [C].
+
+    wind_speed : numeric, default 1.0
+        Wind speed in m/s measured at the same height for which the wind loss
```
cwhanse, line 481:
```suggestion
        Wind speed measured at the same height for which the wind loss
```

pvlib/temperature.py
```diff
@@ -457,6 +457,108 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    u1 : numeric, default 6.84
+        Combined heat loss factor influenced by wind. The default value is one
+        determined by Faiman for 7 silicon modules.
+        :math:`\left[ \frac{\text{W}/\text{m}^2}{\text{C}\ \left( \text{m/s} \right)} \right]`
+
+    sky_view : numeric, default 1.0
+        Effective view factor limiting the radiative exchange between the
+        module and the sky. For a tilted array the expressions
+        (1 + 3*cos(tilt)) / 4 can be used as a first estimate for sky_view
+        as discussed in [4]_. The default value is for a horizontal module. [-]
```
cwhanse, line 503: Prefer [unitless]

pvlib/temperature.py
```diff
@@ -457,6 +457,108 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+        determined by Faiman for 7 silicon modules.
+        :math:`\left[\frac{\text{W}/{\text{m}^2}}{\text{C}}\right]`
+
+    u1 : numeric, default 6.84
+        Combined heat loss factor influenced by wind. The default value is one
+        determined by Faiman for 7 silicon modules.
+        :math:`\left[ \frac{\text{W}/\text{m}^2}{\text{C}\ \left( \text{m/s} \right)} \right]`
+
+    sky_view : numeric, default 1.0
+        Effective view factor limiting the radiative exchange between the
```
cwhanse, line 500: Is "effective" adding something here? `sky_view` looks to be the fraction of the sky dome "seen" by the module, so that's the usual view factor.  I wonder if the formula could be considered part of the model and the input would be `surface_tilt`, but I'll defer to the model author.

cwhanse (commented):
I think we should have a test that checks the output of this function matches the output of faiman, for default parameters.

pvlib/temperature.py
adriesse: I snuck in the "effective" because the infra-red sky is very non-isotropic and that needs to be taken into account. The formula is really just a starting point for future work, not adequately explored and validated.

pvlib/temperature.py
adriesse: Ok.

pvlib/temperature.py
adriesse: No, it's for converting to Kelvin.

cwhanse: Lost me on that intent, on line 553. I think it would be more clear to just convert `temp_air` to `tempK` or whatever name you prefer.

cwhanse:
> > I think we should have a test that checks the output of this function matches the output of faiman, for default parameters.
> 
> You mean, the test should call faiman to determine the expected value?

Yes

pvlib/temperature.py
adriesse: I agree that was not so clear.  Perhaps the rename will help.

adriesse:
Additional reviewer(s) invited!

cwhanse (approved):

kanderso-nrel (commented):
@adriesse I've tagged this for 0.9.4 if you're ready to add a what'snew entry for this PR

pvlib/temperature.py
```diff
@@ -419,7 +419,7 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
     u1 : numeric, default 6.84
         Combined heat loss factor influenced by wind. The default value is one
         determined by Faiman for 7 silicon modules.
-        :math:`\left[ \frac{\text{W}/\text{m}^2}{\text{C}\ \left( \text{m/s} \right)} \right]`
+        :math:`\left[ \frac{\text{W}/\text{m}^2}{\text{C}\ \left( \text{m/s} \right)} \right]`  # noQA: E501
```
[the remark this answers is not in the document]
adriesse: Thanks, I didn't know what was going on there.

pvlib/temperature.py
```diff
@@ -457,6 +457,106 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    if ir_down is None:
+        qrad_sky = np.array(0.0)
+    else:
+        ir_up = kstefbolz * ((temp_air - abs_zero)**4)
+        qrad_sky = emissivity * sky_view * (ir_up - ir_down)
+
+    heat_input = poa_global - qrad_sky
+    total_loss_factor = u0 + u1 * wind_speed
+    temp_difference = heat_input / total_loss_factor
+    return temp_air + temp_difference
```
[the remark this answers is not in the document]
adriesse: I was inclined to do that initially, but felt a bit uneasy about passing the ```heat_input``` to ```poa_global```.

adriesse:
Last chance to comment! @jsstein @mtheristis

docs/sphinx/source/whatsnew/v0.9.4.rst
```diff
@@ -56,4 +58,7 @@ Contributors
 * Kevin Anderson (:ghuser:`kanderso-nrel`)
 * Karel De Brabandere (:ghuser:`kdebrab`)
 * Naman Priyadarshi (:ghuser:`Naman-Priyadarshi`)
+* Anton Driesse (:ghuser:`adriesse`)
+* Cliff Hansen (:ghuser:`cwhanse`)
+* Kevin Anderson (:ghuser:`kanderso-nrel`)
```
kanderso-nrel, line 63: Cliff and I are already listed a bit farther up :)

pvlib/temperature.py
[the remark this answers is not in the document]
kanderso-nrel: This one is still here

kanderso-nrel (approved):

mtheristis:
@adriesse, I am sorry for the delay. 

Out of curiosity, is there a reason we are only focusing on Faiman here?  Or will the others (SAPM, NOCT, PVSyst) come later?

adriesse:
> @adriesse, I am sorry for the delay.
> 
> Out of curiosity, is there a reason we are only focusing on Faiman here? Or will the others (SAPM, NOCT, PVSyst) come later?

Good questions.  There were multiple considerations in my head.  I decided to do one in order to keep the PR to a reasonable size and work through integration issues on one function rather than four at a time.  I picked Faiman because it is the simplest but also because it is used in IEC 61853.  When this feature is released I will point it out to interested parties in the working group for that standard.

If there is user demand I see no problem to implement the other three and maybe enhance the generic linear model too.

docs/sphinx/source/whatsnew/v0.9.4.rst
adriesse: I guess that must have happened in the merge.

mtheristis:
> > @adriesse, I am sorry for the delay.
> > Out of curiosity, is there a reason we are only focusing on Faiman here? Or will the others (SAPM, NOCT, PVSyst) come later?
> 
> Good questions. There were multiple considerations in my head. I decided to do one in order to keep the PR to a reasonable size and work through integration issues on one function rather than four at a time. I picked Faiman because it is the simplest but also because it is used in IEC 61853. When this feature is released I will point it out to interested parties in the working group for that standard.
> 
> If there is user demand I see no problem to implement the other three and maybe enhance the generic linear model too.

Understood.

I am just thinking that we mind end up with X versions of a same base model, and the users might be confused with their differences. Is there a way to have an option when running faiman to say "include_radiative_losses = True/False"? Or is this bad practice?

adriesse:
> I am just thinking that we mind end up with X versions of a same base model, and the users might be confused with their differences. 

Well, it has taken a lot of years for the first enhancement to arrive and I'm not planning another one!  Presenting this as an enhancement of something that people feel familiar and comfortable with is a strategy to gain acceptance more readily.  It could
easily have been packaged as a single new model.

> Is there a way to have an option when running faiman to say "include_radiative_losses = True/False"? Or is this bad practice?

Perhaps in the system class or model chain something like this is possible.  I'm not sure.

pvlib/temperature.py
```diff
@@ -442,7 +444,8 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
     .. [3] "IEC 61853-3 Photovoltaic (PV) module performance testing and energy
        rating - Part 3: Energy rating of PV modules". IEC, Geneva, 2018.
 
-    '''
+    '''                                                            # noQA: E501
```
AdamRJensen, line 447:
```suggestion
    '''  # noqa: E501
```

pvlib/temperature.py
```diff
@@ -457,6 +460,110 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+
+    Notes
+    -----
+    All arguments may be scalars or vectors. If multiple arguments
+    are vectors they must be the same length.
+
+    If the input `ir_down` is `None` the model output is the same as the
+    original Faiman model. Setting `ir_down` to zero does not have the same
+    meaning; it leads to very high radiative loss and low module temperature.
+
```
AdamRJensen, line 528:
```suggestion

    See Also
    --------
    pvlib.temperature.faiman

```

AdamRJensen: I would add a similar See Also section to the original Faiman function. For the original Faiman function I would probably also refer to the other module temperature functions.

pvlib/temperature.py
```diff
@@ -457,6 +460,110 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    numeric, values in degrees Celsius
+
+    Notes
+    -----
+    All arguments may be scalars or vectors. If multiple arguments
+    are vectors they must be the same length.
+
+    If the input `ir_down` is `None` the model output is the same as the
+    original Faiman model. Setting `ir_down` to zero does not have the same
+    meaning; it leads to very high radiative loss and low module temperature.
```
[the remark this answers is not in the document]
adriesse: Yes, I'll add some warning about the values.  We had some discussion already about this.  The current defaults are there to fall back to the standard Faiman model.  My second favorite option would be to have no defaults.

pvlib/temperature.py
adriesse: I will add the Faiman reference for sure.  If you would like all temperature models to point to each other that would make a lot of changes, both now and in the future. But it could be done...

pvlib/temperature.py
```diff
@@ -457,6 +460,110 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+    '''                                                            # noQA: E501
+
+    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Nov., 2022
+
+    u0 = np.asanyarray(u0)
+    u1 = np.asanyarray(u1)
+    emissivity = np.asanyarray(emissivity)
+
+    abs_zero = np.array(-273.15)
+    kstefbolz = np.array(5.670367e-8)
```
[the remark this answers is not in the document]
adriesse: Actually this is a different constant, but it is also in scipy so I could import it.  I recall some discussion about such imports in the past but don't recall the conclusion.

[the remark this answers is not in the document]
adriesse: ```temp_air``` could be a scalar or list

pvlib/tests/test_temperature.py
```diff
@@ -131,7 +131,32 @@ def test_faiman_ndarray():
...
+    expected = [-4.071, -2.036, 0.000]
+    result = temperature.faiman_rad(0, 0, 0, ir_down=200,
+                                    sky_view=sky_view)
+    assert_allclose(result, expected, atol=0.001)
+
+    emissivity = [1.0, 0.88, 0.5, 0.0]
+    expected = [-4.626, -4.071, -2.313, 0.000]
+    result = temperature.faiman_rad(0, 0, 0, ir_down=200,
+                                    emissivity=emissivity)
+    assert_allclose(result, expected, atol=0.001)
```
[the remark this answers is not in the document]
adriesse: It's a possibility, but what kind of error would it catch that the others don't?

adriesse (commented):
Thanks for reviewing, Adam.  I'm glad you found the work interesting!  
I responded with some comments and will hold off on code changes to see what other comments might appear...

PS. Sorry, I thought these would appear as replies to your comments.  :(

pvlib/temperature.py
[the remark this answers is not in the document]
AdamRJensen: I would be ok with no defaults. But still, it would need a note saying that the coefficients are note the same when you add the long-wave irradiance.

pvlib/temperature.py
kanderso-nrel: > all temperature models to point to each other that would make a lot of changes, both now and in the future

Proliferation of "see also" has crossed my mind as well, in the context of `pvlib.iam`.  Nowadays I lean towards only referencing functions that are more closely related than just being in the business of modeling the same phenomenon.

pvlib/temperature.py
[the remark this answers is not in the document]
AdamRJensen: Ahh I see, the dear Boltzmann was a productive man. Though, now I might perhaps suggest just calling it sigma, which is the common symbol used to represent it. This [PR](https://github.com/pvlib/pvlib-python/pull/1083/files#diff-5b6b0d8a6e8ee2af1716d84ea1922d3c0e8c9f431024c1626117b87a91f8fc87R755) also refers to it as sigma. As a side note, *k* is typically used for the Boltzmann constant:

image

pvlib/temperature.py
AdamRJensen: Fair point. Let's just keep it at the suggested edit then.

pvlib/temperature.py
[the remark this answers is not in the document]
AdamRJensen: But couldn't the remaining input parameters then also be a list, e.g., ``poa_global``?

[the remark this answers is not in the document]
cwhanse: As a general rule, I don't think it is worth the effort to accommodate type `list` as input. Scalar, numpy.array and Series should be acceptable.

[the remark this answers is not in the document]
cwhanse: The function wouldn't return the same type, and we may unintentionally create an expectation that other functions would also support `list` inputs.

It's not a blocker here for me.

pvlib/temperature.py
```diff
@@ -457,6 +466,119 @@ def faiman(poa_global, temp_air, wind_speed=1.0, u0=25.0, u1=6.84):
...
+
+    u0 = np.asanyarray(u0)
+    u1 = np.asanyarray(u1)
+    emissivity = np.asanyarray(emissivity)
+
+    abs_zero = np.array(-273.15)
+    sigma = np.array(scipy.constants.Stefan_Boltzmann)
+
+    if ir_down is None:
+        qrad_sky = np.array(0.0)
```
AdamRJensen, line 571:
```suggestion
    poa_global = np.asanyarray(poa_global)
    temp_air = np.asanyarray(temp_air)
    wind_speed = np.asanyarray(wind_speed)
    ir_down = np.asanyarray(ir_down)
    u0 = np.asanyarray(u0)
    u1 = np.asanyarray(u1)
    sky_view = np.asanyarray(sky_view)
    emissivity = np.asanyarray(emissivity)

    abs_zero = -273.15
    sigma = scipy.constants.Stefan_Boltzmann

    if ir_down is None:
        qrad_sky = 0.0
```

I see that the usage of np.asanyarray is very carefully chosen such that all inputs can be passed as lists and directly or indirectly be converted to arrays. The indirect conversion to arrays (by multiplying a list something array-like) seems undesirable and unintuitive to the user. Why not instead just convert all the inputs to array-like in the beginning and be straightforward about it?

AdamRJensen (approved):

pvlib/temperature.py
adriesse: On the whole I believe this function behaves like numpy would.  The reason for the indirect and partial approach is to allow pandas and xarray data structures to be propagated from the first four arguments.  Try it, you might like it!

wholmgren: What about this function requires any of the `np.asanyarray` calls?

wholmgren: Ah, now I see the hidden conversation about allowing `list` input. That is not needed and comes at a cost of disallowing arbitrary array-like input that doesn't fully support the asanyarray machinery (e.g. dask, or whatever the pydata library of the month is).

adriesse: Could you elaborate on what doesn't work here?

adriesse: Overall I'm getting more push-back than expected on something which I though was a very simple technique to make the function very flexible.  If there is a concrete down-side I'd like to learn about it, especially the risks associated with ```asanyarray``` (which could be replaced with ```asarray``` in this particular function).  I've use this technique before and am likely to use it again unless there really is a problem with it.

wholmgren: The function is already very flexible because you did a good job writing it :). It doesn't need `asanyarray`! Here's a concrete downside for `asanyarray`:

```python
In [1]: import numpy as np

In [2]: import dask.array as da

In [3]: da.from_array([0, 1])
Out[3]: dask.array<array, shape=(2,), dtype=int32, chunksize=(2,), chunktype=numpy.ndarray>

In [4]: np.asanyarray(da.from_array([0, 1]))
Out[4]: array([0, 1])

In [5]: np.asarray(da.from_array([0, 1]))
Out[5]: array([0, 1])
```

adriesse: > `np.asanyarray` is not robust.

I am curious to learn more about this...

adriesse:
Thank for all the reviews!!

Issue #329, opened 2017-05-29
define/improve documentation structure


wholmgren:
We should consider implementing the documentation structure ideas described in this pycon 2017 talk and associated blog post:

https://www.youtube.com/watch?v=azf6yzuJt54

https://www.divio.com/en/blog/documentation/

My hope is that a more clearly defined documentation structure will help users identify how they can easily contribute to the documentation. 

If people are interested and willing to commit the time to produce more how-to and tutorial content, I would volunteer to improve the documentation infrastructure and then help users get content into the documentation. Ideally, adding to the documentation would be as easy as adding a new jupyter notebook, markdown, or rst file to a source/howtos or source/tutorials directory. This would be a post-PVSC project.

cwhanse:
@wholmgren what do you think about [SunPy's ](https://docs.sunpy.org/en/stable/) documentation structure? Looking around at other projects to find a model for pvlib-python

wholmgren:
I like it. I suggest we pursue that refactoring once #846 is merged.

wholmgren:
I also like where MetPy is going here https://github.com/Unidata/MetPy/pull/1457

kanderso-nrel:
I just stumbled on this -- too late for this year I guess, but maybe next year: https://developers.google.com/season-of-docs

mikofski:
There's also [outreachy](https://www.outreachy.org/docs/applicant/#what-is-outreachy):
>Outreachy internship projects may include programming, user experience, **documentation**, illustration, graphical design, or data science.

Altho I believe Google Season of Docs recruits professional tech writers who probably need considerably less oversight than an intern.

kanderso-nrel:
FYI the Divio link is now broken (but is still on [archive.org](https://web.archive.org/web/20170529181245/https://www.divio.com/en/blog/documentation/)).  Here is an updated link that seems to have more detail than the original page did: https://documentation.divio.com/

I had a hard time understanding the differences between "Tutorials" and "How-To Guides" until I saw this table again:

|  | Most useful when we’re studying | Most useful when we’re coding |
| -- | -- | -- |
| **Practical steps** | Tutorials | How-to guides |
| **Theoretical knowledge** | Explanation | Reference |

So at a basic level, "how-to" guides are for people that are *already writing code* but want something to copy/paste for their specific task.  Tutorials are for people who are looking to learn and the exact content doesn't need to be immediately applicable.

[...]

PR #1598, opened 2022-11-22
Restructure API Reference content to improve left sidebar

kanderso-nrel:
- [ ] Closes #xxxx
 - [ ] Tests added

The sidebar in the pydata-sphinx-theme reflects the document hierarchy; it shows the tree relationship between parent pages and children pages.  Unfortunately, unlike the RTD theme we used to use, it doesn't seem to have any ability to group siblings by what section of the page their toctree entry appears in.  That results in unwieldy layers of the tree with 50+ elements.  I think this is what @wholmgren was referring to in https://github.com/pvlib/pvlib-python/issues/329#issuecomment-1295437513.

I have looked into this at some length and I don't think there's any way to somehow include the in-page header/layout information in the left sidebar.  It is possible to show captions for the top-level tree nodes as I did for [bifacial_radiance](https://bifacial-radiance.readthedocs.io/en/latest/manualapi.html), but I don't think that can be used anywhere but the top level, plus they can't expand/collapse their lists.  The only way I have found to more or less replicate our sidebar from the RTD theme is to rearrange the content such that the document (folder) hierarchy reflects the desired structure.  The first commit on this PR branch tries it out for the PV Modeling subsection of our API Reference:

| `main` ([link](https://pvlib-python.readthedocs.io/en/stable/reference/pv_modeling.html)) | this PR ([link](https://pvlib-python--1598.org.readthedocs.build/en/1598/reference/pv_modeling/index.html))|
|------|------|
| image | image | 

Unfortunately this also means that things that used to be all on one page (e.g. the IAM and Temperature toctrees) are now separate pages.  I don't think there's any way around this with the pydata theme; if we want a tree-like sidebar then we have to have a tree-like page hierarchy.  

Switching to another theme is another possibility, but I'm not enthusiastic about that idea. 

For now I've only changed the "PV Modeling" section in order to get feedback on this approach.  If it is accepted then I will proceed with restructuring the other sections as well.


adriesse:
I think it is an improvement.  Independent of this, there are some awkward categories and this may or may not be the right time or place to discuss them.  Some things are categorized by what they model; others by how they model something.  I'm not sure, for example, where empirical models of PV module behavior would fit.

AdamRJensen:
After having looked into this a bit, there doesn't seem to be a better solution than what @kanderso-nrel proposes. Thus, I am in favor of changing the document/file structure

mikofski:
@kanderso-nrel I like the new layout too.

mikofski:
I like it a lot!

kanderso-nrel:
Thanks everyone for taking a look.  I broke up the `pv_modeling`, `effects_on_pv_system_output`, and `irradiance` pages since they had in-page headers that could be recast as directory structure.  `modelchain` is another candidate but I thought it would be better to keep all that in one page since those class methods are more related to each other than the functions on the other pages are.  The other pages didn't have in-page headers so I left them alone.

As always there's more that could be done to improve the Reference section, but I'm inclined to call this an improvement as-is and leave other changes to the future.  Test failure is unrelated.

cwhanse:
@kanderso-nrel this is a significant improvement to the API reference. I have more hope that I can locate the page for a specific function.

kanderso-nrel:
Ok, glad to hear it.  I'm waiting to finalize this PR until #1602 is merged; better to resolve these pesky conflicts here in this PR.  After that I'll submit a PR for the release itself.

adriesse:
One thing that is not clear is where various existing and future fitting functions should go.  In general I think the best place for them would be as close as possible to the models they serve, both in the code (except if they are really long) and in the API reference.  There could be a section listing just fitting functions as there is now, but I would see that as secondary.

kanderso-nrel:
I'm pretty sure I've correctly fixed up all the merge conflicts here, but of course we should expect additional conflicts in any open PRs that touch this part of the docs.  I won't bother folks for a green checkmark before merging since several people have already expressed support.

Let's take up @adriesse's point about where functions should live and be listed in a separate issue (Anton, can you open one?).

PR #1617, opened 2022-12-19
Use scipy.constants

### LINKED ISSUE #483: Add a public constants module and include scipy as base requirement for pvlib

cwhanse:
- [x] Closes #483 

In review of #1573 @markcampanelli pointed out that the precisely calculated solutions to the single diode equation should use the latest values of physical constants (Boltzmann and elementary charge). In 2019, the various committees agreed that the [published values](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.330-2019.pdf) of these two constants were _exact_, without uncertainty. (I link to the U.S. agency NIST, other agencies also published the same values).

With this agreement I see no reason for pvlib to use approximate values anywhere. Scipy 1.4.0 published the agree exact values in `scipy.constants`. Advancing to scipy 1.4.0 does not require pvlib to also advance the numpy version.

### DIFF
--- a/ci/requirements-py3.10.yml
+++ b/ci/requirements-py3.10.yml
@@ -23,7 +23,7 @@ dependencies:
     - python=3.10
     - pytz
     - requests
-    - scipy >= 1.2.0
+    - scipy >= 1.4.0
     - shapely  # pvfactors dependency
     # - siphon  # conda-forge
     - statsmodels
--- a/ci/requirements-py3.7-min.yml
+++ b/ci/requirements-py3.7-min.yml
@@ -17,7 +17,7 @@ dependencies:
         - h5py==3.1.0
         - numpy==1.16.0
         - pandas==0.25.0
-        - scipy==1.2.0
+        - scipy==1.4.0
         - pytest-rerunfailures # conda version is >3.6
         - pytest-remotedata # conda package is 0.3.0, needs > 0.3.1
         - requests-mock
--- a/ci/requirements-py3.7.yml
+++ b/ci/requirements-py3.7.yml
@@ -23,7 +23,7 @@ dependencies:
     - python=3.7
     - pytz
     - requests
-    - scipy >= 1.2.0
+    - scipy >= 1.4.0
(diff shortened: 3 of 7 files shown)


kanderso-nrel:
The new minimum scipy must also be specified in setup.py's `INSTALL_REQUIRES`.

I didn't do an exhaustive search, but at least `pvlib.temperature.fuentes` has a hardcoded `boltz` variable that could be updated here as well.  One reason to keep the current value in that case is to match the original model implementation from which ours was derived, but I wouldn't object to using scipy's value.

cwhanse:
I searched but am not confident, because I didn't find that instance in temperature.py. I'm inclined to leave that one as is, with a comment.

kanderso-nrel:
One more place to update the minimum scipy version: `/benchmarks/asv.conf.json`.

wholmgren (approved):
Should we note the minimum version change in the release notes? I can go either way.

kanderso-nrel (approved):

markcampanelli:
Thanks everyone for seeing this through, including chasing down all the related “homework” items.

PR #1628, opened 2022-12-29
Migrate spectral modifier functions to pvlib.spectrum

kandersolar:
- ~[ ] Closes #xxxx~
 - [x] Tests added

As suggested in https://github.com/pvlib/pvlib-python/pull/1062#issuecomment-696220404 and maybe elsewhere, this PR moves the functions for First Solar and SAPM spectral modifiers to the `pvlib.spectrum.mismatch` module.

~I also thought as long as we are moving things, it would be good to rename `sapm_spectral_loss` to `sapm_spectral_correction`.~ ~Currently, the two functions are renamed to just `spectrum.sapm` and `spectrum.first_solar`.~ Now renamed to `spectrum.spectral_factor_firstsolar` and `spectrum.spectral_factor_sapm`  Do others agree with this change?  Should `pvlib.pvsystem.PVSystem.sapm_spectral_loss` be renamed as well?


adriesse:
Good move!  

Still thinking about the names and wondering what naming patterns they conform to and which patterns are currently gaining or losing popularity.  Subspecies (?) is often at the end.

docs/sphinx/source/reference/effects_on_pv_system_output/spectrum.rst
```diff
@@ -10,3 +10,5 @@ Spectrum
    spectrum.get_example_spectral_response
    spectrum.get_am15g
    spectrum.calc_spectral_mismatch_field
+   spectrum.first_solar_spectral_correction
```
cwhanse, line 13: Should this be `spectrum.mismatch.first_solar_spectral_correction` ?  Or does the import in `spectrum.__init__` handle the path?

docs/sphinx/source/reference/effects_on_pv_system_output/spectrum.rst
```diff
@@ -10,3 +10,5 @@ Spectrum
...
+   spectrum.sapm_spectral_correction
```
cwhanse, line 14: +1 on the rename.

docs/sphinx/source/reference/effects_on_pv_system_output/spectrum.rst
kanderso-nrel: > Or does the import in `spectrum.__init__` handle the path?

Yep.  I think sphinx's autodoc doesn't fetch docstrings by scanning `.py` files, but rather by importing the specified object and accessing its `__doc__` attribute.  So as long as the name is importable with the specified name it shouldn't matter where the original definition lives.  The choice here is whether we want `mismatch` to be included in the display text for the link.

adriesse:
I'm not in favor of shortening the function names to just the author names.  I though we had some other discussion recently where that was discouraged actually.

mikofski:
I had a similar thought recently, thanks for pulling this together!

kanderso-nrel:
> I'm not in favor of shortening the function names to just the author names. I though we had some other discussion recently where that was discouraged actually.

My thought was that function names in other modules don't include the type of the model, e.g. it's not `iam.sapm_reflectance` or somesuch, so it makes sense to remove `spectral_modifier` from these names.  But that's a separate consideration from whether it's desirable to choose `sapm` and `first_solar` as the model names here.  Suggestions for alternatives are welcome :)

cwhanse:
I don't know if this scales, but maybe we can reach out to authors and ask them what they'd like the function name to be. 

I agree with shortening the function names when the module conveys the function purpose, but, that creates a pit for new users to fall into:

```
from pvlib.iam import sapm
from pvlib.spectrum import sapm
```

etc.

Another thought to toss out: we could add a publication year to the function name to create space for e.g. another model from First Solar.

adriesse:
I think the year may be added on occasion, but I don't think this is a global solution.  

There are few modules/namespaces as focused and conveniently short as ```iam```; because of these two characteristics it is not onerous to import iam and use iam.author().  We could try to create more of those, but again it would not be a global solution.

Spectrum does not contain only one type of function, so I think the functions need a suitable prefix.  The latest addition to this module got a nice long name: [calc_spectral_mismatch_field](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.calc_spectral_mismatch_field.html#pvlib.spectrum.calc_spectral_mismatch_field) so if we take that as a starting point the new ones could be:

```calc_spectral_factor_sapm```
```calc_spectral_factor_firstsolar```

I somewhat hesitate to use *spectral_mismatch* in the names of the empirical models or approximations.

I'm also not sure about the ```calc_``` prefix because it doesn't seem to add much information, but I would go with the flow on that one.  Maybe we could introduce a prefix ```est_``` for functions that merely estimate things.  :)

**Comment by cwhanse** [comment]
```
spectral_factor_sapm
spectral_factor_firstsolar
```

These work for me.

> I somewhat hesitate to use spectral_mismatch in the names

Ditto.

**Comment by kanderso-nrel** [comment]
I have updated to `spectral_factor_sapm` and `spectral_factor_firstsolar`, but still grumble to myself that there is no `irradiance.transposition_perez` or `irradiance.decomposition_erbs` ;)

**Comment by adriesse** [comment]
I'll be anticipating some PR's with renames then!

File: pvlib/atmosphere.py [hunk 3]
```diff
@@ -336,175 +336,10 @@ def gueymard94_pw(temp_air, relative_humidity):
...
-    # Evaluate Spectral Shift
-    coeff = coefficients
-    ama = airmass_absolute
-    modifier = (
-        coeff[0] + coeff[1]*ama + coeff[2]*pw + coeff[3]*np.sqrt(ama) +
-        coeff[4]*np.sqrt(pw) + coeff[5]*ama/np.sqrt(pw))
-
-    return modifier
+first_solar_spectral_correction = deprecated(
+    since='0.9.6',
```
cwhanse, line 340: Don't forget to update the `since` line if we're going to 0.10.0

cwhanse (approved):

kandersolar:
Looking back through the discussion, it seems like the current state of the PR is acceptable to those who expressed opinions on the matter.  I'll go ahead and merge to prevent it from languishing again.

Codecov failure can be ignored.

PR #1661, opened 2023-02-13
WIP: speed up singlediode._lambertw

### LINKED ISSUE #1649: Accelerate singlediode._lambertw
I've found a way to replace the calculation of the maximum power point in `singlediode._lambertw` with a more analytic approach. The speedup is not quite 10x, and would make `pvsystem.singlediode(...method='lambertw')` and `pvsystem.singlediode(...method='newton')` comparable.

I'm not sure if the maths are new, they might be. In any case, it would not be difficult to document on pvpmc.sandia.gov, if not in a conference paper.

We could remove the golden mean search, or retain it. I lean towards retaining it. The new way would use `scipy.optimize.newton` which should converge with quite high confidence, but, not guaranteed for some pathological case. golden mean search would always converge.

[gist](https://gist.github.com/cwhanse/25b91e01d7de594399eb97fb010499ac) with new method.

cwhanse:
- [X] Closes #1649 
 - [ ] Tests added

Implements and documents the solution technique illustrate in this [gist](https://gist.github.com/cwhanse/25b91e01d7de594399eb97fb010499ac) for solving for the maximum power point using the Lambert W method.

### DIFF
--- a/pvlib/singlediode.py
+++ b/pvlib/singlediode.py
@@ -4,7 +4,6 @@ Low-level functions for solving the single diode equation.
 
 from functools import partial
 import numpy as np
-from pvlib.tools import _golden_sect_DataFrame
 
 from scipy.optimize import brentq, newton
 from scipy.special import lambertw
@@ -640,17 +639,18 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
     v_oc = _lambertw_v_from_i(resistance_shunt, resistance_series, nNsVth, 0.,
                               saturation_current, photocurrent)
 
-    params = {'r_sh': resistance_shunt,
-              'r_s': resistance_series,
-              'nNsVth': nNsVth,
-              'i_0': saturation_current,
-              'i_l': photocurrent}
+    params = (photocurrent, saturation_current, resistance_series,
+              resistance_shunt, nNsVth)
 
-    # Find the voltage, v_mp, where the power is maximized.
-    # Start the golden section search at v_oc * 1.14
-    p_mp, v_mp = _golden_sect_DataFrame(params, 0., v_oc * 1.14,
-                                        _pwr_optfcn)
-
+    # Compute maximum power point quantities
+    imp_est = 0.8 * photocurrent
+    args, i0 = _prepare_newton_inputs((), params, imp_est)
+    i_mp = newton(func=_imp_zero, x0=i0, fprime=_imp_zero_prime,
+                 args=args)
+    v_mp = _lambertw_v_from_i(resistance_shunt, resistance_series, nNsVth,
+                              i_mp, saturation_current, photocurrent)
+    p_mp = i_mp * v_mp
+    
     # Find Imp using Lambert W
     i_mp = _lambertw_i_from_v(resistance_shunt, resistance_series, nNsVth,
                               v_mp, saturation_current, photocurrent)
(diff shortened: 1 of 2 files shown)


cwhanse:
Coverage drops because I have not (yet) removed the now-unused private function which was maximized to find the MPP with the golden mean search. Assuming this PR is approved, that helper function can be removed before merging.

cwhanse:
@pvlib/pvlib-maintainer I can't get the environment to solve to run the benchmark. Is it simple for one of you to run the new benchmark and suggest any edits/corrections?

kanderso-nrel:
Just looking at the CI, the `quick-benchmarks` jobs is erroring with `ImportError: cannot import name 'Generator' from 'numpy.random'` with numpy=1.16.5.  It looks like `Generator` was added in numpy 1.17.0, so I guess we'll need to either bump the numpy version in the ASV configuration, or not use `Generator` in the benchmarks.  Either way is fine with me.

cwhanse:
I can rewind to numpy 1.16, it shouldn't affect the time for the benchmarks.

cwhanse:
> Just looking at the CI, the `quick-benchmarks` jobs is erroring with `ImportError: cannot import name 'Generator' from 'numpy.random'` with numpy=1.16.5. It looks like `Generator` was added in numpy 1.17.0, so I guess we'll need to either bump the numpy version in the ASV configuration, or not use `Generator` in the benchmarks. Either way is fine with me.

I can't find where that error was happening - the CI report for the asv action looks green to me. Did you run that locally (whcih I haven't been able to do), or where should I be looking?

It doesn't appear that this new benchmark is being run yet.

kanderso-nrel:
I almost never run the benchmarks locally.  I vaguely recall having problems getting them working on Windows.  I'm pretty sure I've run them locally on Linux before.  The nightly benchmark server runs some flavor of Linux. 

Here's the ASV run for 063abf80e6a18316ccfd4a61cb4cf6c94cb8cc7f which has the error I mentioned: https://github.com/pvlib/pvlib-python/actions/runs/4168577294/jobs/7215540457

I guess the job is (mis)configured in such a way that python errors in the benchmarks don't propagate out to become a CI job failure, so we always get the green checkmark here on the PR.  That needs to be fixed.

> It doesn't appear that this new benchmark is being run yet.

I think asv identifies benchmarks by looking for functions/methods whose name begins with `time_`, so renaming the benchmarks from `bishop88` -> `time_bishop88` should get it working.

benchmarks/benchmarks/singlediode.py
```diff
@@ -0,0 +1,41 @@
...
+    ix = _singlediode.bishop88_i_from_v(vmp/2., *params)
+    ixx = _singlediode.bishop88_i_from_v((voc + vmp)/2., *params)
+    return imp, vmp, pmp, isc, voc, ix, ixx
+
+
+class SingleDiode:
+
+    def setup(self):
+        seed = 11471
+        rng = RandomState(seed)
```
kanderso-nrel, line 26: Do we expect this to produce the same random values as we increase our numpy requirement?  If so, do we expect the benchmark times to change if the inputs change?

kanderso-nrel (commented):
Haven't inspected the math yet

benchmarks/benchmarks/singlediode.py
cwhanse: I expect the random values will be the same (same algorithm, but faster implementation). The timing may be very slightly different - most of the execution time is in computing the 10k IV curves.

pvlib/singlediode.py
```diff
@@ -679,6 +682,105 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
...
+        The value of W(psi)
+
+    '''
+    with np.errstate(over='ignore'):
+        argW = (io / (gsh * a) *
+                np.exp((-i + il + io) /
+                       (gsh * a)))
+    lambertwterm = np.array(lambertw(argW).real)
+
+    idx_inf = np.logical_not(np.isfinite(lambertwterm))
```
kanderso-nrel, line 715: This will cause `nan` to be recalculated, which I'm guessing isn't desirable if the goal is to recalculate for `inf`

pvlib/singlediode.py
```diff
@@ -679,6 +682,105 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
...
+
+def _imp_zero_prime(i, il, io, rs, gsh, a):
+    ''' Derivative of _imp_zero with respect to current i
+    '''
+    idx_p, idx_z = _split_on_gsh(gsh)
+    res = np.full_like(i, np.nan, dtype=np.float64)
+    if np.any(idx_z):
+        # explicit solution for gsh=0
+        t = il[idx_z] + io[idx_z] - i[idx_z]
+        res[idx_z] = 2. / t + i[idx_z] / t**2.
```
kanderso-nrel, line 773: Depending on whether the previous comment is in error, this may also need an additional term

kanderso-nrel (commented):
I have now worked through (most of) the math myself.  

Shall we delete `_golden_sect_DataFrame` and `_pwr_optfcn` if they aren't used anywhere anymore?

Also can we change the URLs for reference 3 in `user_guide/singlediode.rst`?  One of them seems to be behind a Sandia login page and the other is Research Gate.  Better alternatives:
- https://doi.org/10.2172/1177157
- https://www.osti.gov/servlets/purl/1177157

cwhanse:
> Shall we delete `_golden_sect_DataFrame` and `_pwr_optfcn` if they aren't used anywhere anymore?

I vote 'yes'

pvlib/singlediode.py
```diff
@@ -679,6 +682,105 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
...
+def _imp_zero(i, il, io, rs, gsh, a):
+    ''' Root of this function is at imp
+    '''
+    idx_p, idx_z = _split_on_gsh(gsh)
+    res = np.full_like(i, np.nan, dtype=np.float64)
+
+    if np.any(idx_z):
+        # explicit solution for gsh=0
+        t = il[idx_z] + io[idx_z] - i[idx_z]
+        res[idx_z] = i[idx_z] / t - np.log(t / io[idx_z])
```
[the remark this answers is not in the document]
cwhanse: You're correct, I dropped a sign change.

cwhanse:
Moving this back to WIP, because I'm finding convergence problems with corner cases: Rsh=infinity, IL=0. Back to paper and pen.

cwhanse:
Superceded by #2571

Issue #1309, opened 2021-09-15
Add fitting function for Anton's inverter model


Peque:
pvlib includes de `inverter.adr()` model, but it still does not provide a way to infer the required parameter for the model given, for example, some known datasheet parameters. The fitting function would be a great addition to be able to work with Anton's model.

I think @adriesse had a MATLAB implementation, so this issue is to discuss its addition and track progress when/if Anton finds some time to implement the fitting function in Python. :blush:

adriesse:
I have finally taken some time to work in this topic, which has been at the bottom of my to-do list for too long.  You can now find a ready-to-use function to convert from Sandia model parameters to ADR model parameters here:

https://github.com/adriesse/pvpltools-python

There are also helper functions that can be used with suitable measured data, the CEC matrix, or other tabulated specifications.

Related issue: https://github.com/pvlib/pvlib-python/issues/317.

adriesse:
Does this seem useful to anyone?

cwhanse:
> Does this seem useful to anyone?

For this specific issue, the two `fit_xxx` functions in `power_conversion.py` would be valuable to add to `pvlib.inverter` if you are offering that code.

adriesse:
It's written with pvlib in mind, but I'm still thinking about some improvements.

PR #1695, opened 2023-03-12
Update adrinverter parameter database

### LINKED ISSUE #317: Update adrinverter parameter database
I generated the current parameter database from the CEC inverter data in October 2013.  I now have the data to August 2015 and will produce an update.

adriesse:
- [x] Closes #317
 - [x] Tests added

The previous ADR inverter parameter database files were generated using the original lab measurements published by the CEC and the resulting parameters would occasionally predict slightly different AC power compared to the Sandia model and parameters derived from the same measurements.  The question of which was better was not that important because the differences were small.

Since the CEC no longer publishes those measurements, this update simply fits the ADR model to the Sandia model output and the differences in AC output power are now even smaller.  With the ADR coefficients rounded to 5 decimal places (so they don't look too silly) the absolute worst case difference in AC output power between the two models among _all_ matrix points and _all_ inverters is approximately 50 ppm.


adriesse:
Could someone give me a hint how to update test_pvsystem.py?  It's not clear to me where the adr model parameters are coming from...

kanderso-nrel:
I think this is the relevant line for the test failure: https://github.com/adriesse/pvlib-python/blob/adr_db_update/pvlib/tests/test_modelchain.py#L139

Do we have a policy/convention for whether changing entry names in these databases constitutes a breaking change as it relates to semver?  I'm wondering if this needs to wait for 0.10.0.

adriesse:
> Do we have a policy/convention for whether changing entry names in these databases constitutes a breaking change as it relates to semver? I'm wondering if this needs to wait for 0.10.0.

@pvlib/pvlib-maintainer this seems like a call for opinions.

I could append the new data to the old to make it non-breaking.

cwhanse:
As far as changing the keys (names) in the data files, my view: we copy from SAM who imports from the CEC. SAM constructs the product key from the manufacturer and product strings that the CEC publish, which in turn are submitted to the CEC by manufacturers and test labs. As a consequence, product keys change in the SAM file. Until pvlib (and SAM) can unhitch from the CEC's list (which is maintained for different purposes), managing changes in key names is out of our hands. I think the best we can do is to keep some consistency between pvlib and SAM.

As far as appending old data to a new file: I think this is a good idea provided we can screen to avoid having duplicate entries.

kanderso-nrel:
> managing changes in key names is out of our hands

Managing the changes is not _entirely_ out of our hands; we have the option to at least delay distributing database updates in pvlib.

What's not clear to me is whether database names should be considered part of the pvlib API.  On the one hand, distributing tables with changed/removed keys seems as breaking a change as any we make, in that it turns working code into non-working code.  On the other hand we know that these tables are going to continue evolving and I don't think the eventual pvlib 1.0 should have to increment its version number from 1.x to 2.x just to allow an update of the CEC module database.

adriesse:
Pending other inputs I would then tend toward the append solution.  There will be no duplicate records.

adriesse:
I'm afraid I have to ask for help again as I do not understand how the model_chain tests work.  :(

adriesse:
Append complete.

docs/sphinx/source/whatsnew/v0.9.5.rst
```diff
@@ -29,6 +29,8 @@ Enhancements
   :py:func:`pvlib.bifacial.infinite_sheds.get_irradiance_poa`
   to enable use of the hay-davies sky diffuse irradiance model
   instead of the default isotropic model. (:pull:`1668`)
+* Update the ADR inverter database by appending the ADR equivalents of all
+  inverters in the current (2019) Sandia inverter database. (:pull:`1695`)
```
cwhanse, line 33: Did you mean "CEC inverter database"?

adriesse: Actually I meant "Sandia inverter model database".  I'll fix that.

cwhanse: Ah, right. The CEC file contains parameters for the Sandia model. Glad that's not at all confusing :)

adriesse:
Any more volunteers to review this one?  Only one line of code!

cwhanse:
@adriesse if we merge this, can we anticipate having a `fit_adr` function added in the near future?

adriesse:
Here is a link to a fitting function and more:

https://github.com/adriesse/pvpltools-python/blob/master/pvpltools/power_conversion.py

Fitting is basically just throwing the matrix data at scipy curvefit.
To verify a set of parameters, just create a matrix of inputs, run both models and compare.

kanderso-nrel:
I think some rows might have gotten lost in the append, perhaps due to duplicate `Name` entries.  Check for example `Name=='Dow Chemical:  240V [CEC 2011]'`.  Also some columns (`Source`, `Vintage`, `TambLow`, `TambHi`, `Weight`) are all blank in the new rows, is that intentional?

adriesse:
The old table had 12 rows with duplicated keys, which I dropped.  The Dow Chemical one appeared four times actually.  The information for the additional fields is unfortunately not available in my source, the Sandia inverter model parameter database, so the blank fields are indeed sort of intentional.

kanderso-nrel:
The `Name`s are duplicated, but the parameters (at least for the 4x Dow) are not identical, so I'd lean towards keeping them if possible since all four get returned by `pvlib.pvsystem.retrieve_sam('adrinverter')` with pvlib 0.9.4.

adriesse:
I asked myself what anyone would ever do with these duplicates keys except swear at them when the usual syntax for extracting a single set of parameters failed, and drew a blank.  If I was actually tempted to delete the fourth one as well.  But maybe there are other angles I haven't considered.

kanderso-nrel:
Fair.  I was thinking more about backwards compatibility than future utility, but I won't harp on it.  Should we delete the older file and save 300 kB?  Otherwise I think LGTM.

> can we anticipate having a fit_adr function added in the near future?

Tracked in #1309

adriesse:
We could keep the old file for a bit, kind of like a deprecation cycle.  But I'm not particularly attached to it.

kanderso-nrel (approved):
No objection from me.  I'd lean towards deleting the old file but it's not a blocker.

@adriesse FYI there's a merge conflict

cwhanse:
I would delete the old file. We don't keep old versions of the files from SAM.

kanderso-nrel:
Thanks @adriesse!

adriesse:
One less item on my bucket list!

PR #1708, opened 2023-03-24
Update detect_clearsky( )

eccoope:
Modified algorithm in detect_clearsky( ) to support missing timestamps (in a DatetimeIndex of evenly spaced intervals). Supporting functions were also modified to accommodate this change.


pvlib/clearsky.py
```diff
@@ -594,30 +595,48 @@ def _slope_nstd_windowed(slopes, data, H, samples_per_window, sample_interval):
...
+    if samples_per_window % 2 == 0:
+        center_row = samples_per_window//2 - 1
+    else:
+        center_row = samples_per_window//2
+
+    try:
+        # Maintain tz that is stripped when idx is put in H
+        if idx.tz is not None:
+            c = pd.DatetimeIndex(idx.values[H][center_row, :],
+                                 tz = 'UTC').tz_convert(idx.tz)
```
cwhanse, line 626:
```suggestion
                                 tz='UTC').tz_convert(idx.tz)
```

mikofski:
Would it be possible to run the [`detect_clearsky` benchmarks](https://github.com/pvlib/pvlib-python/blob/main/benchmarks/benchmarks/detect_clearsky.py) with and without this PR on the same machine for comparison?

eccoope:
> Would it be possible to run the [`detect_clearsky` benchmarks](https://github.com/pvlib/pvlib-python/blob/main/benchmarks/benchmarks/detect_clearsky.py) with and without this PR on the same machine for comparison?

Yes - just did this on my local machine. The old function takes 0.4933785000030184 seconds to run 10 times and the version in the PR takes 0.9015975999936927 seconds to do the same.

pvlib/clearsky.py
```diff
@@ -635,16 +654,26 @@ def _clear_sample_index(clear_windows, samples_per_window, align, H):
...
+    # Boolean mask for column indices of intervals with temporal gaps
+    gap_cols = [True if c not in gaps else False for c in range(0,
+                len(clear_windows) - (samples_per_window - 1))]
+    mask = np.logical_and(clear_cols, gap_cols)
+    clear_samples = np.unique(H[:, mask])
     return clear_samples
 
 
-def detect_clearsky(measured, clearsky, times=None, window_length=10,
+def detect_clearsky(measured, clear_sky, times=None, window_length=10,
```
cwhanse, line 676: Is the arg name change `clearsky` to `clear_sky` intentional? If we keep this, need to modify the docstring.

pvlib/clearsky.py
```diff
@@ -765,10 +797,17 @@ def detect_clearsky(measured, clearsky, times=None, window_length=10,
...
-        clear = pd.Series(clearsky, index=times)
+    if not isinstance(clear_sky, pd.Series):
+        clear = pd.Series(clear_sky, index=times)
+    # This clause is designed to address cases where measured has missing time
+    # steps - if this is the case, clear should be set to have the same
+    # missing time intervals as measured. Not doing this may cause issues with
+    # arrays of different lengths when evaluating comparison criteria and
+    # when indexing the Hankel matrix to construct clear_samples
+    elif len(clear_sky.index) != len(times):
+        clear = pd.Series(clear_sky, index=times)
```
cwhanse, line 808: Untested, and I suspect this will fail - `clear_sky` is a `Series` of different length than `times` so it can't get indexed by `times`.

cwhanse (approved):
IMO the performance hit is worth the enhancement of handling missing timesteps.

pvlib/clearsky.py
eccoope: I was trying to avoid overloading the clearsky namespace in my development environment, as I also had clearsky defined with `from pvlib import clearsky`. If this change is okay with the maintainers, I will modify the docstring.

cwhanse: I think we want to avoid changing the arg name here.

pvlib/clearsky.py
eccoope: Just added a test for this in test_clearsky.py

pvlib/clearsky.py
eccoope: Reverted to original name

pvlib/tools.py
```diff
@@ -7,6 +7,7 @@
 import pandas as pd
 import pytz
 import warnings
+from .conftest import DATA_DIR
```
cwhanse, line 10: This line should be `test_tools`, that will fix the doc build error.

eccoope: Thanks @cwhanse !

AdamRJensen:
@cwhanse is this still be worked on from Sandias side?

cwhanse:
> @cwhanse is this still be worked on from Sandias side?

Let me talk a look at it. I had forgotten.

Issue #1722, opened 2023-05-03
stickler-ci is shutting down


kandersolar:
It seems the stickler-ci service is shutting down: https://stickler-ci.com/shutting-down

>  Stickler CI will remain available in a read only fashion until June 30 2023.. After that, login will cease to work and the site will fall off the internet.

That website seems to suggest that it should still be working for the moment, but @AdamRJensen noticed that it's not reporting a status to #1395, so it might be inactive already for us.

Thinking about how we might replace it: instead of using an external service, [pvanalytics](https://github.com/pvlib/pvanalytics/blob/main/.github/workflows/lint-and-test.yml#L63) and [twoaxistracking](https://github.com/pvlib/twoaxistracking/blob/main/.github/workflows/lint.yml#L9) both DIY with a custom flake8 action.  This approach doesn't comment in-line like stickler does, which is a downside especially for new contributors.  Probably there are nicer pre-built linter services we could switch to.

mikofski:
[Pycodestyle](https://pypi.org/project/pycodestyle/) returns line item results we could pipe that to an artifact. No preference between this and flake8

i also notice there are many GitHub actions for python code style checking. IDK if any will offer the same level of feedback, but maybe?
* this action created stickler like GitHub annotations from flake8 output: https://github.com/marketplace/actions/flake8-annotator
* here’s a long list of flake 8 actions including the annotation one I just mentioned, some that are defunct, & some that are more popular than others: https://github.com/marketplace?type=actions&query=Flake8
* Here’s a list of a few pycodestyle actions: https://github.com/marketplace?type=actions&query=Pycodestyle
* Here’s one for pep8: https://github.com/marketplace?type=actions&query=pep8+
* Here’s another random python code style checker: https://github.com/marketplace?type=actions&query=python+code+style+checker+

kandersolar:
> it's not reporting a status to https://github.com/pvlib/pvlib-python/pull/1395, so it might be inactive already for us.

Well I see it's running on other PRs now.  Not sure why it's not running on #1395.  Anyway I guess we don't have to replace it immediately.

adriesse:
> It seems the stickler-ci service is shutting down

Yay, no more annoying stickler error messages!!  :)

wholmgren:
[...]

PR #1723, opened 2023-05-03
Remove Stickler-CI integration files

AdamRJensen:
As described in #1722, the stickler-ci service is no longer working and is shutting down.

Thus, this PR removes superfluous files associated with stickler-ci.


kandersolar (approved):
LGTM!

PR #1748, opened 2023-05-25
Improve `pvlib.spa` speed

kandersolar:

Some functions in `pvlib.spa` are doing quite a lot of unnecessary computation:  

- For the table-based computations, our code combines arrays of different lengths into a single 2d array, thus padding the shorter arrays with zero, and creating a bunch of unnecessary busy work to process those dummy entries.
- The `longitude_nutation` and `obliquity_nutation` functions perform the same expensive calculation
- To much lesser effect, there are several instances of a single function containing multiple calls to `np.radians(x)`.

 This PR prevents those unnecessary computations and reduces overall calculation time by 40-50% for both numpy and numba.  The table below compares runtimes in seconds:

|       n |   numpy, PR |   numpy, main |   numba, PR |   numba, main |
|--------:|------------:|--------------:|------------:|--------------:|
|      10 |       0.002 |         0.004 |       0     |         0     |
|     100 |       0.003 |         0.004 |       0     |         0.001 |
|    1000 |       0.006 |         0.011 |       0.001 |         0.002 |
|   10000 |       0.041 |         0.069 |       0.009 |         0.017 |
|  100000 |       0.4   |         0.657 |       0.083 |         0.161 |
| 1000000 |       4.654 |         8.492 |       0.864 |         1.633 |

Regarding the table-length issue, it might be better to  do away  with the combined `HELIO_LONG_TABLE` et al. entirely and just keep the arrays as separate module-level variables.


kandersolar:
Ready for review.  The drop in reported coverage should be ignored; it's in the part of `pvlib.spa` that is run in numba-mode only, and coverage tracking doesn't work with numba functions.

cwhanse (approved):

pvlib/spa.py
```diff
@@ -468,60 +456,32 @@ def julian_ephemeris_millennium(julian_ephemeris_century):
     return jme
 
 
+@jcompile(nopython=True)
```
wholmgren, line 459: Do we need to declare types in this jcompile like we do in most of the other functions in this module? I haven't kept up with numba development.

docs/sphinx/source/whatsnew/v0.9.6.rst
```diff
@@ -30,6 +30,9 @@ Enhancements
   the option of controlling what NSRDB endpoint is used. (:pull:`1736`)
 * :py:func:`pvlib.iotools.get_psm3` now uses the new NSRDB 3.2.2 endpoint for
   hourly and half-hourly single-year datasets. (:issue:`1591`, :pull:`1736`)
+* :py:mod:`pvlib.spa`, and :py:func:`pvlib.solarposition.get_solarposition`/
+  :py:meth:`pvlib.location.Location.get_solarposition` with ``method='nrel_numpy'``
+  or ``method='nrel_numba'``, are now 50-100% faster than before. (:pull:`1748`)
```
wholmgren, line 35: There's also `pvlib.solarposition.spa_python` and `ModelChain.run_model*`. Maybe less is more in this situation: `The default solar position algorithm (NREL SPA) is now 50-100% faster.`

wholmgren (approved):
Great work @kandersolar!

pvlib/spa.py
kandersolar: I'm not an expert here either, but I think specifying types is not necessary in general; when using numba for JIT compiling like we do, it can infer types as needed.  

In this particular case, I did try to specify types just for consistency with the rest of the code.  However it seems that specifying the appropriate type for module-level array constants is not possible using the shorthand string notation.  It is possible by importing numba, but I didn't see a clean way to no-op the type declaration when numba isn't installed, so I opted to just leave the type signature empty here instead.  I've added a comment along these lines (should have done that originally).

Issue #1056, opened 2020-09-09
Organize pvlib.data


cwhanse:
pvlib.data currently contains 1) databases for module and inverter models; 2) Linke turbidity values; 3) data files for tests and examples and 4) `variable_style_rules.csv`. Accurately described as a "junk drawer."  The majority of the data files are in category 3, supporting tests.

As a start, maybe create `data.tests` and perhaps subfolders within `tests` that mirror the subfolders in `pvlib.tests` And perhaps add a prefix or other text to file names to help identify where or how it is used, e.g., `PVsyst_demo.csv` becomes `test_sdm_pvsyst_demo.csv`.

wholmgren:
Mirroring the structure would be an improvement.

Another option to consider is moving the subpackage tests into the subpackage, along with a data subdirectory within that test directory. For example:

```
pvlib/data  # databases, linke turbidity, anything a user might need
pvlib/tests  # test_atmosphere.py, etc
pvlib/tests/data  # singleaxis_tracker_wslope.csv, etc
pvlib/iotools/tests  # test_tmy.py, etc
pvlib/iotools/tests/data  # pvgis_tmy_test.dat, etc
pvlib/ivtools/tests
pvlib/ivtools/tests/data
```

I proposed a similar structure when we created `pvlib/tests/iotools`. I was outvoted but I still think it's better!

cwhanse:
I'm in favor of `pvlib/tests/data`, etc. rather than `pvlib/data/tests`.

echedey-ls:
68 files to organize... I propose splitting up the work to make it manageable. I'd say the first step is to categorize each file. Feel free to react to this message with a 👍 if you want to contribute to that. I'll split the work in ranges for each one to work on it by mentioning you in this message next week.

but for those brave enough, you can work on it now: https://docs.google.com/spreadsheets/d/12LeEFa9-wRqc3v7utfgcTk96KTmaWfhHSPkLx6K0QhY/edit?usp=sharing

There are five categories, four for what Cliff said in this issue plus one if it's unknown; multiple can be selected for each entry. My way to go would be to look up where this files are mentioned and select the appropriate labels. Could be automated, but I don't feel like overengineering today - and that wouldn't take into account files not mentioned anywhere.

React 👎  if you are against doing it this way (and potentially have a better idea)

cwhanse:
@echedey-ls can you add "Lookup table" or something like that to the pull down options? For files like the CEC module parameters.

[...]

Issue #473, opened 2018-06-06
Include tests in PyPI source tarball


jluttine:
It would be nice if the PyPI source tarball included the tests so one can run them after installing to check that it works. Running package tests is extremely useful and important when packaging for a Linux distro as I'm doing at the moment, so I'd love to have the tests included. I think this is good practice.

wholmgren:
I'm not opposed but is this the standard for other "pydata" projects? I usually look to pandas (big), xarray (medium), metpy (smaller) as examples to try to follow. How do they manage this?

jluttine:
Pandas and xarray include tests in PyPI source tarball. For some reason, metpy doesn't.

Some references I found quickly:

- https://docs.python.org/2/distutils/sourcedist.html#specifying-the-files-to-distribute
- https://stackoverflow.com/a/20180000

mikofski:
I usually put tests and documentation source in the `sdist` or source distribution which is uploaded to PyPI as a tarball or zip file, but not in the wheels or other binary distributions.

If the tests include a lot of test data then it would be an unnecessary burden to put the in wheels, but I suppose if they are smell enough then it doesn't matter.

There was a discussion about this on the SciPy Dev mail list, in their case the tests were so big that they're considering moving the test data to a separate repo altogether, and either accessing it through an API or having testers clone the repo. There is an upside to have a shared public test data repo for multiple projects with similar test data, then the data can also be used in the docs as examples and the examples can be more realistic or rocket in their demonstrative power

wholmgren:
With only a quick read of our manifest/setup.py and comparing with xarray/pandas, I am surprised that our tests are not already bundled. What do we need to do differently? Pull request welcome. Or edit the release procedures wiki.

Note that the ``test_forecast.py`` tests frequently fail due to issues with the data on the server (I think) so you'd probably want to exclude those and/or improve them.

A separate test repo sounds like a lot of work to me.

mikofski:
There already is a shared test data repo for R, I think maybe it's [datasets](http://stat.ethz.ch/R-manual/R-devel/library/datasets/html/00Index.html) there are efforts in Python to read them I think? As I said you can read [the thread in SciPy devs starting in March](https://mail.python.org/pipermail/scipy-dev/2018-March/thread.html)

wholmgren:
closed by #580

PR #1763, opened 2023-06-07
Add function to retrieve example dataset paths

### LINKED ISSUE #924: Add function to retrieve example datasets
It would be nice to have a clean way for gallery examples to get access to the test data files.  See https://github.com/pvlib/pvlib-python/pull/860#discussion_r379847062 

For a function like `load_dataset('greensboro-tmy')`:

Pros:
- No need to monkey around with filepaths, especially ones that aren't really meant to be public
- Associating files with keys means we can move and rename test data files without it being a breaking change
- Simplifies example code

Cons:
- The data files are in several formats (csv, json, nc, h5 etc), so this function would either have to know the appropriate reading method for each file (complicated) or just return a file handle and let the user parse the contents (less useful).

echedey-ls:
Done: 
 - [x] Closes #924
 - [x] Tests added
     ~About this one, there isn't a `tools.rst` file, so I believe it doesn't need referencing.~

I'll wait for milestone assignment:

Adds a `get_example_dataset_path()` function to `pvlib.tools` that returns a filepath to a dataset bundled with pvlib. See linked issue above.

Here is the [documentation link](https://pvlib-python--1763.org.readthedocs.build/en/1763/reference/generated/pvlib.tools.get_example_dataset_path.html) [UPDATED 2023-06-18].


echedey-ls:
This issue is related to #1056. However, I haven't addressed any of it since in first place I don't know which purpose each file has.
Tweaking this new function would leverage work put into changing the structure.

pvlib/tests/test_tools.py
```diff
@@ -95,3 +96,17 @@ def test_degrees_to_index_1():
     'latitude' or 'longitude' is passed."""
     with pytest.raises(IndexError):  # invalid value for coordinate argument
         tools._degrees_to_index(degrees=22.0, coordinate='width')
+
+
+def test_dataset_passes():
+    expected_dataset = '723170TYA.CSV'
+    assert tools.dataset(expected_dataset).endswith(expected_dataset)
+    assert tools.dataset(pathlib.Path(expected_dataset)) \
+        .endswith(expected_dataset)
```
kandersolar, line 105: I suggest checking that something exists at the returned file system location

pvlib/tools.py
```diff
@@ -469,3 +472,15 @@ def _first_order_centered_difference(f, x0, dx=DX, args=()):
...
+
+def dataset(dataset):
+    """
+    Return a filepath to a dataset bundled with PVLIB with name `dataset`.
+    This utility is intended to be used in tests and examples.
+    """
+    dataset = os.path.join(pvlib.__path__[0], 'data', dataset)
+    if not os.path.exists(dataset):
+        raise IOError("Dataset has not been found in pvlib. "
+                      "Please check dataset name.")
```
kandersolar, line 485: I think I'd make this a `ValueError`

pvlib/tools.py
```diff
@@ -469,3 +472,15 @@ def _first_order_centered_difference(f, x0, dx=DX, args=()):
...
     df = f(x0+dx, *args) - f(x0-dx, *args)
     return df / 2 / dx
+
+
+def dataset(dataset):
+    """
+    Return a filepath to a dataset bundled with PVLIB with name `dataset`.
+    This utility is intended to be used in tests and examples.
+    """
+    dataset = os.path.join(pvlib.__path__[0], 'data', dataset)
```
kandersolar, line 482: Should we use pathlib for this?  Doesn't really matter, but might as well follow best practice (which I think is pathlib nowadays)

docs/examples/soiling/plot_greensboro_kimber_soiling.py
```diff
@@ -30,17 +30,16 @@
...
-import pathlib
 from matplotlib import pyplot as plt
 from pvlib.iotools import read_tmy3
 from pvlib.soiling import kimber
-import pvlib
+from pvlib.tools import dataset
 
 # get full path to the data directory
-DATA_DIR = pathlib.Path(pvlib.__file__).parent / 'data'
+DATASET_DIR = dataset('723170TYA.CSV')
```
kandersolar, line 39: Looking at this example code, I think we should consider more descriptive names.  I'm worried that people will see it in an example and think that it's for retrieving all datasets (whatever that means), not just the example datasets included in `pvlib.data.`  `locate_example_dataset()`, maybe?

I'd also name this variable `tmy_filepath` or something like that, and update the comment since it's now a specific file instead of a directory.

kandersolar (commented):
Thanks @echedey-ls :)

> there isn't a tools.rst file, so I believe it doesn't need referencing.

It's true that we don't usually document the functions in `pvlib.tools`, but in this case I think we probably should so that curious users can see what the function does and doesn't do.  How about a new entry list at the bottom of the [iotools page](https://pvlib-python.readthedocs.io/en/stable/reference/iotools.html), something like this:

```
Functions for locating the example data files included in pvlib.

.. autosummary::
   :toctree: generated/

   tools.locate_example_dataset
```

If we do that, the function will need a complete docstring (parameter descriptions, etc).

pvlib/tests/test_tools.py
echedey-ls: Good point.

pvlib/tools.py
echedey-ls: I had in mind keeping it as simple as possible, but this [StackOverflow post](https://stackoverflow.com/questions/67112343/pathlib-vs-os-path-join-in-python) suggests using pathlib too. I'll change it.

docs/examples/soiling/plot_greensboro_kimber_soiling.py
echedey-ls: Fair point, although in that sense I think (English is not my mother tongue) something along `get_dataset_path` makes more sense, maybe? Just to bring up other options, I'll stick to your suggestion for now.

Thanks for spotting that of the comments and the var name.

echedey-ls:
Here is the [documentation link](https://pvlib-python--1763.org.readthedocs.build/en/1763/reference/generated/pvlib.tools.locate_example_dataset.html#pvlib.tools.locate_example_dataset).

docs/examples/soiling/plot_greensboro_kimber_soiling.py
kandersolar: Now I am wondering if having `locate` in the function name might make people think it has something to do with `pvlib.location.Location`...

I like that `get_dataset_path` makes it clear that the function returns a path to a dataset (not the dataset itself), but I still think the name should also make it clear that it is just for example/test datasets.  `get_test_dataset_path`?

echedey-ls: That's a good point. I like `get_test_dataset_path`.
If I or anyone addresses #1056, another function called `get_example_dataset_path` would also make sense. Do you think that this one should be the latter, since the current path is considered to remain for example files (from [this comment](https://github.com/pvlib/pvlib-python/issues/1056#issuecomment-689714408))? And leave the former for test files?

I will rename it tomorrow.

kandersolar: Good question.  On first thought, I don't think we should have a function for each folder, since depending on where we go with #1056, we might end up with multiple `data` folders inside `tests`.  What do you think about having a single function, but with an optional `location='data'` parameter to allow finding the file in a custom folder?  For example `get_test_dataset_path(dataset='723170TYA.CSV', location='tests/data')`?  And then the path would look like `dataset = pathlib.Path(pvlib.__path__[0], location, dataset)`.

echedey-ls: I couldn't like that approach more.
Just a quick detail, should `location` default to user example data? Or do we prefer being verbose on that?

kandersolar: Yeah, it makes sense to me for it to default to `data` so files will be found in `pvlib/data` by default.  That means for now the `location` parameter probably won't be useful to anyone, since all files are in `pvlib/data` right now. But if we expand to have multiple `data` folders in the future, then it will be useful.

I guess that means maybe we shouldn't bother adding the parameter yet, since there is no use for it today?  We could leave that for future work, and just leave a note about it in #1056.

echedey-ls:
This should be ok now.
Please have a look at the docs. I've created a table to leverage work on the other issue. Maybe it's better to position it on another paragraph.

kandersolar (approved):
This looks great!  @echedey-ls can you resolve the merge conflict?  After that I think this will be ready to go.

Maybe @wholmgren would like to review as well.

wholmgren (commented):
I'm concerned about using 1 public function to do three things:

1. load resources for pvlib code
2. load resources for test code
3. load resources for documentation code

The function name `get_test_dataset_path` implies that it's only for tests, yet it's used for all 3 of those items. Seems ripe for problems down the road.

I also don't think we need this, at least not until we clean up our data directories and make it clear what is public and what is not. I think the "preferably one way to do it" is now to use [`importlib.resources.files`](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files). Example below

```python
In [3]: import importlib.resources

In [6]: importlib.resources.files('pvlib') / 'data'
Out[6]: WindowsPath('C:/Users/WILHOL/Miniconda3/envs/pvdeg/lib/site-packages/pvlib/data')

In [7]: tmy3_filepath = importlib.resources.files('pvlib') / 'data' / '723170TYA.CSV'

In [8]: tmy3_filepath.exists()
Out[8]: True
```

kandersolar:
Good to know about `importlib.resources.files`!  It does require python 3.9 though.

> The function name get_test_dataset_path implies that it's only for tests, yet it's used for all 3 of those items. 

I suggested including `test` in the function name in an attempt to prevent confusing it for something like an `iotools.get_` function; the thought was to try to make it clear that it's limited to only a select set of special files, not a general purpose data tool.  But if we're looking to draw a line between test (as in `pvlib.tests`) and other types of data then I can see that it could be confusing.

But I don't see why one function for three types of data files might cause problems later on.  If all the function does is `figure_out_where_pvlib_is() + path_parameter`, what's it matter whether the file is `LinkeTurbidities.h5` or a testing file?

wholmgren:
There's an [importlib-resources](https://importlib-resources.readthedocs.io/en/latest/) backport for older python. We [already use](https://github.com/pvlib/pvlib-python/blob/763dd0a7d0bc0f195315717bdf2a6d8b4c7169fc/setup.py#L43) this approach for `importlib-metadata` for python < 3.8. This [realpython](https://realpython.com/python37-new-features/#importing-data-files-with-importlibresources) example might be helpful too.

I think your [original list](https://github.com/pvlib/pvlib-python/issues/924#issue-574040355) of pros to a function are still valid. There the focus is on a function for example code and maybe user onboarding that separates public resources from package structure. If the user-facing function only does that then it might be a local win. On the other hand, it can also be good to propagate standard python patterns in example code so it's less clear that it's a global win. I think the internal code should keep it simple and stick to importlib-resources.

echedey-ls:
I understand the issue. We can limit the scope of this function until #1056 is addressed.
I agree with @wholmgren criteria on that this PR should only satisfy user needs. I will revert unrelated changes.

> I suggested including test in the function name in an attempt to prevent confusing it for something like an `iotools.get_` function

@kandersolar what is your opinion on renaming it to `get_example_dataset_path` (or `get_pvlib_example_dataset_path`, `get_bundled_dataset_path`, maybe consider other synonym of 'bundled')?

I'd like to bring attention to two points:
* Importlib imports the whole module with `importlib.resources.files('pvlib')`. I don't know how much overhead that introduces compared to using `__path__[0]` or `__file__,` but I expect it to be a little higher. Not much, but just to consider the micro optimization.
* Tests are being shipped with pvlib when it's installed. Is that on purpose?

echedey-ls:
This should be fine to go. Renamed to `get_example_dataset_path` and use cases are only in the examples.

cwhanse (approved):

kandersolar:
> Importlib imports the whole module with importlib.resources.files('pvlib'). I don't know how much overhead that introduces compared to using `__path__[0]` or `__file__`, but I expect it to be a little higher. Not much, but just to consider the micro optimization.

It seems to use the same `sys.modules` cache as `import` does, so as long as we are assuming `pvlib` is already imported, or will be imported later, then I don't think there is any significant penalty.  

> Tests are being shipped with pvlib when it's installed. Is that on purpose?

Yes, see #473.  There is a non-pvlib discussion [here](https://discuss.python.org/t/should-sdists-include-docs-and-tests/14578) on whether including tests in sdists and wheels is a good idea in general (tl;dr opinions vary).

echedey-ls:
That's good to know @kandersolar !

echedey-ls:
Closing this due to inactivity. If there's interest in the future, feel free to ping me.

PR #1766, opened 2023-06-08
Remove `pvlib.forecast`

kandersolar:
- [x] Closes #1596 
 - [x] Closes #1417
 - [x] Closes #918
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [x] Tests added
 - [x] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - ~[ ] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.~
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

Following the thumbs up on https://github.com/pvlib/pvlib-python/issues/1735#issuecomment-1564605782, this PR removes `pvlib.forecast`.  Removing `pvlib.forecast`, combined with #1654, means that several optional dependencies (siphon, netcdf4, cftime) are no longer used anywhere and can also be removed.  This PR removes them here, but we should also remember to remove them from our conda-forge [feedstock](https://github.com/conda-forge/pvlib-python-feedstock/blob/main/recipe/meta.yaml) when we release v0.10.0.

Issue #1420, opened 2022-03-03
Ensure map_variables returns altitude and not elevation


AdamRJensen:
**Describe the bug**
A number of the iotools do not map `elevation` to `altitude` as is preferred (see this [comment](https://github.com/pvlib/pvlib-python/pull/1374#discussion_r818867940)).

For example, ``_parse_pvgis_tmy_csv`` and ``_parse_pvgis_tmy_csv`` includes elevation as a dict entry and so does ``_parse_pvgis_tmy_json`` and surely other PVGIS functions behind the scenes, without necessarily being specified in our code. ``read_surfrad`` also returns elevation. 

Also, as `altitude` is the preferred pvlib terminology, it should also be added to the [Variables and Symbols](https://pvlib-python.readthedocs.io/en/latest/user_guide/variables_style_rules.html#variables-style-rules) list.

soma2000-lang:
@AdamRJensen Please guide how to go about working on these kinds of issues. I would like to contribute more to this repo.

Issue #911, opened 2020-02-25
Sphinx emits lots of warnings when building docs


kanderso-nrel:
Sphinx currently emits a few types of warnings when building the docs, none are really a big deal but it clutters the output and makes #909 less effective.  I'll open a PR shortly to address:

Docutils warning
----------------------
```
/usr/share/miniconda/envs/test_env/lib/python3.8/site-packages/sphinx/util/nodes.py:94: FutureWarning: 
   The iterable returned by Node.traverse()
   will become an iterator instead of a list in Docutils > 0.16.
```

I'm not 100% sure what this one is, but I assume it is related to the newest version of docutils deprecating something and sphinx hasn't caught up yet.  [example](https://dev.azure.com/solararbiter/pvlib%20python/_build/results?buildId=2804&view=logs&j=7e620c85-24a8-5ffa-8b1f-642bc9b1fc36&t=2febcae8-7f6a-51bb-45aa-d6e72c9ea7e2&l=128)

I'll play around with pinning the docutils version and see if that fixes it. 

Improper RST in gallery example
------------------------------------------

The I-V curve example generates `WARNING: Title underline too short.` because the page title got updated without extending the underline (my bad).  Easy fix. [link](https://dev.azure.com/solararbiter/pvlib%20python/_build/results?buildId=2804&view=logs&j=7e620c85-24a8-5ffa-8b1f-642bc9b1fc36&t=2febcae8-7f6a-51bb-45aa-d6e72c9ea7e2&l=187) 

First-line references
-------------------------

The first lines of docstrings get orphaned when they are copied into the API Reference docs page and Sphinx emits a bunch of warnings about it: [example](https://dev.azure.com/solararbiter/pvlib%20python/_build/results?buildId=2804&view=logs&j=7e620c85-24a8-5ffa-8b1f-642bc9b1fc36&t=2febcae8-7f6a-51bb-45aa-d6e72c9ea7e2&l=154)

There were a few votes in #909 to remove references from first lines of docstrings to prevent these warnings.  If nobody objects, I'll move references into the second part of the docstring, for instance:

Current:
```
def nrel_earthsun_distance(time, how='numpy', delta_t=67.0, numthreads=4):
    """
    Calculates the distance from the earth to the sun using the
    NREL SPA algorithm described in [1]_.

    Parameters
    ----------
    time : pandas.DatetimeIndex
```

New:
```
def nrel_earthsun_distance(time, how='numpy', delta_t=67.0, numthreads=4):
[...]

PR #1767, opened 2023-06-10
Add retrieval functions for daily precip, snow, and temperature data from NOAA RCC ACIS

### LINKED ISSUE #1293: get PRISM rainfall data
**Problem**
[PRISM](https://prism.oregonstate.edu/) doesn't have an intuitive API. It has rainfall and other info, but he interface is an AJAX call to a PHP backend which is hidden by JavaScript [here](https://prism.oregonstate.edu/explorer/dataexplorer/js/explorer.js) which I can't figure it out. Maybe there's a way to use the JavaScript to get the data and add it to iotools. Then the data could be used for soiling. See [HSU soiling example](https://pvlib-python.readthedocs.io/en/latest/auto_examples/plot_fig3A_hsu_soiling_example.html)

**Solution**
in iotools, `get_prism_rainfall()` that uses the JavaScript https://prism.oregonstate.edu/explorer/dataexplorer/js/explorer.js to get the rainfall

Just point users to the website, it also has a bulk download tool. Maybe they don't want folks to use the site programmatically?

**Additional context**
JavaScript modules used:
* https://prism.oregonstate.edu/explorer/dataexplorer/js/explorer.js
* https://prism.oregonstate.edu/explorer/dataexplorer/js/data_stability.js

base URL of AJAX post call:
* https://prism.oregonstate.edu/explorer/dataexplorer/rpc.php
but it returns 500 when I post anything, since I don't know what to post

there's also a web service to download an entire monthly or daily grid:
* https://prism.oregonstate.edu/documents/PRISM_downloads_web_service.pdf
or bulk
* https://prism.oregonstate.edu/documents/PRISM_downloads_FTP.pdf

maybe that's the better way to go?

kandersolar:
- [x] Closes #1293
 - [x] Tests added

ACIS provides precipitation estimates from PRISM, plus three other gridded datasets.  No registration needed.

ACIS has another API endpoint for station-recorded (not gridded) data that supplies snowfall, among other variables.  ~Possible future PR for that.~ now included in this PR


kandersolar:
Test failures are something to do with forecasts which I'm not excited about debugging given #1766.

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,105 @@
+import requests
+import pandas as pd
+import numpy as np
+
+
+def get_acis_precipitation(latitude, longitude, start, end, dataset,
+                           url="https://data.rcc-acis.org/GridData", **kwargs):
```
AdamRJensen, line 7: Is there a preference for having the URL defined as a module-level parameter vs. having it as a default value?

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    Retrieve daily gridded precipitation data from the Applied Climate
```
AdamRJensen, line 9: By saying "gridded" I'm getting the impression that I can retrieve multiple locations at the same time (gridded data). Perhaps replace gridded with modeled (if it is modeled data)?

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,105 @@
...
+        {"name": "pcpn", "interval": "dly", "units": "mm"},
+    ]
+    params = {
+        'loc': f"{longitude},{latitude}",
+        'sdate': pd.to_datetime(start).strftime('%Y-%m-%d'),
+        'edate': pd.to_datetime(end).strftime('%Y-%m-%d'),
+        'grid': str(dataset),
+        'elems': elems,
+        'output': 'json',
+        'meta': ["ll"],  # "elev" should work, but it errors for some databases
```
AdamRJensen, line 86: Can you elaborate on this?

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,105 @@
...
+    .. [5] `PRISM <https://prism.oregonstate.edu/>`_
+    """
+    elems = [
+        # other variables exist, but are not of interest for PV modeling
+        {"name": "pcpn", "interval": "dly", "units": "mm"},
+    ]
+    params = {
+        'loc': f"{longitude},{latitude}",
+        'sdate': pd.to_datetime(start).strftime('%Y-%m-%d'),
+        'edate': pd.to_datetime(end).strftime('%Y-%m-%d'),
```
AdamRJensen, line 82: Why not just `start.strftime('%Y-%m-%d')`?

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,105 @@
...
+    .. [2] `ACIS Gridded Data <http://www.rcc-acis.org/docs_gridded.html>`_
+    .. [3] `NRCC <http://www.nrcc.cornell.edu/>`_
+    .. [4] `Multisensor Precipitation Estimates
+           <https://www.weather.gov/marfc/Multisensor_Precipitation>`_
+    .. [5] `PRISM <https://prism.oregonstate.edu/>`_
+    """
+    elems = [
+        # other variables exist, but are not of interest for PV modeling
+        {"name": "pcpn", "interval": "dly", "units": "mm"},
+    ]
```
AdamRJensen, line 78: Could it be an idea to make this a module-level constant? A lot of non-PV people use the pvlib iotools so they may want other parameters.

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,105 @@
...
+
+def get_acis_precipitation(latitude, longitude, start, end, dataset,
+                           url="https://data.rcc-acis.org/GridData", **kwargs):
+    """
+    Retrieve daily gridded precipitation data from the Applied Climate
+    Information System (ACIS).
+
+    The Applied Climate Information System (ACIS) was developed and is
+    maintained by the NOAA Regional Climate Centers (RCCs) and brings together
+    climate data from many sources.
```
AdamRJensen, line 14: Is it only US coverage?

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,105 @@
+import requests
+import pandas as pd
+import numpy as np
+
+
+def get_acis_precipitation(latitude, longitude, start, end, dataset,
```
AdamRJensen, line 6: Could we recommend a default dataset?

AdamRJensen (commented):
Awesome contribution!

pvlib/iotools/acis.py
kandersolar: For the convenience of allowing `start` to be a string like `'2021-01-01'`.  I wish all `iotools` functions did this :)

I've added a comment to explain.

pvlib/iotools/acis.py
kandersolar: Elaborated

kandersolar:
After further consideration, I think there are enough differences between datasets that it makes sense to rewrite this PR so that there is a 1:1 correspondence between `get_` functions here and datasets in the API.  Otherwise the code becomes a mess trying to handle all the datasets in a single function.  Marking this PR back to draft for now.

kandersolar:
@AdamRJensen this is ready for another look.  In addition to splitting up the function as mentioned above, I also added functions for the weather station dataset.

A discussion point: currently, `get_acis_station_data` returns precipitation in mm but snow in cm, for consistency with relevant PV models (`pvlib.soiling` takes rainfall in mm; `pvlib.snow` takes snowfall in `cm`).  Is that the right choice?

Test failures are `pvlib.forecast` and can be ignored here.

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+    df = pd.DataFrame(payload['data'], columns=columns)
+    df = df.set_index('date')
+    df.index = pd.to_datetime(df.index)
+    df.index.name = None
+
+    metadata = payload['meta']
+    # for StnData endpoint, unpack combination "ll" into lat, lon 
+    ll = metadata.pop('ll', None)
+    if ll:
+        metadata['lon'], metadata['lat'] = ll
```
AdamRJensen, line 56:
```suggestion
    if 'll' in metadata.keys():
         metadata['lon'], metadata['lat'] = metadata.pop('ll')
```
Perhaps a bit simple?

*Update:* Why not use try/except as done below?

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
+import requests
+import pandas as pd
+import numpy as np
+
+
+VARIABLE_MAP = {
+    # time series names
+    'pcpn': 'precipitation',
+    'maxt': 'temp_air_max',
+    'avgt': 'temp_air_average',
```
AdamRJensen, line 10: Perhaps remove `_average` if this is almost always the temperature of interest for PV applications? For example, the bsrn function doesn't append `_average` but does append `_min` and `_max`.

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+    'obst': 'temp_air_observation',
+    'mint': 'temp_air_min',
+    'snow': 'snowfall',
+    'snwd': 'snowdepth',
+
+    # metadata names
+    'lat': 'latitude',
+    'lon': 'longitude',
+    'elev': 'altitude',
```
AdamRJensen, line 19: I guess whether to map it to elevation or altitude is arbitrary in this case? I think pvlib uses both conventions?

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+        First day of the requested period
+    end : datetime-like
+        Last day of the requested period
+    map_variables : bool, default True
+        When True, rename data columns and metadata keys to pvlib variable
+        names where applicable. See variable :const:`VARIABLE_MAP`.
+    url : str, default: 'https://data.rcc-acis.org/GridData'
+        API endpoint URL
+    kwargs:
+        Optional parameters passed to ``requests.get``.
```
AdamRJensen, line 99:
```suggestion
        Optional parameters passed to ``requests.post``.
```

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+    Raises
+    ------
+    requests.HTTPError
+        A message from the ACIS server if the request is rejected
+
+    Notes
+    -----
+    PRISM data is aggregated from 12:00 to 12:00 UTC, meaning data labeled
+    May 26 reflects to the 24 hours ending at 7:00am Eastern Standard Time
+    on May 26.
```
AdamRJensen, line 117: Just to make sure I understand it correct, it goes from the previous day at 12:00 UTC to the current day 12:00 UTC?

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+        Last day of the requested period
+    map_variables : bool, default True
+        When True, rename data columns and metadata keys to pvlib variable
+        names where applicable. See variable :const:`VARIABLE_MAP`.
+    trace_val : float, default 0.001
+        Value to replace "trace" values in the precipitation data
+    url : str, default: 'https://data.rcc-acis.org/GridData'
+        API endpoint URL
+    kwargs:
+        Optional parameters passed to ``requests.get``.
```
AdamRJensen, line 329:
```suggestion
        Optional parameters passed to ``requests.post``.
```

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+        in decimal degrees, between -180 and 180, east is positive
+    start : datetime-like, optional
+        If specified, return only stations that have data between ``start`` and
+        ``end``.  If not specified, all stations in the region are returned.
+    end : datetime-like, optional
+        See ``start``
+    url : str, default: 'https://data.rcc-acis.org/StnMeta'
+        API endpoint URL
+    kwargs:
+        Optional parameters passed to ``requests.get``.
```
AdamRJensen, line 418:
```suggestion
        Optional parameters passed to ``requests.post``.
```

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+    grid : int
+        Options are either 1 (for "NRCC Interpolated") or 3
+        (for "NRCC Hi-Resolution").  See [2]_ for details.
+    map_variables : bool, default True
+        When True, rename data columns and metadata keys to pvlib variable
+        names where applicable. See variable :const:`VARIABLE_MAP`.
+    url : str, default: 'https://data.rcc-acis.org/GridData'
+        API endpoint URL
+    kwargs:
+        Optional parameters passed to ``requests.get``.
```
AdamRJensen, line 178:
```suggestion
        Optional parameters passed to ``requests.post``.
```

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+        First day of the requested period
+    end : datetime-like
+        Last day of the requested period
+    map_variables : bool, default True
+        When True, rename data columns and metadata keys to pvlib variable
+        names where applicable. See variable :const:`VARIABLE_MAP`.
+    url : str, default: 'https://data.rcc-acis.org/GridData'
+        API endpoint URL
+    kwargs:
+        Optional parameters passed to ``requests.get``.
```
AdamRJensen, line 254:
```suggestion
        Optional parameters passed to ``requests.post``.
```

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+                metadata[VARIABLE_MAP[key]] = metadata.pop(key)
+
+    return df, metadata
+
+
+def get_acis_prism(latitude, longitude, start, end, map_variables=True,
+                   url="https://data.rcc-acis.org/GridData", **kwargs):
+    """
+    Retrieve estimated daily precipitation and temperature data from PRISM
+    via the Applied Climate Information System (ACIS).
```
AdamRJensen, line 78:
```suggestion
    Retrieve estimated daily precipitation and temperature data from PRISM
    [1]_ via the Applied Climate Information System (ACIS) [2]_, [3]_.
```

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+
+    Geographical coverage: approximately -130° to -65° in longitude and
+    0° to 50° in latitude.
```
AdamRJensen, line 81:
```suggestion
    Geographical coverage: approximately 0° to 50° in latitude and
    -130° to -65° in longitude.
```

To most folks, it might be nice to state it a bit more explicit:

    Geographical coverage: US, Central America, and part of South America.
    Approximately 0° to 50° in latitude and -130° to -65° in longitude.

Same comment below.

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,483 @@
...
+
+    References
+    ----------
+    .. [1] `PRISM <https://prism.oregonstate.edu/>`_
+    .. [2] `ACIS Gridded Data <http://www.rcc-acis.org/docs_gridded.html>`_
+    .. [3] `ACIS Web Services <http://www.rcc-acis.org/docs_webservices.html>`_
+
+    Examples
+    --------
+    >>> df, meta = get_acis_prism(40, -80, '2020-01-01', '2020-12-31')
```
AdamRJensen, line 127:
```suggestion
    >>> df, meta = pvlib.iotools.get_acis_prism(
    >>>     latitude=40, longitude=-80, start='2020-01-01', end='2020-12-31')
```
The users who find examples useful might easily not be aware that the function is in the iotools module. In any case, I think it's nice that it works out of the box without special imports.

AdamRJensen (commented):
I'll play more around with the functions in the next couple of days.

My main concern is that it is a lot of functions (clutter) compared to the previous state of the PR which had it combined into one function. But it looks solid!

pvlib/iotools/acis.py
kandersolar: I was following #1420, does it not apply here?

AdamRJensen: Congratulations, you're paying attention! 😉It certainly does, as long as everyone agrees with the choice of altitude over elevation.

pvlib/iotools/acis.py
kandersolar: I lean towards keeping the `_average` suffix here.  Daily values seem different enough from the higher-resolution values that we call `temp_air` that consistency with the other names in this dict seems more important to me.  It doesn't make sense to feed these values into `ModelChain.run_model`, for example.

pvlib/iotools/acis.py
kandersolar: We generally avoid citing the references in the first sentence of the docstring (see #911), so I've added second sentences with a bit more info and these citations.

pvlib/iotools/acis.py
kandersolar: I think so.  The wording here is a shortened form of this text from https://prism.oregonstate.edu/explorer/:

>  Note: Rather than calculating days based on the 24-hour period ending at midnight local time, PRISM defines a "day" as the 24 hours ending at 12:00 Greenwich Mean Time (GMT, or 7:00am Eastern Standard Time). This means that PRISM data for May 26, for example, actually refers to the 24 hours ending at 7:00am EST on May 26.

kandersolar (commented):
Thanks @AdamRJensen!

> My main concern is that it is a lot of functions (clutter) compared to the previous state of the PR which had it combined into one function.

Yes, it does add a bit of clutter to the API listing.  But there are enough differences between the datasets that I thought splitting it up would make it easier to use and more straightforward to maintain, which seemed more important. 

Any thoughts on this question from https://github.com/pvlib/pvlib-python/pull/1767#issuecomment-1594990009?

> A discussion point: currently, get_acis_station_data returns precipitation in mm but snow in cm, for consistency with relevant PV models (pvlib.soiling takes rainfall in mm; pvlib.snow takes snowfall in cm). Is that the right choice?

pvlib/tests/iotools/test_acis.py
```diff
@@ -0,0 +1,206 @@
...
+def test_get_acis_prism():
+    # map_variables=True
+    df, meta = get_acis_prism(40.001, -80.001, '2020-01-01', '2020-01-02')
+    expected = pd.DataFrame(
+        [
+            [0.5, 5, 0, 2.5, 0, 62, 0],
+            [0, 5, -3, 1, 0, 64, 0]
+        ],
+        columns=['precipitation', 'temp_air_max', 'temp_air_min',
+                 'temp_air_average', 'cdd', 'hdd', 'gdd'],
```
wholmgren, line 27: mild preference for mapping `cdd` to `cooling_degree_days` and similar.

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,510 @@
...
+        Value to replace "trace" values in the precipitation data
+    url : str, default: 'https://data.rcc-acis.org/GridData'
+        API endpoint URL
+    kwargs:
+        Optional parameters passed to ``requests.post``.
+
+    Returns
+    -------
+    data : pandas.DataFrame
+        Daily precipitation [mm], temperature [Celsius], and snow [cm] data
```
wholmgren, line 358: DataFrame now contains more than this, right?

pvlib/iotools/acis.py
```diff
@@ -0,0 +1,510 @@
...
+        names where applicable. See variable :const:`VARIABLE_MAP`.
+    url : str, default: 'https://data.rcc-acis.org/GridData'
+        API endpoint URL
+    kwargs:
+        Optional parameters passed to ``requests.post``.
+
+    Returns
+    -------
+    data : pandas.DataFrame
+        Daily precipitation [mm] and temperature [Celsius] data
```
wholmgren, line 196: DataFrame now contains cooling, heating, and growing days, right?

wholmgren (approved):
thanks @kandersolar - looking forward to using this!

As for snow units, I think I'd rather keep them in mm here and make other functions consistent with that. But ask me again tomorrow and I might have a different opinion.

AdamRJensen:
> As for snow units, I think I'd rather keep them in mm here and make other functions consistent with that. But ask me again tomorrow and I might have a different opinion.

I think I agree with Will - we should avoid having too many different units.

kandersolar:
See a bit of discussion in #1792 about snow units.  Keeping them as cm seems like the best option to me, at least for now.  I think that was the last open item here, so in it goes!

PR #1781, opened 2023-06-20
Add mismatch.py and function to combine curves

ajonesr:
 - [ ] Tests added

Adds a function combine_curves that combines curves (in series) and a function that prepares inputs for combine_curves. This is part of the work to move some of the functionality of PVMismatch over to pvlib.

### DIFF
--- a/pvlib/ivtools/mismatch.py
+++ b/pvlib/ivtools/mismatch.py
@@ -1,0 +1,76 @@
+import pvlib
+import numpy as np
+
+
+def prepare_curves(params, num_pts):
+    # params is an n x 5 array where each row corresponds to one of
+    # the n curves' five defining parameters (photocurrent, saturation
+    # current, series resistance, shunt resistance, nNsVth)
+    # num_pts is the number of points you want calculated for each curve
+    # (this will also be the number of points in the aggregate curve)
+
+    # in case params is a list containing scalars, add a dimension
+    if len(np.shape(params)) == 1:
+        params = params[np.newaxis,:]
+
+    # get range of currents from 0 to max_isc
+    max_isc = np.max(pvlib.singlediode.bishop88_i_from_v(0.0, *params.T))
+    currents = np.linspace(0, max_isc, num=num_pts, endpoint=True)
+
+    # prepare inputs for bishop88
+    bishop_inputs = np.array([[currents[idx]]*len(params) for idx in
+                    range(num_pts)])
+    # each row of bishop_inputs contains n copies of a single current
+    # value, where n is the number of curves being added together
+    # there is a row for each current value
+
+    # get voltages for each curve
+    # transpose result so each row contains voltages for a single curve
+    voltages = pvlib.singlediode.bishop88_v_from_i(bishop_inputs, *params.T).T
+
+    return currents, voltages
+
+
+def combine_curves(currents, voltages, breakdown_voltage=-0.5):
+    # currents is a 1D array that contains a range from 0 to max_isc
+    # voltages is a 2D array where each row corresponds to the
+    # associated voltages of a single curve (should be n by
+    # len(currents), where n is the number of curves we're summing over)
+    # breakdown_voltage is the asymptote we use left of the y-axis
+
+    currents = np.asarray(currents)
+    voltages = np.asarray(voltages)
+    assert currents.ndim == 1
+    assert voltages.ndim == 2
+
+    # any voltages in array that are smaller than breakdown_voltage are
+    # clipped to be breakdown_voltage
+    clipped_voltages = np.clip(voltages, a_min=breakdown_voltage, a_max=None)
+
+    # for each current, add the associated voltages of all the curves
+    # in our setup, this means summing each column of the voltage array
+    combined_voltages = np.sum(clipped_voltages, axis=0)
+
+    # combined_voltages should now have same shape as currents
+    assert np.shape(combined_voltages) == np.shape(currents)
+
+    # find max power point (in the dataset)
+    powers = currents*combined_voltages
+    mpp_idx = np.argmax(powers)
+    vmp = combined_voltages[mpp_idx]
+    imp = currents[mpp_idx]
+    pmp = powers[mpp_idx]
+
+    # get isc
+    # np.interp requires second argument is increasing
+    assert np.all(np.diff(combined_voltages) < 0)
+    # FIXME not sure if I can guarantee that combined_voltages is
+    # decreasing for all inputs
+    isc = np.interp(0., np.flip(combined_voltages), np.flip(currents))
+
+    # get voc
+    voc = combined_voltages[0]
+
+    return {'i_sc': isc, 'v_oc': voc, 'i_mp': imp, 'v_mp': vmp, 'p_mp': pmp,
+            'i': currents, 'v': combined_voltages}
+


pvlib/ivtools/mismatch.py
```diff
@@ -0,0 +1,188 @@
...
+
+import numpy as np
+from pvlib.singlediode import bishop88_i_from_v, bishop88_v_from_i
+
+
+def prepare_curves(params, num_pts, breakdown_voltage=-0.5):
+    """
+    Calculates currents and voltages on IV curves with the given
+    parameters, using the single diode equation. Returns values
+    in format needed for inputs to :func:`combine_curves`.
```
cwhanse, line 14:
```suggestion
    Calculates currents and voltages on IV curves with the given
    parameters, using the single diode equation, and a simple
    model for reverse bias behavior.
    
    The current values are linearly spaced from 0 to the maximum Isc for all curves.
    All curves have the same current values.
    
    Returns values in format needed for inputs to :func:`combine_curves`.
```

pvlib/ivtools/mismatch.py
```diff
@@ -0,0 +1,188 @@
...
+            A 2D array of voltage values, where each row corresponds to
+            a single IV curve. Has shape (:math:`n`, ``num_pts``), where
+            :math:`n` is the number of IV curves passed in.
+
+    Notes
+    -----
+    This function assumes a simplified reverse bias model. When using
+    :func:`pvlib.singlediode.bishop88_v_from_i`, ``breakdown_factor`` is
+    left at the default value, which excludes the reverse bias term from
+    the model. Instead, any returned voltages that are less than
```
cwhanse, line 64:
```suggestion
    the calculation. Instead, any returned voltages that are less than
```

pvlib/ivtools/mismatch.py
```diff
@@ -0,0 +1,188 @@
...
+def combine_curves(currents, voltages):
+    """
+    Combines IV curves in series.
+
+    Parameters
+    ----------
+    currents : array-like
+        A 1D array-like object. Its first element must be zero, and it
+        should be increasing.
+
```
cwhanse, line 109: Could we perhaps reverse this array order? Its more natural to order current from Isc to 0, I think.

ajonesr: Sure thing. Just pushed this change.

pvlib/ivtools/mismatch.py
```diff
@@ -0,0 +1,193 @@
...
+            :math:`n` is the number of IV curves passed in.
+
+    Notes
+    -----
+    This function assumes a simplified reverse bias model. When using
+    :func:`pvlib.singlediode.bishop88_v_from_i`, ``breakdown_factor`` is
+    left at the default value, which excludes the reverse bias term from
+    the calculation. Instead, any returned voltages that are less than
+    ``breakdown_voltage`` are replaced by it, yielding a vertical line
+    at ``breakdown_voltage``.
```
cwhanse, line 72:
```suggestion
    This function assumes a simplified reverse bias model. IV curves are
    solved using :func:`pvlib.singlediode.bishop88_v_from_i` with
    ``breakdown_factor`` left at the default value of 0., which excludes
    the reverse bias term from the calculation. Here, voltages that are
    less than ``breakdown_voltage`` are set equal to ``breakdown_voltage``
    effectively modeling the reverse bias curve by a vertical line
    at ``breakdown_voltage``.
```

pvlib/ivtools/mismatch.py
```diff
@@ -0,0 +1,193 @@
...
+def prepare_curves(params, num_pts, breakdown_voltage=-0.5):
+    """
+    Calculates currents and voltages on IV curves with the given
+    parameters, using the single diode equation, and a simple
+    model for reverse bias behavior.
+
+    The current values are linearly spaced from the maximum Isc for all
+    curves to 0. All curves have the same current values.
+
+    Returns values in format needed for inputs to
```
cwhanse, line 19:
```suggestion
    Returns currents and voltages in the format needed for input to
```

pvlib/ivtools/mismatch.py
```diff
@@ -0,0 +1,193 @@
...
+    :func:`combine_curves`.
+
+    Parameters
+    ----------
+    params : array-like
+        An array of parameters representing a set of :math:`n` IV
+        curves. The array should contain :math:`n` rows and five
+        columns. Each row contains the five parameters needed for a
+        single curve. The parameters should be in the following order:
```
cwhanse, line 28:
```suggestion
        single curve in the following order:
```

pvlib/tests/ivtools/test_mismatch.py
```diff
@@ -0,0 +1,125 @@
...
+            np.array([1, 0.9, 0.]),
+            np.array([[0., 0.5, 1], [0, 0.5, 1]]),
+            {'i_sc': 1, 'v_oc': 2, 'i_mp': 0.9, 'v_mp': 1, 'p_mp': 0.9,
+             'i': np.array([1, 0.9, 0.]),
+             'v': np.array([0, 1, 2])
+            }
+        )
+    ])
+
+
```
cwhanse, line 54:
```suggestion
```

pvlib/tests/ivtools/test_mismatch.py
```diff
@@ -0,0 +1,125 @@
...
+            [1, 3e-08, 1., 300, 1.868364353685363],
+            (
+                np.array([0.99667772, 0.74750829, 0.49833886, 0.24916943, 0.]),
+                np.array([2.42028619e-14, 2.81472975e+01, 3.01512748e+01,
+                          3.12974337e+01, 3.21521313e+01])
+            )
+        )
+    ])
+
+
```
cwhanse, line 115:
```suggestion
```

cwhanse (commented):
Just need to add two tests that raise the ValueErrors.

AdamRJensen:
Would it be possible to add a simple example to the documentation?

ajonesr:
@AdamRJensen I've added an example in docs/examples/iv-modeling, though I'm not sure that's the right place to put it. Let me know if I should move it somewhere else

AdamRJensen:
> @AdamRJensen I've added an example in docs/examples/iv-modeling, though I'm not sure that's the right place to put it. Let me know if I should move it somewhere else

Thanks a lot @ajonesr !
Could you add a legend for the individual lines, e.g., just panel 1, panel 2, and panel 3. Also, if you could add just a tiny bit more text in the example about what is going on and why it is useful :)

ajonesr:
> > @AdamRJensen I've added an example in docs/examples/iv-modeling, though I'm not sure that's the right place to put it. Let me know if I should move it somewhere else
> 
> Thanks a lot @ajonesr ! Could you add a legend for the individual lines, e.g., just panel 1, panel 2, and panel 3. Also, if you could add just a tiny bit more text in the example about what is going on and why it is useful :)
> 
> You also need to move the whatsnew entry from v0.10.0 to v0.10.2.

Sure thing! Just pushed those changes.

kandersolar:
I can do a review (sorry for the wait!) but I'm a little hesitant about the current approach this PR takes.  To me, this code feels more like an application than it does reusable functions.  For example, `combine_curves` is pretty specific in its requirements, which makes it suitable for use with `prepare_curves`, but difficult to use if the curves come from anywhere else.  

I wonder if these functions would make more sense as a gallery example page than functions in pvlib itself?

cwhanse:
@kandersolar I'm interested to hear your thoughts on different ways to organize this capability.

I don't think a gallery example is enough here. We've heard several requests for capabilities native to pvlib to model mismatch. It is clear that pvmismatch, while useful, has some limits and is not going to be further developed. It seems reasonable to include functions in pvlib that one could use to get beyond the limits of pvmismatch.

 Curves are added by aligning current and summing the corresponding voltage. The functions in this PR meet one use case: I have diode equation parameters for the series-connected modules/cells and I want the string-level IV curve.I can see a second use case: I have non-aligned IV curves (measured perhaps) and want the string-level curve. These functions don't meet the measured-curve use case.

I think separate function(s) that align the currents are useful, so that the function which adds curves is more reusable. We would want to align currents starting either from diode equation values (`prepare_curves` does this now) and from curve data itself (by interpolation, likely with extrapolation to negative voltages). 

Maybe `combine_curves` should be `add_curves`. That would free the more general `combine_curves` name for the measured-curve use case, `combine_curves(current_array, voltage_array)`. In that context, `combine_curves` would handle interpolation to align currents, then call `add_curves`.

kandersolar:
I think my hesitation was centered around not being able to use the addition function with the existing methods for generating IV curves.  It felt somewhat disconnected from the rest of pvlib.  A function to align curves on one of the two dimensions would address that.  

The tight coupling of `prepare_curves` to the SDE makes me wonder if it should live in `pvlib.singlediode` as a new `bishop88_*` function.  

Series vs parallel connection is something else to consider when choosing names to leave room for future functionality.

adriesse:
> I don't think a gallery example is enough here.

It could be a start though...

PR #1786, opened 2023-06-26
Python Flake8 linter workflow to replace stickler-ci

### LINKED ISSUE #1722: stickler-ci is shutting down

reepoi:
- [x] Closes #1722 
 - ~[ ] Tests added~

Adds GitHub workflow to replace stickler-ci for Python code linting with flake8. The new workflow annotates the PR changes using the errors outputted by flake8. Here is an example showing what the annotations look like: https://github.com/reepoi/pvlib-python/pull/1/files

The workflow implementation is based on the GitHub action here: https://github.com/pr-annotators/flake8-pr-annotator.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -19,7 +19,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
-        run: echo CHANGED_FILENAMES="$(gh pr view --repo reepoi/pvlib-python ${{ env.PR_NUMBER }} --json files -q '.files[].path')"
+        run: echo CHANGED_FILENAMES="$(gh pr view --repo reepoi/pvlib-python ${{ env.PR_NUMBER }} --json files -q '.files[].path' | xargs)"
              >> ${{ github.env }}
       - name: Run Flake8 linter
         run: flake8
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -1,6 +1,5 @@
 name: Python Flake8 Linter
 on:
-  push:
   pull_request:
 jobs:
   build:
@@ -17,7 +16,11 @@ jobs:
       - name: Setup Flake8 output matcher for PR annotations
         run: echo '::add-matcher::.github/workflows/flake8-pr-annotation-matcher.json'
       - name: Run Flake8 linter
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
         run: flake8
+               --filename $(gh pr view --repo reepoi/pvlib-python ${{ env.PR_NUMBER }} --json files -q '[.files[].path] | @csv')
                --exclude pvlib/version.py
                --ignore E201,E241,E226,W503,W504
                --max-line-length 79
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -2,7 +2,7 @@ name: Python Flake8 Linter
 on:
   pull_request:
 jobs:
-  build:
+  lint:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout source
@@ -23,7 +23,7 @@ jobs:
              >> ${{ github.env }}
       - name: Run Flake8 linter
         run: flake8
-               --filename ${{ env.CHANGED_FILENAMES }}
+               ${{ env.CHANGED_FILENAMES }}
                --exclude pvlib/version.py
                --ignore E201,E241,E226,W503,W504
                --max-line-length 79
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -1,0 +1,24 @@
+name: Python Flake8 Linter
+on:
+  push:
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+      - name: Install Python 3.11
+        uses: actions/setup-python@v4
+        with:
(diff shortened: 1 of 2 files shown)


reepoi:
Stickler CI appears to only lint the functions changed inside a file, but this workflow lints the entire file. I am working on changing it to only add annotations for changed functions.

echedey-ls:
Personally, I don't like excluding errors or warnings for all files, especially [`W504`](https://www.flake8rules.com/rules/W504.html). 

For reference, these are currently excluded in this PR:
* [`E201`](https://www.flake8rules.com/rules/E201.html), Whitespace after '(' 
* [`E226`](https://www.flake8rules.com/rules/E226.html), Missing whitespace around arithmetic operator 
* [`E241`](https://www.flake8rules.com/rules/E241.html), Multiple spaces after ',' 
* [`W503`](https://www.flake8rules.com/rules/W503.html), Line break occurred before a binary operator 
* [`W504`](https://www.flake8rules.com/rules/W504.html), Line break occurred after a binary operator 

flake8 does not provide a way to disable errors for code blocks, but per file with a configuration file (or per line as we know).
I'm not suggesting changing current behaviour, I only want to point that out.

kandersolar (commented):
This PR looks great!  I didn't know (or maybe forgot) that jobs could create annotations like this.  Very cool.  Seems to work as intended.

One thing to note is that this job flags all errors in the changed files, whereas I think (although my memory is a bit hazy) our former setup with stickler-ci wouldn't flag unchanged lines.  The old behavior had pros and cons: it wouldn't pester the contributor with errors unrelated to their PR, but it also wouldn't necessarily catch things like unused imports.  In our case, flagging all errors means making a tiny change to some files will result in dozens of errors.  See for example: https://github.com/pvlib/pvlib-python/actions/runs/5476853711/jobs/9975004553?pr=1786

Is there some way to not report those unrelated errors?  Unless we go through and fix all of those (which I'm not sure is a good idea), I think they will make this CI job a lot less useful for many PRs.

Something else is that there is a maximum of 10 annotations.  Not really a big deal.  You can always look at the job text output to see the full list of lint errors.

> For reference, these are currently excluded in this PR

I think @reepoi just copied those exclusions over from our current lint configuration in `.stickler.yml`.  We can certainly talk about changing those rules, but let's focus on just getting the linter running again in this PR.

wholmgren:
do we want to use an older version of flake8 that supports the deprecated `--diff` option?

reepoi:
Using an older version of flake8 that supports the `--diff` option is a good idea, and I changed the workflow to do that. If we didn't, we'd need to find the changed Python files and filter their linting errors by line number. The easiest I found to get the line numbers of changes in a file is by parsing the output of `git diff`, which is what the older versions of flake8 do.

As a bonus, flake8's `--diff` option appears to ignore changes in files that are not Python files automatically. The workflow's code to do that is removed now.

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Install Flake8 5.0.4 linter
+        run: pip install flake8==5.0.4  # use this version for --diff option
+      - name: Setup Flake8 output matcher for PR annotations
+        run: echo '::add-matcher::.github/workflows/flake8-linter-matcher.json'
+      - name: Fetch pvlib main branch
+        run: |
+          git remote add upstream https://github.com/pvlib/pvlib-python.git
+          git fetch upstream main
+      - name: Run Flake8 linter
+        run: git diff upstream/main HEAD | flake8
```
kandersolar, line 23:
```suggestion
        run: git diff upstream/$GITHUB_BASE_REF HEAD | flake8
```

Should we do this so that the diff is always taken with respect to the branch the PR is opened against?  We rarely (never?) open PRs to any branch except `main`, so maybe doesn't matter, but I could see us using release branches and such at some point in the future.

Docs: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

kandersolar (approved):
LGTM!  Does anyone else have any comments before we merge this?  

@reepoi can you make a short 0.10.2 whatsnew entry and list yourself as a contributor?  And remove the demonstration lint error of course :)

kandersolar:
I think whitespace is often but not always an improvement.  For example, I prefer no whitespace around the `**` operators here:

https://github.com/pvlib/pvlib-python/blob/6606af7775ac18ee1334ff3b1ad5589fef1f4e04/pvlib/temperature.py#L752

I lean towards continuing to ignore that linter rule, but I don't feel strongly about it.

echedey-ls:
@kandersolar @cwhanse  Idk if Kevin's point is only for exponentiation operator, but E226 [does not raise warnings for operator `**`](https://www.flake8rules.com/rules/E226.html). I've also tested it to be sure with latest flake8.

Just wanted to point that out, with reviewers doing their job there shouldn't be any unreadable code pushed.

kandersolar:
Thanks @echedey-ls, that's good to know!  The `**` operator was the first that came to mind, but I expect there are other cases where we'd like to not be hassled about operator whitespace too. 

Anyway I'm going to go ahead and merge this so that we have an operational linter again.  Thanks @reepoi for setting this up and everyone else for reviewing!

PR #691, opened 2019-04-23
DOC: fix napoleon, render sphinx docs right

mikofski:
pvlib python pull request guidelines
====================================

Thank you for your contribution to pvlib python! You may delete all of these instructions except for the list below.

You may submit a pull request with your code at any stage of completion.

The following items must be addressed before the code can be merged. Please don't hesitate to ask for help if you're unsure of how to accomplish any of the items below:

 - [x] Closes issue #690 
 - [x] I am familiar with the [contributing guidelines](http://pvlib-python.readthedocs.io/en/latest/contributing.html).
 - [x] Fully tested. Added and/or modified tests to ensure correct behavior for all reasonable inputs. Tests (usually) must pass on the TravisCI and Appveyor testing services.
 - [ ] Updates entries to `docs/sphinx/source/api.rst` for API changes.
 - [ ] Adds description and name entries in the appropriate `docs/sphinx/source/whatsnew` file for all changes.
 - [x] Code quality and style is sufficient. Passes LGTM and SticklerCI checks.
 - [x] New code is fully documented. Includes sphinx/numpydoc compliant docstrings and comments in the code where necessary.
 - [x] Pull request is nearly complete and ready for detailed review.

Brief description of the problem and proposed solution (if not already fully described in the issue linked to above):

* add napoleon extension to sphinx source `conf.py`
* fix see also in `solar_zenith_analytical`

PR #1790, opened 2023-06-28
Refactor docstring params

### LINKED ISSUE #1574: "default None" in docstring parameter descriptions considered harmful
A common python idiom for situations where a function's parameter can be left unspecified is to set its default value to `None`.  In this usage `None` is understood not as a valid value for that parameter but rather as an indication to the function's body that it should treat the parameter as having taken no value.  We could have chosen any python value for this task; we use `None` because it is convenient and idiomatic.  One can view this pattern as a workaround to python's requirement that all parameters be specified either explicitly by the caller or implicitly by its default value.  Take Javascript as a point of comparison, where unspecified parameters are automatically set to `undefined` without the function signature having to specify that default value.

As such, I don't think there is a good reason to say `default None` in docstring parameter descriptions in these situations; `None` is more of an implementation detail than a meaningful default value that the reader would be interested in.  For example, consider `get_total_irradiance`'s `airmass` parameter:
https://github.com/pvlib/pvlib-python/blob/bdbaf4c238ce2b373d04af48193b14659662cef1/pvlib/irradiance.py#L344

Yes it's true that this parameter's default value is `None`, but I don't think that information is useful to the reader.  Either `airmass` is required and they need to pass a `numeric`, or it isn't required and they don't need to touch that parameter.  There is no situation where the user needs to say `airmass=None` in their own code.  I think it would be better to use this style:

```
     airmass : numeric, optional
```
The useful information is still there, but without the distraction of `None`.  It also renders fine in sphinx docs.

`default None` is fairly widespread in pvlib: `$ git grep -n -i -F "default None" | wc -l` returns 81.  I propose we not use `default None` in docstrings in the future (except of course in the rare case when `None` is a legitimate value) and eventually edit existing docstrings to say `optional` instead.

P.S. for anyone who doesn't get the title of this issue: https://en.wikipedia.org/wiki/Considered_harmful

echedey-ls:
- [x] Closes #1574 among other updates. See bottom of PR body.
 - ~[ ] Tests added~
 - [x] Current code includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments everywhere.

I just want the docstrings to be homogeneus and compliant with numpydoc.

## Changes in this PR
1. Adds a space before colon in parameters and returns
  `solar_azimuth: numeric` -> `solar_azimuth : numeric`
1. Changes excessive use of parentheses in defaults
  `return_components : bool (optional, default=False)` -> `return_components : bool, default False`
1. Removes colon when default is to be specified
  `logical_records : list or tuple, default: ('0100',)` -> `logical_records : list or tuple, default ('0100',)`
1. default None -> optional (#1574)
2. `See also` -> `See Also`

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.10.0.rst
+++ b/docs/sphinx/source/whatsnew/v0.10.0.rst
@@ -29,6 +29,11 @@ Deprecations
 
 Enhancements
 ~~~~~~~~~~~~
+* The return values of :py:func:`pvlib.pvsystem.calcparams_desoto`,
+  :py:func:`pvlib.pvsystem.calcparams_cec`, and
+  :py:func:`pvlib.pvsystem.calcparams_pvsyst` are all numeric types and have
+  the same Python type as the `effective_irradiance` and `temp_cell` parameters. (:issue:`1626`, :pull:`1700`)
+
 * Added `map_variables` parameter to :py:func:`pvlib.iotools.read_srml`
   and :py:func:`pvlib.iotools.read_srml_month_from_solardat` (:pull:`1773`)
 * Allow passing keyword arguments to :py:func:`scipy:scipy.optimize.brentq` and
--- a/pvlib/atmosphere.py
+++ b/pvlib/atmosphere.py
@@ -361,7 +361,7 @@ def bird_hulstrom80_aod_bb(aod380, aod500):
     aod_bb : numeric
         Broadband AOD.  [unitless]
 
-    See also
+    See Also
     --------
     pvlib.atmosphere.kasten96_lt
 
@@ -407,7 +407,7 @@ def kasten96_lt(airmass_absolute, precipitable_water, aod_bb):
     lt : numeric
         Linke turbidity. [unitless]
 
-    See also
+    See Also
     --------
     pvlib.atmosphere.bird_hulstrom80_aod_bb
     pvlib.atmosphere.angstrom_aod_at_lambda
@@ -488,7 +488,7 @@ def angstrom_aod_at_lambda(aod0, lambda0, alpha=1.14, lambda1=700.0):
     aod1 : numeric
         AOD at desired wavelength ``lambda1``. [unitless]
 
-    See also
+    See Also
     --------
     pvlib.atmosphere.angstrom_alpha
 
@@ -528,7 +528,7 @@ def angstrom_alpha(aod1, lambda1, aod2, lambda2):
         Angstrom :math:`\alpha` exponent for wavelength in
         ``(lambda1, lambda2)``. [unitless]
 
-    See also
+    See Also
     --------
     pvlib.atmosphere.angstrom_aod_at_lambda
     """
--- a/pvlib/bifacial/infinite_sheds.py
+++ b/pvlib/bifacial/infinite_sheds.py
@@ -284,7 +284,7 @@ def get_irradiance_poa(surface_tilt, surface_azimuth, solar_zenith,
        Photovoltaic Specialists Conference (PVSC), 2019, pp. 1282-1287.
        :doi:`10.1109/PVSC40753.2019.8980572`.
 
-    See also
+    See Also
     --------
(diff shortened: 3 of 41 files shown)


kandersolar:
Thanks for submitting this PR @echedey-ls!

I think the main priority for docstring formatting is that the sphinx HTML output is correctly rendered.  Since spaces before colons and `See Also` versus `See also` don't make any difference in the HTML, I don't think those are really needed.  The cost of changing them is that it clutters the `git blame`, which is only a minor problem, but still enough to want to avoid making these kinds of sweeping changes without a good reason for it IMHO.  

For changing `default None` to `optional`, I think that is good.  But probably some of the parameter descriptions should be updated as well.  For example in `pvlib.clearsky.detect_clearsky` it now says:
```
    times : DatetimeIndex, optional
        Times of measured and clearsky values. If None the index of measured
        will be used.
```
Since `None` is no longer mentioned as the default value, I think the description needs to change to `If not specified, ...` instead.  There are other cases like this too (e.g. `surface_type` in `pvlib.irradiance.get_ground_diffuse` and many parameters in `pvlib.modelchain.basic_chain`).

echedey-ls:
Good points. I forgot to add an explanation to what I really want to address since I've been travelling a bit and didn't bother to check what I had written. 

I intend to open a PR to hint contributors of docstring errors. I've come up to two options that may be useful. Both may be used at the same time:
1. suggest running the command `python -m numpydoc --validate pvlib.module.func` in the PR template and/or the contributing guidelines, so it can be tested quickly. From numpydoc validation, [this section](https://numpydoc.readthedocs.io/en/latest/validation.html#docstring-validation-using-python).
2. in the sphinx-build job (well, in the `conf.py` file). See this [section](https://numpydoc.readthedocs.io/en/latest/validation.html#docstring-validation-during-sphinx-build) of numpydoc.

Also, they recently added a pre-commit hook, although I feel that is a bit overwhelming to add to pvlib since it is currently not being used, I believe.

Anyway, for choice 1, it doesn't mind having these "issues" in the docstrings, since contributor would only apply the checker to their function/class/method... The problem with that is that I haven't found a way to include/exclude rules (it checks all of them) and that's definitely not desired. Some rules just don't apply to the pvlib style, like mandatory examples section, some punctuation or capitalisation, but users may not know of that. Adding a `grep` pipe to the output might be the hardcoded way to do so but it may be OS dependent.
Some work on this may be found at [numpy/numpydoc#459](https://github.com/numpy/numpydoc/issues/459)

On the other hand, choice 2 allows to check for a custom subset of rules, so there is a cleaner output. However, it checks all sphinx docstrings. This PR would allow using the most of them that seem useful. For example, the "See Also" change allows using the rule that checks that all sections exists (it raises a warning for all of them). Adding a space before a colon and removing parentheses makes the default validator parse the argument name and the type correctly. To sum it up, most of the changes in this PR remove false positives and typos that mess with the validator, as a previous step. Personally, I like this option the most.

And these superfluous changes, well, since they almost don't have an impact on the docs, I thought of doing them via an independent PR. So checking another PR with the tool is not a pain (I'm aware this one is).

I will check the `default None` and clean those docstrings in the following days.

kandersolar:
Thanks for this in-depth research @echedey-ls :)

One thing relevant especially for option 2 is that we don't actually use numpydoc, we use the [napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) sphinx extension (#691).  I'm guessing that enabling both of them in the build would cause problems, but I don't know that for sure.

echedey-ls:
>  I'm guessing that enabling both of them in the build would cause problems, but I don't know that for sure.

Good point, it does.

I will pursue choice 1, but I'd like to wait for the numpydoc issue to get resolved then.

Thanks for the review. Which changes should I keep? Changes 2 to 4 in the "changes" section from PR body?

echedey-ls:
Continued work over #1828

PR #1797, opened 2023-07-01
Replace `pvfactors` with `solarfactors`

### LINKED ISSUE #1796: latest version of pvfactors cannot be installed with pvlib 0.10.0
The last several versions of pvfactors require `pvlib<0.10` to preempt possible breaking changes.  In fact this turned out to be unnecessary; to the best of my knowledge, there is no actual incompatibility between pvfactors==1.5.2 (the latest) and pvlib==0.10.0.

However, that upper limit does prevent pip from installing `pvfactors==1.5.2` and `pvlib==0.10.0` simultaneously.  What it does instead is go back to `pvfactors==1.4.1`, the last version that doesn't require `pvlib<0.10`.  The problem is that this version of pvfactors is old enough to not contain a compatibility fix with the latest shapely, resulting in a broken installation that errors when you try to actually use pvfactors.  See for example this failed docs build: https://readthedocs.org/projects/pvlib-python/builds/21171307/

Users have two options to work around this problem in the immediate term.  Option 1: downgrade shapely to a version that works with `pvfactors==1.4.1`:

```bash
$ pip install pvlib==0.10.0 pvfactors  # installs pvfactors 1.4.1 and shapely 2.x
$ pip install 'shapely<2.0'  # might need to use double quotes in some terminals
```

Option 2: tell pip to ignore pvfactors when it says it isn't compatible with pvlib 0.10.0.  This will allow you to keep using pvfactors 1.5.2:
```bash
$ pip install pvfactors==1.5.2  # installs pvlib 0.9.5, and shapely 1.8.5
$ pip install --no-deps pvlib==0.10.0  # installs pvlib 0.10.0 alongside pvfactors 1.5.2
```

For a more permanent fix, the best solution would be for pvfactors to release a new version that doesn't have this unnecessary `pvlib<0.10` requirement.  But since that project is no longer maintained, we can't expect that to happen.

In the absence of that, I see two options, both of which require going to 0.10.1:
- A simple solution would be to add `shapely<2.0` to our `optional` requirements, effectively making Option 1 above automatic.  This has the downside of using an older version of pvfactors, which is missing a minor bugfix that a small minority of users might care about.
- Another option is to use this as motivation to get [solarfactors](https://github.com/pvlib/solarfactors) sufficiently up to speed such that we could switch pvlib to depend on it instead of pvfactors.  I will look into how much work that would take.

---

An aside:  although I did make a local `v0.10.0` tag and ensure that the tests passed before making the official 0.10.0 release, I neglected to install `pvlib[optional]` in the fresh test environment, so the pvfactors tests were skipped and this problem did not arise.  Had I been more thorough, this problem could have been found out before pvlib 0.10.0 was actually released.

kandersolar:
- [x] Closes #1796
 - ~[ ] Tests added~
 - [x] TODO: update `pvfactors_timeseries` docstring

I've got a first release of [solarfactors](https://github.com/pvlib/solarfactors) up on PyPI, with the important feature of not requiring `pvlib<0.10`.  Other than `pip install`ing a different name, it should be a transparent replacement for pvfactors as far as pvlib is concerned.  So hopefully just by changing the dependency name, the tests and docs build should pass without any other changes being needed.

If so, I think we could proceed with making a v0.10.1 release with this update.

### DIFF
--- a/ci/requirements-py3.10.yml
+++ b/ci/requirements-py3.10.yml
@@ -25,4 +25,4 @@ dependencies:
     - statsmodels
     - pip:
         - nrel-pysam>=2.0
-        - pvfactors==1.5.2
+        - solarfactors
--- a/ci/requirements-py3.11.yml
+++ b/ci/requirements-py3.11.yml
@@ -25,4 +25,4 @@ dependencies:
     - statsmodels
     - pip:
         - nrel-pysam>=2.0
-        - pvfactors==1.5.2
+        - solarfactors
--- a/ci/requirements-py3.7.yml
+++ b/ci/requirements-py3.7.yml
@@ -25,4 +25,4 @@ dependencies:
     - statsmodels
     - pip:
         - nrel-pysam>=2.0
-        - pvfactors==1.5.2
+        - solarfactors
--- a/ci/requirements-py3.8.yml
+++ b/ci/requirements-py3.8.yml
@@ -25,4 +25,4 @@ dependencies:
     - statsmodels
     - pip:
         - nrel-pysam>=2.0
-        - pvfactors==1.5.2
+        - solarfactors
--- a/ci/requirements-py3.9.yml
+++ b/ci/requirements-py3.9.yml
@@ -25,4 +25,4 @@ dependencies:
     - statsmodels
     - pip:
         - nrel-pysam>=2.0
-        - pvfactors==1.5.2
+        - solarfactors
--- a/docs/examples/bifacial/plot_bifi_model_mc.py
+++ b/docs/examples/bifacial/plot_bifi_model_mc.py
@@ -18,6 +18,13 @@ Example of bifacial modeling using pvfactors and ModelChain
 #
 # Future versions of pvlib may make it easier to do bifacial modeling
 # with ``ModelChain``.
+#
+# .. attention::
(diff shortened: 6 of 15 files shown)


kandersolar:
@pvlib/pvlib-core this is ready for review.  I suggest we merge this and release v0.10.1.

wholmgren (approved):
Works as expected!

```
$ conda create -n pvlib101 python=3.11
$ conda activate pvlib101
$ pip install "pvlib[optional] @ git+ssh://git@github.com/kandersolar/pvlib-python.git@solarfactors"
$ python
>>> import pvlib  # this takes ~10-20 seconds in a new environment!?
>>> pvlib.__version__  # probably need a git fetch --tags
'0.9.2a3.dev106+g8d38d65'
>>> import solarfactors
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'solarfactors'
>>> import pvfactors
>>> pvfactors.__version__
'1.5.3'
```

AdamRJensen (approved):
This is a bit complicated lingo: "pip install solarfactors installs a package that is still accessed in code with the original import pvfactors."

Also, I think most users would appreciate a more explicit reason for why we switched to pvfactors - i.e., state that pvfactors is not being maintained

PR #1821, opened 2023-08-01
Update `singlediode.bishop88` parameter types in docs

### LINKED ISSUE #1820: `breakdown_factor` type in `singlediode.bishop88` docs
**Describe the bug**
The docs of the `singlediode.bishop88*` functions say that `breakdown_factor` can be numeric, but these functions crash when `breakdown_factor` is an array of length two or more. This is caused by the if-statement [here](https://github.com/pvlib/pvlib-python/blob/cfd9fcc924686f0b9dbcc942c427568678deb2cd/pvlib/singlediode.py#L165).

**Expected behavior**
The functions can be modified to support the numeric type for `breakdown_factor`, or the docs can be updated to say that `breakdown_factor` must be a float.

**Versions:**
 - ``pvlib.__version__``: 0.10.1

reepoi:
- [x] Closes #1820
 - ~~[ ] Tests added~~
 - ~~[ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.~~

Update docs of `singlediode.bishop88*` functions to say that the parameters related to the bishop88 model ~~and PVSyst~~ are float types, not numeric.

I explored the option of modifying these functions to support a numeric `breakdown_factor` and found that it significantly slows down the common case when `breakdown_factor = 0`. Here is the Jupyter notebook I used to measure bishop88 performance: https://github.com/reepoi/pvlib-python/blob/bishop88-if/bishop88_speed.ipynb

The notebook compares three implementations of bishop88:
- bishop88
  - the original
- bishop88_no_if
  - always calls np.power (https://github.com/reepoi/pvlib-python/blob/bishop88-if/pvlib/singlediode.py#L59)
- bishop88_where
  - always calls np.power but use its `where` parameter to have numpy skip computing np.power when breakdown_factor is zero

There are two tests in the notebook:
- Breakdown_factor is a scalar
  - call the functions with 20 IV curves and a single scalar breakdown_factor for all of them
- Breakdown_factor is an array
  - call the functions with 20 IV curves and a breakdown_factor array with length 20. I try a breakdown_factor array with all zeros, half of its entries are zero, and all entries are nonzero. The original bishop88 implementation is excluded from this test because it doesn’t work when breakdown_factor is an array.


cwhanse:
+1 for the reverse bias parameters `breakdown_factor`, `breakdown_voltage` and `breakdown_exp` being floats. The common reason reason to compute the negative voltage portion of IV curves is to get to an IV curve for a series-connected devices, and it's hard to imagine having different diode behavior among the series-connected components.

 I'm not sure about making the recombination current term parameters `NsVbi` and `d2mutau` to floats. The code lets one use an array. My impression is that these parameters are seldom used, but when they are used, they are probably set to PVsyst's constant defaults. So I don't think it does any harm to expect `float`.

reepoi:
Since the code already works when `NsVbi` and `d2mutau` are arrays, they should probably be left as numeric.

docs/sphinx/source/whatsnew/v0.10.2.rst
```diff
@@ -31,6 +31,11 @@ Documentation
 ~~~~~~~~~~~~~
 * Removed Stickler-CI integration as the service has ceased June 2023.
   (:issue:`1722`, :pull:`1723`)
+* Restricted :py:func:`pvlib.singlediode.bishop88`,
+  :py:func:`pvlib.singlediode.bishop88_i_from_v`, and
+  :py:func:`pvlib.singlediode.bishop88_v_from_i` parameters `breakdown_factor`,
+  `breakdown_voltage`, and `breakdown_exp` to be floats.
+  (:issue:`1820`, :pull:`1821`)
```
cwhanse, line 38:
```suggestion
* Specified that :py:func:`pvlib.singlediode.bishop88`,
  :py:func:`pvlib.singlediode.bishop88_i_from_v`, and
  :py:func:`pvlib.singlediode.bishop88_v_from_i` parameters `breakdown_factor`,
  `breakdown_voltage`, and `breakdown_exp` should be floats.
  (:issue:`1820`, :pull:`1821`)
```

cwhanse (approved):
Thanks @reepoi

kandersolar (approved):
Thanks @reepoi!

PR #1827, opened 2023-08-03
Add functions to fit and convert IAM models

### LINKED ISSUE #1824: Functions to fit and convert IAM models
It would be useful to be able to convert parameters between IAM models (e.g., physical to martin_ruiz). To convert from an interpolation approach to a model described by an equation, it would also be useful to have a "fit" function.

Add `fit` and `convert` functions to `pvlib.iam.py`. `convert` could be implemented as

```
target_params = convert(source_model, source_params, target_model)
```

Add an example to the gallery showing how to use these functions, and demonstrating that a converted model yields similar effective insolation for a representative system.

An example showing how to do this. The example would contain the fit/convert code, which seems like a less-than-useful implementation.

ajonesr:
- [x] Closes #1824
 - [x] Tests added


pvlib/iam.py
```diff
@@ -940,3 +941,360 @@ def schlick_diffuse(surface_tilt):
...
+    # that case aoi is defined internally)
+    weight = weight * np.logical_and(aoi >= 0, aoi <= 90).astype(int)
+
+    diff = np.abs(source_iam - np.nan_to_num(target(aoi, *target_params)))
+    return np.sum(diff * weight)
+
+
+def _get_ashrae_int(b):
+    # find x-intercept of ashrae model
+    return np.rad2deg(np.arccos(b / (1 + b)))
```
kandersolar, line 997: `pvlib.tools.acosd`

pvlib/iam.py
```diff
@@ -940,3 +941,360 @@ def schlick_diffuse(surface_tilt):
...
+def fit(measured_aoi, measured_iam, target_name, options=None):
+    """
+    Finds parameters for target model that best fit the
+    measured data.
+
+    Parameters
+    ----------
+    measured_aoi : array-like
+        Angle of incidence values associated with the
+        measured IAM values.
```
kandersolar, line 1218:
```suggestion
        measured IAM values. [degrees]
```

pvlib/iam.py
```diff
@@ -940,3 +941,360 @@ def schlick_diffuse(surface_tilt):
...
+
+    measured_iam : array-like
+        IAM values.
```
kandersolar, line 1221:
```suggestion
        IAM values. [unitless]
```

docs/examples/reflections/plot_convert_iam_models.py
```diff
@@ -0,0 +1,190 @@
...
+Illustrates how to convert from one IAM model to a different model using
+pvlib.iam.convert, and how to fit an IAM model to data using
+pvlib.iam.fit
+
+"""
+
+# %%
+# An incidence angle modifier (IAM) model quantifies the fraction of direct
+# irradiance is that is reflected away from a module's surface. Three popular
+# IAM models are Martin-Ruiz py:func:`~pvlib.iam.martin_ruiz`, physical
```
kandersolar, line 15:
```suggestion
# IAM models are Martin-Ruiz :py:func:`~pvlib.iam.martin_ruiz`, physical
```

And the same elsewhere

docs/examples/reflections/plot_convert_iam_models.py
```diff
@@ -0,0 +1,190 @@
+
+"""
+IAM Model Conversion and Fitting
+================================
+
+Illustrates how to convert from one IAM model to a different model using
+pvlib.iam.convert, and how to fit an IAM model to data using
+pvlib.iam.fit
```
kandersolar, line 8: I think fitting is distinct enough from converting to warrant a separate example page

docs/examples/reflections/plot_convert_iam_models.py
```diff
@@ -0,0 +1,190 @@
...
+# Compute IAM values using the martin_ruiz model.
+aoi = np.linspace(0, 90, 100)
+martin_ruiz_params = {'a_r': 0.16}
+martin_ruiz_iam = martin_ruiz(aoi, **martin_ruiz_params)
+
+# Get parameters for the physical model and compute IAM using these parameters.
+physical_params = convert('martin_ruiz', martin_ruiz_params, 'physical')
+physical_iam = physical(aoi, **physical_params)
+
+# Get parameters for the ASHRAE model and compute IAm using these parameters.
```
kandersolar, line 51:
```suggestion
# Get parameters for the ASHRAE model and compute IAM using these parameters.
```

docs/examples/reflections/plot_convert_iam_models.py
```diff
@@ -0,0 +1,190 @@
...
+plt.show()
+
+
+# %%
+# The weight function
+# -------------------
+# Both :py:func:`pvlib.iam.convert` and :py:func:`pvlib.iam.fit` use
+# a weight function when computing residuals between the two models, or
+# between a model and data. The default weight
+# function is $1 - \sin(aoi)$. We can instead pass a custom weight function
```
kandersolar, line 111:
```suggestion
# function is :math:`1 - \sin(aoi)`. We can instead pass a custom weight function
```

pvlib/iam.py
```diff
@@ -940,3 +941,360 @@ def schlick_diffuse(surface_tilt):
...
+    weight = weight_function(aoi, **weight_args)
+
+    # if aoi contains values outside of interval (0, 90), annihilate
+    # the associated weights (we don't want IAM values from AOI outside
+    # of (0, 90) to affect the fit; this is a possible issue when using
+    # `iam.fit`, but not an issue when using `iam.convert`, since in
+    # that case aoi is defined internally)
+    weight = weight * np.logical_and(aoi >= 0, aoi <= 90).astype(int)
+
+    diff = np.abs(source_iam - np.nan_to_num(target(aoi, *target_params)))
```
kandersolar, line 991: Thinking more about user-defined functions, I think it's worth considering allowing the user to define the entire objective function instead of just the weight component of it.  Of course motivated users can always write their own versions of these functions, but it struck me as strange to allow custom weights but hard-code the rest of the objective function.

pvlib/iam.py
```diff
@@ -940,3 +941,360 @@ def schlick_diffuse(surface_tilt):
...
+                Requirements:
+                    1. Must have ``aoi`` as the first positional argument.
+                    2. Any other arguments must be keyword arguments,
+                       defined in ``weight_args``.
+                    3. Must return a float or an array-like object with the
+                       same shape as ``aoi``.
+
+            weight_args : dict, optional
+                A dictionary containing keyword arguments for the
+                weight function.
```
kandersolar, line 1128: What's the idea behind supporting `weight_args`?  It seems like a user can achieve the same thing by defining a wrapper around their weight function and handing that to pvlib.  If so, I'd say let's require `weight_function` to be a single-parameter function and get rid of `weight_args` to simplify the API here.

kandersolar (commented):
Initial comments

adriesse:
I guess there will be a SAND report on this some day--is there a draft available by chance?  There should really be more high-level information about the methods available to reviewers. Perhaps you could add a few paragraphs of notes in the doc strings.  That will be helpful to users later as well

cwhanse:
There is a Sandia report in preparation which will be the reference. I hope to share that report in the next few weeks.

pvlib/iam.py
cwhanse: Replaced `options` with a kwarg `weight` that looks for a function value.

cwhanse:
Nearing read for review again. There are failures in the py3.7min environment that look like they result from using older scipy functions. I'll investigate. https://github.com/pvlib/pvlib-python/actions/runs/6539998013/job/17759124948?pr=1827#step:9:67

There's also an unrelated failure appearing for certain environments that appears to be a pandas issue. https://github.com/pvlib/pvlib-python/actions/runs/6539998013/job/17759122369?pr=1827#step:9:57

cwhanse:
The new functions require scipy>1.5.0. I'm not enamored with how scipy<1.5.0 is being handled, if you have suggestions.

kandersolar:
> The new functions require scipy>=1.5.0. I'm not enamored with how scipy<1.5.0 is being handled, if you have suggestions.

I suggest we keep things simple and just increase our minimum requirement to 1.5.0.

docs/examples/reflections/plot_convert_iam_models.py
AdamRJensen: Agreed!

pvlib/iam.py
```diff
@@ -940,3 +942,365 @@ def schlick_diffuse(surface_tilt):
...
+    # of (0, 90) to affect the fit; this is a possible issue when using
+    # `iam.fit`, but not an issue when using `iam.convert`, since in
+    # that case aoi is defined internally)
+    weights = weights * np.logical_and(aoi >= 0, aoi <= 90).astype(int)
+
+    diff = np.abs(source_iam - np.nan_to_num(target(aoi, *target_params)))
+    return np.sum(diff * weights)
+
+
+def _get_ashrae_int(b):
```
AdamRJensen, line 992: I would write out intercept as "int" makes me think of integer

pvlib/iam.py
```diff
@@ -940,3 +942,365 @@ def schlick_diffuse(surface_tilt):
...
+        return _process_return(target_name, optimize_result)
+
+    else:
+        return {}
+
+
+def fit(measured_aoi, measured_iam, target_name, weight=None, xtol=None):
+    """
+    Finds parameters for target model that best fit the
+    measured data.
```
AdamRJensen, line 1228: I suggest renaming ``target_name`` to ``model_name`` (as the fit function doesn't have a source model then it seems silly calling it a target model)

pvlib/tests/test_iam.py
```diff
@@ -8,7 +8,7 @@
 import pandas as pd
 
 import pytest
-from .conftest import assert_series_equal
+from .conftest import (assert_series_equal, requires_scipy_150)
```
AdamRJensen, line 11:
```suggestion
from .conftest import assert_series_equal, requires_scipy_150
```

docs/examples/reflections/plot_fit_iam_models.py
```diff
@@ -0,0 +1,98 @@
...
+
+Illustrates how to fit an IAM model to data using :py:func:`~pvlib.iam.fit`
+
+"""
+
+# %%
+# An incidence angle modifier (IAM) model quantifies the fraction of direct
+# irradiance is that is reflected away from a module's surface. Three popular
+# IAM models are Martin-Ruiz :py:func:`~pvlib.iam.martin_ruiz`, physical
+# :py:func:`~pvlib.iam.physical`, and ASHRAE py:func:`~pvlib.iam.ashrae`.
```
AdamRJensen, line 14:
```suggestion
# :py:func:`~pvlib.iam.physical`, and ASHRAE :py:func:`~pvlib.iam.ashrae`.
```

docs/examples/reflections/plot_convert_iam_models.py
```diff
@@ -0,0 +1,154 @@
...
+Illustrates how to convert from one IAM model to a different model using
+:py:func:`~pvlib.iam.convert`
+
+"""
+
+# %%
+# An incidence angle modifier (IAM) model quantifies the fraction of direct
+# irradiance is that is reflected away from a module's surface. Three popular
+# IAM models are Martin-Ruiz :py:func:`~pvlib.iam.martin_ruiz`, physical
+# :py:func:`~pvlib.iam.physical`, and ASHRAE py:func:`~pvlib.iam.ashrae`.
```
AdamRJensen, line 15:
```suggestion
# :py:func:`~pvlib.iam.physical`, and ASHRAE :py:func:`~pvlib.iam.ashrae`.
```

docs/examples/reflections/plot_fit_iam_models.py
```diff
@@ -0,0 +1,98 @@
...
+plt.title('Fitting the physical model to data')
+plt.legend()
+plt.show()
+
+
+# %%
+# The weight function
+# -------------------
+# :py:func:`pvlib.iam.fit` uses a weight function when computing residuals
+# between the model abd data. The default weight
```
AdamRJensen, line 67:
```suggestion
# between the model and data. The default weight
```

AdamRJensen (commented):
Very well done!

One question, when fitting an IAM curve to data is there a way to account for potentially uneven sampling of IAM measurements with respect to AOI values?

pvlib/iam.py
```diff
@@ -940,3 +942,368 @@ def schlick_diffuse(surface_tilt):
...
+def _min_scipy():
+    '''convert and fit require scipy>=1.5.0
+    '''
+    from scipy import __version__
+    major, minor = __version__.split('.')[:2]
+    if (int(major) >= 1) & (int(minor) >= 5):
+        return True
+    else:
+        warnings.warn('iam.convert and iam.fit require scipy>=1.5.0')
+        return False
```
kandersolar, line 1106: This check logic doesn't work in all cases; consider an eventual scipy `2.0`.  But weren't we going to get rid of this version check and increase the overall minimum to 1.5 anyway?

pvlib/iam.py
```diff
@@ -940,3 +942,368 @@ def schlick_diffuse(surface_tilt):
...
+
+def _get_model(model_name):
+    # check that model is implemented
+    model_dict = {'ashrae': ashrae, 'martin_ruiz': martin_ruiz,
+                  'physical': physical}
+    try:
+        model = model_dict[model_name]
+    except KeyError:
+        raise NotImplementedError(f"The {model_name} model has not been \
+                                    implemented")
```
kandersolar, line 955:
```suggestion
        raise NotImplementedError(f"The {model_name} model has not been "
                                  "implemented")
```

The backslash approach results in a bunch of whitespace inside the error message string.

pvlib/iam.py
```diff
@@ -940,3 +942,368 @@ def schlick_diffuse(surface_tilt):
...
+
+
+def _check_params(model_name, params):
+    # check that the parameters passed in with the model
+    # belong to the model
+    exp_params = _IAM_MODEL_PARAMS[model_name]
+    if set(params.keys()) != exp_params:
+        raise ValueError(f"The {model_name} model was expecting to be passed \
+                         {', '.join(list(exp_params))}, but \
+                         was handed {', '.join(list(params.keys()))}")
```
kandersolar, line 967: Same here about backslashes

pvlib/iam.py
```diff
@@ -940,3 +942,368 @@ def schlick_diffuse(surface_tilt):
...
+
+            If source model is ``'physical'``, the dictionary must
+            contain the keys ``'n'``, ``'K'``, and ``'L'``.
+
+    target_name : str
+        Name of the target model. Must be ``'ashrae'``, ``'martin_ruiz'``, or
+        ``'physical'``.
+
+    weight : function, optional
+        A single-argument function of AOI that calculates weights for the
```
kandersolar, line 1137:
```suggestion
        A single-argument function of AOI (in degrees) that calculates weights for the
```

pvlib/iam.py
```diff
@@ -940,3 +942,368 @@ def schlick_diffuse(surface_tilt):
...
+    pvlib.iam.physical
+    """
+    if _min_scipy():
+
+        source = _get_model(source_name)
+        target = _get_model(target_name)
+
+        # if no options were passed in, we will use the default arguments
+        if weight is None:
+            weight = _sin_weight
```
kandersolar, line 1191: Rather than default to `None` and overwrite the default here, I think it should work to just assign `_sin_weight` as the default parameter value in the function signature.

pvlib/iam.py
```diff
@@ -940,3 +942,368 @@ def schlick_diffuse(surface_tilt):
...
+
+    measured_iam : array-like
+        IAM values. [unitless]
+
+    model_name : str
+        Name of the model to be fit. Must be ``'ashrae'``, ``'martin_ruiz'``,
+        or ``'physical'``.
+
+    weight : function, optional
+        A single-argument function of AOI that calculates weights for the
```
kandersolar, line 1245: degrees here as well

pvlib/iam.py
```diff
@@ -940,3 +942,368 @@ def schlick_diffuse(surface_tilt):
...
+    pvlib.iam.martin_ruiz
+    pvlib.iam.physical
+    """
+    if _min_scipy():
+
+        target = _get_model(model_name)
+
+        # if no options were passed in, we will use the default arguments
+        if weight is None:
+            weight = _sin_weight
```
kandersolar, line 1285: default value here as well

kandersolar (commented):
A quick look.  ~Haven't examined the tests yet.~ tests look great!

kandersolar:
> > One question, when fitting an IAM curve to data is there a way to account for potentially uneven sampling of IAM measurements with respect to AOI values?
>
> The fitting doesn't require evenly spaced IAM measurements.

In the case of nonuniform samples, more weight being given to the regions of denser samples, I think.  Not necessarily a problem, but something to consider documenting.

pvlib/iam.py
cwhanse: #1918 do you want to address and merge that first? Or just remove this check in anticipation?

kandersolar: I was thinking to do it all in this PR, but perhaps it's cleaner to do it separately.  I'll submit a PR to increase the minimum scipy.

docs/examples/reflections/plot_fit_iam_models.py
```diff
@@ -0,0 +1,98 @@
...
+# Fitting an IAM model to data
+# ----------------------------
+#
+# Here, we'll show how to fit an IAM model to data.
+# We'll generate some data by perturbing output from the Martin-Ruiz model to
+# mimic measured data and then we'll fit the physical model to the perturbed
+# data.
+
+# Create and perturb IAM data.
+aoi = np.linspace(0, 90, 100)
```
adriesse, line 42:
```suggestion
aoi = np.linspace(0, 85, 18)
```
This would more representative for measurements.

docs/examples/reflections/plot_fit_iam_models.py
```diff
@@ -0,0 +1,98 @@
...
+plt.ylabel('IAM')
+plt.title('Fitting the physical model to data')
+plt.legend()
+plt.show()
+
+
+# %%
+# The weight function
+# -------------------
+# :py:func:`pvlib.iam.fit` uses a weight function when computing residuals
```
adriesse, line 66: I suggest you say a bit more about the shape and purpose of these functions.  I see there is more in the other example, but I think it is more relevant or important here.

pvlib/iam.py
```diff
@@ -940,3 +942,383 @@ def schlick_diffuse(surface_tilt):
...
+        return True
+    else:
+        warnings.warn('iam.convert and iam.fit require scipy>=1.5.0')
+        return False
+
+
+def convert(source_name, source_params, target_name, weight=_sin_weight,
+            fix_n=True, xtol=None):
+    """
+    Convert a source IAM model to a target IAM model.
```
adriesse, line 1112:
```suggestion
    Convert source IAM model parameters to target IAM model parameters.
```

pvlib/iam.py
```diff
@@ -940,3 +942,383 @@ def schlick_diffuse(surface_tilt):
...
+            the key ``'b'``.
+
+            If target model is ``'martin_ruiz'``, the dictionary will
+            contain the key ``'a_r'``.
+
+            If target model is ``'physical'``, the dictionary will
+            contain the keys ``'n'``, ``'K'``, and ``'L'``.
+
+    Note
+    ----
```
adriesse, line 1172: I know the reference explains everything, but being a little more loquacious here wouldn't hurt.  Same in the other function's note.

pvlib/iam.py
cwhanse: I'll quibble with this edit, we're not converting `b` to `a_r`, we're estimating `a_r` from the source model's output.

pvlib/iam.py
cwhanse: Some hints what you hoped to see here?

pvlib/iam.py
adriesse: Don't worry about it.

pvlib/iam.py
adriesse: No, it just struck me as terse.  Don't worry about it.

docs/examples/reflections/plot_convert_iam_models.py
```diff
@@ -99,7 +99,8 @@
 physical_iam_default = physical(aoi, **physical_params_default)
 
 
-# ... using a custom weight function.
+# ... using a custom weight function. The weight function must take ``aoi``
+# as it's argument and return a vector of the same length as ``aoi``.
```
adriesse, line 103:
```suggestion
# as its argument and return a vector of the same length as ``aoi``.
```

docs/examples/reflections/plot_fit_iam_models.py
```diff
@@ -69,7 +69,8 @@
 # function to :py:func:`pvlib.iam.fit`.
 #
 
-# Define a custom weight function.
+# Define a custom weight function.  The weight function must take ``aoi``
+# as it's argument and return a vector of the same length as ``aoi``.
```
adriesse, line 73:
```suggestion
# as its argument and return a vector of the same length as ``aoi``.
```

pvlib/iam.py
```diff
@@ -940,3 +941,362 @@ def schlick_diffuse(surface_tilt):
...
+    Target model parameters are determined by minimizing
+
+    .. math::
+
+        \\sum_{\\theta=0}^{90} weight \\left(\\theta \\right) \\times
+        \\| source \\left(\\theta \\right) - target \\left(\\theta \\right) \\|
+
+    The sum is over :math:`\\theta = 0, 1, 2, ..., 90`.
```
kandersolar, line 1167: Not quite true; the code actually uses `np.linspace(0, 90, 100)`

kandersolar (approved):
One more minor doc comment...

@cwhanse good to merge?

pvlib/iam.py
kandersolar: Pedantic nitpick: `linspace(0, 90, 90)` produces `[0., 1.0112, 2.0225, ... , 87.9775, 88.9888, 90.]`.  To produce `theta = 0, 1, 2, ..., 90` as the docstring states, I think we need `linspace(0, 90, 91)`.

cwhanse: Not a nit, rather, that's the intent, because it feels natural to step by 1.0000000000000...

kandersolar: I agree it feels natural to step by unity.  I was trying to point out that the current code (`np.linspace(0, 90, 90)`) steps by 90/89 $\approx$ 1.011236.  To step by unity from 0 to 90 inclusive, it should be changed to `linspace(0, 90, 91)`.

cwhanse:
@kandersolar codecov error.

kandersolar:
Thanks @ajonesr and @cwhanse for this substantial contribution!

PR #1828, opened 2023-08-04
Refactor docstring params (2)

### LINKED ISSUE #1574: "default None" in docstring parameter descriptions considered harmful

echedey-ls:
- [x] Closes #1574 among other updates. See bottom of PR body.
 - ~[ ] Tests added~
 - [x] Current code includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments everywhere.

Cherry-picked commits from #1790 as proposed by @kandersolar 

## Changes in this PR
1. Changes excessive use of parentheses in defaults
  `return_components: bool (optional, default=False)` -> `return_components : bool, default False`
1. Removes colon when default is to be specified
  `logical_records: list or tuple, default: ('0100',)` -> `logical_records : list or tuple, default ('0100',)`
1. default None -> optional (#1574)
1. `If None` -> `If not specified`
1. Remove trailing dots from type definition


echedey-ls:
TODO: regex ```DOI: ...``` to ``` :doi:`...` ```

pvlib/irradiance.py
```diff
@@ -574,7 +574,7 @@ def get_ground_diffuse(surface_tilt, ghi, albedo=.25, surface_type=None):
         the reflection coefficient. Must be >=0 and <=1. Will be
         overridden if surface_type is supplied.
 
-    surface_type: None or string, default None
+    surface_type : string, optional
         If not None, overrides albedo. String can be one of 'urban',
```
cwhanse, line 578:
```suggestion
        If supplied, overrides ``albedo``. String can be one of 'urban',
```

pvlib/irradiance.py
```diff
@@ -1621,7 +1621,7 @@ def disc(ghi, solar_zenith, datetime_or_doy, pressure=101325,
         pd.DatetimeIndex.dayofyear, or pd.DatetimeIndex.
 
     pressure : None or numeric, default 101325
-        Site pressure in Pascal. If None, relative airmass is used
+        Site pressure in Pascal. If not specified, relative airmass is used
```
cwhanse, line 1624:
```suggestion
        Site pressure in Pascal. If specified, relative airmass is used 
```

pvlib/irradiance.py
```diff
@@ -1776,7 +1776,7 @@ def dirint(ghi, solar_zenith, times, pressure=101325., use_delta_kt_prime=True,
         GHI points is 1.5 hours or greater. If use_delta_kt_prime=True,
         input data must be Series.
 
-    temp_dew : None, float, or array-like, default None
+    temp_dew : None, float, or array-like, optional
         Surface dew point temperatures, in degrees C. Values of temp_dew
```
cwhanse, line 1780: Here and throughout, ``temp_dew`` in the docstrings. This is an optional nit, decline if you feel it's out of scope for the PR.

pvlib/modelchain.py
```diff
@@ -138,12 +138,12 @@ def basic_chain(times, latitude, longitude,
     airmass_model : str, default 'kastenyoung1989'
         Passed to atmosphere.relativeairmass.
 
-    altitude : None or float, default None
-        If None, computed from pressure. Assumed to be 0 m
+    altitude : float, optional
+        If not specified, computed from pressure. Assumed to be 0 m
         if pressure is also None.
```
cwhanse, line 143:
```suggestion
        If not specified, computed from ``pressure``. Assumed to be 0 m
        if ``pressure`` is None.
```

pvlib/modelchain.py
```diff
@@ -138,12 +138,12 @@ def basic_chain(times, latitude, longitude,
...
 
-    pressure : None or float, default None
-        If None, computed from altitude. Assumed to be 101325 Pa
+    pressure : float, optional
+        If not specified, computed from altitude. Assumed to be 101325 Pa
         if altitude is also None.
```
cwhanse, line 147:
```suggestion
        If not specified, computed from ``altitude``. Assumed to be 101325 Pa
        if ``altitude`` is None.
```

pvlib/pvsystem.py
```diff
@@ -129,17 +129,17 @@ class PVSystem:
         Azimuth angle of the module surface.
         North=0, East=90, South=180, West=270.
 
-    albedo : None or float, default None
+    albedo : float, optional
         Ground surface albedo. If ``None``, then ``surface_type`` is used
```
cwhanse, line 133:
```suggestion
        Ground surface albedo. If not specified then ``surface_type`` is used
```

pvlib/pvsystem.py
```diff
@@ -129,17 +129,17 @@ class PVSystem:
...
         to look up a value in ``irradiance.SURFACE_ALBEDOS``.
         If ``surface_type`` is also None then a ground surface albedo
```
cwhanse, line 135:
```suggestion
        If ``surface_type`` is not specified then a ground surface albedo
```

pvlib/pvsystem.py
```diff
@@ -148,7 +148,7 @@ class PVSystem:
          Describes the module's construction. Valid strings are 'glass_polymer'
          and 'glass_glass'. Used for cell and module temperature calculations.
 
-    module_parameters : None, dict or Series, default None
+    module_parameters : None, dict or Series, optional
         Module parameters as defined by the SAPM, CEC, or other.
 
     temperature_model_parameters : None, dict or Series, default None.
```
cwhanse, line 154:
```suggestion
    temperature_model_parameters : None, dict or Series, optional
```

cwhanse (commented):
Thanks @echedey-ls 

There's more markdown to be done in the descriptions of parameters this touches, if you feel it is in scope. I stopped suggesting it.

@pvlib/pvlib-maintainer it is questionable in my mind whether `module_parameters`, `inverter_parameters` and the like should be optional for instantiating a `ModelChain`. The `ModelChain` is created, true, but nothing runs.

pvlib/irradiance.py
echedey-ls: I think you misread this statements. Mind double-checking it?

pvlib/modelchain.py
echedey-ls: I've changed `is also None` to `is also unspecified`. What do you think about it?

pvlib/modelchain.py
echedey-ls: See comment on line 142

pvlib/pvsystem.py
echedey-ls: Gotta account for "also" in the regexes...

pvlib/pvsystem.py
echedey-ls: Thanks for spotting that None, I've regexed all the files to find and replace them.

pvlib/irradiance.py
echedey-ls: Done!

pvlib/pvsystem.py
echedey-ls: I ended up doing that

cwhanse:
@echedey-ls thanks for the work on this. If we don't get all instances of `default None` in this PR, we can do more later.

pvlib/modelchain.py
echedey-ls: [This](https://github.com/echedey-ls/pvlib-python/commit/6171e77f3831fe0184e375ee45188ebf644018dd) is the commit.

echedey-ls:
I had to iterate a bit more... At first glance I thought very constraining regexes were enough, but even trailing dots were messing with that. I think all have been addressed for now.

pvlib/irradiance.py
```diff
@@ -574,8 +574,8 @@ def get_ground_diffuse(surface_tilt, ghi, albedo=.25, surface_type=None):
         the reflection coefficient. Must be >=0 and <=1. Will be
         overridden if surface_type is supplied.
 
-    surface_type: None or string, default None
-        If not None, overrides albedo. String can be one of 'urban',
+    surface_type : string, optional
+        If supplied, overrides ``albedo``. String can be one of 'urban',
         'grass', 'fresh grass', 'snow', 'fresh snow', 'asphalt', 'concrete',
         'aluminum', 'copper', 'fresh steel', 'dirty steel', 'sea'.
```
cwhanse, line 580:
```suggestion
        If supplied, overrides ``albedo``. `surface_type` can be one of 
        'urban', 'grass', 'fresh grass', 'snow', 'fresh snow', 'asphalt',
        'concrete', 'aluminum', 'copper', 'fresh steel', 'dirty steel',
        'sea'.
```

pvlib/irradiance.py
```diff
@@ -1341,7 +1341,7 @@ def perez_driesse(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
         and <=360. The azimuth convention is defined as degrees east of
         north (e.g. North = 0, East = 90, West = 270).
 
-    airmass : numeric (optional, default None)
+    airmass : numeric, optional
         Relative (not pressure-corrected) airmass values. If airmass is a
```
cwhanse, line 1345:
```suggestion
        Relative (not pressure-corrected) airmass values. If ``airmass`` is a
```

pvlib/irradiance.py
cwhanse: The logic in the code is perhaps not as clear as it could be. Relative air mass is used when ``pressure`` is None. Let's  clear this up in a new PR.

```suggestion
        Site pressure in Pascal. If not specified, relative airmass is used.
```

pvlib/location.py
```diff
@@ -173,10 +173,10 @@ def get_solarposition(self, times, pressure=None, temperature=12,
         ----------
         times : pandas.DatetimeIndex
             Must be localized or UTC will be assumed.
-        pressure : None, float, or array-like, default None
-            If None, pressure will be calculated using
+        pressure : float, or array-like, optional
+            If not specified, pressure will be calculated using
```
cwhanse, line 177:
```suggestion
            If not specified, ``pressure`` is calculated using
```

pvlib/solarposition.py
```diff
@@ -312,7 +312,7 @@ def spa_python(time, latitude, longitude,
         *Note: delta_t = None will break code using nrel_numba,
         this will be fixed in a future version.*
         The USNO has historical and forecasted delta_t [3]_.
-    atmos_refrac : None or float, optional, default None
+    atmos_refrac : float, optional, optional
```
cwhanse, line 315:
```suggestion
    atmos_refrac : float, optional
```

cwhanse (approved):

pvlib/irradiance.py
```diff
@@ -575,7 +575,7 @@ def get_ground_diffuse(surface_tilt, ghi, albedo=.25, surface_type=None):
         overridden if surface_type is supplied.
 
     surface_type : string, optional
-        If supplied, overrides ``albedo``. `surface_type` can be one of 
+        If supplied, overrides ``albedo``. `surface_type` can be one of
```
cwhanse, line 578:
```suggestion
        If supplied, overrides ``albedo``. ``surface_type`` can be one of
```

pvlib/irradiance.py
```diff
@@ -1620,8 +1621,8 @@ def disc(ghi, solar_zenith, datetime_or_doy, pressure=101325,
         Day of year or array of days of year e.g.
         pd.DatetimeIndex.dayofyear, or pd.DatetimeIndex.
 
-    pressure : None or numeric, default 101325
-        Site pressure in Pascal. If None, relative airmass is used
+    pressure : numeric, default 101325
+        Site pressure in Pascal. If not specified, relative airmass is used.
```
cwhanse, line 1625: I think the text below would clear up my confusion. I don't know if the alternative "If `None`" requires changing the parameter descriptor to `numeric or None`,  

```suggestion
        Site pressure in Pascal. If numeric, absolute airmass is used.
        If ``None``, relative airmass is used.
```

echedey-ls: English is not my mother tongue, but I really like how it currently is. Anyways, we could change "If None" to "Otherwise, ...". I don't dislike your suggestion.

Or to "If not provided". I think the latter makes more sense. It's up to you.

cwhanse: My problem is with `numeric, default 101325`. So it has a value (a number). If I don't provide a different value, 101325 is used. Then what triggers the "not specified" condition?

Maybe it would be better left as original, but perhaps `numeric or None, default 101325`.

echedey-ls: Ah, I see... from the coded behaviour it needs to be `None` to use relative AM. Also, see [current docs](https://github.com/pvlib/pvlib-python/blob/57d61af5aba87e7089a78cfae0b26d153a130fc7/pvlib/irradiance.py#L1604-L1610). `pressure=None` is mentioned. All I see are reasons to stick to the original. Thanks for spotting that.

echedey-ls: How do you feel about this text?

```
pressure : numeric or None, default 101325
        Site pressure in Pascal. Uses absolute (pressure-corrected) airmass
        by default.
        Set to ``None`` to use relative airmass.
```

I think it's safe to assume that if someone inputs the pressure it means they will know that it is using it to compute absolute airmass. Or we may clear that up with:

```
pressure : numeric or None, default 101325
        Site pressure in Pascal. Values to compute absolute
        (pressure-corrected) airmass. Uses sea-level pressure by default.
        Set to ``None`` to use relative airmass.
```

A bit more descriptive than it currently is IMO.

cwhanse: I like the first, that is clear to me. Thanks for your patience.

echedey-ls:
This should be ready to go, unless anything else wants to be done!

kandersolar (approved):
LGTM, thanks @echedey-ls!  Might be worth leaving a comment on this PR with a list of the final set of regexes you used for this PR, in case we want to re-run something like this in the future.  

> it is questionable in my mind whether module_parameters, inverter_parameters and the like should be optional for instantiating a ModelChain. The ModelChain is created, true, but nothing runs.

@cwhanse I'm not sure what you mean.  `ModelChain` doesn't take `module_parameters` or `inverter_parameters`.

echedey-ls:
None regexes

```
These go by order:
F: (?<spaces>\s{8}|\s{4})(?<name>\w*)\s?: (?<type>.*) \(optional, default=(?<default>.*)\)$
Manually fix every match. This is a very constraining regex.

F1: (?<spaces> {8}| {4})(?<name>\w*) ?: (?<type>.*), default:? None\.?
F2: (?<spaces> {8}| {4})(?<name>\w*) ?: [Nn]one(?:,| or) (?<types>.*)
F3: (?<spaces> {8}| {4})(?<name>\w*) ?: (?<type>.*) or [nN]one, optional
R: $1$2 : $3, optional
Note: F3 requires special attention to commas.

F: (?<spaces> {8}| {4})(?<name>\w*) ?: [Nn]one, (?<type>.*), default:? (?<def>.*)\.?
R: $1$2 : $3, default $4

F: (?<spaces> {8}| {4})(?<name>\w*) ?: (?<types>.*) or [Nn]one(?<trailing>.*)
R: $1$2 : $3$4

F: If None
R: If not specified,
```

DOI regexes:
```
Regex find: (?:doi|DOI):(?!`)\s?(.*?)(\.\n|\n)
Replace: :doi:`$1`$2
```

Some of them could be changed to be more flexible, but I wanted the least number of false positives. I'm looking at the excessive-parentheses regex, they are really ugly in the web.

kandersolar:
With no objections, time to merge!  Thanks again @echedey-ls :)

PR #1829, opened 2023-08-06
Lazify various calls to all() by using generators instead of comprehensions

eumiro:
- ~~[ ] Closes #xxxx~~
 - [ ] Tests added

### DIFF
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -1984,7 +1984,7 @@ def _irrad_for_celltemp(total_irrad, effective_irradiance):
 
     """
     if isinstance(total_irrad, tuple):
-        if all(['poa_global' in df for df in total_irrad]):
+        if all('poa_global' in df for df in total_irrad):
             return _tuple_from_dfs(total_irrad, 'poa_global')
         else:
             return effective_irradiance
--- a/docs/sphinx/source/whatsnew/v0.10.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.10.2.rst
@@ -64,3 +64,4 @@ Contributors
 * Anton Driesse (:ghuser:`adriesse`)
 * Lukas Grossar (:ghuser:`tongpu`)
 * Areeba Turabi (:ghuser:`aturabi`)
+* Miroslav Šedivý (:ghuser:`eumiro`)
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -1296,20 +1296,18 @@ class Array:
         """
 
         # get relevent Vmp and Imp parameters from CEC parameters
-        if all([elem in self.module_parameters
-                for elem in ['V_mp_ref', 'I_mp_ref']]):
+        if all(elem in self.module_parameters
+               for elem in ['V_mp_ref', 'I_mp_ref']):
             vmp_ref = self.module_parameters['V_mp_ref']
             imp_ref = self.module_parameters['I_mp_ref']
 
         # get relevant Vmp and Imp parameters from SAPM parameters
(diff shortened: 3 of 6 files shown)


pvlib/tests/test_pvsystem.py
```diff
@@ -174,7 +174,7 @@ def test_retrieve_sam_cecinverter():
     ]
     inverter = 'Yaskawa_Solectria_Solar__PVI_5300_208__208V_'
     assert inverter in data
-    assert set(data[inverter].keys()) == set(keys)
+    assert set(data[inverter]) == set(keys)
```
kandersolar, line 177: For the two cases in this file, we are actually dealing with pandas Series, not dicts, so the semantics are different: `set(Series)` returns a set of the values instead of the keys.  So these two lines should keep the `.keys()` (or be changed to use `assert_index_equal` or something)`.

pvlib/modelchain.py
```diff
@@ -2034,11 +2034,9 @@ def _all_same_index(data):
 def _common_keys(dicts):
     """Return the intersection of the set of keys for each dictionary
     in `dicts`"""
-    def _keys(x):
-        return set(x.keys())
     if isinstance(dicts, tuple):
-        return set.intersection(*map(_keys, dicts))
-    return _keys(dicts)
+        return set.intersection(*map(set, dicts))
+    return set(dicts)
```
kandersolar, line 2039: Here also, we need to support `pd.Series`, so we need to keep the `.keys()`.

pvlib/modelchain.py
```diff
@@ -861,7 +861,7 @@ def ac_model(self, model):
 
     def infer_ac_model(self):
         """Infer AC power model from system attributes."""
-        inverter_params = set(self.system.inverter_parameters.keys())
+        inverter_params = set(self.system.inverter_parameters)
```
kandersolar, line 864: And here, I expect

kandersolar (commented):
Thanks @eumiro for this PR!  And apologies for the delayed review.

I am generally okay with this PR, but a few changes of the changes need to be reverted in order to preserve compatibility pandas Series objects.  Details below.

cwhanse:
@eumiro you'll see test failures in unrelated code. Those are from a scipy issue, and you may ignore them.

kandersolar (approved):
Thanks @eumiro!

adriesse:
This is another example of a PR where I have to go to the code changes to get any idea of what its purpose is and even then I only half understand it.  I recommend putting descriptions in PR's and/or associated issues.

mikofski:
A jargon glossary would be helpful. “Lazy” means evaluated at the last possible conceivable instant, in this case because `all()` will fail if any object in the potential list evaluates as `False` then it’s unnecessary to evaluate all of the objects in the list or even expand a list into an array. Using a Python generator is “lazy” because it only evaluates one item at a time from the potential list, so that as soon as a `False` item is evaluated it can return `False` from the `all()` or at least that’s the theory. In practice the only advantage may be more terse code with fewer characters. Hard to know unless you run the numbers. There’s also a memory advantage usually too

adriesse:
I understand the concepts actually, just think if it's worth the trouble to make the change it should be worth a few sentences from the contributor to describe it.  The benefit of a generator for two simple items must surely be very small.

kandersolar:
I agree that the benefit of this change is minor at best, but as long as the PR was already submitted I didn't see any harm in merging it.  I wouldn't suggest anyone spend time on these kinds of changes in the future though :)

I also agree that a bit of explanation would have been useful here.  I did edit the title to be a bit more helpful but did not think to add explanation in the body.  I will do that now!

PR #1834, opened 2023-08-14
Implement C1 continuous version of the Erbs diffuse-fraction/decomposition model

### LINKED ISSUE #1755: Erbs diffuse fraction model has discontinuities at transition points
If you look closely at the piecewise definition of the Erbs diffuse fraction model you will see that it is discontinuous at the two transition points.  This is not usually a problem, but when the Erbs model happens to be used in some kind of numerical optimization the discontinuities could cause problems.

Add a new diffuse fraction function with a piecewise definition that closely matches the original Erbs model but is continuous. And while we're at it we can make the first derivative continuous too.

Modify the existing model function.

**Additional context**
This has been implemented and tested off line but I don't have the bandwidth for a PR currently.  I thought I should mention it since it is somewhat relevant to #1753.

adriesse:
- [x] Closes #1755 
 - [x] Tests added

### DIFF
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -2272,7 +2272,7 @@ def erbs_driesse(ghi, zenith, datetime_or_doy=None, dni_extra=None,
     r"""
     Estimate DNI and DHI from GHI using the continuous Erbs-Driesse model.
 
-    The Erbs-Driesse model [1]_ is a new formulations of the original Erbs
+    The Erbs-Driesse model [1]_ is a reformulation of the original Erbs
     model [2]_ that provides continuity of the function and its first
     derivative at the two transition points.
 
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -2273,8 +2273,8 @@ def erbs_driesse(ghi, zenith, datetime_or_doy=None, dni_extra=None,
     Estimate DNI and DHI from GHI using the continuous Erbs-Driesse model.
 
     The Erbs-Driesse model [1]_ is a new formulations of the original Erbs
-    model [2]_ that provides continuity of the function and its first derivative
-    at the two transition points.
+    model [2]_ that provides continuity of the function and its first
+    derivative at the two transition points.
 
     .. math::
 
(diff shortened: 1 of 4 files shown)


adriesse:
This new model/function is using the same structure as all the other diffuse-fraction/decomposition functions with one exception: it has an optional input for dni_extra, which makes datetime_or_doy also optional.

pvlib/irradiance.py
```diff
@@ -2267,6 +2267,138 @@ def erbs(ghi, zenith, datetime_or_doy, min_cos_zenith=0.065, max_zenith=87):
...
+        If neither datetime_or_doy nor dni_extra is provided.
+
+    Notes
+    -----
+    The diffuse fraction of the Erbs-Driesse model deviates from the original
+    Erbs model by less than 0.0005.
+
+    References
+    ----------
+    .. [1] A. Driesse, A. Jensen, R. Perez, A Continuous Form of the Perez
```
cwhanse, line 2335: This says "Perez" model, is that a typo? Or is there also a continuous Perez model forthcoming?

cwhanse (approved):

pvlib/irradiance.py
adriesse: No. Yes.  The modification is introduced along with the Perez model.  They are used together for reverse transposition in the paper.

adriesse:
@kandersolar you have a very good point that I have also thought about.   Here are a two options that are not mutually exclusive: 

- delay the merge and reconsider the question just before the next release
- give the PR reviewers access to our manuscript

Here is a plot from the manuscript:

erbs_discontinuity

cwhanse:
@kandersolar we added the `sandia_multi` inverter function before the reference paper was published. The extension of `iam.physical` to handle multiple interfaces was originated in pvlib discussions, if I'm not mistaken. So we've made exceptions before.

pvlib/irradiance.py
```diff
@@ -2267,6 +2267,136 @@ def erbs(ghi, zenith, datetime_or_doy, min_cos_zenith=0.065, max_zenith=87):
     return data
 
 
+def erbs_driesse(ghi, zenith, datetime_or_doy=None, dni_extra=None,
+                 min_cos_zenith=0.065, max_zenith=87):
+    r"""
+    Estimate DNI and DHI from GHI using the continuous Erbs-Driesse model.
+
+    The Erbs-Driesse model [1]_ is a new formulations of the original Erbs
```
wholmgren, line 2275: "a *new* formulation" may eventually feel not so new. "a reformulation" or something like that might be more enduring.

wholmgren (approved):
Thanks @adriesse! 

I don't think we need to wait for a publication in this case.

pvlib/irradiance.py
adriesse: "New" vs. "original" should be enduring, but I have nevertheless reformulated the string and made a new commit.

adriesse:
Does anyone have an idea why these test failures started to pop up?

kandersolar:
I suspect `RuntimeError: Optimal parameters not found: The maximum number of function evaluations is exceeded.` errors are due to https://github.com/scipy/scipy/issues/19103.

kandersolar (approved):
Thanks @adriesse!

adriesse:
Ok, seems ready to merge then--or do we have to wait for scipy to sort itself out?

kandersolar:
> do we have to wait for scipy to sort itself out?

Nope, no reason to hold this up for that :)

adriesse:
Thanks @ all!

PR #1851, opened 2023-09-11
Miniconda devcontainer

nicomt:
- [ ] Closes #xxxx
 - [ ] Tests added

Adding [devcontainer](https://code.visualstudio.com/docs/devcontainers/containers) files to automate the development environment in VSCode and GitHub [Codespaces](https://github.com/features/codespaces).

I got a new computer and had to set up my pvlib development environment again.
I have been using these at work and find them very useful. 
It is handy when working with tricky dependencies like Numba, Cython, etc. 
Also, setting up Codespaces will allow less experienced developers to start contributing from the browser without all the boilerplate setup.

I am leaving this PR as a draft because I should add some docs before releasing it. But I wanted to get some feedback first.


echedey-ls:
Just a quick note on Python pinned version, isn't it a better idea to pin the minimum suppported version (3.7)?

So tests don't fail because newer features are being used.

nicomt:
@echedey-ls, that's a valid point. But it feels a little icky to use a Python version past its [end-of-life support](https://devguide.python.org/versions/).
It seems like the tests are passing for other versions of Python in Ubuntu.
Should we pin somewhere in the middle 3.8, 3.9?

kandersolar:
Regarding 3.7 being EOL, the main reason I see to keep it for now is that it's still the top version by downloads we get on PyPI: https://pypistats.org/packages/pvlib

So I think it makes sense to continue testing on it until it becomes a maintenance or development pain for us.

Back to this PR, I am generally not convinced it makes sense to include these types of files in the main pvlib repository.  My main objection is that everyone has their own favorite dev environment, and I don't want to see a proliferation of configuration files for spyder, pycharm, vim, sublime text et al in the repo.  A second point is that the `environment.yml` file is another thing to have to keep updated and is mostly redundant with the dependencies listed in `setup.py`.

However, I am interested in learning about what this could enable with codespaces.  @nicomt could you say a bit more about that?

nicomt:
Hi @kandersolar,  here is some extra info

## environment.yml
I agree about redundancy in the packages. The intention for environment.yml was to add missing packages like flake8 or any package you wouldn't want in the setup.py. I included the standard `pip install -e .` [startup command](https://github.com/nicomt/pvlib-python/blob/devcontainer/.devcontainer/devcontainer.json#L30), which should pull all the other dependencies.
I made it confusing because I was using `pip install -e .` instead of `pip install -e .[all]` to speed up the initial build and startup. But if we release it, we could change it to install all and trim redundancies in the environment.yml.

## IDE Vendor Files
Yes, it goes against standard practices to include VSCode specific configuration, which is a significant consideration before merging this. One difference is that Microsoft is trying to make the devcontainer an [open standard](https://containers.dev/implementors/spec/) that works with other vendors. It recently got [adopted by IntelliJ](https://www.jetbrains.com/help/idea/connect-to-devcontainer.html), and others may join. But that might not pan out, and it still can be a deal breaker.

## Codespaces
Codespaces is enabled by default for any user logged in on the GitHub website. You get 60 hours free each month, and it's [$0.18 for any additional hour](https://github.com/pricing/calculator) if you use their smallest 2-core VMs.
You can start the codespace from the repo.
image
You can start a codespace even if the repo doesn't have a devcontainer definition. It will start a generic devconainter that you can configure yourself.
To try the config, open a codespace on the [PR branch](https://github.com/nicomt/pvlib-python/tree/devcontainer). It takes a while to build, but I think that can be solved by caching with GitHub Actions.

nicomt:
FYI the devcontainer is failing to build after switching to Python 3.7.
I will report back once is fixed

nicomt:
Okay, is working again.
I had to change to bare Python because Conda dropped support for Python 3.7. https://github.com/conda/conda/issues/12436
I also enabled [codespaces pre-build](https://github.com/nicomt/pvlib-python/actions/runs/6154729968) so it should start faster now.
Let me know what you think and if is worth configuring further.

PR #1863, opened 2023-09-16
Add loss input to example in modelchain.with_pvwatts docs

AdamRJensen:
- ~[ ] Closes #xxxx~
 - ~[ ] Tests added~

I had some difficulties figuring out what losses were used in the ``modelchain.with_pvwatts``, I am making this PR to make it clearer and demonstrate how to change the loss factors.

My motivation for this PR is that I think it's important to encourage users to consider what loss assumptions are made and make it easier to modify them.

Also, it's typically new users who would use the ``modelchain.with_pvwatts`` model, which is why I have added the necessary imports such that the example can be run exactly as copied without having users to figure out where the different modules are imported from (something that would be difficult for a new user).

### DIFF
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -582,17 +582,27 @@ class ModelChain:
             constructor and take precedence over the default
             configuration.
 
+        .. warning::
+            The PVWatts model assumes 14% total system losses. The loss assumptions
+            can be modified as shown in the example below.
+
         Examples
         --------
+        >>> from pvlib import temperature, pvsystem, location, modelchain
         >>> module_parameters = dict(gamma_pdc=-0.003, pdc0=4500)
         >>> inverter_parameters = dict(pdc0=4000)
-        >>> tparams = TEMPERATURE_MODEL_PARAMETERS['sapm']['open_rack_glass_glass']
-        >>> system = PVSystem(surface_tilt=30, surface_azimuth=180,
-        ...     module_parameters=module_parameters,
-        ...     inverter_parameters=inverter_parameters,
-        ...     temperature_model_parameters=tparams)
-        >>> location = Location(32.2, -110.9)
-        >>> ModelChain.with_pvwatts(system, location)
+        >>> tparams = temperature.TEMPERATURE_MODEL_PARAMETERS['sapm']['open_rack_glass_glass']
+        >>> pvwatts_losses = {'soiling': 2, 'shading': 3, 'snow': 0, 'mismatch': 2,
+        >>>                   'wiring': 2, 'connections': 0.5, 'lid': 1.5,
+        >>>                   'nameplate_rating': 1, 'age': 0, 'availability': 30}
+        >>> system = pvsystem.PVSystem(
+        >>>     surface_tilt=30, surface_azimuth=180,
+        >>>     module_parameters=module_parameters,
+        >>>     inverter_parameters=inverter_parameters,
+        >>>     temperature_model_parameters=tparams,
+        >>>     losses_parameters=pvwatts_losses)
+        >>> loc = location.Location(32.2, -110.9)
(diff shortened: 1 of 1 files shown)


pvlib/modelchain.py
```diff
@@ -582,17 +582,28 @@ def with_pvwatts(cls, system, location,
...
-        >>> system = PVSystem(surface_tilt=30, surface_azimuth=180,
-        ...     module_parameters=module_parameters,
-        ...     inverter_parameters=inverter_parameters,
-        ...     temperature_model_parameters=tparams)
-        >>> location = Location(32.2, -110.9)
-        >>> ModelChain.with_pvwatts(system, location)
+        >>> tparams = temperature.TEMPERATURE_MODEL_PARAMETERS['sapm']['open_rack_glass_glass']
+        >>> pvwatts_losses = {'soiling': 2, 'shading': 3, 'snow': 0, 'mismatch': 2,
+        >>>                   'wiring': 2, 'connections': 0.5, 'lid': 1.5,
+        >>>                   'nameplate_rating': 1, 'age': 0, 'availability': 30}
```
kandersolar, line 598: Maybe we should have two examples here, one that keeps things as simple as possible, and one that gets more into the weeds (like specifying losses)?  `with_pvwatts` is an easy point of entry for ModelChain, and PVWatts isn't really the model to use if you care a lot about the details, so I'm hesitant to make the only usage example more complicated.

IMHO there is also an argument to be made that PVWatts losses aren't really legitimate on their own and are better regarded as a single combined fudge factor, but maybe that's just my editorializing.

AdamRJensen: I don't think users should ever run a model without any considerations for what losses are specified or at least be aware that losses are accounted for by the model? Acknowledging and accepting the default values is perfectly resonable, but awareness of what they are seems pertinent.

pvlib/modelchain.py
```diff
@@ -582,17 +582,28 @@ def with_pvwatts(cls, system, location,
             constructor and take precedence over the default
             configuration.
 
+        Warning
+        -------
+        The PVWatts model defaults to 14 % total system losses. The loss
+        assumptions can be modified as shown in the example below.
```
AdamRJensen, line 588:
```suggestion
        The PVWatts model defaults to 14 % total system losses. The loss
        assumptions represent annual averages and can be modified as
        shown in the example below.
```

pvlib/modelchain.py
AdamRJensen: I have created two examples, one simple and one that builds upon the simple one where the losses are modified.

pvlib/modelchain.py
kandersolar: Two thoughts, neither of which I feel too strongly about:
- I think "annual averages" is true but not as specific as it could be.  How about "...represent constant DC-side loss fractions and can be modified..."?  As an aside, constant losses means all simulation timesteps are affected equally (on a relative basis), so the "snow loss" applies to July just as much as it does to December...
- Re "not recommended": I certainly don't disagree with the sentiment, but this struck me as too subjective a statement to be appropriate for pvlib's documentation.  If we want to make novice users aware that PVWatts v5 isn't a very detailed model, I think it would be better to describe what the model does (or doesn't) do, and let the user judge suitability for themselves.

cwhanse: Bikeshedding: "The PVWatts losses are fractions of DC power and can be modified..."

pvlib/modelchain.py
```diff
@@ -582,17 +582,38 @@ def with_pvwatts(cls, system, location,
...
```
AdamRJensen, line 588:
```suggestion
        The PVWatts model defaults to 14 % total system losses. The PVWatts
        losses are fractions of DC power and can be modified, as shown in the
        example below.
```

kandersolar (approved):

PR #1876, opened 2023-09-26
Continuous version of the Perez transposition model implementation

### LINKED ISSUE #1841: Continuous version of the Perez transposition model
Due to the empirical look-up tables, the Perez transposition model produces small jumps in tilted irradiance under gradually evolving sky conditions, and creates a difficult challenge for reverse transposition algorithms.

Create a new version of the model having smooth so transitions, but without changing the overall performance w.r.t. the original.

**Additional context**
A publication describing a solution is forthcoming, and this will be implemented here in parallel.

adriesse:
- [x] Closes #1841
 - [x] Tests added

Create a new version of the Perez model having smooth transitions, but without changing the overall performance w.r.t. the original.  The solution uses quadratic splines as described in recent presentations and posters, and documented in a paper just accepted for publication in the Solar Energy Journal.  (Link will be provided later; the pre-print is available to PR reviewers on request.)

### DIFF
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -483,8 +483,7 @@ def get_sky_diffuse(surface_tilt, surface_azimuth,
     elif model == 'perez-driesse':
         # perez_driesse will calculate its own airmass if needed
         sky = perez_driesse(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
-                            solar_zenith, solar_azimuth, airmass,
-                            model=model_perez)
+                            solar_zenith, solar_azimuth, airmass)
     else:
         raise ValueError(f'invalid model selection {model}')
 
@@ -1303,7 +1302,7 @@ def _f(i, j, zeta):
 
 def perez_driesse(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
                   solar_zenith, solar_azimuth, airmass=None,
-                  model='allsitescomposite1990', return_components=False):
+                  return_components=False):
     '''
     Determine diffuse irradiance from the sky on a tilted surface using
     the continuous Perez-Driesse model.
@@ -1349,11 +1348,6 @@ def perez_driesse(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
         inputs. The kastenyoung1989 airmass calculation is used internally
         and is also recommended when pre-calculating airmass because
         it was used in the original model development.
-
-    model : string (optional, default='allsitescomposite1990')
-        A string which selects the desired set of Perez coefficients.
-        This argument is only for compatibility with the `perez` function.
-        If a value other than the default value is provided, it is ignored.
 
     return_components: bool (optional, default=False)
         Flag used to decide whether to return the calculated diffuse components
@@ -1405,11 +1399,6 @@ def perez_driesse(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
     king
     '''
     # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Oct., 2023
-
-    if model not in ('allsitescomposite1990', '1990'):
-        warnings.warn("The 'model' parameter is ignored. "
-                      "Only 'allsitescomposite1990' is available.",
-                      UserWarning)
 
     delta = _calc_delta(dhi, dni_extra, solar_zenith, airmass)
     zeta = _calc_zeta(dhi, dni, solar_zenith)
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -10,8 +10,10 @@ from functools import partial
 
 import numpy as np
 import pandas as pd
+from scipy.interpolate import splev
 
 from pvlib import atmosphere, solarposition, tools
+from pvlib.atmosphere import get_relative_airmass
 import pvlib  # used to avoid dni name collision in complete_irradiance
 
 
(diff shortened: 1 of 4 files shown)


adriesse:
Ok, you can run it now if you want!

I have made the airmass argument optional and could do the same in the original perez.  Opinions?

We have recommended an upper limit of 0.9 on F1 in our paper.  I think I should put it in the original perez too.  Opinions?

I will probably just copy the whole suite of tests for the original perez. Opinions?

cwhanse:
> We have recommended an upper limit of 0.9 on F1 in our paper. I think I should put it in the original perez too. Opinions?

Would F1 <= 0.9 be correcting an error or omission in the original paper, or is this a model improvement that has emerged from the recent work?

adriesse:
> > We have recommended an upper limit of 0.9 on F1 in our paper. I think I should put it in the original perez too. Opinions?
> 
> Would F1 <= 0.9 be correcting an error or omission in the original paper, or is this a model improvement that has emerged from the recent work?

Quoting from the paper:

>While preparing Figure 2 we noticed that the value of F1
could exceed 1.0 for zenith angles less than 30°, which would
be physically incoherent and produce a negative irradiance for
the isotropic sky component. However, in our validation data
sets, we observed only isolated cases with F1 over 0.7, and for
one equatorial location (the BSRN station DWN) the natural
limit was approximately 0.9. Consequently, there appears to
be a low likelihood of reaching 1.0. Nevertheless, to guarantee
physically coherent transposition results, we recommend that a
limit of 0.9 be imposed on F1 in implementations of both the
original and the continuous models

cwhanse:
Let's modify the `perez` function in its own PR. It would be nice to have (in GH comments, if not more formally) a comparison of model output before and after the change, to answer "how much difference does this make"

adriesse:
Ok, the main code is a ready.  I will not touch the existing perez function.  

Comments on whether to duplicate the whole test suite are still welcome.

cwhanse:
Share my thoughts: the implementation changes relatively few lines from `irradiance.perez`. My reaction was "can this be a kwarg-activated variant in `irradiance.perez`?" However, I view the Perez-Driesse calculation as distinct from the original Perez model, rather than a variation on the Perez model. The two share a conceptual structure, but differ (importantly) in their parameterization. For this reason, I think it is better to have both `irradiance.perez` and `irradiance.perez_driesse` in the API.

adriesse:
The duplicated code in the two function bothers me a bit as well.  I would say a future move of the common code to a helper function could be considered, which would not affect the API.

adriesse:
Ok, I've copied all tests but one from perez and added some specific ones for perez_driesse.  

Ready for complete review!

pvlib/irradiance.py
```diff
@@ -351,7 +355,8 @@ def get_total_irradiance(surface_tilt, surface_azimuth,
         the list of accepted values.
     model : str, default 'isotropic'
         Irradiance model. Can be one of ``'isotropic'``, ``'klucher'``,
-        ``'haydavies'``, ``'reindl'``, ``'king'``, ``'perez'``.
+        ``'haydavies'``, ``'reindl'``, ``'king'``, ``'perez'``,
+        ``'perez-driesse'``.
```
cwhanse, line 359: This might be the only instance of a hyphenated value. Pointing it out in case there's some reason, I don't know, to avoid this.

pvlib/irradiance.py
```diff
@@ -1212,6 +1225,233 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+                  solar_zenith, solar_azimuth, airmass=None,
+                  model='allsitescomposite1990', return_components=False):
+    '''
+    Determine diffuse irradiance from the sky on a tilted surface using
+    the continuous Perez-Driesse model.
+
+    The Perez-Driesse model [1]_ is a reformulation of the 1990 Perez
+    model [2]_ that provides continuity of the function and of its first
+    derivatives.  This is achieved by replacing the look-up table of
+    coefficients with quadratic splines.  Seet Notes for more information.
```
cwhanse, line 1312:
```suggestion
    coefficients with quadratic splines.
```

pvlib/irradiance.py
```diff
@@ -1212,6 +1225,233 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    surface_azimuth : numeric
+        Surface azimuth angles in decimal degrees. surface_azimuth must
+        be >=0 and <=360. The azimuth convention is defined as degrees
+        east of north (e.g. North = 0, South=180 East = 90, West = 270).
+
+    dhi : numeric
+        Diffuse horizontal irradiance in W/m^2. DHI must be >=0.
+
+    dni : numeric
+        Direct normal irradiance in W/m^2. DNI must be >=0.
```
cwhanse, line 1330: This isn't important: I don't think `DHI>=0` or `DNI>=0` are required. This code uses `np.where` and `np.clip` etc. and looks like it can't produce negative irradiance. The capitalization makes me wonder if this text is held over from the original Matlab code. If so, we should delete it.

pvlib/irradiance.py
```diff
@@ -1212,6 +1225,233 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+            * circumsolar
+            * horizon
+
+    Notes
+    -----
+    The Driesse-Perez model can be considered a plug-in replacement for the
+    1990 Perez model using the ``'allsitescomposite1990'`` coefficient set.
+    Deviations between the two are very small, as demonstrated in [1]_.
+    Support for other coefficient sets is not provided because the 1990
+    set is based on the largest and most diverse set of empirical data.
```
cwhanse, line 1384:
```suggestion
    Other coefficient sets for the 1990 Perez model are not supported because the 1990
    set is based on the largest and most diverse set of empirical data.
```

cwhanse (approved):
I suggest that we add a See Also block to the `perez` function.

pvlib/irradiance.py
adriesse: Negative irradiance would also affect delta and zeta (or epsilon), so I think the requirement is ok.

pvlib/irradiance.py
adriesse: Scipy has a variety of hyphenated method identifiers.

adriesse:
> I suggest that we add a See Also block to the `perez` function.

There are several little things that could be mirrored from this function to the old one, I think.  A new issue and PR might be best for this.

pvlib/irradiance.py
```diff
@@ -1212,6 +1225,235 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    '''
+    dhi = np.asarray(dhi)
+    dni = np.asarray(dni)
+
+    with np.errstate(invalid='ignore'):
+        zeta = dni / (dhi + dni)
+
+    zeta = np.where(dhi == 0, 0.0, zeta)
+
+    if use_kappa:
```
kandersolar, line 1260: I guess `use_kappa=False` switches back to the 1987 model?  Unless I missed it, the reference makes no mention of the kappa term being optional in the Perez-Driesse model.  I suggest we remove the `use_kappa` parameter and make the calculations in this function structured more similarly to the equations in the reference.

kandersolar (commented):
One comment about the `zeta` calculation below but otherwise LGTM!  There's now also a merge conflict.

pvlib/irradiance.py
adriesse: The option is left over from development and I thought it might be interesting for other experimenters.  I can take it out if you want. Not sure what you mean by the last bit though.

kandersolar: The last bit meant that it took me a few moments to mentally rework Eqs 6, 7, and 13 in the reference to match the calculations being done here.  Not a big deal :)

If you'd rather keep the code the way it is now, maybe a brief explanatory comment would save future readers the minute or two it took me to convince myself I understood this function.

pvlib/irradiance.py
```diff
@@ -1212,6 +1225,235 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    klucher
+    reindl
+    king
+    '''
+    # Contributed by Anton Driesse (@adriesse), PV Performance Labs. Oct., 2023
+
+    if model not in ('allsitescomposite1990', '1990'):
+        warnings.warn("The 'model' parameter is ignored. "
+                      "Only 'allsitescomposite1990' is available.",
+                      UserWarning)
```
AdamRJensen, line 1410: I have a mild preference for not having the `model` parameter in the Perez-Driesse model (and state in the docs that it's developed based on `allsitescomposite1990`).

If future model options are developed, it would be easy to add them (but this is not foreseen at the moment).

pvlib/irradiance.py
```diff
@@ -1212,6 +1225,235 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    diffuse_components : OrderedDict (array input) or DataFrame (Series input)
+        Keys/columns are:
+            * sky_diffuse: Total sky diffuse
+            * isotropic
+            * circumsolar
+            * horizon
+
+    Notes
+    -----
+    The Driesse-Perez model can be considered a plug-in replacement for the
```
AdamRJensen, line 1380:
```suggestion
    The Perez-Driesse model can be considered a plug-in replacement for the
```

AdamRJensen (approved):

pvlib/irradiance.py
```diff
@@ -1212,6 +1225,235 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
+    Compute the delta parameter, which represents sky dome "brightness"
+    in the Perez and Perez-Driesse models.
+
+    Helper function for perez_driesse transposition.
+    '''
+    if airmass is None:
+        # use the same airmass model as in the original perez work
+        airmass = get_relative_airmass(solar_zenith, 'kastenyoung1989')
+
+    max_airmass = get_relative_airmass(90, 'kastenyoung1989')
```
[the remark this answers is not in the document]
adriesse: I'm curious what motivates this one...

pvlib/irradiance.py
adriesse: This was motivated by the "plug-in" compatibility claim.  If the upper classes can live without it, so can I.

pvlib/irradiance.py
adriesse: I will simplify this code but keep the direct calculation.

pvlib/irradiance.py
adriesse: Any more input from those who use and maintain the classes?

adriesse:
I'm happy to entertain more comments and reviews, but if there are none in the pipeline I would suggest merging so I can continue with things that depend on this.

adriesse:
@kandersolar @cwhanse Regarding the calculations in _calc_zeta, do you think I should add these equations in the journal paper (if still possible).

pvlib/irradiance.py
kandersolar: Regarding the classes: it seems `ModelChain` has no provision for specifying alternative parameter sets, but `PVsystem` allows `model` to be specified via **kwargs by the user.  No big deal to not have `model` for the P-D model in my view.

I suppose I'd lean towards removing `model`, for simplicity and so that those who use alternative parameter sets (which I speculate is rather uncommon) are forced to consider the consequences of switching to the P-D model.

kandersolar:
> Regarding the calculations in _calc_zeta, do you think I should add these equations in the journal paper (if still possible).

I like the paper the way it is now.  I think the code comments are sufficient to link these calculations to the paper.

cwhanse:
> @kandersolar @cwhanse Regarding the calculations in _calc_zeta, do you think I should add these equations in the journal paper (if still possible).

Yes that would be of value.

adriesse:
I get a merge conflict on those two opinions.  :)  I'll see if it can be done when @AdamRJensen gets back from Nepal.

pvlib/irradiance.py
adriesse: Ok, I'll remove it.

adriesse:
Are any further changes needed before merging?

kandersolar:
Thanks @adriesse! (and sorry for needing a reminder to merge this)

Issue #974, opened 2020-05-22
Speed up Brent how-methods in single-diode using Scipy Cython Optimize


mikofski:
### Problem
As described in #971 `singlediode.bishop88_i_from_v()` and `singlediode.bishop88_v_from_i()` can be very slow using how `method=brentq`

### Solution
Scipy made changes to access Brent using Cython which can be compiled without the GIL and with `prange`, a Cython command that parallelizes long loops. Compiled code is already typically very fast, so this could result in 10-100x speed improvements. Here's [a gist with a Jupyter notebook benchmark](https://gist.github.com/mikofski/ac30065073d0d32d6ea3569f6e24e5ec) that demonstrates about 30x reduction in runtime, but [YMMV](https://en.wiktionary.org/wiki/your_mileage_may_vary).

### Alternatives
1. An alternative to Brent is Newton which is already vectorized in Scipy using NumPy arrays. A Newton how-method is already available in `singlediode` but Newton is unbounded, and so it's not guaranteed to converge like Brent, which is a bounded search method.
2. write a custom vectorized bisection method - actually there already is one, I think Rob Andrews may have authored it in `pvlib.tools`: [`_golden_sect_DataFrame`](https://github.com/pvlib/pvlib-python/blob/e04d959042e8dda9440a07dc9709e35b3f479200/pvlib/tools.py#L362), I think it's a golden search algorithm, not as fast as Brent, but still very good. Should we test it? Might need modification to work because I think it was specifically written for solving LambertW problems.
3. is there a [numba](http://numba.pydata.org/) way to solve this problem?
4. is there a [pythran](https://pythran.readthedocs.io/en/latest/) solution to this?
5. wait til next summer and make this a GSoC project :wink:

**Additional context**
There might be some unpleasant hurdles:
- how do you maintain cython code in your repo?
- this should work fine for conda-forge & pvlib Anaconda channels, but what about PyPI users? We don't have a binary distribution on PyPI,
- The closest experience we've had is with the NREL SPA algorithm, which we've sort of abandoned because the NumPy & numba versions are so much faster and easier
- would we consider making binaries for deployment on multiple platforms like Windows (win32) & Mac OS X (darwin)
- would we make Linux users build their own wheels or figure out how to get involved with [manylinux](https://github.com/pypa/manylinux)?

kanderso-nrel:
[...]

PR #1884, opened 2023-10-06
Add "logwright" SDE solver functions

### LINKED ISSUE #1856: Add an SDE solver using the "LogWright" function
The `lambertw` method has to take some pains to appropriately handle numerical overflow in an intermediate calculation.  

An implementation of [1,2], which reformulates the SDE so that it uses a different form of the Lambert W equation and does not require calculating such large intermediate values.  

Additionally, according to [3] and also my own brief testing, it is also computationally faster than the `lambertw` method.

**Additional context**
[1] "On Calculating the Current-Voltage Characteristic of Multi-Diode Models for Organic Solar Cells": https://arxiv.org/abs/1601.02679
[2] "A Robust Approximation to a Lambert-Type Function": https://arxiv.org/abs/1504.01964
[3] "Solar Cells, Lambert W and the LogWright Functions": https://arxiv.org/abs/2307.08099

(I notice also that @markcampanelli is mentioned in the acknowledgements of [2])

Here is a rough implementation of the method, which I observe to be 2x faster than `_lambertw_v_from_i` while staying within $\pm$ 1e-10 of its output.

```python
def _g(x, n=2):
    y = np.where(x <= -np.e, x, 1)
    y = np.where((-np.e < x) & (x < np.e), -np.e + (1 + np.e) * (x + np.e) / (2 * np.e), y)
    np.log(x, out=y, where=x >= np.e)
    
    for _ in range(n):
        ey = np.exp(y)
        ey_plus_one = 1 + ey
        y_ey_x = y + ey - x
        y = y - 2 * (y_ey_x) * (ey_plus_one) / ((2 * (ey_plus_one)**2 - (y_ey_x)*ey))
    
    return y


def _logwright_v_from_i(current, photocurrent, saturation_current,
                        resistance_series, resistance_shunt, nNsVth):
    output_is_scalar = all(map(np.isscalar,
                               (current, photocurrent, saturation_current,
                                resistance_series, resistance_shunt, nNsVth)))

    I, IL, I0, Rs, Rsh, a = \
        np.broadcast_arrays(current, photocurrent, saturation_current,
                            resistance_series, resistance_shunt, nNsVth)

    x = np.log(I0 * Rsh / a) + Rsh * (-I + IL + I0) / a
    V = -I * Rs - a * np.log(I0 * Rsh / a) + a * _g(x)

    if output_is_scalar:
        return V.item()
    else:
        return V
```

kandersolar:
- [x] Closes #1856
 - [ ] Tests added

This PR implements two new functions in `pvlib.singlediode`: `_logwright_v_from_i` and `_logwright_i_from_v`.  The reference from #1856 ([link](https://arxiv.org/pdf/1601.02679.pdf)) provides the `v_from_i` equation.  I used the same technique to derive the `i_from_v` equation.  I plan to document the equations in our `singlediode.rst` page.

The new base functions are then integrated into the rest of the SDE solving machinery.  Here are some runtimes for the `pvsystem.i_from_v`, `v_from_i`, and `singlediode` functions comparing `method='lambertw'` with `method='logwright'`:

image

### DIFF
--- a/docs/sphinx/source/user_guide/singlediode.rst
+++ b/docs/sphinx/source/user_guide/singlediode.rst
@@ -49,6 +49,8 @@ Then the module current can be solved using the Lambert W-function,
        - \frac{n Ns V_{th}}{R_s} W \left(z \right)
 
 
+TODO
+
 Bishop's Algorithm
 ------------------
 The function :func:`pvlib.singlediode.bishop88` uses an explicit solution [4]
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -2405,7 +2405,8 @@ def singlediode(photocurrent, saturation_current, resistance_series,
 
     method : str, default 'lambertw'
         Determines the method used to calculate points on the IV curve. The
-        options are ``'lambertw'``, ``'newton'``, or ``'brentq'``.
+        options are ``'lambertw'``, ``'logwright'``, ``'newton'``, or
+        ``'brentq'``.
 
     Returns
     -------
@@ -2440,6 +2441,8 @@ def singlediode(photocurrent, saturation_current, resistance_series,
     If the method is ``'lambertw'`` then the solution employed to solve the
     implicit diode equation utilizes the Lambert W function to obtain an
     explicit function of :math:`V=f(I)` and :math:`I=f(V)` as shown in [2]_.
+
+    If the method is ``'logwright'`` TODO
 
     If the method is ``'newton'`` then the root-finding Newton-Raphson method
     is used. It should be safe for well behaved IV-curves, but the ``'brentq'``
@@ -2482,8 +2485,8 @@ def singlediode(photocurrent, saturation_current, resistance_series,
             resistance_shunt, nNsVth)  # collect args
     # Calculate points on the IV curve using the LambertW solution to the
     # single diode equation
-    if method.lower() == 'lambertw':
-        out = _singlediode._lambertw(*args, ivcurve_pnts)
+    if method.lower() in ['lambertw', 'logwright']:
+        out = _singlediode._lambertw(*args, ivcurve_pnts, how=method)
         points = out[:7]
         if ivcurve_pnts:
             ivcurve_i, ivcurve_v = out[7:]
@@ -2651,8 +2654,8 @@ def v_from_i(current, photocurrent, saturation_current, resistance_series,
         0 < nNsVth
 
     method : str
-        Method to use: ``'lambertw'``, ``'newton'``, or ``'brentq'``. *Note*:
-        ``'brentq'`` is limited to 1st quadrant only.
+        Method to use: ``'lambertw'``, ``'logwright'``, ``'newton'``,
+        or ``'brentq'``. *Note*: ``'brentq'`` is limited to 1st quadrant only.
 
     Returns
     -------
@@ -2668,6 +2671,8 @@ def v_from_i(current, photocurrent, saturation_current, resistance_series,
             resistance_series, resistance_shunt, nNsVth)
     if method.lower() == 'lambertw':
         return _singlediode._lambertw_v_from_i(*args)
+    elif method.lower() == 'logwright':
+        return _singlediode._logwright_v_from_i(*args)
     else:
         # Calculate points on the IV curve using either 'newton' or 'brentq'
         # methods. Voltages are determined by first solving the single diode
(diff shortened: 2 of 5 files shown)


kandersolar:
@cwhanse I wonder if you can provide some insight on the calculation result differences I am seeing, relative to the values from `method='lambertw'`.  In particular the identified vmp and its downstream values are much less accurate than the other values, so I suspected something with the golden section search may be the cause.  However, setting a stricter tolerance in the golden section search doesn't make much difference.  I also looked at the differences from `logwright` to `newton` and `lambertw` to `newton`, and they're all about the same as this:

image

cwhanse:
@kandersolar you are plotting Vmp (logWright) - Vmp (lambertw)?

What happens if you increase the iterations in `tools._logwrightomega`?

cwhanse:
The curves in the `precise_iv_curve*.json` files are precise (to 15 significant figures), maybe we can benchmark the three solvers against those (lambertw, log Wright, netwon). Sorry if the format is inconvenient.

kandersolar:
> @kandersolar you are plotting Vmp (logWright) - Vmp (lambertw)?

Yep. 

> What happens if you increase the iterations in tools._logwrightomega?

No meaningful change.  I also tried switching to scipy's implementation of the Wright Omega function, and no change there either.  

> The curves in the precise_iv_curve*.json files are precise (to 15 significant figures), maybe we can benchmark the three solvers against those (lambertw, log Wright, netwon).

Good idea!  I will try that.

kandersolar:
Here is a comparison of the various methods, calculating error relative to the "Case 1" known values in https://github.com/cwhanse/ivcurves/tree/main/ivcurves/test_sets:

image

So newton and brentq do quite well, but the two lambertw-style approaches have some trouble finding the MPP.  

Putting aside logwright for the moment, I think I have narrowed the MPP issue down to the golden section search function performing poorly at the peak of the P-V curve, which is relatively flat and broad.  Specifically, P varies by only an ULP or two (~10^-15) over a voltage range of order 10^-7, so unsurprisingly the golden section search has trouble navigating that region.  Here is a plot showing the P-V curve and comparing identified MPPs (horizontal lines show a 1-ULP step, since matplotlib refuses to show tick labels at such a small scale):

image

Just to see if another search method might work better, I swapped scipy's powell in place of our golden section search and got somewhat better convergence in this single example, but still not as good as the bishop/brentq method.  I am curious how it is that brentq seems to handle this problem so reliably.  

Anyway, this is starting to sound like a bug report for `_golden_sect_DataFrame`.  Perhaps this should be a new issue?

cwhanse:
Nicely laid out. Your findings are consistent with my own assessments.

The lambert W methods find the maximum of the (V, P) curve. The bishop88 methods find the zero of the dP/dV curve. I'm not sure the difference is a bug in the usual sense. When that approach was taken, I'm sure we thought 8 digits of accuracy sufficed, and may not have realized the shortcoming in that approach.

adriesse:
I seem to recall another discussion about (vectorized?) root finding recently, but I can't remember where.  Could someone point me to it?

cwhanse:
#1661

kandersolar:
Or #974.

> If the blue line above from the precise IV curve, would it be possible to make it even more precise, or perhaps more ideal, by removing the dithering so it steps up on the left and down on the right of Pmp?

Shame on me for not labeling it!  The blue line in that plot was from `lambertw_i_from_v`.  The precise curves would presumably not suffer from such numerical artifacts, but they also doesn't have nearly the voltage resolution needed to show them anyway.

adriesse:
> Or #974.

That's the one, thanks!

mikofski:
there was a similar [issue in pvmismatch](https://github.com/SunPower/PVMismatch/pull/68) which was using `argmax` to find the max power point (MPP), but in issue 65 there were discontinuities/jumps similar to the issue described for LambertW approach. A simple solution was to calculate the derivative around MPP using central difference and interpolate to find a better zero.

echedey-ls:
Mmmm, can I be of help with this PR? I see the discussion revolves around the precision of each algorithm, but that doesn't invalidate the fact that logwright is still faster (if I understood correctly).

kandersolar:
The convergence error discussed above is large enough to make me uncomfortable, but probably not too large to keep the method from being useful.  The other thing is that only the `v_from_i` equation has a reference; I derived the `i_from_v` equation myself.  

This PR currently shows that the logwright method has promise, but IMHO a short paper defining the complete method and comparing speed+accuracy with other methods needs to be written before this PR gets picked up again.

Issue #1403, opened 2022-02-13
More common terms for solar zenith and solar azimuth


cwhanse:
In pvlib, two solar zenith quantities are of interest: true (geometric) zenith angle, and apparent (refraction corrected) zenith angle. Currently, `zenith`, `solar_zenith` and `apparent_zenith` all appear in different places, and the meaning of each isn't consistent throughout. For example, [solarposition/spa_c](https://github.com/pvlib/pvlib-python/blob/95e25fd7a258ae97bcffde947c0980fa3a737536/pvlib/solarposition.py#L131) uses `zenith` (true) and `apparent_zenith` (refraction-corrected), and `Location` methods expect these terms. But both `apparent_zenith` and `solar_zenith` are used in [SingleAxisTracker](https://github.com/pvlib/pvlib-python/blob/0fd5c43acc8a90972bf5018d3c07b6ad8b93a5fd/pvlib/tracking.py#L13) methods (the latter without being specific in the docstring), and in places where I think refraction-corrected zenith should be expected, e.g., [PVSystem.get_aoi](https://github.com/pvlib/pvlib-python/blob/6c857ae64fe46959ea0518be30c01929e3dc2be7/pvlib/pvsystem.py#L315).

For solar azimuth, there is not counterpart to true vs. refraction-corrected. But for the solar azimuth, both `azimuth` and `solar_azimuth` are being used.
 
**Describe the solution you'd like**

1. Agreement on the pvlib meaning of `zenith` and `solar_zenith`. `apparent_zenith`'s meaning is clear. The other two terms leave room for uncertainty.

2. Agreement on one term to use for solar azimuth: `azimuth` or `solar_azimuth`.

**Describe alternatives you've considered**
It makes sense to have a term that is not specific to true or apparent zenith. Perhaps `zenith` for this use, and `solar_zenith` specifically for true zenith?

mikofski:
Thanks for raising this issue. I have noticed the various terms usage as well, and have been careful to read the docs (especially in the irradiance module) to make sure to use what I think is supposed to be the correct intended input.

[...]

PR #1904, opened 2023-11-06
Projected zenith convenience function

### LINKED ISSUE #1734: make projected solar zenith a convenience function
>A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

As mentioned in https://github.com/pvlib/pvlib-python/pull/1725#discussion_r1190000002 the projected solar zenith (PSZ) is used more than once, and may benefit from having a standalone function for reuse to remove redundant code.

>A clear and concise description of what you want to happen.

a new function perhaps in tools or tracking like this:
```python
sp = pvl.solarposition.get_solarposition(times, latitude, longitude, ...)
proj_zenith = np.degrees(np.arctan2(
    np.sin(np.radians(sp.azimuth))*np.sin(np.radians(sp.apparent_zenith)),
    np.cos(np.radians(sp.apparent_zenith))))
```

>A clear and concise description of any alternative solutions or features you've considered.

For linear shade calculation in #1725 the user is required to calculate PSZ externally.

**Additional context**
>Add any other context or screenshots about the feature request here.

one version of the PSZ calculation is currently in `tracking.py` based on the solar vector as seen in the orientation of the tracker axes, that's why axis tilt appears in this "transformed" solar vector:
```python
xp = x*cos_axis_azimuth - y*sin_axis_azimuth
zp = (x*sin_axis_tilt*sin_axis_azimuth
          + y*sin_axis_tilt*cos_axis_azimuth
          + z*cos_axis_tilt)
```
https://github.com/pvlib/pvlib-python/blob/275e6718caf7486cb5b7dcf29acd59499ad51f7f/pvlib/tracking.py#L432

echedey-ls:
- [x] Closes #1734 
 - [x] Tests added

Adds a function that calculates the projected solar zenith.

Pending tasks:
 - [x] Add paper references
 - ~[ ] Add diagram?~
 - [x] Generate test data from `singleaxis`, change it's implementation to use `projected_solar_zenith_angle` and change tests accordingly.

<details><summary>Test data generation script for the PSZ implementation port from `singleaxis` to `projected_solar_zenith_angle`</summary>
<p>

```python
# %%
import pvlib
import pandas as pd

from datetime import timedelta, timezone

axis_tilt_angle = 12.224
axis_azimuth_angle = 187.2

times = pd.date_range(
    # these limits avoid NaN results in singleaxis
    "2024-01-25 08:40",
    "2024-01-25 18:20",
    freq="20min", tz=timezone(timedelta(hours=1))
)
solar_pos = pvlib.solarposition.get_solarposition(times, 40.4165, -3.70256, 700)

tracker_data = pvlib.tracking.singleaxis(
    solar_pos["apparent_zenith"],
    solar_pos["azimuth"],
    axis_tilt=axis_tilt_angle,
    axis_azimuth=axis_azimuth_angle,
    backtrack=False,
)

psz_angle = pvlib.shading.projected_solar_zenith_angle(
    axis_tilt=axis_tilt_angle,
    axis_azimuth=axis_azimuth_angle,
    solar_zenith=solar_pos["apparent_zenith"],
    solar_azimuth=solar_pos["azimuth"],
)

# %%
tracker_thetas = tracker_data["tracker_theta"]

# %%
# optional, just to remove singleaxis nans
psz_angle = psz_angle[tracker_thetas.notna()]
tracker_thetas = tracker_thetas[tracker_thetas.notna()]

# %%
psz_angle == tracker_thetas

# %%
tracker_thetas.to_numpy()

``` 

</p>
</details> 

### Documentation links
[Link to function](https://pvlib-python--1904.org.readthedocs.build/en/1904/reference/generated/pvlib.shading.projected_solar_zenith_angle.html).


kandersolar:
> about the diagram, should I replicate it? 

Such a diagram might be helpful, but no need in this PR IMHO.  I'm not sure where it would live anyway.  I don't think we currently have any diagrams in function-level documentation.  I guess it could be in a gallery example (or a new User's Guide page) but unless it improves on the references, I'd say let's not bother :)

> may be better to write pure NumPy python so anyone can numba-decorate it 

Is there a reason this function needs JIT more than the rest of pvlib does?  We use `sind` et al for code readability, so unless this function is special somehow, I think we should stick with that convention.  (as an aside, I'd be surprised if numba couldn't handle `pvlib.tools.sind`, but I don't know for sure either)

> About the tests, do you have any data handy for that 

Slope-Aware Backtracking has a small validation dataset in a table at the end.  It's not enough on its own for this PR, but it's at least a start, and can be easily extended in at least one way (subtract 180 from `axis_azimuth` and negate the expected PSZAs).  I'm not aware of any other existing datasets to use.  The `tracker_theta` output of `pvlib.tracking.singleaxis` with `backtrack=False` and `max_angle=180` can be used to calculate expected values for additional test cases.

echedey-ls:
> subtract 180 from axis_azimuth and negate the expected PSZAs

I can't make this work. Could you double check if that's supposed to happen?

```python
    # test by changing array azimuth
    psz = psz_func(
        array_tilt,
        array_azimuth-180,
        timedata["Apparent Elevation"],
        timedata["Solar Azimuth"],
    )
    assert_allclose(psz, -timedata["True-Tracking"], atol=1e-3)
```

kandersolar:
>  > subtract 180 from axis_azimuth and negate the expected PSZAs
>
> I can't make this work

My mistake, I forgot those test cases used a nonzero `axis_tilt`.  Negate the axis tilt as well and I think that little trick should work.

echedey-ls:
I've left out whether the result is valid like `pvlib.tracking.singleaxis` [does](https://github.com/pvlib/pvlib-python/blob/4e55d500821847b1876c443cbaef4ecc4fe0b05e/pvlib/tracking.py#L175C1-L175C1).

pvlib/shading.py
```diff
@@ -232,3 +232,64 @@ def sky_diffuse_passias(masking_angle):
...
+    r"""
+    Calculate projected solar zenith angle in degrees.
+
+    This is common in track and shadow computation [1]_ [2]_ [3]_.
+
+    Parameters
+    ----------
+    surface_tilt : numeric
+        Axis tilt angle in degrees. From horizontal plane to array plane.
+    surface_azimuth : numeric
```
kandersolar, line 248: All of the `surface_*` in this function should be `axis_*`; `surface` refers to orientation of a PV module while `axis` refers to the rotation axis around which a module rotates.

pvlib/shading.py
```diff
@@ -232,3 +232,64 @@ def sky_diffuse_passias(masking_angle):
...
+        Axis azimuth angle in degrees.
+        North = 0°; East = 90°; South = 180°; West = 270°
+    solar_apparent_elevation : numeric
+        Sun's apparent elevation in degrees.
```
kandersolar, line 252: Let's switch to taking `solar_zenith` as input, and calculate elevation internally.  Also no need for `apparent` in the parameter name (#1403)

pvlib/tests/test_shading.py
```diff
@@ -104,3 +107,87 @@ def test_sky_diffuse_passias_scalar(average_masking_angle, shading_loss):
...
+        timedata["Apparent Elevation"],
+        timedata["Solar Azimuth"],
+    )
+    assert_allclose(psz, timedata["True-Tracking"], atol=1e-3)
+    # test equivalence against pvlib.tracking.singleaxis
+    singleaxis = pvlib.tracking.singleaxis(timedata["Apparent Zenith"],
+                                           timedata["Solar Azimuth"],
+                                           axis_tilt, axis_azimuth,
+                                           backtrack=False)
+    assert_allclose(psz, singleaxis["tracker_theta"])
```
kandersolar, line 157: It's true that `singleaxis` currently implements the same calculation, but the idea is to follow this PR with another that has `singleaxis` call this new function instead of implementing the same calculations.  So testing against a call to `singleaxis` makes sense right now, but pretty soon it won't make sense (it will essentially just test this new function against itself).

I think what we should do is use `singleaxis` to calculate some test values today, and hardcode those values into these tests.

kandersolar (commented):
A few initial notes

pvlib/tests/test_shading.py
echedey-ls: Ah! That makes sense, I feel stupid for not copying that implementation from the start.

pvlib/shading.py
echedey-ls: Was [`variables_style_rules.csv`](https://github.com/pvlib/pvlib-python/blob/4e55d500821847b1876c443cbaef4ecc4fe0b05e/pvlib/data/variables_style_rules.csv#L23C1-L27) updated according to that conversation?
I see

```
surface_tilt;tilt angle of the surface
surface_azimuth;azimuth angle of the surface
solar_zenith;zenith angle of the sun in degrees
apparent_zenith;refraction-corrected solar zenith angle in degrees
solar_azimuth;azimuth angle of the sun in degrees East of North
```
At least `solar_zenith` suggests me it's true solar zenith, since apparent is mentioned right after and it does not have "refraction-corrected" in the definition.
I've kept "apparent" in the parameter description just in case.

pvlib/tests/test_shading.py
echedey-ls: Wouldn't it be better to just copy it? I mean, it already works and is commented.

pvlib/shading.py
```diff
@@ -232,3 +232,66 @@ def sky_diffuse_passias(masking_angle):
        Available at https://www.nrel.gov/docs/fy18osti/67399.pdf
     """
     return 1 - cosd(masking_angle/2)**2
+
+
+def projected_solar_zenith_angle(axis_tilt, axis_azimuth,
+                                 solar_zenith, solar_azimuth):
+    r"""
+    Calculate projected solar zenith angle in degrees.
```
adriesse, line 240: Could we have description in terms of a vector and a surface, which I'm guessing the projection is all about?

pvlib/tests/test_shading.py
kandersolar: The code seems essentially the same between the two implementations, so I don't think it matters.  The new one might be slightly easier to trace to the reference.  The comments would be nice to copy over though (assuming they are still correct!)

pvlib/shading.py
```diff
@@ -232,3 +232,66 @@ def sky_diffuse_passias(masking_angle):
...
+
+    This is common in track and shadow computation [1]_ [2]_ [3]_.
+
+    Parameters
+    ----------
+    axis_tilt : numeric
```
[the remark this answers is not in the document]
echedey-ls: There's a comment up there from @kandersolar asking for changes from `surface_*` to `axis_*` prefixes. I think `surface` is a close synonym of `plane`? So I'm unsure of applying that suggestion.

I'll request reviews for the current reviewers hoping they also give some insight regarding this. Anyway, if you are sure of it, I change it without no problems.

pvlib/shading.py
```diff
@@ -232,3 +232,84 @@ def sky_diffuse_passias(masking_angle):
...
     """
     return 1 - cosd(masking_angle/2)**2
+
+
+def projected_solar_zenith_angle(axis_tilt, axis_azimuth,
+                                 solar_zenith, solar_azimuth):
+    r"""
+    Calculate projected solar zenith angle in degrees.
+    This is the solar zenith angle projected onto the tracker rotation plane or
+    the plane defined by its normal vector and the azimuth.
```
cwhanse, line 242:
```suggestion
    This solar zenith angle is projected onto the plane whose normal vector is defined by
    ``axis_tilt`` and ``axis_azimuth``.  The normal vector is in the direction of ``axis_azimuth``
    (clockwise from north) and tilted from horizontal by ``axis_tilt``. See Figure 5 in [1]_.
```

One picture would be worth many words here.

pvlib/shading.py
```diff
@@ -232,3 +232,84 @@ def sky_diffuse_passias(masking_angle):
...
+    -------
+    Projected_solar_zenith : numeric
+        In degrees.
+
+    References
+    ----------
+    .. [1] K. Anderson and M. Mikofski, 'Slope-Aware Backtracking for
+       Single-Axis Trackers', National Renewable Energy Lab. (NREL), Golden,
+       CO (United States); Det Norske Veritas Group, Oslo (Norway),
+       NREL/TP-5K00-76626, Jul. 2020. :doi:`10.2172/1660126`.
```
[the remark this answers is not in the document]
echedey-ls: Thanks for letting me know!

pvlib/shading.py
```diff
@@ -232,3 +232,84 @@ def sky_diffuse_passias(masking_angle):
...
+    )
+    # The ideal tracking angle wid is the rotation to place the sun position
+    # vector (xp, yp, zp) in the (x, z) plane, which is normal to the panel and
+    # contains the axis of rotation.  wid = 0 indicates that the panel is
+    # horizontal. Here, our convention is that a clockwise rotation is
+    # positive, to view rotation angles in the same frame of reference as
+    # azimuth. For example, for a system with tracking axis oriented south, a
+    # rotation toward the east is negative, and a rotation to the west is
+    # positive. This is a right-handed rotation around the tracker y-axis.
+    # Eq. (5); angle between sun's beam and surface
```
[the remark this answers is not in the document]
echedey-ls: Makes sense, my bad.

pvlib/shading.py
echedey-ls: Thanks for upgrading the definition.

Regarding copyright, is it possible to add that Figure 5 here (if you consider it a good addition)? Or is it better to do something new by myself? I had previously started doing a diagram:

image

It still lacks some vectors and angles, and IMO has an excess of details, anyway. Just in case it is a starting point.

cwhanse: We can add that Figure here, with a citation. I think your version with the colored planes would be easier to read but: 1) theta should be an angle in the purple plane, it looks different here 2) indicate the plane's normal vector 3) indicate the solar zenith angle 4) show the tilt and azimuth relative to a N-E-S-W coordinate system 5) label the sun, and 6) add direction to the angles so the sign convention is clear.

pvlib/shading.py
adriesse: Sorry, I think it's not clear.  I think you have to say at least you're projecting onto a vertical plane. (Or maybe that's only for the non-tracking case?)

pvlib/shading.py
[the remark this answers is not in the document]
kandersolar: Is the issue that `axis_tilt` doesn't suggest that the equations are equally applicable to fixed tilt arrays?

I'm inclined (pun intended) to keep the term `axis_tilt` here.  I think it's a useful, although perhaps not obvious at first, point of view to consider FT as a subset/special case of SAT anyway, in which case `axis_tilt` is the relevant quantity even for fixed tilt.  

Perhaps a good compromise is to clarify that point in the parameter description/Notes?

pvlib/shading.py
echedey-ls: Yes, I see my figure is completely wrong for this purpose. I've got some issues regarding the comprehension of this PSZ concept. If you are ok with it, I'll be adding said Figure 5.

echedey-ls: Image added!

https://pvlib-python--1904.org.readthedocs.build/en/1904/reference/generated/pvlib.shading.projected_solar_zenith_angle.html

pvlib/shading.py
echedey-ls: Did you check the last change to the definition proposed by Cliff? IIRC, there wasn't much time between the time I made the change and your comment ([docs](https://pvlib-python--1904.org.readthedocs.build/en/1904/reference/generated/pvlib.shading.projected_solar_zenith_angle.html)).

Anyway, I'm pretty sure (although my conception may be wrong), the projection plane isn't vertical for all cases. Figure 5 from [1] (in the docs too) may be of help.

If you still think the definition can be improved, can you state some concrete changes? I've got some difficulties seeing where to improve it.

adriesse: After reading a bit more I realize that it's not always projecting onto a vertical plane.

echedey-ls: Glad to hear that! I'll close this conversation then 😉

pvlib/shading.py
```diff
@@ -232,3 +232,76 @@ def sky_diffuse_passias(masking_angle):
...
+
+def projected_solar_zenith_angle(axis_tilt, axis_azimuth,
+                                 solar_zenith, solar_azimuth):
+    r"""
+    Calculate projected solar zenith angle in degrees.
+
+    This solar zenith angle is projected onto the plane whose normal vector is
+    defined by ``axis_tilt`` and ``axis_azimuth``. The normal vector is in the
+    direction of ``axis_azimuth`` (clockwise from north) and tilted from
+    horizontal by ``axis_tilt``. See Figure 5 in [1]_:
```
adriesse, line 245: Does that mean the axis of tracker rotation ***is*** the surface normal vector we're talking about?

echedey-ls: From Figure 5 caption:
> AB shows a tracker axis with its tracker rotation plane defined by the 𝒕𝒙 and 𝒕𝒛 axes.

I'm pretty sure it is. Do you think this line of text is worth to be added to the current caption? I didn't want to copy-paste the paper content.

adriesse: How about: 
```suggestion
    This solar zenith angle is projected onto the plane whose normal vector is the torque tube? 
```
 That puts a picture in my mind!

echedey-ls: That's more focused on tangible parts, I like it.

Nonetheless, I'm a bit hesitant to change that since there is [ongoing discussion](https://github.com/pvlib/pvlib-python/pull/1904#discussion_r1396262134) regarding the tracking or fixed support nomenclature. And that suggests SAT.

pvlib/shading.py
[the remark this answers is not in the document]
echedey-ls: I feel like these two terms refer to pure maths, and mapping those words to concise real-world systems is a bit counter-productive due to the variability. Is there some generic -maths- term we can use for a general description of this, a projection? Like `basis`, maybe? There's always a beautiful open paper backing the procedure and explaining what it does thoroughly. How the tracker works is up to the user or the other functions.

Just to weight in some more options outside of the current frame:
`tracker_*` (but this suggests everything except FT, right?), `collector_*`, `reference_*`, `system_*`, `projection_system_*`, `coordinate_system_*`, `reference_system_*`, `basis_*`, `projection_basis_*`... whatever other constructs you may think of.

> Perhaps a good compromise is to clarify that point in the parameter description/Notes?

I don't dislike this possibility if we don't agree on anything else eventually, but we are overcomplicating the docs IMHO.

BTW, I'm +1 for the `basis_*`, `projection_basis_*` options. Anyway, remember I have 0 experience in PV outside of this repo.

pvlib/tests/test_shading.py
```diff
@@ -104,3 +107,155 @@ def test_sky_diffuse_passias_scalar(average_masking_angle, shading_loss):
...
+        axis_azimuth - 180,
+        timedata["Apparent Zenith"],
+        timedata["Solar Azimuth"],
+    )
+    assert_allclose(psz, -timedata["True-Tracking"], atol=1e-3)
+
+    # test implementation port from tracking.singleaxis
+    axis_tilt, axis_azimuth, singleaxis = \
+        singleaxis_psz_implementation_port_data
+    psz = pvlib.tracking.singleaxis(
```
kandersolar, line 224: I'm not really convinced that this kind of  "port" test is necessary.  In theory, the existing correctness tests for `tracking.singleaxis` should be sufficient.  (if they aren't, then of course we should supplement them!)  

If we do need the "port" test, it should be in `test_tracking.py` rather than `test_shading.py`.

pvlib/tests/test_shading.py
```diff
@@ -104,3 +107,155 @@ def test_sky_diffuse_passias_scalar(average_masking_angle, shading_loss):
     for angle, loss in zip(average_masking_angle, shading_loss):
         actual_loss = shading.sky_diffuse_passias(angle)
         assert np.isclose(loss, actual_loss)
+
+
+@pytest.fixture
+def true_tracking_angle_and_inputs_NREL():
```
kandersolar, line 113: This test is good for normal inputs.  I think it would be good to complement it with another test that checks edge cases.  Here are some test edge cases that I came up with (please verify that they make sense!):

```python
df = pd.DataFrame(
    data=[
     [0, 0, 0, 0, 0],
     [0, 180, 0, 0, 0],
     [0, 0, 0, 180, 0],
     [0, 180, 0, 180, 0],
     [45, 0, 0, 180, 0],
     [45, 90, 0, 180, -45],
     [45, 270, 0, 180, 45],
     [45, 90, 90, 180, -90],
     [45, 270, 90, 180, 90],
     [45, 90, 90, 0, 90],
     [45, 270, 90, 0, -90],
     [45, 45, 90, 180, -135],
     [45, 315, 90, 180, 135]],
    columns=['solar_zenith', 'solar_azimuth', 'axis_tilt', 'axis_azimuth', 'psza']
)
```

pvlib/shading.py
```diff
@@ -232,3 +232,76 @@ def sky_diffuse_passias(masking_angle):
        Available at https://www.nrel.gov/docs/fy18osti/67399.pdf
     """
     return 1 - cosd(masking_angle/2)**2
+
+
+def projected_solar_zenith_angle(axis_tilt, axis_azimuth,
+                                 solar_zenith, solar_azimuth):
```
kandersolar, line 238: Can we swap the order around to `solar_zenith, solar_azimuth, axis_tilt, axis_azimuth` to match `pvlib.tracking.singleaxis`?

pvlib/shading.py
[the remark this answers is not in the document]
kandersolar: I admit to suffering from tunnel vision for trackers, but in my view `axis_` is still the best suggested prefix thus far.  I think no matter what prefix we choose, we're going to have to include some kind of explanation/clarification.  So IMHO we might as well choose `axis_` for consistency with the reference and other pvlib functions. 

How about:

```
    axis_tilt : numeric
        Tilt from horizontal of the coordinate axis pointing along the tracker
        torque tube (or along the length of a fixed-tilt row). [degrees]
```

Similar for `axis_azimuth`.  And then a Notes section:

```
    Notes
    -----
    Although the cited reference derives the equations in the context of
    single-axis tracking arrays, the math is also applicable to fixed-tilt
    arrays.  For example, a south-facing fixed-tilt array on level
    terrain would have ``axis_tilt=0`` and ``axis_azimuth=90``.
```

kandersolar (commented):
Sorry for the slow review here @echedey-ls.  I think this is getting close!

pvlib/tests/test_shading.py
echedey-ls: I forgot I left that there ☠️

echedey-ls:
@kandersolar , GitHub does not allow me to reply to your last review message, so here it goes:

> we might as well choose ``axis_`` for consistency with the reference 

I agree it's a good point. It's difficult to understand exactly the vectors and the projection plane without reading it. In fact, such a recommendation would be a nice addition the notes section.

> we're going to have to include some kind of explanation/clarification

I also agree with that. But regarding the suggested explanations, I think it would be better to describe the use cases with real examples in general. If I didn't already know about the internals and purposes, I wouldn't understand why `the math is also applicable to fixed-tilt arrays` since the paper only talks about SATs and the optimal rotation angle. I would ask myself what would be the rotation angle in a fixed-tilt array. My suggestion, feel free to edit as much as yall want:

```
    Notes
    -----
    This projection has a variety of applications in PV. For example:

    - Given a single-axis tracker, a plane whose normal vector's
      ``axis_azimuth`` equals the torque tube direction and its ``axis_tilt``
      is the zero rotation angle, this function yields the optimal angle to
      point to in order to maximize direct irradiance capture (this is called
      :ref:`true-tracking
      <sphx_glr_examples_plot_single_axis_tracking.py:True-tracking>`).

    - Given rows of PV arrays (or other sunlight obstacles), a projection plane
      that contains the module's normal vector (i.e., its ``axis_azimuth`` is
      the module's normal vector's azimuth shifted ``-90º``) and its
      ``axis_rotation`` is zero, the result is the smallest angle between the
      shadow plane and a horizontal plane. 
      See :py:func:`pvlib.shading.tracker_shaded_fraction`.
```

I wouldn't change the current arguments definitions (almost due to the 1:1 relation with the paper). Maybe move your suggested clarifications to the notes section, although I'm not sure of how to integrate them.

I'd like to clear up that I don't have a problem applying your suggestions straight forward if you want to, specially due to my lack of experience in PV and the language barrier. And that I hope this comment helps improving the quality of the PR and doesn't add noise to the conversation.

echedey-ls:
I'd like to move forward this PR another bit, specially the docstring. Would you like me to push @kandersolar suggestions, or give another suggestions? Feel free to request anything from my side.

kandersolar:
Thanks @echedey-ls for keeping this going :) 

Describing some real use cases is a great idea, and I like your proposed text.  What do you think about a combination of `Notes` and `Examples` like this?

```
    Notes
    -----
    This projection has a variety of applications in PV. For example:

    - Projecting the sun's position onto the plane perpendicular to
      the axis of a single-axis tracker (i.e. the plane
      whose normal vector coincides with the tracker torque tube)
      yields the tracker rotation angle that maximizes direct irradiance
      capture.  This tracking strategy is called
      :ref:`true-tracking
      <sphx_glr_examples_plot_single_axis_tracking.py:True-tracking>`.

    - Self-shading in large PV arrays is often modeled by assuming
      a simplified 2-D array geometry where the sun's position is
      projected onto the plane perpendicular to the PV rows.
      The projected zenith angle is then used for calculations
      regarding row-to-row shading.

    Examples
    --------

    Calculate the ideal true-tracking angle for a horizontal north-south
    single-axis tracker:

    >>> rotation = projected_solar_zenith_angle(solar_zenith, solar_azimuth,
    >>>                                         axis_tilt=0, axis_azimuth=180)

    Calculate the projected zenith angle in a south-facing fixed tilt array
    (note: the ``axis_azimuth`` of a fixed-tilt row points along the length
    of the row):

    >>> psza = projected_solar_zenith_angle(solar_zenith, solar_azimuth,
    >>>                                     axis_tilt=0, axis_azimuth=90)
```

echedey-ls:
I like that pretty much, thanks for the improvements. I will add the still inexistent `pvlib.shading.tracker_shaded_fraction` to the see also section and the notes section in its PR (#1962). Just in case anyone heard of the PSZA for the calculation of shadows so that s/he finds the upcoming feature.

You can check the docs out [here](https://pvlib-python--1904.org.readthedocs.build/en/1904/reference/generated/pvlib.shading.projected_solar_zenith_angle.html).

echedey-ls:
I had some trouble finding the example's link. I can add a section link too but reading it all seems a better choice.

From my side, this looks like it can be merged!

kandersolar (approved):
One trivial suggestion so that the image doesn't take up so much webpage space, but otherwise LGTM!

kandersolar:
Thanks for this contribution and the patience @echedey-ls :)

PR #1923, opened 2023-12-01
Maximum power point tracking for mismatched devices (singlediode model)

markcampanelli:
- [ ] Closes #xxxx
 - [ ] Tests added

I would like some early feedback on this idea and code.


mikofski:
Thanks Mark!  I think there’s an issue for this, if I recall correctly it’s off MPP inverter operation or something.

planted-jason-hoving:
@markcampanelli thanks for pointing me to this! It generally works well for me but I noticed a couple corner cases that need to be handled differently.

With ~zero irradiance at night the photocurrent drops to zero and the shunt resistance goes to inf, or at least that is the case for the CEC model. During the optimization search `pvsystem.v_from_i` will in turn use `singlediode._lambert2w_v_from_i` and that results in a log(negative number) error as it tries current values less than zero. Could be fixed with constraints on the `minimize` call. What I did for now was just return zero power if any photocurrent is zero, but I know there might be some cases where people want to still drive current through a shaded panel or something.

At very low irradiance, even with the constraints above added, I am getting some convergence issues on the scipy call. I was able to resolve it by adding a tolerance (say 1e-6A) that was fine for me.

PR #1940, opened 2023-12-22
Add Huld PV model

cwhanse:
~~- [ ] Closes #xxxx~~
 - [x] Tests added

Adds Huld's model for DC power. This model is used in PVGIS.


pvlib/pvsystem.py
```diff
@@ -2845,6 +2845,80 @@ def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+    r"""
+    Power (DC) using the Huld model.
+
+    The Huld model [1]_ is given by
+
+
+    .. math::
+
+        P_{dc} = G` ( P_{dc0} k_1 log(G') + k_2 log^2 (G') + k_3 T' log(G') +
+                 k_5 T' log^2 (G') + k_6 T'^2)
```
kandersolar, line 2869:
```suggestion
        P_{dc} = G' ( P_{dc0} + k_1 \log(G') + k_2 \log^2 (G') + k_3 T' +
                 k_4 T' \log(G') + k_5 T' \log^2 (G') + k_6 T'^2)
```

pvlib/pvsystem.py
```diff
@@ -2845,6 +2845,80 @@ def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+        DC power. [W]
+
+    Raises
+    ------
+    ValueError
+        If neither `k` nor `cell_type` are specified.
+
+    References
+    ----------
+    .. [1] T. Huld, G. Friesen, A. Skoczek, K. R. Kenny, T. Sample, M. Field,
```
kandersolar, line 2902:
```suggestion
    .. [1] T. Huld, G. Friesen, A. Skoczek, R. Kenny, T. Sample, M. Field,
```

pvlib/pvsystem.py
```diff
@@ -2845,6 +2845,80 @@ def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+
+
+    .. math::
+
+        P_{dc} = G` ( P_{dc0} k_1 log(G') + k_2 log^2 (G') + k_3 T' log(G') +
+                 k_5 T' log^2 (G') + k_6 T'^2)
+
+        G' = \frac{G_{poa eff}}{1000}
+
+        T' = T_{mod} - 25 \degree C
```
kandersolar, line 2873:
```suggestion
        T' = T_{mod} - 25 °C
```

I think mathjax doesn't have the degree symbol.  I think using unicode should be okay here?

docs/sphinx/source/whatsnew/v0.10.4.rst
```diff
@@ -7,6 +7,7 @@ v0.10.4 (Anticipated March, 2024)
 
 Enhancements
 ~~~~~~~~~~~~
+* Added the Huld PV model (:pull:`XXXX`)
```
kandersolar, line 10:
```suggestion
* Added the Huld PV model (:pull:`1940`)
```

pvlib/pvsystem.py
```diff
@@ -2845,6 +2845,80 @@ def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+
+
+    Parameters
+    ----------
+    effective_irradiance : numeric
+        The irradiance that is converted to photocurrent. [W/m^2]
+    temp_mod: numeric
+        Module back-surface temperature [C].
```
AdamRJensen, line 2881:
```suggestion
        Module back-surface temperature. [C]
```

pvlib/pvsystem.py
```diff
@@ -2845,6 +2845,80 @@ def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+    ------
+    ValueError
+        If neither `k` nor `cell_type` are specified.
+
+    References
+    ----------
+    .. [1] T. Huld, G. Friesen, A. Skoczek, K. R. Kenny, T. Sample, M. Field,
+           E. Dunlop. A power-rating model for crystalline silicon PV modules.
+           Solar Energy Materials and Solar Cells 95, (2011), pp. 3359-3369.
+           10.1016/j.solmat.2011.07.026.
```
AdamRJensen, line 2905:
```suggestion
           :doi:`10.1016/j.solmat.2011.07.026`.
```
I think this is the correct formatting of DOIs

pvlib/pvsystem.py
```diff
@@ -2845,6 +2845,80 @@ def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+    Parameters
+    ----------
+    effective_irradiance : numeric
+        The irradiance that is converted to photocurrent. [W/m^2]
+    temp_mod: numeric
+        Module back-surface temperature [C].
+    pdc0: numeric
+        Power of the modules at 1000 W/m^2 and cell reference temperature. [W]
+    k : tuple, optional
+        Length 6. If `k` is not provided, `cell_type` must be specified.
```
AdamRJensen, line 2885:
```suggestion
    k : tuple, optional
        Empirical coefficients of the relative efficiency used as input to the
        module power model. Length 6. If `k` is not provided, `cell_type`
        must be specified.
```

pvlib/pvsystem.py
```diff
@@ -2845,6 +2845,80 @@ def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+                   'CdTe': (-0.046689, -0.072844, -0.002262, 0.000276,
+                            -0.000159, -0.000006)}
+    return huld_params[cell_type]
+
+
+def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
+    r"""
+    Power (DC) using the Huld model.
+
+    The Huld model [1]_ is given by
```
AdamRJensen, line 2863:
```suggestion
    The Huld model [1]_ is used by PVGIS and is given by
```

AdamRJensen (commented):
I think it will be great addition having the Huld/PVGIS model available!

pvlib/pvsystem.py
cwhanse: Technically, this form uses `k` as coefficients of a power model. I'll modify the suggestion, thanks.

adriesse:
FYI: This and other efficiency models are also available in https://github.com/adriesse/pvpltools-python.

cwhanse:
Would `pvsystem.pvgis` be more widely recognized that `huld`?

mikofski:
> Would `pvsystem.pvgis` be more widely recognized that `huld`?

I’m not fan of pvgis as a subheading, module, subpackage, namespace.

also sometimes I use `$25^{\circ}$`

pvlib/tests/test_pvsystem.py
```diff
@@ -2549,3 +2549,25 @@ def test_Array_temperature_missing_parameters(model, keys):
...
+    res = pvsystem.huld(100, 30, pdc0, k=(1, 1, 1, 1, 1, 1))
+    exp_100 = 0.1 * (pdc0 + np.log(0.1) + np.log(0.1)**2 + + 5 + 5*np.log(0.1)
+                     + 5*np.log(0.1)**2 + 25)
+    assert np.isclose(res, exp_100)
+    # Series input
+    eff_irr = pd.Series([1000, 100])
+    tm = pd.Series([25, 30])
+    expected = pd.Series([pdc0, exp_100])
+    res = pvsystem.huld(eff_irr, tm, pdc0, k=(1, 1, 1, 1, 1, 1))
+    assert_allclose(res, expected)
```
kandersolar, line 2570: This checks that the values are correct, but doesn't require that `res` is a Series.  Use `assert_series_equal` (`from .conftest import assert_series_equal`) instead?

adriesse:
> Would `pvsystem.pvgis` be more widely recognized that `huld`?

I think so, in the same way `pvsyst` is more recognizable than... what's his name again?

AdamRJensen:
> > Would `pvsystem.pvgis` be more widely recognized that `huld`?
> 
> I think so, in the same way `pvsyst` is more recognizable than... what's his name again?

I'm inclined to disagree as I wouldn't be surprised if pvgis one day switched to another PV performance model.

mikofski:
Desoto is a woman

adriesse:
> Desoto is a woman

?? Mermoud is a man

kandersolar:
@cwhanse I noticed that the current implementation doesn't exhibit the temperature dependence that I expect.  Here's the temperature coefficient of power at STC:
```python
pdc0 = 300
dT = 0.1
gamma = (huld(1000, 25 + dT/2, pdc0, cell_type='cSi') - huld(1000, 25 - dT/2, pdc0, cell_type='cSi')) / dT / pdc0
print(gamma)  # -1.56733e-05
```

I think this is because the `k` coefficient values from the [PVGIS website](https://joint-research-centre.ec.europa.eu/photovoltaic-geographical-information-system-pvgis/getting-started-pvgis/pvgis-data-sources-calculation-methods_en) are for a different form of Huld's equation which uses them differently from the 2011 paper.  Compare:

- PVGIS website: $P(G, T) = G' \cdot A \cdot \text{eff}_\text{nom} \cdot (1 + k_1 \log G' + k_2 \log^2 G' + ...)$
- Huld et al. 2011: $P(G, T) = G' \cdot (P_\text{stc} + k_1 \log G' + k_2 \log^2 G' + ...)$

The two versions of $k_n$ differ by a factor of $P_\text{stc} = A \cdot \text{eff}_\text{nom}$.  Right now we're using the website's coefficients with the paper's equation, thus the strange behavior.  We need to choose which of the two forms to adopt here.  Not sure which one is better.

AdamRJensen:
It would be nice, as a minimum, to be able to replace the website/API calculations.

cwhanse:
> The two versions of $k_n$ differ by a factor of $P_\text{stc} = A \cdot \text{eff}_\text{nom}$. 

Good catch. I noticed the notation difference $k$ and $k'$ in the paper but overlooked what it meant.

> Not sure which one is better.

I prefer Huld's paper, since the PVGIS technical document is harder to locate.

cwhanse:
> It would be nice, as a minimum, to be able to replace the PVGIS website/API calculations.

I agree, and my attempt fails. I don't know how to go about diagnosing where this code diverges from PVGIS.

```
from pvlib.iotools import pvgis
from pvarray import huld
from pvlib.temperature import faiman


res = pvgis.get_pvgis_hourly(45, 2, components=False, surface_tilt=35,
                             usehorizon=False, pvcalculation=True,
                             peakpower=1, pvtechchoice='crystSi')

data = res[0]

geff = data['poa_global']
ta = data['temp_air']
ws = data['wind_speed']
pdc = data['P']

tmod = faiman(geff, ta, ws, u0=26.9, u1=6.2)
expected = huld(geff, tmod, 1000, cell_type='csi')

print(pdc[:16])
print(expected[:16])
```

kandersolar:
IAM and SMM, I'm guessing?  Their [docs](https://joint-research-centre.ec.europa.eu/photovoltaic-geographical-information-system-pvgis/getting-started-pvgis/pvgis-data-sources-calculation-methods_en) say PVGIS uses Martin-Ruiz for IAM, so at least that is easily recreated.  However, the spectral model (SPECMAGIC?) looks decidedly nontrivial to recreate, at least from a quick skim.  https://doi.org/10.3390/rs4030622

kandersolar:
Indeed including IAM gets a lot closer:

image

Code:
<details>

```python
from pvlib.iotools import pvgis
from pvarray import huld
from pvlib.temperature import faiman
from pvlib.iam import martin_ruiz


# use surface_tilt = 0 so that zenith = aoi.  Means we don't have to compute solar position for IAM
df, _, meta = pvgis.get_pvgis_hourly(45, 2, components=True, surface_tilt=0,
                                     usehorizon=False, pvcalculation=True,
                                     peakpower=1, pvtechchoice='crystSi')

# %%

fig, axes = plt.subplots(1, 2, sharex=True, sharey=True, figsize=(8, 4))

for (iam, label), ax in zip([(1, "no IAM"),
                             (martin_ruiz(90 - df['solar_elevation']), "with Martin-Ruiz")], axes):
    smm = 1
    geff = smm * (iam * df['poa_direct'] + df['poa_sky_diffuse'] + df['poa_ground_diffuse'])
    ta = df['temp_air']
    ws = df['wind_speed']
    pdc = df['P']
    
    tmod = faiman(geff, ta, ws, u0=26.9, u1=6.2)
    expected = huld(geff, tmod, 1000, cell_type='csi')
    
    ax.scatter(expected, pdc, s=1)
    ax.axline((0, 0), slope=1, c='k', ls=':')
    ax.set_title(label)

axes[0].set_xlim(0, 1000)
axes[0].set_ylim(0, 1000)
axes[0].set_ylabel('PVGIS $P_{dc}$')
for ax in axes:
    ax.set_xlabel('pvlib $P_{dc}$')

fig.tight_layout()
```

</details>

cwhanse:
The reference (Huld's 2010 paper) defines G as "in-plane irradiance W/m2" and later, in its discussion of differences between model and measurement, discusses reflections ("the irradiance should be corrected for this effect before applying it to the power model") and spectral effects (no recommendation given). The paper states that the irradiance values used to derive the k' values for cSi modules (Table 7 in the paper, same as in the PVGIS documentation) are from a cSi reference cell. So, I agree - in this model, "G" is effective irradiance, not POA irradiance which is what I assumed, and what I believe the PVGIS API is returning.

cwhanse:
@AdamRJensen I don't think we can reproduce the PVGIS API's results, since the spectral adjustment is not exposed nor documented in detail (it appears that PVGIS has an internal look-up table of sorts for spectral adjustment factors). The scope of `pvarray.huld` is only the conversion of `effective_irradiance` and `cell_temp` to `pdc`. Perhaps a full replication of PVGIS could be realized later, with something like `ModelChain.with_pvwatts`. Agree?

AdamRJensen:
> @AdamRJensen I don't think we can reproduce the PVGIS API's results, since the spectral adjustment is not exposed nor documented in detail (it appears that PVGIS has an internal look-up table of sorts for spectral adjustment factors). The scope of `pvarray.huld` is only the conversion of `effective_irradiance` and `cell_temp` to `pdc`. Perhaps a full replication of PVGIS could be realized later, with something like `ModelChain.with_pvwatts`. Agree?

Agreed! I imagine that some future users will ask why they don't match. So I suggest adding an admonition stating that the PVGIS API also applies a spectral correction.

pvlib/pvarray.py
```diff
@@ -223,3 +223,122 @@ def adr_wrapper(xdata, *params):
...
+
+    Raises
+    ------
+    ValueError
+        If neither ``k`` nor ``cell_type`` are specified.
+
+    Notes
+    -----
+    The equation for :math:`P_{dc}` is from [1]_. The expression used in PVGIS
+    documentation is differs by factoring :math:`P_{dc0}` out of the
```
kandersolar, line 290:
```suggestion
    documentation differs by factoring :math:`P_{dc0}` out of the
```

kandersolar (approved):

adriesse:
A thought for consideration:  As Huld notes in his paper, the model can also be expressed as an efficiency model.  A function ```pvefficiency_huld()``` would be a nice complement to the existing functions

pvlib/pvarray.py
```diff
@@ -223,3 +223,122 @@ def adr_wrapper(xdata, *params):
...
+
+    Parameters
+    ----------
+    effective_irradiance : numeric
+        The irradiance that is converted to photocurrent. [W/m^2]
+    temp_mod: numeric
+        Module back-surface temperature. [C]
+    pdc0: numeric
+        Power of the modules at 1000 W/m^2 and cell reference temperature. [W]
+    k : tuple, optional
```
adriesse, line 270: Most model functions have separate arguments for each parameter.  Are there pros and cons?

pvlib/pvarray.py
```diff
@@ -223,3 +223,122 @@ def adr_wrapper(xdata, *params):
...
+    different cell types. The parameters :math:`k'` differ from the parameters
+    :math:`k` expected by :py:func:`huld` by the factor ``pdc0``, that is,
+
+    .. math::
+
+        k = P_{dc0} k'
+
+    :py:func:`huld` is a component of the PV performance model implemented in
+    PVGIS. Among other components, the full PVGIS model includes:
+        - the Faiman model for cell temperature
```
adriesse, line 308: Does the model use cell or module temperature?

pvlib/pvarray.py
```diff
@@ -223,3 +223,122 @@ def adr_wrapper(xdata, *params):
...
+    """
+    if k is None:
+        if cell_type is not None:
+            k = _infer_k_huld(cell_type, pdc0)
+        else:
+            raise ValueError('Either k or cell_type must be specified')
+
+    gprime = effective_irradiance / 1000
+    tprime = temp_mod - 25
+    # accomodate gprime<=0
```
adriesse, line 336: Technically logG tends to -inf as G tends to 0.

pvlib/pvarray.py
cwhanse: My personal preference is to use tuples for polynomial coefficients. It would be tedious to list k1, k2, etc. and define each as the "nth of 6 coefficients..."

pvlib/pvarray.py
```diff
@@ -223,3 +223,122 @@ def adr_wrapper(xdata, *params):
...
+                           0.000170, 0.000005),
+                   'cis': (-0.005554, -0.038724, -0.003723, -0.000905,
+                           -0.001256, 0.000001),
+                   'cdte': (-0.046689, -0.072844, -0.002262, 0.000276,
+                            0.000159, -0.000006)}
+    k = tuple([x*pdc0 for x in huld_params[cell_type.lower()]])
+    return k
+
+
+def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
```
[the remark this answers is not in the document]
cwhanse: I'm not in favor of this renaming. Efficiency can be easily calculated, and this function can be wired into `ModelChain` in the same way that other DC models are.

pvlib/pvarray.py
cwhanse: Module. Good catch.

cwhanse, line 308:
```suggestion
        - the Faiman model for module temperature
```

pvlib/pvarray.py
kandersolar: One consideration: ModelChain's ability to infer models based on supplied parameters relies on the models having distinct sets of parameter names.  `k1`, `k2` etc seems less likely to run into future conflict than a single `k` does.  But neither seems particularly likely to me.

I also favor a single `k`.  The individual values are not very meaningful on their own (IMHO), so a single bundled parameter seems appropriate to me.

pvlib/pvarray.py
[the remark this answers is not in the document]
adriesse: Wasn't there was some discussion about avoiding function names consisting only of an author's name?

pvlib/pvarray.py
```diff
@@ -223,3 +223,122 @@ def adr_wrapper(xdata, *params):
...
+
+
+    Parameters
+    ----------
+    effective_irradiance : numeric
+        The irradiance that is converted to photocurrent. [W/m^2]
+    temp_mod: numeric
+        Module back-surface temperature. [C]
+    pdc0: numeric
+        Power of the modules at 1000 W/m^2 and cell reference temperature. [W]
```
adriesse, line 269:
```suggestion
        Power of the modules at reference conditions 1000 W/m^2 and 25°C. [W]
```

pvlib/pvarray.py
```diff
@@ -223,3 +223,122 @@ def adr_wrapper(xdata, *params):
...
+
+
+    Parameters
+    ----------
+    effective_irradiance : numeric
+        The irradiance that is converted to photocurrent. [W/m^2]
+    temp_mod: numeric
+        Module back-surface temperature. [C]
+    pdc0: numeric
+        Power of the modules at reference conditions 1000 W/m^2 and 25°C. [W]
```
cwhanse, line 269:
```suggestion
        Power of the modules at reference conditions :math:`1000 W/m^2`
        and :math:`25^{\circ}C`. [W]
```

pvlib/pvarray.py
```diff
@@ -223,3 +223,122 @@ def adr_wrapper(xdata, *params):
...
+
+        G' &= \frac{G_{poa eff}}{1000}
+
+        T' &= T_{mod} - 25^{\circ}C
+
+
+    Parameters
+    ----------
+    effective_irradiance : numeric
+        The irradiance that is converted to photocurrent. [W/m^2]
```
cwhanse, line 265:
```suggestion
        The irradiance that is converted to photocurrent. [:math:`W/m^2`]
```

pvlib/pvarray.py
```diff
@@ -223,3 +223,123 @@ def adr_wrapper(xdata, *params):
...
+
+
+    Parameters
+    ----------
+    effective_irradiance : numeric
+        The irradiance that is converted to photocurrent. [:math:`W/m^2`]
+    temp_mod: numeric
+        Module back-surface temperature. [C]
+    pdc0: numeric
+        Power of the modules at reference conditions :math:`1000 W/m^2`
```
cwhanse, line 269:
```suggestion
        Power of the modules at reference conditions 1000 :math:`W/m^2`
```

kandersolar:
I notice that small but nonzero irradiance results in negative power values:

image

I'm not immediately seeing any provision for handling this in the reference.  We could constrain the output to be non-negative.  Either way I think it deserves a docstring note.

docs/sphinx/source/whatsnew/v0.10.4.rst
```diff
@@ -7,6 +7,7 @@ v0.10.4 (Anticipated March, 2024)
 
 Enhancements
 ~~~~~~~~~~~~
+* Added the Huld PV model (:pull:`1940`)
```
AdamRJensen, line 10:
```suggestion
* Added the Huld PV model used by PVGIS (:pull:`1940`)
```

pvlib/pvarray.py
```diff
@@ -296,6 +296,10 @@ def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
         P_{dc} = G' P_{dc0} (1 + k'_1 \log(G') + k'_2 \log^2 (G') + k'_3 T' +
                  k'_4 T' \log(G') + k'_5 T' \log^2 (G') + k'_6 T'^2)
 
+
+    Users should be aware that at very low irradiance, i.e.,
+    :math:`G' < 20 W/m^2`, :math:`P_{dc}` may be negative.
```
adriesse, line 301:
```suggestion
    :math:`G' < 20 W/m^2`, :math:`P_{dc}` may be negative.
    This is caused by the :math:`\log(G')` term, which tends toward 
    negative infinity as irradiance approaches zero.
```
I can provide a reference that discusses this, if it is of interest.

kandersolar:
Discussion has died down, so I assume no further changes are needed.  Thanks @cwhanse and reviewers!

PR #1941, opened 2023-12-25
Remove download badges from README

AdamRJensen:

There are currently two download badges on the pvlib python README page:
image

The first badge is from PyPi and the second from conda forge. Both badges show approximately the same number of downloads, but the first is per month and the second is totals - that seems inconsistent although I realize that they're different sources. I simply suggest to declutter the badge section and remove this downloads as they are dominated by automated tests downloading the packages and doesn't add any meaningful information/value (i.e., they say nothing about our actual user base).

There are already badges for both PyPi and conda forge at the top of the bad section in the "Latest release" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,11 @@
     <a href="https://pypi.org/project/pvlib/">
     <img src="https://img.shields.io/pypi/v/pvlib.svg" alt="latest release" />
     </a>
-    <a href="https://anaconda.org/conda-forge/pvlib-python">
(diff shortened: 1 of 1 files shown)


kandersolar (approved):
OK with me.

Tangentially related: can we updated all the conda-forge links in the first row to use `pvlib` instead of `pvlib-python`?  The latter is just for backwards compatibility, so better to direct people to the former.

Issue #1060, opened 2020-09-15
reevaluate numba minimum, usage patterns in spa.py


wholmgren:
`conda search -c main numba | grep py36` shows that numba 0.36.1 is the oldest package compatible with python 3.6 and numpy 1.12. The numba git history says 0.36.1 was released on Dec 7, 2017. (0.17.0 was released Feb 3, 2015.) I'll add that to the asv conf.

I'd be fine with also adding a minimum numba requirement to setup.py, but I think that should be done in combination with changes to the import logic in spa.py. In particular, I don't like the numpy fallback - that led to hard to track down errors when developing this. It would also be worth reviewing modern numba best practices. If we're adding a minimum numba requirement to setup.py then we should probably be testing against it too. More than I want to tackle in this PR.

_Originally posted by @wholmgren in https://github.com/pvlib/pvlib-python/pull/1059#discussion_r488775560_

wholmgren:
@kanderso-nrel looked into adding numba support to `temperature.fuentes` in #1037. Whatever we do in `spa.py` should simplify that process.

kanderso-nrel:
I think my hesitation in #1037 around numba was in part because the SPA code switches numba on/off based on what the user has requested (`how='numba'`), not just whether numba is installed or not.  I think it's valuable to be able to turn numba off somehow, but I don't think adding a `how` parameter to every numba-accelerated function is a good idea.  And although I like being able to disable numba, I would guess that not many people want to enable/disable numba for specific function invocations, and the extra complexity is hard to reason about, at least for me.  See also #401. 

If we're willing to give up the call-by-call numba control, I propose to enable or disable numba (by using a dummy `@jit` decorator) when pvlib is imported according to the behavior in this table.  The default is to use numba if it's available, but that can be overridden by an environment variable if needed:

| `PVLIB_USE_NUMBA` | numba installed? | result |
|-|-|-|
| no value | no | dummy |
| no value | yes | numba jit |
| 0 | no | dummy |
| 0 | yes| dummy |
| 1 | no | error |
| 1 | yes| numba jit |

Curious what people think about this.  Here's an example decorator definition for reference:

<details>
  <summary>Click to expand!</summary>

```python
import os
use_numba = os.getenv('PVLIB_USE_NUMBA', None)

def dummy_jit(func=None, *args, **kwargs):
    # accommodate using as either `@jit` or `@jit()`
    def wrapper(func):
        return func
    if func is not None:
        return func
    return wrapper

[...]

PR #1944, opened 2024-01-04
Simplify and future-proof numba import in spa.py

kandersolar:
- ~[ ] Closes #xxxx~
 - [ ] Tests added

`spa.py` currently inspects numba's version string to determine whether it meets the required minimum of 0.17.0:

https://github.com/pvlib/pvlib-python/blob/16017609ea2765c4651dbc13b4743578363b641a/pvlib/spa.py#L23-L40

However, the way it performs the check isn't quite correct.  Consider what will happen when `numba.__version__` eventually ticks over from `0.x.y` to `1.0.0`.  That code on line 32 will compare `int("1" + "0")` with 17 and incorrectly decide that the installed version of numba isn't new enough.  

Rather than fix the version check, I think it makes sense to just remove that check in `spa.py` altogether in favor of specifying the minimum in setup.py as usual.  Even that is probably unnecessarily cautious; numba 0.17.0 is ancient and I think it would be safe to assume that any version of numba that pvlib might encounter today would be much more recent than that, but no harm in setting the minimum anyway.


kandersolar:
To-do: make the same change to `has_numba` in `conftest.py`

pvlib/spa.py
```diff
@@ -21,23 +21,15 @@ def nocompile(*args, **kwargs):
 
 if os.getenv('PVLIB_USE_NUMBA', '0') != '0':
```
AdamRJensen, line 22:
```suggestion
if os.getenv('PVLIB_USE_NUMBA') is not None:
```

pvlib/spa.py
```diff
@@ -21,23 +21,15 @@ def nocompile(*args, **kwargs):
...
-        else:
-            warnings.warn('Numba version must be >= 0.17.0, falling back to ' +
-                          'numpy')
-            jcompile = nocompile
-            USE_NUMBA = False
+        jcompile = jit
+        USE_NUMBA = True
 else:
     jcompile = nocompile
     USE_NUMBA = False
```
AdamRJensen, line 35: Is there a reason for inconsistent usage of upper vs. lower case, e.g., `USE_NUMBA` vs. `has_numba`?

AdamRJensen (approved):
I'm all for simplifying things!

LGTM

pvlib/spa.py
kandersolar: The reason I see is is that `USE_NUMBA` is a module-level variable used inside a function far away from where it is defined while `has_numba` is used once in the same scope immediately after its definition.  But of course it's just style :)

pvlib/spa.py
kandersolar: I think this change would require some corresponding changes in the part of `solarposition.py` that controls how `spa.py` is imported (by setting this environment variable to `'0'` or `'1'`), so I'm inclined to leave this as-is for now.  It will be addressed someday with #1060.

AdamRJensen: I didn't realize that it could be defined as zero (I just assumed it would be undefined). Fine with me to just leave it.

AdamRJensen:
Thanks @kandersolar for your continuous commitment to keeping pvlib in good shape!

PR #1725, opened 2023-05-04
Linear shade gh1690

mikofski:
- [x] Closes #1689, #1690
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [x] Tests added
 - [x] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [x] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

add functions to get shaded fraction and calculate linear shade loss
@williamhobbs,

PR #1962, opened 2024-02-03
Shaded Fraction on Sloped Terrains and Linear Loss model - PR1725 continuation

### LINKED ISSUE #1689: Add shaded fraction (FS) calculation for true-tracking on flat terrain
It would be useful to have a shaded fraction calculation for true-tracking on flat terrain. It could be used in modeling thin film projects with (approximately) linear power loss due to self-shade.

image

An implementation of equations 1-6 in Lorenzo 2011 [1], where inputs are:
- tracker rotation angle (true-tracking, not backtracking)
- ground coverage ratio
- solar azimuth
- solar elevation

And the output is shaded fraction (0-1) of each shaded row.

Custom implementations would not be difficult, but a built-in function would be nice. More robust versions that can account for non-horizontal tracker rotation axis, non-flat terrain, etc., would be nice, but are more difficult.

**Additional context**
Here's the description we used in a 2023 PVRW poster on the topic [2], adapted from [1]:

Calculate the ideal tracker angle, 

`ω = arctan(cos(θA) * sin(θel) / sin(θA))`

where `θel` is solar elevation and `θA` is solar azimuth. Then calculate the shaded fraction of the array, `FS`, as 

`FS = max[0, (1-cos(ω)/GCR)]`

where `GCR` is the row spacing divided by tracker table width. 

[1] E. Lorenzo, L Narvarte, and J Muñoz. 2011. "Tracking and back-tracking". Progress in
Photovoltaics: Research and Applications 19: 747-753. https://doi.org/10.1002/pip.1085
[2] I. Azad, W. Hobbs, "Improved PV expected energy modeling with a simple self-shading model". NREL PVRW 2023. (proceedings pending)
[3] K. Anderson, W. Hobbs. "Improved CdTe PLR Estimates: Self-Shading and Spectral Mismatch". NREL PVRW, 2022. https://www.osti.gov/biblio/1846944

### LINKED ISSUE #1690: Add linear shade loss model for thin film
Related to https://github.com/pvlib/pvlib-python/issues/1689, it would be nice to have a linear self-shade power loss model. 

Once a shaded fraction is calculated, power loss can be calculated by modifying equation 6 from [1] and assuming that power loss is linear (as is done in SAM for thin film [2]).

For a linear shade loss, shade-corrected power, `Psc`, can be related to estimated power with no shade, `Pns`, with 

`Psc = Pns * (1 – FS * (1 – DF))`

where `DF` is diffuse fraction (I've used a fixed value, i.e., 0.2, but it could be estimated more precisely in each timestep using satellite data, dirint or gti_dirint, etc.[3]).

This approach works well when using measured (unshaded) POA irradiance.

...

**Additional context**
I'm not sure where the equation for `Psc` came from. I first implemented it in 2021, and more recently thought that it came directly from Lorenzo 2011 [1], but now I don't think that's the case. And I think SAM simply reduces beam irradiance for the linear shade model and already has diffuse irradiance separately, so does not need to calculate a diffuse fraction. 

This method does not explicitly account for reductions in diffuse irradiance due to row to row shade. 

This was presented on at PVRW 2023 [4].

[1] E. Lorenzo, L Narvarte, and J Muñoz. 2011. "Tracking and back-tracking". Progress in
Photovoltaics: Research and Applications 19: 747-753. https://doi.org/10.1002/pip.1085
[2] Gilman, P. et al.., (2018) SAM Photovoltaic Model Technical Reference Update. 93 pp.; NREL/TP-6A20-67399 
[3] K. Anderson, W. Hobbs. Improved CdTe PLR Estimates: Self-Shading and Spectral Mismatch. NREL PVRW, 2022. https://www.osti.gov/biblio/1846944 
[4] I. Azad, W. Hobbs, "Improved PV expected energy modeling with a simple self-shading model". NREL PVRW 2023. (proceedings pending)

echedey-ls:
- [x] Closes #1689
 - [x] Closes #1690
 - [x] Closes/substitutes #1725
 - [x] Tests added
 - [ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``). TODO.

Required actions to take before merging:
 - [ ] Rebase from main. Required since this branch has been rebased to add the PSZ feature. _Note for future self: I'm following [this guide](https://softwareengineering.stackexchange.com/questions/310427/in-github-flow-is-it-ok-to-base-feature-branch-on-another-feature-branch)._

See #1725 for context.

### Link to docs

Update when available.


kandersolar:
> Does, for example, a N-S SAT tracker azimuth follow the East - West direction or the torque tube? I would say the azimuth and elevation define the normal vector of any PV panel on it, but I'm not at all sure.

A distinction needs to be drawn between the angles that define the orientation of tracker and the angles that define the orientation of the modules on the tracker.  A single-axis tracker with N-S axis has constant `axis_azimuth=180` (or zero, but 180 is better).  The modules on the tracker will have `surface_azimuth=90` in the morning and `surface_azimuth=270`.

I will take a closer look at this PR in the coming days.  Thanks for your patience @echedey-ls!

echedey-ls:
Lot of thanks for clearing this up. Now I feel like its obvious.

Don't worry about the review. I know it's premature to open a PR based on another one (I've got yet another one in the drawer that needs this implementation to work, sorry 😬 )

williamhobbs:
I'm late jumping in here, but one thought I have: does this need to be limited to trackers? Or can the function name and variable names be updated to more generally cover fixed or tracking rows?

mikofski:
It should work for fixed tilt just fine, so not limited to trackers

williamhobbs:
Thanks, @mikofski. It seems like it would be helpful to change:
- the function name from `tracker_shaded_fraction` to `shaded_fraction`,
- the variables _tracker_tilt_ and _tracker_azimuth_ to _surface_tilt_ and _surface_azimuth_, and
- references in the docs to “trackers” to “rows”

or something like that.

mikofski:
Maybe I misspoke. Sorry, I meant to say that it *could* be repurposed for fixed tilt with some care, but, I’m not sure if we should make it all purpose for both fixed tilt & SAT. For one, tracker azimuth and surface azimuth mean different things, so we can’t just swap them.

I think it would work if we treat fixed tilt as a tracker aligned east to west and set the tracker azimuth to either 90° or 270° measured clockwise from north, or 90°from the surface azimuth.

Anyway I apologize for being too glib, I should have been more careful in my replies, again sorry.

echedey-ls:
I agree with @williamhobbs. `surface_tilt` and `surface_azimuth` terms seem to apply to all contexts independently of the used tracker, so they are less prone to be misunderstood by a user.

echedey-ls:
Rebase conflicts solved. Ready for review!

AdamRJensen:
> Reference in n

@echedey-ls could you update the code to use ``surface_x`` instead of ``tracker_``?

echedey-ls:
@AdamRJensen I'm pleased to do so. Should I change the function name too?

mikofski:
Beware, `trackers_x` doesn’t always mean the same as `surface_x`. 

`tracker_theta` = `surface_tilt` only if tracker axis tilt is zero.

`tracker_azimuth - 90°` = `surface_azimuth` in the morning 

`tracker_azimuth + 90°` = `surface_azimuth` in the afternoon

the terms are not interchangeable

kandersolar:
A few thoughts:
- I propose renaming `tracker_shaded_fraction` to `shaded_fraction1d`.  As pointed out above, the math is equally applicable to fixed-tilt arrays.  And there are many ways of calculating a "shaded fraction" of an array, many of which calculate a 2-D area fraction instead of a 1-D length fraction like this equation does.
- I think we need a more substantive reference.  For the shaded fraction, doesn't the Slope-Aware Backtracking report solve the same problem?  Is there another reference for the linear shade loss model?  The SAM technical reference describes a linear loss model along these lines, but it deals with lost irradiance, not lost power...
- We should not forget the question about whether the math uses left- or right-handed rotations.  I think we should strive for pvlib to use right-handed rotations everywhere.
- Just an FYI: @adamrjensen and I have a paper under review containing an improved shaded fraction equation that allows the tracker casting the shadow to be at a different rotation from the tracker receiving the shadow.  It won't be published in time for 0.10.4, and I don't want to hold up this PR in the mean time, but to me it makes sense to plan to extend this function with some additional optional parameters in the near future.

cwhanse:
> I propose renaming tracker_shaded_fraction to shaded_fraction1d. 

+1 here

> We should not forget the question about whether the math uses left- or right-handed rotations. I think we should strive for pvlib to use right-handed rotations everywhere.

++1 here.

adriesse:
> > We should not forget the question about whether the math uses left- or right-handed rotations. I think we should strive for pvlib to use right-handed rotations everywhere.
> 
> ++1 here.

Isn't that a bit hard with the way we use azimuth angles and positive heights above ground?

adriesse:
> > I propose renaming tracker_shaded_fraction to shaded_fraction1d.
> 
> +1 here

Perhaps the one dimension should be identified/named?

adriesse:
> > Perhaps the one dimension should be identified/named?
> 
> Perhaps! Any suggestions?

I'm probably not the best person to generate such a suggestions...

The doc says "The fraction of the collector *width* shaded by an adjacent row."  I guess we could raise the question here as well as to whether it should be width, length, height, or whatever--naming is hard.  I think something that suggests the shade line moves vertically would be best, maybe `shaded_fraction_lower`.

echedey-ls:
>  I think something that suggests the shade line moves vertically would be best

Then vertical / horizontal as you say; only these two (frontal surface) dimensions can be shaded. And to be honest, the horizontal shade effect is of less importance compared to the vertical, so I don't think it is a priority to include it for now. Maybe in the future. I will keep the proposed `shaded_fraction1d` name and update the docs according to this possibility. If we end up with a `shaded_fraction1d_horizontal` we can consider changing (with a deprecation) the former name.

@kandersolar @cwhanse I'm applying your suggestions slowly, I need to fix the tests with the `tracker_` -> `surface_` changes.

echedey-ls:
Ready for review 👀 

Few doubts, notes and ideas:
* I've changed the figure and added the reference of backtracking slope-aware, I hadn't seen the angle convention in [this file](https://github.com/pvlib/pvlib-python/blob/92c0e5bcfca717056971c27c556ff443d07d4e4c/docs/sphinx/source/_images/ground_slope_angle_convention.png). Since the implementation relies on the original Mikofski paper and notation, I've left it there with a comment.

* >  The SAM technical reference describes a linear loss model along these lines, but it deals with lost irradiance, not lost power...

    The most common pattern for models is to modify the incident irradiance, right? Since the irradiance to power conversion is completely linear as the model, I lean toward just documenting that it can be applied to any of those values, with the respective updates to doctest examples.

* > an improved shaded fraction equation that allows the tracker casting the shadow to be at a different rotation from the tracker receiving the shadow

     I'd love to facilitate future implementation of it. How do you feel about making `gcr` and/or `cross_axis_slope` keyword-only parameters? That way the code will be more readable and new parameters can have their place after solar position, which I find as the best choice. They can be appended to the end of the function too, but I dislike having orientation params so far from each other.


Link to [`shaded_fraction1d`](https://pvlib-python--1962.org.readthedocs.build/en/1962/reference/generated/pvlib.shading.shaded_fraction1d.html) docs.

williamhobbs:
> * I think we need a more substantive reference.  For the shaded fraction, doesn't the Slope-Aware Backtracking report solve the same problem?  Is there another reference for the linear shade loss model?

@kandersolar, just to clarify, you are saying that we need a substantive reference that says linear power loss models are valid for thin film modules. Is that right?

If so, I would guess that First Solar published something on it. I found this https://energy.sandia.gov/wp-content/gallery/uploads/26-Littmann-Tracking-and-Diffuse-Shading.pdf0_.pdf, but that's maybe not as substantive as would be ideal. Maybe https://ieeexplore.ieee.org/abstract/document/6744267 (I don't have free access) or something else by Ngan (Nelson), Panchula, or Littman?

adriesse:
> @kandersolar, just to clarify, you are saying that we need a substantive reference that says linear power loss models are valid for thin film modules. Is that right?

Evidence and illustrations for thin-film aSi are provided in this paper:
https://www.researchgate.net/publication/267261857_Non-Uniform_Conditions_and_Performance_in_Parallel-Only_and_SeriesParallel_Array_Configurations

The term "Linear loss" is another way of saying "no mismatch loss", it seems.

kandersolar:
> you are saying that we need a substantive reference that says linear power loss models are valid for thin film modules. Is that right?

What I'd really like to reference is a paper that deals with specifically the equation implemented in this PR: `power_loss_fraction = shaded_fraction1d * (1 - diffuse_fraction)`.  

>  Since the irradiance to power conversion is completely linear as the model

Only the simplest PV performance models assume that power is exactly linear with irradiance.  Here's a demonstration of how the CEC model is nonlinear for a First Solar thin film module:

<details>

```python
import pvlib
import numpy as np

cecdb = pvlib.pvsystem.retrieve_sam('cecmod')
module_params = cecdb['First_Solar__Inc__FS_6430']
module_params = module_params[['alpha_sc', 'a_ref', 'I_L_ref', 'I_o_ref', 'R_sh_ref', 'R_s', 'Adjust']]

effective_irradiance = np.linspace(0, 1200)
temperature_cell = np.array([25] * len(effective_irradiance))

sde_params = pvlib.pvsystem.calcparams_cec(effective_irradiance, temperature_cell, **module_params)
mpp = pvlib.pvsystem.max_power_point(*sde_params)

plt.plot(effective_irradiance, mpp['p_mp'] / effective_irradiance)
plt.ylabel('Pmp divided by Geff')
plt.xlabel('Geff')
```
</details>

So it does make a (small) difference whether the shaded fraction loss is applied to the irradiance or to the power.

Conceptually, applying the loss to the irradiance makes more sense to me.  And that variant of the linear model has the SAM document as a reference.  Should we switch to calling it an irradiance loss?

williamhobbs:
> Only the simplest PV performance models assume that power is exactly linear with irradiance

🤯 I’m not sure why I had never thought of it that way, but thanks for helping it “click” for me!

kandersolar:
There are a couple problems with this shaded fraction calculation.  One is that the meaning of `cross_axis_slope` changes depending on `surface_azimuth`.  That means that for a tracking system, you get problems like this, where a cross-axis slope increases (or decreases) the shading in *both* halves of the day instead of increasing in one half and decreasing in the other:

image

I think the only way to get expected behavior is for the user to calculate and pass in a time-varying `cross_axis_slope`.  That seems like an unnecessary burden that will inevitable trip people up.

Another problem is that shaded fraction goes to zero when the sun goes below the slope plane.  Seems to me to be better to keep it at 100% in those times, but maybe that's a matter of preference.

I think we should switch to calculating shaded fraction using the method in the slope-aware backtracking report.  I think it would fix the above issues, and have the additional advantages of supporting nonzero `axis_tilt` and being more easily generalized in the future due to using tracker rotations and axis orientation angles instead of surface angles.  @echedey-ls what do you think?

kandersolar:
@echedey-ls I think the issues mentioned above are big enough that they won't be resolved in time for v0.10.4, so I retagged this PR for v0.10.5.  

Also, the generalized shaded fraction calculation I mentioned in https://github.com/pvlib/pvlib-python/pull/1962#issuecomment-1991805047 is now accepted by the journal.  It might make sense for `shaded_fraction1d` to just implement that equation from the start instead of finalizing the current equation and then adding the generalized form as an extension later, since I think the latter will be extra work.  Sorry for the confusing and conflicting guidance about this feature submission!  At least we are learning useful things about how these functions should look like and how to document them well.

echedey-ls:
Thanks Kevin for the in-depth review. I agree 100% on all your points. Maybe some tricks could be made by checking whether the Sun is on the front or the back of the array, but if you already have a paper accepted that's definitely better. Can you let me know once it gets published?

echedey-ls:
I've changed the scope of this PR to only address the shaded fraction topic. The linear loss model will be left to another PR I'll open in the following days.

echedey-ls:
Fixed a few things regarding the new implementation 🔧 . Ready to be re-reviewed!

pvlib/shading.py
```diff
@@ -342,3 +342,144 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    Parameters
+    ----------
+    solar_zenith : numeric
+        Solar position zenith, in degrees.
+    solar_azimuth : numeric
+        Solar position azimuth, in degrees.
+    trackers_axis_azimuth : numeric
+        In degrees. North=0º, South=180º, East=90º, West=270º.
+    shaded_tracker_tilt : numeric
+        Tilt of the tracker receiving the shade in degrees.
```
kandersolar, line 380: pvlib's convention is to use `tilt` as a positive (or zero) number, but the angles these equations accept can be positive or negative.  Can we call them `rotation` instead of `tilt`?

pvlib/shading.py
```diff
@@ -342,3 +342,144 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    if shading_tracker_tilt is None:
+        shading_tracker_tilt = shaded_tracker_tilt
+
+    # projected solar zenith:
+    # consider the angle the sun direct beam has on the vertical plane which
+    # contains the row's normal vector, with respect to a horizontal line
+    projected_solar_zenith = projected_solar_zenith_angle(
+        solar_zenith,
+        solar_azimuth,
+        0,  # no rotation from the horizontal plane
```
kandersolar, line 453: Does this have to be zero?  Can't we accept `axis_tilt` as a parameter as well?

pvlib/shading.py
```diff
@@ -342,3 +342,144 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+
+    .. versionadded:: 0.10.5
+
+    Parameters
+    ----------
+    solar_zenith : numeric
+        Solar position zenith, in degrees.
+    solar_azimuth : numeric
+        Solar position azimuth, in degrees.
+    trackers_axis_azimuth : numeric
```
kandersolar, line 377: `axis_azimuth`, to match other `pvlib.tracking` functions?

pvlib/shading.py
```diff
@@ -342,3 +342,144 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+        In degrees. North=0º, South=180º, East=90º, West=270º.
+    shaded_tracker_tilt : numeric
+        Tilt of the tracker receiving the shade in degrees.
+    collector_width : numeric
+        Vertical length of a tilted tracker. The returned ``shaded fraction``
+        is the ratio of the shadow over this value.
+    row_pitch : numeric
```
kandersolar, line 384: `pitch`, to match `pvlib.bifacial.infinite_sheds`?

pvlib/shading.py
kandersolar: I think we should also mention somewhere that both rotation parameters are right-handed rotations according to the specified `axis_azimuth` and `axis_tilt`.

pvlib/shading.py
```diff
@@ -342,3 +342,144 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    solar_zenith,
+    solar_azimuth,
+    trackers_axis_azimuth,
+    shaded_tracker_tilt,
+    *,
+    collector_width,
+    row_pitch,
+    surface_to_axis_offset=0,
+    cross_axis_slope=0,
+    shading_tracker_tilt=None,
```
kandersolar, line 357: I'm not necessarily opposed to using keyword-only parameters here, but can you explain the motivation for using them here?  And also the specific position, i.e. why are `collector_width` and `row_pitch` kw-only, but not (for example) `trackers_axis_azimuth`?

pvlib/shading.py
```diff
@@ -342,3 +342,144 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    | :math:`l`        | ``collector_width``        | all these           |
+    +------------------+----------------------------+ parameters, e.g.    |
+    | :math:`z_0`      | ``surface_to_axis_offset`` | :math:`m`.          |
+    +------------------+----------------------------+---------------------+
+    | :math:`f_s`      | Return value               | Dimensionless       |
+    +------------------+----------------------------+---------------------+
+
+    See also
+    --------
+    pvlib.tracking.projected_solar_zenith_angle
```
kandersolar, line 433:
```suggestion
    pvlib.shading.projected_solar_zenith_angle
```

pvlib/shading.py
```diff
@@ -342,3 +342,144 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    +==================+============================+=====================+
+    | :math:`\theta_1` | ``shading_tracker_tilt``   |                     |
+    +------------------+----------------------------+                     |
+    | :math:`\theta_2` | ``shaded_tracker_tilt``    | Degrees             |
+    +------------------+----------------------------+ :math:`^{\circ}`    |
+    | :math:`\beta_c`  | ``cross_axis_slope``       |                     |
+    +------------------+----------------------------+---------------------+
+    | :math:`p`        | ``row_pitch``              | Any consistent      |
+    +------------------+----------------------------+ length unit across  |
+    | :math:`l`        | ``collector_width``        | all these           |
```
kandersolar, line 424:
```suggestion
    | :math:`\ell`        | ``collector_width``        | all these           |
```

This is the latex command for the same L character used in the diagram.  Hopefully it displays correctly with mathjax :)

pvlib/shading.py
echedey-ls: It doesn't. I just mixed concepts. Fixing it ASAP.

pvlib/shading.py
echedey-ls: Yay!! It does! https://pvlib-python--1962.org.readthedocs.build/en/1962/reference/generated/pvlib.shading.shaded_fraction1d.html

pvlib/shading.py
echedey-ls: Nice spot

pvlib/shading.py
echedey-ls: It obviously is to avoid allowing magic numbers in user's code.

I admit what I classify as a potential magic number parameter is subjective. But my reasoning is the following:
- `solar_zenith, solar_azimuth,` are usually provided from a variable in an explicit way
- `trackers_axis_azimuth, shaded_tracker_tilt,` I think from the context it is easy to know which one is from the values.
- `collector_width, row_pitch, surface_to_axis_offset, cross_axis_slope,` these are the real candidates to be magic numbers
- `shading_tracker_tilt` in fact, I did not know what to do with this one, but given the amount of parameters in degrees, excluding this one knowing which one is any of `trackers_axis_azimuth, shaded_tracker_tilt,` seems easier.

I don't have a strong opinion other than wanting to avoid debugging unreadable code. I'm up to change this completely, I know it introduces a new way of exposing the API.

echedey-ls:
> maybe we should include an Examples section with one or two basic cases (e.g. south-facing fixed tilt on flat terrain)?

That's an amazing idea. How do you feel about a gallery example, to replicate some 2-3 test cases in https://zenodo.org/doi/10.5281/zenodo.10513987 , per south-facing, north-facing, N-S HSAT configurations.

I say both south- and north- facing because I'd like to:
- show examples with the different right-handed rotations explicitly (even thou they don't change too much)
- avoid being the FOSS package taken as an example of the scientific bias 🙃 - I was responsible for it in the PSZA PR btw :grimacing:
   - `south-facing` yields 9 results
   - `north-facing` yields 0 results

kandersolar:
Sounds good to me, although I'm not sure the test cases in the zenodo archive are necessarily the best beginner examples (for example, there is no "simple" case where `zL=zR=0`), so maybe we should just invent some new example cases.  Anyway let's add a gallery example in a separate PR. 

Another question for this PR: I imagine most users will want to use this function with long time series inputs and get time series output back.  With this function, when using different shaded and shading rotations, it is complicated by having to swap the two rotations depending on the time of day.  Should the function perform that swapping automatically somehow?  Should we just document it and put the burden on the user?  Maybe @AdamRJensen has a comment?

pvlib/shading.py
echedey-ls: I'd like to hear opinions regarding this 🤔

echedey-ls:
> maybe we should include an Examples section with one or two basic cases (e.g. south-facing fixed tilt on flat terrain)?

I've added three and a half examples to the docstring 🫡. I have the gut feeling of having missed some points, please have a look at it.

>  With this function, when using different shaded and shading rotations, it is complicated by having to swap the two rotations depending on the time of day

I think I don't understand quite enough your point. Do you mean that, say, an HSAT with two rows A & B, A is shading B in the morning, and B is shading A in the afternoon, and we should return the shaded fraction for any of them? If so, right now we are focusing on calculating the shaded fraction for a specific row. I prefer that since one of the key points of this model is the independence between rows angles. And perhaps row-level modelling may be more intuitive than system-wide?

I don't have an issue with calculating a shaded fraction per row, if that's your point.

echedey-ls:
Sorry for the confusion, you are right . It isn't calculating the shaded fraction for a single row, it does however assume the shaded tracker and the shading one have the same role during all the day, when it is not for N-S trackers, for example.

As stated previously, I'd like to stick to row-focused computations. They seem more intuitive to me. And if we have three rows A, B & C, we could calculate both A shading B and C shading B and get B's shaded fraction with a `max(SF_A_shading_B, SF_C_shading_B)`

I don't know how to approach this issue right now. Maybe by following the paper implementation and changing to `theta_1` and `theta_2` notation? But how would they be defined then... 🤔

pvlib/shading.py
kandersolar: I am tentatively in favor of using more keyword-only parameters in new functions.  I think we can always convert them to "normal" parameters later without breaking anything, so it's not a big risk in terms of API stability.  For old functions, I am not sure it is worth the deprecation and change, but perhaps it could be evaluated on a case-by-case basis.

Here are some references I found useful:
- https://docs.scipy.org/doc/scipy/dev/missing-bits.html#required-keyword-names
- https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep009/proposal.html

Anyway no changes needed here, I think.  Thanks for the explanation!

echedey-ls:
Ready for another review!

Changes highlights:
+ Added the gallery example. [Link](https://pvlib-python--1962.org.readthedocs.build/en/1962/gallery/shading/plot_shaded_fraction1d_ns_hsat_example.html#sphx-glr-gallery-shading-plot-shaded-fraction1d-ns-hsat-example-py).
   - Switches have been made on the output of `shaded_fraction1d`
   - All rotations have been assumed to be the same. I'm up for changing that. For example, I can change that some row broke and goes 10 degrees behind the others, or that it doesn't move. To show other applications. I'm afraid that it could also be confusing. Another thought, maybe another section?
+ Documentation notes expanded plus links
+ Admonition added to docstring

[Current docs](https://pvlib-python--1962.org.readthedocs.build/en/1962/reference/generated/pvlib.shading.shaded_fraction1d.html#pvlib-shading-shaded-fraction1d).

Feel free to make any objections.

> I am not sure it is as simple as this. The function assumes that the tracker corresponding to shading_tracker_rotation is always "in front" (towards the sun) relative to the primary tracker. It's not possible to tell the function that the shading tracker is behind the primary tracker. So I don't think calculating two shaded fractions and taking the maximum is going to work.

I tried to mean that the function could be changed to check the PSZA sign and provide some interface that would make it easy to "max" the outputs of the shaded fractions cast by the two neighbouring rows. This would only work if the implementation knew which row is east/west. As you state, I agree on not scope creeping the use cases.

pvlib/shading.py
```diff
@@ -342,3 +342,144 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+):
+    r"""
+    Shaded fraction in the vertical dimension of tilted rows.
+
+    If ``shading_tracker_tilt`` isn't provided, assumes both the shaded
+    row and the one blocking the direct beam
+    share the same tilt and azimuth values.
+
+    If only the GCR is known, feed GCR into ``trackers_vertical_length`` and
+    specify ``trackers_row_pitch=1``.
```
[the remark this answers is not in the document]
echedey-ls: Forgot to address this😬 , has been addressed now

pvlib/shading.py
```diff
@@ -342,3 +342,213 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    +------------------+----------------------------+---------------------+
+    | :math:`f_s`      | Return value               | Dimensionless       |
+    +------------------+----------------------------+---------------------+
+
+    Examples
+    --------
+    Fixed-tilt south-facing array on flat terrain
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    Tilted tracker with a pitch of :math:`1.5m`, a collector width of
+    :math:`2m`, and tracker rotations of :math:`30^{\circ}`. In the morning.
```
kandersolar, line 447: If this is supposed to be a fixed-tilt array (with `axis_azimuth=180`), I think we should not use the word "tracker" to describe it.  Same for the next example.  How about `row` instead?

pvlib/shading.py
```diff
@@ -342,3 +342,213 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+
+    >>> shaded_fraction1d(solar_zenith=80, solar_azimuth=104.5,
+    ...     axis_azimuth=90, shaded_tracker_rotation=30,
+    ...     shading_tracker_rotation=30, collector_width=2, pitch=1,
```
kandersolar, line 451: I think most (all?) realistic scenarios will have `pitch > collector_width`.  For fixed-tilt arrays it is technically possible to have `pitch < collector_width`, but I am not sure anyone builds such a system.  For trackers it is not possible, as the modules will collide with each other at `rotation=0` :)

Let's increase these `pitch` values to 3 or 4 meters.  

(note also that the text above the code says it is 1.5 m, not 1)

pvlib/shading.py
```diff
@@ -342,3 +342,213 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+
+    Parameters
+    ----------
+    solar_zenith : numeric
+        Solar position zenith, in degrees.
+    solar_azimuth : numeric
+        Solar position azimuth, in degrees.
+    axis_azimuth : numeric
+        In degrees. North=0º, South=180º, East=90º, West=270º.
+    shaded_tracker_rotation : numeric
```
kandersolar, line 385: I know we have gone back and forth with these parameter names previously, but a new option occurs to me: `shaded_row_rotation`.  The advantage is that `row` is generic and applies equally to tracking and fixed-tilt arrays.  

@echedey-ls what do you think?

kandersolar (commented):
This is looking really good @echedey-ls!  A few more comments below.  I think we are getting close to merging this.

pvlib/shading.py
```diff
@@ -342,3 +342,213 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    +------------------+----------------------------+ parameters, e.g.    |
+    | :math:`z_0`      | ``surface_to_axis_offset`` | :math:`m`.          |
+    +------------------+----------------------------+---------------------+
+    | :math:`f_s`      | Return value               | Dimensionless       |
+    +------------------+----------------------------+---------------------+
+
+    Examples
+    --------
+    Fixed-tilt south-facing array on flat terrain
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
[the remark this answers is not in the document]
echedey-ls: I couldn't make it work with bulleted lists, so I've just set the titles bold.

pvlib/shading.py
echedey-ls: Yup, I didn't think much about it.

pvlib/shading.py
echedey-ls: > I know we have gone back and forth with these parameter names previously

No worries, I rather address these improvements now than seeing that comment after the merge. 

> The advantage is that row is generic and applies equally to tracking and fixed-tilt arrays.

I also 100% agree on that.

echedey-ls:
Thanks for the input @kandersolar 😄 

Ready for another review 👀

kandersolar (approved):
The docs build is failing because the example gallery is still using the old parameter names.  Other than that, LGTM!

FYI I believe @AdamRJensen plans to review as well.

echedey-ls:
Well spotted, I always miss something...

> FYI I believe @AdamRJensen plans to review as well.

All feedback is welcome, even better if it comes from one of the authors 😉

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
+"""
+shaded_fraction1d N-S horizontal single-axis example
```
AdamRJensen, line 2:
```suggestion
Shaded fraction of a horizontal single-axis tracker
```

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+====================================================
+
+This example illustrates how to calculate the shaded fraction of three rows
+in an N-S HSAT configuration.
```
AdamRJensen, line 6:
```suggestion
in a North-South horizontal single axis tracker (HSAT) configuration.
```

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+"""
+
+# %%
+# :py:func:`pvlib.shading.shaded_fraction1d` exposes a useful method for the
+# calculation of the shaded fraction of the width of a solar collector. Here,
+# the width is defined as the dimension perpendicular to the axis of rotation.
+# This method for calculating the shaded fraction also applies to fixed-tilt
+# systems with little changes.
```
AdamRJensen, line 14:
```suggestion
# This method for calculating the shaded fraction only requires minor
# modifications to be applicable for fixed-tilt systems.
```

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+#
+# Reading its documentation is recommended to understand the parameters and
+# the method capabilities.
+#
```
AdamRJensen, line 18:
```suggestion
# It is highly recommended to read the :py:func:`pvlib.shading.shaded_fraction1d` documentation
# to understand the input parameters and the method's capabilities.
#
```

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+# Let's start by obtaining the true-tracking angles for each of the rows and
+# limiting the angles to the range of -50 to 50 degrees. This decision is
+# done to allow significant shade to be used as an example.
```
AdamRJensen, line 21:
```suggestion
# done to create an example scenario with significant shading.
```

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+import pandas as pd
+import matplotlib.pyplot as plt
+from matplotlib.dates import DateFormatter
+
+# Define the solar system parameters
+latitude, longitude = 28.51, -13.89
+altitude = pvlib.location.lookup_altitude(latitude, longitude)
+
+axis_tilt = 3  # degrees
+axis_azimuth = 180  # degrees
```
AdamRJensen, line 45:
```suggestion
axis_azimuth = 180  # degrees, N-S tracking axis
```

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+collector_width = 3.2  # m
+pitch = 4.15  # m
+gcr = collector_width / pitch
+cross_axis_slope = -5  # degrees
```
AdamRJensen, line 49: @kandersolar Is this allowed to be negative?

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+import numpy as np
+import pandas as pd
+import matplotlib.pyplot as plt
+from matplotlib.dates import DateFormatter
+
+# Define the solar system parameters
+latitude, longitude = 28.51, -13.89
+altitude = pvlib.location.lookup_altitude(latitude, longitude)
+
+axis_tilt = 3  # degrees
```
AdamRJensen, line 44: Could you add a comment about what this represents?

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+    axis_tilt,
+    axis_azimuth,
+    max_angle=(-50, 50),  # (min, max) degrees
+    backtrack=False,
+    gcr=gcr,
+    cross_axis_tilt=cross_axis_slope,
+)["tracker_theta"]
+
+# %%
+# The next step is to calculate the shaded fraction. Special care must be taken
```
AdamRJensen, line 80:
```suggestion
# Once the tracker angles have been obtained, the next step is to calculate
# the shaded fraction. Special care must be taken
```

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+        pitch=pitch,
+        surface_to_axis_offset=surface_to_axis_offset,
+        cross_axis_slope=cross_axis_slope,
+        shading_row_rotation=rotation_angle,
+    ),
+    0,  # no shaded fraction in the evening
+)
+
+# %%
+# Plot the shaded fraction result per row
```
AdamRJensen, line 167:
```suggestion
# Plot the shaded fraction result for each row:
```

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
...
+plt.plot(times, eastmost_shaded_fraction, label="East-most", color="blue")
+plt.plot(
+    times,
+    middle_shaded_fraction,
+    label="Middle",
+    color="green",
+    linewidth=3,
+    linestyle="--",
+)
```
AdamRJensen, line 176:
```suggestion
plt.plot(times, middle_shaded_fraction, label="Middle", color="green",
             linewidth=3, linestyle="--")
```

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,185 @@
+"""
+shaded_fraction1d N-S horizontal single-axis example
+====================================================
+
+This example illustrates how to calculate the shaded fraction of three rows
```
AdamRJensen, line 5:
```suggestion
This example illustrates how to calculate the 1D shaded fraction of three rows
```

AdamRJensen (commented):
An initial review of the gallery example

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
kandersolar: Yes, it is

pvlib/shading.py
```diff
@@ -342,3 +342,214 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    | :math:`f_s`      | Return value               | Dimensionless       |
+    +------------------+----------------------------+---------------------+
+
+    Examples
+    --------
+
+    **Fixed-tilt south-facing array on flat terrain**
+
+    Tilted row with a pitch of :math:`3m`, a collector width of
+    :math:`2m`, and row rotations of :math:`30^{\circ}`. In the morning.
```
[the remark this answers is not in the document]
AdamRJensen: Following SI convention, then there should be a space between the unit and the value. I also see no reason to bold them, i.e., have them in math mode (that's just for symbols in my opinion).

pvlib/shading.py
```diff
@@ -342,3 +342,214 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    ...     surface_to_axis_offset=0.05, cross_axis_slope=0)
+    0.6827437712114521
+
+    **Fixed-tilt north-facing array on sloped terrain**
+
+    Tilted row with a pitch of :math:`4m`, a collector width of
+    :math:`2.5m`, and row rotations of :math:`50^{\circ}` for the shaded
+    row and :math:`30^{\circ}` for the shading row. The rows are on a
+    :math:`10^{\circ}` slope, where their axis is on the most inclined
+    direction (zero cross-axis slope). Shaded in the morning.
```
AdamRJensen, line 464:
```suggestion
    Tilted row with a pitch of 4 m, a collector width of
    2.5 m, and row rotations of 50:math:`^{\circ}` for the shaded
    row and 30:math:`^{\circ}` for the shading row. The rows are on a
    10:math:`^{\circ}` slope, where their axis is on the most inclined
    direction (zero cross-axis slope). Shaded in the morning.
```

pvlib/shading.py
```diff
@@ -342,3 +342,214 @@ def projected_solar_zenith_angle(solar_zenith, solar_azimuth,
...
+    ...     collector_width=2.5, pitch=4, axis_tilt=10,
+    ...     surface_to_axis_offset=0.05, cross_axis_slope=0)
+    0.6975923460352351
+
+    **N-S single-axis tracker on sloped terrain**
+
+    Horizontal trackers with a pitch of :math:`3m`, a collector width of
+    :math:`1.4m`, and tracker rotations of :math:`30^{\circ}` pointing east,
+    in the morning. Terrain slope is :math:`7^{\circ}` west-east (east-most
+    tracker is higher than the west-most tracker).
```
AdamRJensen, line 477:
```suggestion
    Horizontal trackers with a pitch of 3 m, a collector width of
    1.4 m, and tracker rotations of 30:math:`^{\circ}` pointing east,
    in the morning. Terrain slope is 7:math:`^{\circ}` west-east (east-most
    tracker is higher than the west-most tracker).
```

AdamRJensen (commented):
Awesome stuff @echedey-ls !

pvlib/shading.py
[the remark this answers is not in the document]
echedey-ls: Good point.

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
echedey-ls: Let me know your opinion on `axis_tilt = 3  # degrees, positive is upwards in the axis_azimuth direction`

echedey-ls:
Everything's ready IMO 😄 . Feel free to review and propose changes 📝

docs/examples/shading/plot_shaded_fraction1d_ns_hsat_example.py
```diff
@@ -0,0 +1,179 @@
...
+# %%
+# :py:func:`pvlib.shading.shaded_fraction1d` exposes a useful method for the
+# calculation of the shaded fraction of the width of a solar collector. Here,
+# the width is defined as the dimension perpendicular to the axis of rotation.
+# This method for calculating the shaded fraction only requires minor
+# modifications to be applicable for fixed-tilt systems.
+#
+# It is highly recommended to read :py:func:`pvlib.shading.shaded_fraction1d`
+# documentation to understand the input parameters and the method's
+# capabilities.
```
AdamRJensen, line 18:
```suggestion
# It is highly recommended to read :py:func:`pvlib.shading.shaded_fraction1d`
# documentation to understand the input parameters and the method's
# capabilities. For more in depth information, please see the journal paper
# [10.1063/5.0202220](https://doi.org/10.1063/5.0202220) describing the methodology.
```

AdamRJensen (approved):

PR #1968, opened 2024-02-09
Have pvlib.iotools.read_solrad return metadata

AdamRJensen:
- ~~[ ] Closes #xxxx~~
 - [x] Tests added

The ``read_solrad`` function currently does not follow the standard pattern of returning (data, metadata). This PR adds metadata parsing to the function and is thus a breaking change.

### DIFF
--- a/pvlib/iotools/solrad.py
+++ b/pvlib/iotools/solrad.py
@@ -108,7 +108,7 @@ def read_solrad(filename):
     else:
         file_buffer = open(str(filename), 'r')
 
-    # the first has the name of the  station, and the second gives the
+    # The first line has the name of the station, and the second gives the
     # station's latitude, longitude, elevation above mean sea level in meters,
     # and the displacement in hours from local standard time.
     meta['station_name'] = file_buffer.readline().strip()
@@ -116,7 +116,7 @@ def read_solrad(filename):
     meta_line = file_buffer.readline().split()
     meta['latitude'] = float(meta_line[0])
     meta['longitude'] = float(meta_line[1])
-    meta['elevation'] = float(meta_line[2])
+    meta['altitude'] = float(meta_line[2])
     meta['TZ'] = int(meta_line[3])
 
     # read in data
--- a/pvlib/iotools/solrad.py
+++ b/pvlib/iotools/solrad.py
@@ -106,7 +106,8 @@ def read_solrad(filename):
         response = requests.get(filename)
         file_buffer = io.StringIO(response.content.decode())
     else:
-        file_buffer = open(str(filename), 'r')
+        with open(str(filename), 'r') as file_buffer:
+            file_buffer = io.StringIO(file_buffer.read())
 
     # The first line has the name of the station, and the second gives the
     # station's latitude, longitude, elevation above mean sea level in meters,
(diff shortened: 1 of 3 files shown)


pvlib/iotools/solrad.py
```diff
@@ -49,8 +49,15 @@
 
 def read_solrad(filename):
     """
-    Read NOAA SOLRAD fixed-width file into pandas dataframe.  The SOLRAD
-    network is described in [1]_ and [2]_.
+    Read NOAA SOLRAD fixed-width file into pandas dataframe.
+
+    The SOLRAD network is described in [1]_ and [2]_.
+
+    .. versionchanged:: 0.10.4
```
kandersolar, line 56: I like this PR, but it's breaking and should go into v0.11.0.

Should we forget v0.10.4 and make v0.11.0 the next release?

kandersolar: I forgot 0.10.0 included several deprecations marked for removal in 0.11.0.  I agree it would be nice to wait a bit longer for that.

I also agree that this change is small, and in a function likely used by few.  But out of principle I'm reluctant to include any breaking change PR outside of a .0 release.  I suppose that means we put this PR on ice until we are ready to release 0.11.0.

adriesse: I guess I've mellowed with age and wouldn't delay this out of principle.  Would making the returned meta optional be an....option?

pvlib/iotools/solrad.py
```diff
@@ -91,19 +100,28 @@ def read_solrad(filename):
         widths = WIDTHS
         dtypes = DTYPES
 
+    meta = {}
+
+    if str(filename).startswith('ftp') or str(filename).startswith('http'):
+        response = requests.get(filename)
+        file_buffer = io.StringIO(response.content.decode())
+    else:
+        file_buffer = open(str(filename), 'r')
```
kandersolar, line 109: I have a mild allergic reaction to seeing `= open(...)` in code, since it means that the file handle will get left open if an unhandled exception occurs before the `.close()` is called.  It's possible to refactor this code to use `with open(...)` instead, but I think it would be a minor hassle, and make the code a bit harder to follow.  And I realize that `read_surfrad` uses `= open(...)` too.  Not asking for this to be changed, just wanted to point it out.

But: if we're using `open` outside of a context manager, we need to call `file_buffer.close()` when we're done using it (after the call to `pd.read_fwf`)

kandersolar (approved):
I've been persuaded that this minor enough to be more of an enhancement than a breaking change, and so should go in 0.10.4.  I changed the milestone accordingly. 

Nice improvements to the tests by the way!

pvlib/iotools/solrad.py
adriesse, line 109:
```suggestion
        with open(str(filename), 'r') as f:
            file_buffer = io.StringIO(f.read())
```
This would be an easy-to-follow allergy suppression. Faster too.

adriesse (changes requested):

pvlib/iotools/solrad.py
AdamRJensen: Thanks @adriesse 😄

PR #1969, opened 2024-02-10
Add get_solargis iotools function

AdamRJensen:
- ~~[ ] Closes #xxxx~~
 - [x] Tests added

This PR adds a fetching function for Solargis irradiance data. Given that such functions for SolarAnywhere and Solcast has been added in the previous release, it seems fitting to also have it for Solargis.


AdamRJensen:
@kandersolar The data returned from Solargis is in XML format, thus the lxml python package is required. Can that be added to the pvlib's list of optional packages?

mikofski:
I’m okay with including lxml. Another option is the built-in Python module [xml.etree.ElementTree](https://docs.python.org/3/library/xml.etree.elementtree.html) but I don’t recommend it.

AdamRJensen:
> Can we get away with using the standard library by specifying `parser='etree'` in `pd.read_xml`? If not, I'm also fine with adding lxml to the optional dependencies.

Works like a charm! Thanks 😄

pvlib/iotools/solargis.py
```diff
@@ -0,0 +1,171 @@
...
+    'SE': 'solar_elevation',
+    'SA': 'solar_azimuth',
+    'TEMP': 'temp_air',
+    'TD': 'temp_dew',
+    'AP': 'pressure',
+    'RH': 'relative_humidity',
+    'WS': 'wind_speed',
+    'WD': 'wind_direction',
+    'INC': 'aoi',  # angle of incidence of direct irradiance
+    'PWAT': 'precipitable_water',  # [kg/m2]
```
kandersolar, line 29: At least some functions in pvlib expect units of cm for pwat.  Worth looking through pvlib in more detail and seeing if we should convert units here

pvlib/iotools/solargis.py
```diff
@@ -0,0 +1,171 @@
...
+        Dictionary containing metadata.
+
+    Raises
+    ------
+    requests.HTTPError
+        A message from the Solargis server if the request is rejected
+
+    Notes
+    -----
+    Each XMl request is limited to retrieving 31 days of data.
```
kandersolar, line 91:
```suggestion
    Each XML request is limited to retrieving 31 days of data.
```

pvlib/iotools/solargis.py
```diff
@@ -0,0 +1,171 @@
...
+    References
+    ----------
+    .. [1] `Solargis <https://solargis.com>`_
+    .. [2] `Solargis API User Guide
+       <https://solargis.atlassian.net/wiki/spaces/public/pages/7602367/Solargis+API+User+Guide>`_
+
+    Examples
+    --------
+    >>> # Retrieve two days of irradiance data from Solargis
+    >>> data, neta = response = pvlib.iotools.get_solargis(
```
kandersolar:
```suggestion
    >>> data, meta = response = pvlib.iotools.get_solargis(
```

pvlib/iotools/solargis.py
```diff
@@ -0,0 +1,171 @@
...
+                       parser='etree')
+    data.index = pd.to_datetime(data['dateTime'])
+    data = data['values'].str.split(' ', expand=True)
+    data = data.astype(float)
+    data.columns = header['columns'].iloc[0].split()
+
+    if map_variables:
+        data = data.rename(columns=VARIABLE_MAP)
+
+    data = data.replace(-9, pd.NA)
```
kandersolar, line 170: Is -9 the missing value indicator for all variables?  Seems like it could be a valid value for at least some fields.

pvlib/iotools/solargis.py
```diff
@@ -0,0 +1,171 @@
+"""Functions to retrieve and parse irradiance data from Solargis."""
+
+import pandas as pd
+import requests
+import io
+
+URL = 'https://solargis.info/ws/rest/datadelivery/request?'
```
kandersolar, line 7: I suggest leaving the `?` off the URL itself.  Makes more sense to add that in when appending the API key onto the URL IMHO.

kandersolar (commented):
A few initial comments

pvlib/iotools/solargis.py
AdamRJensen: @kandersolar Initially I would agree, but removing the question mark results in the code not working...

pvlib/iotools/solargis.py
AdamRJensen: This is indeed correct. I have gone through all variables that can be returned and looked through the documentation to see which have a missing value indicator of -9. I've implemented a fix to this in the code too.

pvlib/iotools/solargis.py
AdamRJensen: The following functions use or calculate precipitable_water in units of cm: kasten96_lt, gueymard94_pw, spectral_factor_firstsolar, spectral_factor_caballero, simplified_solis, bird.

Notably, the Solcast iotools convert it from kg/m2 to cm. The unit from PSM3 is already in cm. In TMY files it's also in cm. Though in TMY2 it seems like its in mm, but in TMY3 it is in cm!

I have added a conversion similar to what is done in the solcast functions.

AdamRJensen:
Two things I don't like is the summarization naming (e.g., '5_MIN', 'HOURLY'). For the CAMS functions we use the name "time_step". Also, perhaps we should allow for the time steps to follow the ISO 8601 naming convention by mapping with this dictionary:

     {'PT05M': 'MIN_5', 'PT10M': 'MIN_10', 'PT15M': 'MIN_15', 'PT30': 'MIN_30', 'PT60M': 'HOURLY', 'PT1H': 'HOURLY', 'P1D': 'DAILY', 'P1M': 'MONTHLY', 'P1Y': 'YEARLY'}

Last, I don't like how the irradiance units change depend on the time step:

>     The variable units depend on the time frequency, e.g., the unit for sub-hourly irradiance data is W/m^2, for hourly data it is Wh/m^2, and for daily data it is kWh/m^2.

pvlib/iotools/solargis.py
```diff
@@ -0,0 +1,204 @@
...
+    meta_lines = [line.strip() for line in meta_lines]
+    meta = {}
+    for line in meta_lines:
+        if ':' in line:
+            key = line.split(':')[0].lower()
+            if key in METADATA_FIELDS:
+                meta[key] = ':'.join(line.split(':')[1:])
+    meta['latitude'] = float(meta['latitude'])
+    meta['longitude'] = float(meta['longitude'])
+    meta['altitude'] = float(meta.pop('elevation').replace('m a.s.l.', ''))
```
kandersolar, line 176: It would be great to include the data version in the returned metadata too.  Unfortunately it can't be extracted based on colons as you've done for the other fields (it looks like `'#Output from the climate database Solargis v2.2.46`).

kandersolar (commented):
> Thoughts on "summarization" vs "time_step"?

I do not like `summarization`.  `get_solaranywhere` uses `time_resolution` (int).  `get_psm3` uses `period` (int).  Of these, I like the `time_resolution` name best. 

For the ISO 8601 durations idea, I mildly prefer specifying an integer number of minutes, but I don't feel strongly about it.

pvlib/iotools/solargis.py
```diff
@@ -0,0 +1,204 @@
...
+    start : datetime-like
+        Start date of time series.
+    end : datetime-like
+        End date of time series.
+    variables : list
+        List of variables to request, see [2]_ for options.
+    api_key : str
+        API key.
+    summarization : str, {'MIN_5', 'MIN_10', 'MIN_15', 'MIN_30', 'HOURLY', 'DAILY', 'MONTHLY', 'YEARLY'}
+        DESCRIPTION.
```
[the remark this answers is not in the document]
AdamRJensen: What integer would you specify for daily monthly or yearly data then?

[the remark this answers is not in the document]
kandersolar: Is it too inconsistent to allow `{5, 10, 15, 30, 60, 'DAILY', 'MONTHLY', 'ANNUAL'}`?  

The daily/monthly/annual data are of lesser interest to pvlib anyway, so in my view it's no big deal if they are not as "nice" to use.

pvlib/iotools/solargis.py
```diff
@@ -0,0 +1,204 @@
...
+    ParameterMap('TEMP', 'temp_air'),
+    ParameterMap('TD', 'temp_dew'),
+    # surface_pressure (hPa) -> pressure (Pa)
+    ParameterMap('AP', 'pressure', lambda x: x*100),
+    ParameterMap('RH', 'relative_humidity'),
+    ParameterMap('WS', 'wind_speed'),
+    ParameterMap('WD', 'wind_direction'),
+    ParameterMap('INC', 'aoi'),  # angle of incidence of direct irradiance
+    # precipitable_water (kg/m2) -> precipitable_water (cm)
+    ParameterMap('PWAT', 'precipitable_water', lambda x: x*10),
```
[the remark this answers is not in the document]
AdamRJensen: Correct - I'll correct this in the Solcast function also then.

pvlib/iotools/solargis.py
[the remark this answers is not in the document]
AdamRJensen: I have incorporated both conventions in the dictionary

pvlib/iotools/solargis.py
AdamRJensen: I agree it would be nice, but judging it to be out of scope for now.

pvlib/tests/iotools/test_solargis.py
```diff
@@ -0,0 +1,74 @@
...
+def hourly_dataframe(hourly_index):
+    ghi = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 5.0, 73.0, 152.0, 141.0, 105.0,
+           62.0, 65.0, 62.0, 11.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+    dni = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 30.0, 233.0, 301.0, 136.0, 32.0,
+           0.0, 3.0, 77.0, 5.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+    return pd.DataFrame(data={'ghi': ghi, 'dni': dni}, index=hourly_index)
+
+
+@pytest.mark.skipif(pd.__version__ < '1.3.0',
+                    reason='pd.read_xml is new as of 1.3.0')
```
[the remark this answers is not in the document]
AdamRJensen: I don't think there's a good alternative way around it. But yes, seems silly to increase the version by this much because of one function. At the same time, just bumping up the required version is surely the best way to reduce the maintenance burden - so that's my vote.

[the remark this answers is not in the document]
mikofski: I vote bump version & not keep super old minimum

pvlib/iotools/solargis.py
```diff
@@ -0,0 +1,212 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <site id="{site_id}" name="" lat="{latitude}" lng="{longitude}">
+    </site>
+    <processing key="{' '.join(variables)}"
+    summarization="{TIME_RESOLUTION_MAP[time_resolution].upper()}"
+    terrainShading="{str(terrain_shading).lower()}">
+    <timestampType>{timestamp_type.upper()}</timestampType>
+    <timeZone>{tz}</timeZone>
+    </processing>
+    </ws:dataDeliveryRequest>'''
```
adriesse, line 165:
```suggestion
    </ws:dataDeliveryRequest>'''
    
```

pvlib/tests/iotools/test_solargis.py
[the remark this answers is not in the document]
AdamRJensen: Also bumping Numpy:

image

kandersolar (approved):
We need to update the minimum versions in the benchmarks too now, but I'll do that separately.  Thanks @AdamRJensen!

Issue #1952, opened 2024-01-24
Gallery examples of ModelChains for the DOE Solar Data Bounty Prize systems


kandersolar:
The U.S. Department of Energy sponsored a [Solar Data Bounty Prize](https://www.energy.gov/eere/solar/american-made-solar-data-bounty-prize) that resulted in publication of detailed monitoring data and system configuration specs for five commercial PV systems.  I think we should create a section in the example gallery containing "reference models" for these systems.  This section could be expanded over time to have models corresponding to other public PV datasets too (RTC, NIST, etc).

System descriptions:
- https://openei.org/wiki/PVDAQ/Sites/Farm_Solar_Array
- https://openei.org/wiki/PVDAQ/Sites/Maui_Ocean_Center
- https://openei.org/wiki/PVDAQ/Sites/SR_CO
- https://openei.org/wiki/PVDAQ/Sites/Shine_On_Solar_Facility
- https://openei.org/wiki/PVDAQ/Sites/Simon_Solar_Farm

Data available for download here: https://data.openei.org/s3_viewer?bucket=oedi-data-lake&prefix=pvdaq%2F2023-solar-data-prize%2F

It would be great to get these pages created and merged before March 21 so that they can be mentioned in the [Modeling Tools webinar](https://www.energy.gov/eere/solar/solar-energy-technologies-office-events-and-webinars), so I'm tagging this for v0.10.4 (scheduled for mid-March).

AdamRJensen:
Did DOE put funds aside for such work (seems like that would've made sense)?

kandersolar:
To my knowledge, no special funding is currently available for this work.

williamhobbs:
I _really_ like this idea.

SR_CO looks like thin-film with true-tracking, and I think Simon Solar Farm is fixed tilt 2P crystalline, so both could be good for validation of self-shade models.

kandersolar:
A note for whoever makes these models: it's a little hard to tell, but from google street view it looks like SR_CO is a 4L array, so don't be surprised if the simple linear self-shading model doesn't perform very well. 

![image](https://github.com/pvlib/pvlib-python/assets/57452607/1cac5189-2f41-499c-b18b-dec366fcb263) 

![image](https://github.com/pvlib/pvlib-python/assets/57452607/573f272d-c296-453a-b015-f930cac829d2)

[...]

PR #1985, opened 2024-03-07
Add gallery example for OEDI system 9068

kandersolar:
- A step towards #1952
 - ~[ ] Tests added~


AdamRJensen (approved):
I would expose the other loss parameters:

        # Other loss types are left at their default amounts.
        losses_parameters = dict(shading=0, availability=0)

Otherwise things look good to me.
image

I have my doubts if applying a transient operation on the module temperature makes this any more accurate when you use satellite data?

A shame that this great data initiative has so poor metadata - a bit ironic really.

docs/examples/system-models/plot_oedi_9068.py
```diff
@@ -0,0 +1,282 @@
+"""
+SR_CO (OEDI System 9068)
+========================
+
+A basic model of a 4.7 MW single-axis tracking CdTe system located in
+Colorado, United States.
```
AdamRJensen, line 6:
```suggestion
Modeling of a basic model of a 4.7 MW single-axis tracking CdTe system
located in Colorado, United States.
```

docs/examples/system-models/plot_oedi_9068.py
```diff
@@ -0,0 +1,282 @@
+"""
+SR_CO (OEDI System 9068)
```
AdamRJensen, line 2: Can we make a more descriptive title? 4.7 MW CdTe single-axis tracking

kandersolar:
> I have my doubts if applying a transient operation on the module temperature makes this any more accurate when you use satellite data?

5-minute PSM3 data is very "spiky", so using a transient model at least makes the modeled Tcell more realistic/smooth, if not also more accurate.  IMHO it's reasonable to include by default unless we know for some reason that it makes results worse.  I think SAM applies it by default for any dataset with time step below 20 minutes. 

> Maybe add a note about why you're not using the measured irradiance (I think that would be useful for people)

It does say this before fetching the weather dataset:
> The system does have measured plane-of-array irradiance data, but the measurements suffer from row-to-row shading and tracker stalls. In this example, we will use weather data taken from the NSRDB PSM3 for the year 2019.

adriesse:
> The system does have measured plane-of-array irradiance data, but the measurements suffer from row-to-row shading and tracker stalls. In this example, we will use weather data taken from the NSRDB PSM3 for the year 2019.

It could actually be instructive to use that data in the example anyway.

PR #2011, opened 2024-04-12
Update formatting of additional pvlib publications

AdamRJensen:
Not all of the "[additional pvlib python publications](https://pvlib-python.readthedocs.io/en/stable/#citing-pvlib-python)" include a doi reference. This PR adds this and also standardizes the format of the references.

### DIFF
--- a/docs/sphinx/source/index.rst
+++ b/docs/sphinx/source/index.rst
@@ -57,7 +57,7 @@ use pvlib python in a published work, please cite:
   Anderson, K., Hansen, C., Holmgren, W., Jensen, A., Mikofski, M., and Driesse, A.
   "pvlib python: 2023 project update."
   Journal of Open Source Software, 8(92), 5994, (2023).
-  https://doi.org/10.21105/joss.05994
+  :doi:`10.21105/joss.05994`.
 
 
 **Recommended citation for pvlib iotools**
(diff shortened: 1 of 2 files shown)


cwhanse:
LGTM

docs/sphinx/source/index.rst
```diff
@@ -82,26 +82,30 @@ pvlib python that you used. pvlib python DOIs are listed at
 
 Additional pvlib python publications include:
 
-* J. S. Stein, “The photovoltaic performance modeling
-  collaborative (PVPMC),” in Photovoltaic Specialists Conference, 2012.
+* J. S. Stein, "The Photovoltaic Performance Modeling Collaborative (PVPMC),"
+  2012 38th IEEE Photovoltaic Specialists Conference (PVSC),
+  Austin, TX, USA, 2012, pp. 003048-003052, https://doi.org/10.1109/PVSC.2012.6318225.
```
echedey-ls, line 87: Is it more common to add the raw link in the IEEE format or DOI: 10.1109/PVSC.2012.6318225 ?

I think these dois could make use of the role `:doi:`` `

AdamRJensen: DOI is certainly to be preferred.

I've switched to the `doi` role, though I don't exactly know the benefit

echedey-ls: Sorry for not being clear enough regarding the "raw link".

You can either put:
* [DOI: 10.21105/joss.05994](https://dx.doi.org/10.21105/joss.05994)
or
* https://doi.org/10.21105/joss.05994

The first one is with the DOI role and the second one is the one I was referring to with "raw link".

echedey-ls (approved):

docs/sphinx/source/index.rst
AdamRJensen: The HTTP link for DOIs is fairly compact, so I don't have a strong preference for either one. Let's just go with the DOI role for this PR.

mikofski: IMHO it’s not an either/or situation. As the title of this PR suggests there is or should be a standard way to format & render references, which uses the `doi` role that was explicitly created for this purpose. If it is not functional then it should probably be removed from the Sphinx `config` file. Maybe this needs to be documented somewhere?

kandersolar (approved):

PR #2050, opened 2024-05-12
Allow arbitrary IAM function in pvlib.iam.marion_diffuse

### LINKED ISSUE #2049: Allow arbitrary IAM function in pvlib.iam.marion_diffuse
I am working on an application using `pvlib.iam.marion_diffuse`, and I would like to inject my own IAM function (from interpolated IEC 61853-2 data) as a parameter to `pvlib.iam.marion_diffuse`.

Add the option for passing a callable to `pvlib.iam.marion_diffuse` instead of the existing strings for the builtin IAM model functions. (The existing strings would still be accepted.)

Also add a function to create IAM functions as a PCHIP interpolation of data such as from IEC 61853-2. Note that this solution can also be made to add support for passing the existing `pvlib.iam.interp` function (whose `cubic` option is slightly different that `pchip`) to `pvlib.iam.marion_diffuse`.

Rolling my own interpolant each time, and then calling the lower-level `pvlib.iam.marion_integrate` for each of the three components.

**Additional context**
This is a followup to https://github.com/pvlib/pvlib-python/issues/1980, but without any concern to improving the diffuse-horizontal integral.

image

markcampanelli:
- [x] Closes #2049
 - [x] Tests added

### DIFF
--- a/pvlib/iam.py
+++ b/pvlib/iam.py
@@ -11,6 +11,7 @@ irradiance to the module's surface.
 import numpy as np
 import pandas as pd
 import functools
+import scipy.interpolate
 from scipy.optimize import minimize
 from pvlib.tools import cosd, sind, acosd
 
@@ -470,8 +471,6 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
     '''
     # Contributed by Anton Driesse (@adriesse), PV Performance Labs. July, 2019
 
-    from scipy.interpolate import interp1d
-
     # Scipy doesn't give the clearest feedback, so check number of points here.
     MIN_REF_VALS = {'linear': 2, 'quadratic': 3, 'cubic': 4, 1: 2, 2: 3, 3: 4}
 
@@ -483,8 +482,9 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
         raise ValueError("Negative value(s) found in 'iam_ref'. "
                          "This is not physically possible.")
 
-    interpolator = interp1d(theta_ref, iam_ref, kind=method,
-                            fill_value='extrapolate')
+    interpolator = scipy.interpolate.interp1d(
+        theta_ref, iam_ref, kind=method, fill_value='extrapolate'
+    )
     aoi_input = aoi
 
     aoi = np.asanyarray(aoi)
@@ -497,6 +497,39 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
 
     if isinstance(aoi_input, pd.Series):
         iam = pd.Series(iam, index=aoi_input.index)
+
+    return iam
+
+
+def iam_pchip(aoi_data, iam_data):
+    """
+    Generate a piecewise-cubic hermite interpolating polynomial for incident-angle
+    modifier (IAM) data. Assumes aoi_data in [0, 90], in degrees, with normalized
+    iam_data in [0, 1], unitless. Typically, (0, 1) and (90, 0) are included in the
+    interpolation data.
+
+    The resulting function requires aoi to be in [0, 180], in degrees.
+
+    This function can work well for IAM measured according to IEC 61853-2, as it
+    preserves monotonicity between data points.
+
+    Note that scipy.interpolate.PchipInterpolator requires aoi_data be 1D monotonic
+    increasing and without duplicates.
+    """
+
+    iam_pchip_ = scipy.interpolate.PchipInterpolator(
+        aoi_data, iam_data, extrapolate=False
+    )
+
+    def iam(aoi):
+        """Compute unitless incident-angle modifier as a function of aoi, in degrees."""
+        if np.any(np.logical_or(aoi < 0, aoi > 180)):
+            raise ValueError("aoi not between 0 and 180, inclusive")
+
+        iam_ = iam_pchip_(aoi)
+        iam_[90 < aoi] = 0.0
+
+        return iam_
 
     return iam
 
@@ -575,9 +608,9 @@ def marion_diffuse(model, surface_tilt, **kwargs):
 
     Parameters
     ----------
-    model : str
-        The IAM function to evaluate across solid angle. Must be one of
-        `'ashrae', 'physical', 'martin_ruiz', 'sapm', 'schlick'`.
+    model : str or callable
+        The IAM function to evaluate across solid angle. If not a callable, then must be
+        one of `'ashrae', 'martin_ruiz', 'physical', 'sapm', 'schlick'`.
 
     surface_tilt : numeric
         Surface tilt angles in decimal degrees.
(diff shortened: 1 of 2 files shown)


markcampanelli:
Before I work through the full checklist, I'm hoping a maintainer can do an early review of this to ensure it will be accepted.

I have added tests that also show how everything works, including how this adds support for using the existing `pvlib.iam.interp` with `pvlib.iam.marion_diffuse`. cc @adriesse

If this work should proceed, then I would be happy to add usage examples in separate scripts.

cwhanse:
@markcampanelli code structure looks fine to me. I am not sure why adding `iam.pchip` fits in the scope. Accepting a callable doesn't require pchip. It looks to me that pchip is the equivalent of adding a new IAM model (defined by a specific interpolator). Also, what is the reason for needing the `iam.pchip` which is mostly a wrapper on the scipy function - is the scipy code awkward to use?

markcampanelli:
> @markcampanelli code structure looks fine to me. I am not sure why adding `iam.pchip` fits in the scope. Accepting a callable doesn't require pchip. It looks to me that pchip is the equivalent of adding a new IAM model (defined by a specific interpolator). Also, what is the reason for needing the `iam.pchip` which is mostly a wrapper on the scipy function - is the scipy code awkward to use?

@cwhanse Thanks much for taking an early look for purposes of direction and scope. In https://github.com/pvlib/pvlib-python/issues/2049, @adriesse suggesting mking `pchip` another option in the existing `pvlib.iam.interp`. I am ok doing that instead here (or removing `pchip` altogether). Thoughts?

In any event, I will keep the `pchip` function in the tests for `test_marion_diffuse_iam_function_without_kwargs`.

markcampanelli:
@cwhanse @adriesse I decided simply to not add the PCHIP interpolator to the codebase, but it does appear as the main test case in `test_marion_diffuse_iam_function_without_kwargs`.

One issue I hit trying to integrate `pchip` into `interp` is that IAM functions in `interp` appear to NOT set IAM to zero for AOIs >= 90, and I'm am unsure what the justification is here. It also raised questions about the need to validate that IAM(0) > 1 and IAM(90) = 0.

I think my only remaining technical question is do I have to support the full `numeric` type for the `aoi` input in the callable accepted by `marion_diffuse`. I think the answer is yes.

pvlib/iam.py
```diff
@@ -92,7 +145,7 @@ def ashrae(aoi, b=0.05):
     return iam
 
 
-def physical(aoi, n=1.526, K=4.0, L=0.002, *, n_ar=None):
```
markcampanelli, line 95: All this does is require that the `n_ar` argument be named, which seems like an antipattern for this codebase.

cwhanse:
@markcampanelli a few in-flight comments:

- `get_builtin_models` should be private. I'm not sure of the advantage of this function over a constant `dict.
- I'm not sure why the code for fitting and converting models is being edited. That seems unnecessary as the PR isn't adding to the set of models that can be fit/converted.
- I'm OK with the `ModelChain` integration within this PR, but as @echedey-ls pointed out, sometimes with PRs less is more :)

markcampanelli:
@cwhanse @kandersolar The `sapm` IAM function has a different interface from all the remaining IAM functions. I tried to clean this up in the code and make it much more obvious as to the where this inconsistency crops up. I also tried to clarify the handling of the required vs. optional parameters to the builtin IAM functions, after I sorted through all the various function interfaces (esp. adding  `interp` with required and optional model parameters and adding `schlick` without any model parameters). I realize that this is adjacent to the changes I made, but I had to understand how to fit in with this IAM code anyway, so I figured that I'd make the changes.

Question: Are folks open to switching the `sapm` function to the same interface as the other IAM functions, a breaking change?

Note that it has become too much to ask for the `ModelChain` initialization to infer the IAM function from the passed parameters in all cases. I did my best, and IMHO I think this inference should be removed altogether. (Better to just have the user be explicit.)

markcampanelli:
@cwhanse

> * `get_builtin_models` should be private. I'm not sure of the advantage of this function over a constant `dict.

Unfortunately, Python does not have true constant/immutable types, thus Python "constants" carry a risk that they will be mutated when used (often unintentionally, and causing horribly subtle bugs). By using a "getter" function instead, this mutation risk is greatly reduced. In this case, I did not make make it a private function because it gets used (like the pseudo-constant it supersedes) outside the module in which it is defined.

> * I'm not sure why the code for fitting and converting models is being edited. That seems unnecessary as the PR isn't adding to the set of models that can be fit/converted.

The name `_get_model` really misdirected me at first. It cannot be used to get ANY of the models, but only the ones that are fittable/convertible. Since it's a private function, I took the opportunity to give it a more meaningful name that indicates its restricted scope. TBH: The scoping of things in this module was probably the hardest thing for me to grasp as a new contributor. If you don't know the scope of things, then you don't know what you might actually be touching/breaking.

> * I'm OK with the `ModelChain` integration within this PR, but as @echedey-ls pointed out, sometimes with PRs less is more :)

Understood, but I have the present benefit of looking at this code with fresh eyes and seeing how hard it was to track down how different pieces are used in different parts, including outside this immediate Python module in the `ModelChain`. I also know that I don't really have the time to go back with a second PR, so I did it here. These are opportunities for what I call "harmonization", which I find many open source projects tend to miss (although, understandably so).

cwhanse:
I would still make `get_builtin_model` private since it is hard to envisage its use outside the internal plumbing of pvlib. I'm sure it can find _some_ use, but at this point it feels like clutter in the API. Thanks for explaining why the `fit` code is changing; makes sense to me now.

cwhanse:
@markcampanelli I think the scope of this PR is becoming too broad. Can you break the PR into several smaller bites?

markcampanelli:
> @markcampanelli I think the scope of this PR is becoming too broad. Can you break the PR into several smaller bites?

@cwhanse Here is a proposal for two separate PRs:

1. Make the low-level SAPM effective-irradiance-related functions `pvlib.iam.sapm` and `pvlib.spectrum.mismatch.spectral_factor_sapm` consistent with their other respective iam and mismatch functions by taking each polynomial-coefficient parameter explicitly as a separate function parameter instead of the `module` dictionary that contains extra info. (This will not, however, break the higher level interfaces, such as `PVSystem` or `ModelChain`.) Update `pvlb.modelchain.ModelChain.infer_aoi_model` accordingly, using `pvlib.iam._get_builtin_models`. Rename `pvlib.iam._get_model` to `pvlib.iam._get_fittable_or_convertable_model` and update comments.
2. Convert the Marion model for diffuse effective irradiance to take an arbitrary IAM model function that meets the required interface that all "built-in" IAM models also meet (due to 1 above).

Thoughts?

cwhanse:
Do we need to touch spectrum.mismatch.sapm to accomplish the end goal (allow any IAM model to be used with the Marion diffuse calculation)? If we don't need to change it, I would leave it of this initiative.

If we change both IAM and spectrum functions, I would divide #1 into two parts (separate iam and spectrum). We could get by without deprecating the old iam.sapm function since I doubt it gets used. But spectrum.mismatch.sapm probably does get used and should be deprecated until v0.12. 

Overall I support the initiative here, to normalize the interfaces for these lower-level functions.

PR #2070, opened 2024-05-27
Martinez shading factor

### LINKED ISSUE #2063: Shading model that takes into account the bypass diodes and layout of modules
To sum it up, the model [here](https://www.sciencedirect.com/science/article/abs/pii/S092702481000454X) divides the sectors of a PV module protected by bypass modules, and depending on the 2D shaded fraction and the number of shaded "blocks", it calculates a power correction modifier.

A function that implements this model under `shading.py` with an example showcasing different bypass diodes layouts. This example would need the shaded fraction calculated at #1962 

N/A.

**Additional context**
I've already posted this feature in the [mailing list](https://groups.google.com/g/pvlib-python/c/GzZEMxeqLOs/m/i8223h-JAQAJ) - this is part of my GSoC participation.

echedey-ls:
- [x] Closes #2063
 - [x] Tests added

Addresses adding the shading model described in #2063, which divides the modules into *blocks*, i.e. groups of cells protected by bypass diodes. It is applied to the unshaded output power.

Tests have been made with the original paper data, that is not explicitly in the paper.

The most difficult part of this model is documenting the blocks IMO. Please have a look at it.

- Docs links
  -
  - https://pvlib-python--2070.org.readthedocs.build/en/2070/reference/generated/pvlib.shading.direct_martinez.html
  - https://pvlib-python--2070.org.readthedocs.build/en/2070/gallery/shading/plot_martinez_shade_loss.html


pvlib/shading.py
```diff
@@ -553,3 +553,109 @@ def shaded_fraction1d(
...
+    .. versionadded:: 0.11.0
+
+    Parameters
+    ----------
+    shaded_fraction : numeric
+        Surface shaded fraction. Unitless.
+    shaded_blocks : numeric
+        Number of blocks affected by the shadow. Unitless integer.
+        If a floating point number is provided, it will be rounded up.
+    total_blocks : numeric
```
kandersolar, line 573: These parameter names don't match the function definition.  I suggest we look through other pvlib functions to see if we have a convention for how to name "number of X" parameters.  `pvlib.snow.dc_loss_nrel` has a `num_strings` parameter, for example.

pvlib/shading.py
```diff
@@ -553,3 +553,109 @@ def shaded_fraction1d(
...
+    A shading correction factor for the direct and circumsolar incident
+    irradiance of non-monolithic Silicon
+    modules and arrays with an arbitrary number of bypass diodes.
+
+    .. versionadded:: 0.11.0
+
+    Parameters
+    ----------
+    shaded_fraction : numeric
+        Surface shaded fraction. Unitless.
```
kandersolar, line 569: It may be helpful to specify that this model wants a 2-D (area) fraction, which is not always equivalent to what other pvlib functions (`pvlib.shading.shaded_fraction1d`) return.

pvlib/shading.py
```diff
@@ -553,3 +553,109 @@ def shaded_fraction1d(
...
+    2. A module with 3 bypass diodes and 1 junction box is likely to have 3
+       blocks.
+    3. A half-cut module with 3 junction boxes (split junction boxes) is
+       likely to have 3x2 blocks. The number of blocks along the longest
+       side of the module is 2 and along the shortest side is 3.
+    4. A module without bypass diodes doesn't constitute a block, but may be
+       part of one.
+
+    .. figure:: _images/Centralized_and_split_PV_junction_boxes_cesardd.jpg
+       :alt: Centralized and split PV junction boxes.
```
kandersolar, line 621: This image isn't rendering on RTD

pvlib/shading.py
```diff
@@ -553,3 +553,109 @@ def shaded_fraction1d(
...
+    >>> total_blocks = 3  # blocks along the vertical of the module
+    >>> POA_direct = 600  # W
+    >>> POA_diffuse = 80  # W
+    >>> shaded_fraction = shading.shaded_fraction1d(
+    >>>     80, 180, 90, 25,
+    >>>      collector_width=0.5, row_pitch=1, surface_to_axis_offset=0,
+    >>>      cross_axis_slope=5.711, shading_tracker_tilt=50)
+    >>> )
+    >>> shaded_blocks = np.ceil(total_blocks*shaded_fraction)
+    >>> loss_correction = shading.martinez_shade_factor()
```
kandersolar, line 644: Probably some inputs should be passed to this function :)

pvlib/shading.py
```diff
@@ -553,3 +553,109 @@ def shaded_fraction1d(
     )
 
     return np.clip(t_asterisk, 0, 1)
+
+
+def martinez_shade_factor(shaded_fraction, N_shaded_blocks, N_total_blocks):
```
kandersolar, line 558: Typically pvlib puts the model author at the end of the function name. `shade_factor_martinez`?  Or better yet, `direct_martinez`?

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,266 @@
...
+# where the two rows are along the longest side of the module.
+# We can argue that the dimensions of the system changes when you switch from
+# portrait to landscape, but for this example, we will consider it the same.
+#
+# The number of shaded blocks is calculated by rounding up the shaded fraction
+# by the number of blocks along the shaded length. So let's define the number
+# of blocks for each module configuration:
+#
+# - 1 bypass diode: 1 block
+# - 3 bypass diodes: 3 blocks (in portrait; in landscape, it would be 1)
```
kandersolar, line 162:
```suggestion
# - 3 bypass diodes: 3 blocks in landscape; 1 in portrait
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,266 @@
...
+#    Left: common module layout. Right: half-cut cells module layout.
+#    Each module has three bypass diodes. On the left, they connect cell
+#    columns 1-2, 2-3 & 3-4. On the right, they connect cell rows 1-2, 3-4 &
+#    5-6.
+#    *Source: César Domínguez. CC BY-SA 4.0, Wikimedia Commons*
+#
+# .. figure:: ../../_images/Centralized_and_split_PV_junction_boxes_cesardd.jpg
+#    :align: center
+#    :width: 75%
+#    :alt: Centralized and split PV junction boxes
```
kandersolar, line 181: I'm not sure this diagram is helpful.  I suggest either removing it, or adding some explanation of how junction boxes relate to diodes and blocks.

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,266 @@
...
+
+tracker_theta, aoi, surface_tilt, surface_azimuth = (
+    tracking_result["tracker_theta"],
+    tracking_result["aoi"],
+    tracking_result["surface_tilt"],
+    tracking_result["surface_azimuth"],
+)  # unpack for better readability
+
+# Calculate the tracking angles
+rotation_angle = pvlib.tracking.singleaxis(
```
kandersolar, line 121: Are both of these `singleaxis()` calls necessary?  Seems like one is redundant with the other.

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,266 @@
...
+locus = pvlib.location.Location(
+    latitude,
+    longitude,
+    tz="Europe/Madrid",
+    altitude=pvlib.location.lookup_altitude(latitude, longitude),
+)
+
+times = pd.date_range("2001-04-11T03", "2001-04-11T07", periods=24).union(
+    pd.date_range("2001-04-11T16", "2001-04-11T20", periods=24)
+)
```
kandersolar, line 79: Why exclude the middle of the day?

kandersolar (commented):
An initial review

kandersolar:
@williamhobbs I suspect this PR is of interest to you :)

pvlib/shading.py
echedey-ls: Changed param names to `shaded_blocks` and `total_blocks`!

If you prefer that prefix, I don't have an strong opinion on either. It seems there is no convention used here.

Examples:

```
samples_per_window : int
        Number of data points in each window

modules_per_string: int, default 1
        Number of modules per string in the array.

strings: int, default 1
        Number of parallel strings in the array.

cells_in_series : integer
        The number of cells connected in series.

ivcurve_pnts : int, optional
        Number of points in the desired IV curve. If not specified or 0, no
        points on the IV curves will be produced.

num_strings: int
        The number of parallel-connected strings along a row slant height.

snow_events : array-like
        Number of snowfall events each month. Referred to as N in [1]_. [-]

numthreads : int, optional, default 4
        Number of threads to use if how == 'numba'.

npoints : int, default 100
        Number of points for integration. [unitless]

n_pvrows: int, default 3
        Number of PV rows to consider in the PV array

max_rows : int, default 10
        Maximum number of rows to consider in front and behind the current row.

decimals : int, optional
        number of decimal places to which voltage is rounded to remove
        duplicated points. If not specified, no rounding is done.
```

pvlib/shading.py
echedey-ls: Can you give me some insight on why the `direct_` prefix?

pvlib/shading.py
echedey-ls: Removing it too as in https://github.com/pvlib/pvlib-python/pull/2070#discussion_r1623525705, thou IDK if that's a good choice. I don't have better images to describe the blocks.

pvlib/shading.py
kandersolar: Inspired by `pvlib.shading.sky_diffuse_passias`, I was thinking that shade models could be named like `f"{irradiance_component}_{author}"`, where `irradiance_component` is `direct`, `sky_diffuse`, etc.

echedey-ls: Ah! That makes sense. My bad, I didn't update the example accordingly to reflect the modifier is on an irradiance component and not on the output power - a confusion I had some time ago.

docs/examples/shading/plot_martinez_shade_loss.py
echedey-ls: To save some computation time while preserving the output. It doesn't get shaded during those hours. I ridiculously microoptimize too much. Do you think its impact on readability is too significant? I don't have a problem with changing it.

docs/examples/shading/plot_martinez_shade_loss.py
echedey-ls: Definitely not, maybe it was Copilot, but it was stupid me for sure.

docs/examples/shading/plot_martinez_shade_loss.py
kandersolar: I think the benefit (save a few ms at runtime) is not worth the cost (lose ~10s at "readtime") :)

cwhanse (commented):
I haven't edited text in the example to change from "irradiance loss" to "power loss"

echedey-ls:
@cwhanse now I clearly see why I was mistaken at first. I'm afraid you are a victim too 🔪 

Look at the following equations from the paper:
- (2) $P_S = P_{NS} \left(1-F_{ES}\right)$
- (8) $\left(1-\frac{P_S}{P_{NS}}\right) = \left(1 - \frac{\left[(B+C^{DIR})(1-F_{ES}) + D^{ISO}+R\right]}{G} \right)$

Then the model:
- (6) $\left(1-F_{ES}\right) = \left(1-F_{GS}\right) \left(1 - \frac{N_{SB}}{N_{TB}+1} \right)$

It's difficult to understand if this $\left(1-F_{ES}\right)$ refers to the first one (power loss) or the second one (direct irradiance loss).

I think this was a poor notation decision taken by the authors, or a way to express the ambiguity of where to use the model. Nevertheless, they state in the paragraph before (7), which to be fair, makes more sense, so the power loss "depends" on the diffuse fraction:

> In practical terms, $F_{ES}$ must only be applied to the directional components of the in-plane irradiance [16,23,24]: direct, $B$, and circumsolar part of diffuse, $D^{CIR}$. Neither isotropic diffuse, $D^{ISO}$, nor albedo, $R$, are significantly affected by shading.

And finally, I got the original spreadsheet they used to validate the model - I used it to build the test fixture. They calculate the model based on (8) but it is compared with the power loss defined as (2):
image
where $AN4 = AL4/AM4$

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,243 @@
+"""
+Modelling irradiance loss due to module shading in non-monolithic modules
```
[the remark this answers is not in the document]
echedey-ls: Commenting so we don't forget, we should check the name fits the example card.

echedey-ls:
@RDaxini & @IoannisSifnaios wanna review as well?

pvlib/shading.py
```diff
@@ -553,3 +553,101 @@ def shaded_fraction1d(
...
+    >>> POA_direct = 600  # W
+    >>> POA_diffuse = 80  # W
+    >>> shaded_fraction = shading.shaded_fraction1d(
+    >>>     80, 180, 90, 25,
+    >>>     collector_width=0.5, pitch=1, surface_to_axis_offset=0,
+    >>>     cross_axis_slope=5.711, shading_row_rotation=50)
+    >>> shaded_blocks = np.ceil(total_blocks*shaded_fraction)
+    >>> loss_correction = shading.direct_martinez(
+    >>>     shaded_fraction, shaded_blocks, total_blocks)
+    >>> POA_total = POA_direct * loss_correction + POA_diffuse
```
IoannisSifnaios, line 637:
```suggestion
    >>> import numpy as np
    >>> from pvlib import shading
    >>> total_blocks = 3  # blocks along the vertical of the module
    >>> POA_direct = 600  # W
    >>> POA_diffuse = 80  # W
    >>> # calculation of the shaded fraction for the collector
    >>> shaded_fraction = shading.shaded_fraction1d(
    >>>     solar_zenith=80, solar_azimuth=180,
     >>>     axis_azimuth=90, shaded_row_rotation=25,
    >>>     collector_width=0.5, pitch=1, surface_to_axis_offset=0,
    >>>     cross_axis_slope=5.711, shading_row_rotation=50)
    >>> # calculation of the number of shaded blocks 
    >>> shaded_blocks = np.ceil(total_blocks*shaded_fraction)
    >>> # apply the Martinez correction to the calculated shading
    >>> loss_correction = shading.direct_martinez(
    >>>     shaded_fraction, shaded_blocks, total_blocks)
    >>> # calculation of the POA for the corrected direct irradiance
    >>> POA_total = POA_direct * loss_correction + POA_diffuse
```

IoannisSifnaios: I know it is a minimal example, but still I think it is nice to give some info on the steps (just added some comments, feel free to change)

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
+"""
+Modelling irradiance loss due to of direct irradiance in modules with bypass diodes
```
IoannisSifnaios, line 2:
```suggestion
Modelling shading losses in modules with bypass diodes
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+===================================================================================
+"""  # noqa: E501
+
+# %%
+# This example illustrates how to use the proposed by Martinez et al. [1]_.
```
IoannisSifnaios, line 7:
```suggestion
# This example illustrates how to use the model proposed by Martinez et al. [1]_.
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+# The model adjusts the incident direct and circumsolar beam irradiance of a
+# PV module based on the number
+# of shaded *blocks*. A *block* is defined as a group of cells protected by a
+# bypass diode. More information on the *blocks* can be found in the original
```
IoannisSifnaios, line 11:
```suggestion
# bypass diode. More information on *blocks* can be found in the original
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+# paper [1]_ and in :py:func:`pvlib.shading.direct_martinez`
```
IoannisSifnaios, line 12:
```suggestion
# paper [1]_ and in the :py:func:`pvlib.shading.direct_martinez`
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+# documentation.
+#
+# The following key functions are used in this example:
+#
+# 1. :py:func:`pvlib.shading.direct_martinez` to calculate the adjustment
+#    factor for the output power.
```
IoannisSifnaios, line 18: Initally, we say that the model adjusts the irradiance, but here, we say that it adjusts the power output.

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+#    the trackers.
+#
+# .. sectionauthor:: Echedey Luis <echelual (at) gmail.com>
+#
+# References
+# ----------
+# .. [1] F. Martínez-Moreno, J. Muñoz, and E. Lorenzo, 'Experimental model
+#    to estimate shading losses on PV arrays', Solar Energy Materials and
+#    Solar Cells, vol. 94, no. 12, pp. 2298-2303, Dec. 2010,
+#    :doi:`10.1016/j.solmat.2010.07.029`.
```
IoannisSifnaios, line 32: I would put the references to the end

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+#
+# Problem description
+# -------------------
+# Let's consider a PV system with the following characteristics:
+#
+# - Two north-south single-axis tracker with 6 modules each one.
```
IoannisSifnaios, line 38:
```suggestion
# - Two north-south single-axis trackers, each one having 6 modules.
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+# - The rows have the same true-tracking tilt angles. Let's consider
+#   true-tracking so shade is significant for this example.
```
IoannisSifnaios, line 40:
```suggestion
# - The rows have the same true-tracking tilt angles. True tracking
#   is chosen in this example, so shading is significant.
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+
+import pvlib
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.dates import ConciseDateFormatter
+
+pitch = 4  # meters
+width = 1.5  # meters
+gcr = width / pitch
```
IoannisSifnaios, line 61:
```suggestion
gcr = width / pitch  # ground coverage ratio
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+# %%
+# Number of shaded blocks
+# -----------------------
+# The number of shaded blocks depends on the module configuration and number
+# of bypass diodes. For example,
+# modules with one bypass diode will behave like one block.
+# On the other hand, modules with three bypass diodes will have three blocks,
+# except for the half-cut cell modules, which will have six blocks; 2x3 blocks
+# where the two rows are along the longest side of the module.
+# We can argue that the dimensions of the system changes when you switch from
```
IoannisSifnaios, line 138:
```suggestion
# We can argue that the dimensions of the system change when you switch from
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+#    :width: 75%
+#    :alt: Normal and half-cut cells module layouts
+#
+#    Left: common module layout. Right: half-cut cells module layout.
+#    Each module has three bypass diodes. On the left, they connect cell
+#    columns 1-2, 2-3 & 3-4. On the right, they connect cell rows 1-2, 3-4 &
+#    5-6.
+#    *Source: César Domínguez. CC BY-SA 4.0, Wikimedia Commons*
+#
+# In the upper image, each orange U-shaped section of the string is a block.
```
IoannisSifnaios, line 162:
```suggestion
# In the image above, each orange U-shaped string section is a block.
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+    for k, blocks_N in blocks_per_module.items()
+}
+
+# %%
+# Results
+# -------
+# Now that we have the number of shaded blocks for each module configuration,
+# we can apply the model and estimate the power loss due to shading.
+#
+# Note this model is not linear with the shaded blocks ratio, so there is a
```
IoannisSifnaios, line 186:
```suggestion
# Note that this model is not linear with the shaded blocks ratio, so there is a
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,240 @@
...
+ax2.legend()
+ax2.grid()
+ax2.set_xlabel("Time")
+ax2.xaxis.set_major_formatter(
+    ConciseDateFormatter("%H:%M", tz="Europe/Madrid")
+)
+ax2.set_ylabel("Power correction factor")
+ax2.set_title("Per row")
+fig.tight_layout()
+fig.show()
```
IoannisSifnaios, line 231: I guess the block plotting the figure could be hidden (or collapsed) for making the example more compact

IoannisSifnaios (commented):
Nice work @echedey-ls! A few minor things from my side.

I think it is important to decide whether the correction factor is applied in the irradiance or power. Right now, the text is a bit mixed, mentioning both in different sections. My understanding (without being an expert on the topic) is that the factor is applied in power generation (which could be linearly related to irradiance but not necessarily), and the way it is written in the paper is a bit confusing...

docs/examples/shading/plot_martinez_shade_loss.py
echedey-ls: I prefer leaving it as is, so in case of doubt it can be consulted without digging into the repo files. If I knew how to collapse it I would do it, but I think that's impossible. In any case, I see both pros and cons regarding that suggestion, so feedback is more than welcome @AdamRJensen @kandersolar.

echedey-ls:
Thank you very much @IoannisSifnaios , this kind of help is pretty much appreciated ❣️ 

> I think it is important to decide whether the correction factor is applied in the irradiance or power. Right now, the text is a bit mixed, mentioning both in different sections.

Yup, thanks for spotting that. That was left from when I was confused about where to apply the model too. I did the magic shortcut Control+F 🪄  and now should be resolved!

kandersolar:
I think the confusion about power loss vs irradiance loss may be because we are discussing two related but distinct models.  As this PR currently stands, the function signature is this:

```python
def direct_martinez(shaded_fraction, shaded_blocks, total_blocks):
```

With these inputs, the only thing the function can calculate is the factor that is multiplied with the beam+circumsolar irradiance.  I think it is fair to call this function an irradiance loss model.  How could it calculate a power loss if it does not know about the irradiance components?  I think @echedey-ls is talking about *this* model and (fairly) considers it an irradiance loss.

However, I think the function above implements only one part of the overall model described by the reference.  I think the overall model described by the reference is indeed a power loss model.  Equation 8 is calculating a "power decrease" in such a way that just happens (due to the model's assumptions) to correspond to a decrease in direct irradiance.  A function for the overall model would look like this:

```python
def direct_martinez(poa_global, poa_direct, poa_circumsolar, shaded_fraction,
                    shaded_blocks, total_blocks):
```

I think @cwhanse is talking about *this* model and (fairly) considers it a power loss.

I think we should implement the overall model (i.e. power loss), not merely the irradiance loss portion.  I think the power loss model is the more useful of the two, plus the power loss model seems to be the reference's intent -- the text and validation are all in terms of power loss, not irradiance loss.

echedey-ls:
I couldn't have said it better. I don't see much benefit in doing such calculation when the addition of the components will be (usually) done later. Anyway I will work on it. We can always revert commits easily.

echedey-ls:
@kandersolar I'm currently working on what you told me. I don't quite understand the need of three variables here:

```python
def direct_martinez(poa_global, poa_direct, poa_circumsolar, ...
```

I understand the first two make sense from the [Variables and Symbols table](https://pvlib-python.readthedocs.io/en/stable/user_guide/variables_style_rules.html), but the third one is not specified there and I'm not familiar with pvlib widely using it. Is it reasonable to shorten it to 

```python
def direct_martinez(poa_global, poa_direct_and_circumsolar, ...
```

and let the user add them outside if they have both? I prefer that rather than a magic 0.

echedey-ls:
Ready for another review guys!! @kandersolar @cwhanse and everyone else ^.^

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,271 @@
...
+# .. sectionauthor:: Echedey Luis <echelual (at) gmail.com>
+#
+# Problem description
+# -------------------
+# Let's consider a PV system with the following characteristics:
+#
+# - Two north-south single-axis trackers, each one having 6 modules.
+# - The rows have the same true-tracking tilt angles. True tracking
+#   is chosen in this example, so shading is significant.
+#   true-tracking so shade is significant for this example.
```
kandersolar, line 34: duplicate text

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,271 @@
...
+    axis_azimuth,
+    axis_tilt=axis_tilt,
+    shaded_row_rotation=tracker_theta,
+    shading_row_rotation=tracker_theta,
+    collector_width=width,
+    pitch=pitch,
+    cross_axis_slope=cross_axis_tilt,
+)
+
+clearsky = locus.get_clearsky(
```
kandersolar, line 123: Getting the irradiance components doesn't quite fit with the section header.  Maybe give it its own header, or move it down to beginning of the `Results` section, since it's just an example weather dataset?

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,271 @@
...
+    ConciseDateFormatter("%H:%M", tz="Europe/Madrid")
+)
+ax2.set_ylabel(r"$P_{out}$ losses")
+ax2.set_title("Per row")
+fig.tight_layout()
+fig.show()
+
+# %%
+# Note how the half-cut cell module in portrait behaves worse that the
+# normal module with three bypass diodes. This is because the number of shaded
```
kandersolar, line 260:
```suggestion
# Note how the half-cut cell module in portrait performs better than the
# normal module with three bypass diodes. This is because the number of shaded
```

pvlib/tests/test_shading.py
```diff
@@ -327,3 +327,57 @@ def test_shaded_fraction1d_unprovided_shading_row_rotation():
     premises = test_data.drop(columns=["expected_sf"])
     sf = shading.shaded_fraction1d(**premises)
     assert_allclose(sf, expected_sf, atol=1e-2)
+
+
+@pytest.fixture
+def direct_martinez_Table2():
```
kandersolar, line 333: Can we add a few edge case tests also?  For example with zero irradiance, zero shading, etc

pvlib/shading.py
```diff
@@ -553,3 +553,140 @@ def shaded_fraction1d(
...
+    shaded_blocks : numeric
+        Number of blocks affected by the shadow. Unitless.
+        If a floating point number is provided, it will be rounded up.
+    total_blocks : int
+        Number of total blocks. Unitless.
+
+    Returns
+    -------
+    shading_losses : numeric
+        Multiply unshaded output power by 1 minus this value.
```
cwhanse, line 592:
```suggestion
        Fraction of DC power lost due to shading. [unitless]
```

pvlib/shading.py
```diff
@@ -553,3 +553,140 @@ def shaded_fraction1d(
...
+        \left(1 - \frac{P_{S}}{P_{NS}}\right) = \left(1 -
+        \frac{\left[(B + D^{CIR})(1 - F_{ES}) + D^{ISO} + R\right]}{G}\right)
+        \quad \text{(8)}
+
+    In (6), :math:`(1 - F_{ES})` is the correction factor to be multiplied by
+    the direct and circumsolar irradiance, :math:`F_{GS}` is the shaded
+    fraction of the collector, :math:`N_{SB}` is the number of shaded blocks
+    and :math:`N_{TB}` is the number of total blocks.
+
+    In (8), :math:`(1 - \frac{P_{S}}{P_{NS}})` is the power losses factor,
```
cwhanse, line 612:
```suggestion
    In (8), :math:`\frac{P_{S}}{P_{NS}}` is the fraction of DC power lost due to shading,
```

cwhanse (approved):

echedey-ls:
Thank you very much for your reviews @kandersolar & @cwhanse ❣️ 

From my side, I think all comments have been addressed.

pvlib/shading.py
```diff
@@ -553,3 +553,140 @@ def shaded_fraction1d(
...
+
+def direct_martinez(
+    poa_global,
+    poa_direct,
+    shaded_fraction,
+    shaded_blocks,
+    total_blocks,
+):
+    r"""
+    A shading losses power factor for non-monolithic Silicon
```
AdamRJensen, line 566:
```suggestion
    A shading loss power factor for non-monolithic silicon
```

pvlib/shading.py
```diff
@@ -553,3 +553,140 @@ def shaded_fraction1d(
...
+    modules and arrays with an arbitrary number of bypass diodes.
+
+    This experimental model cancels a fraction of the direct and circumsolar
```
AdamRJensen, line 569: I don't understand what "cancels" means in this context

pvlib/shading.py
```diff
@@ -553,3 +553,140 @@ def shaded_fraction1d(
...
+    .. versionadded:: 0.11.0
+
+    Parameters
+    ----------
+    poa_global : numeric
+        Plane of array global irradiance. [W/m²].
+    poa_direct : numeric
+        Plane of array direct and circumsolar irradiance. [W/m²].
+    shaded_fraction : numeric
+        Fraction of module surface area that is shaded. Unitless.
```
AdamRJensen, line 582:
```suggestion
        Fraction of module surface area that is shaded. [Unitless].
```

pvlib/shading.py
```diff
@@ -553,3 +553,140 @@ def shaded_fraction1d(
...
+    In (8), :math:`\frac{P_{S}}{P_{NS}}` is the fraction of DC power lost due
+    to shading, :math:`P_{S}` is the power output of the shaded module,
+    :math:`P_{NS}` is the power output of the non-shaded module,
+    :math:`B + D^{CIR}` is the beam and circumsolar irradiance,
+    :math:`D^{ISO} + R` is the sum of diffuse and albedo irradiances and
+    :math:`G` is the global irradiance.
+
+    **Blocks terminology:**
+
+    [1]_ defines a *block* as a group of solar cells protected by a bypass
```
AdamRJensen, line 621:
```suggestion
    A *block* is defined in [1]_ as a group of solar cells protected by a bypass
```

pvlib/shading.py
```diff
@@ -553,3 +553,140 @@ def shaded_fraction1d(
...
+    diode. Also, a *block* is shaded when at least one of its cells is shaded.
```
AdamRJensen, line 622: If one of its cells is "partly" shaded or "completely" shaded?

pvlib/shading.py
```diff
@@ -553,3 +553,140 @@ def shaded_fraction1d(
...
+
+    The total number of blocks and their layout depend on the module(s) used.
+    Many manufacturers don't specify this information explicitly.
+    However, we can infer these values from:
```
AdamRJensen, line 626:
```suggestion
    However, these values can be inferred from:
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,276 @@
+"""
+Modelling shading losses in modules with bypass diodes
+======================================================
+"""
+
+# %%
+# This example illustrates how to use the model proposed by Martinez et al.
+# [1]_. The model proposes a power output losses factor by adjusting
```
AdamRJensen, line 8:
```suggestion
# This example illustrates how to use the loss model proposed by Martinez et
# al. [1]_. The model proposes a power output losses factor by adjusting
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,276 @@
...
+#
+# Problem description
+# -------------------
+# Let's consider a PV system with the following characteristics:
+#
+# - Two north-south single-axis trackers, each one having 6 modules.
+# - The rows have the same true-tracking tilt angles. True tracking
+#   is chosen in this example, so shading is significant.
+# - Terrain slope is 7 degrees downward to the east.
+# - Rows' axes are horizontal.
```
AdamRJensen, line 35:
```suggestion
# - Row axes are horizontal.
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,276 @@
...
+# - The modules are comprised of multiple cells. We will compare these cases:
+#    - modules with one bypass diode
+#    - modules with three bypass diodes
+#    - half-cut cell modules with three bypass diodes on portrait and landscape
```
AdamRJensen, line 39:
```suggestion
#    - half-cut cell modules with three bypass diodes in portrait and landscape
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,276 @@
...
+
+times = pd.date_range("2001-04-11T04", "2001-04-11T20", freq="10min")
+
+# %%
+# True-tracking algorithm and shaded fraction
+# -------------------------------------------
+# Since this model is about row-to-row shading, we will use the true-tracking
+# algorithm to calculate the trackers rotation. Back-tracking reduces the
+# shading between rows, but since this example is about shading, we will not
+# use it.
```
AdamRJensen, line 76:
```suggestion
# algorithm to calculate the trackers rotation. Back-tracking eliminates
# shading between rows, and since this example is about shading, we will not
# use it.
```

docs/examples/shading/plot_martinez_shade_loss.py
```diff
@@ -0,0 +1,276 @@
...
+    clearsky["dni"],
+)
+dni_extra = pvlib.irradiance.get_extra_radiation(times)
+airmass = pvlib.atmosphere.get_relative_airmass(solar_apparent_zenith)
+sky_diffuse = pvlib.irradiance.perez_driesse(
+    surface_tilt, surface_azimuth, dhi, dni,
+    solar_apparent_zenith, solar_azimuth, dni_extra, airmass,
+)  # fmt: skip
+poa_components = pvlib.irradiance.poa_components(
+    aoi, dni, sky_diffuse, poa_ground_diffuse=0
```
AdamRJensen, line 193:
```suggestion
dni_extra = pvlib.irradiance.get_extra_radiation(times)
airmass = pvlib.atmosphere.get_relative_airmass(solar_apparent_zenith)
sky_diffuse = pvlib.irradiance.perez_driesse(
    surface_tilt, surface_azimuth, clearsky["dhi"], clearsky["dni"],
    solar_apparent_zenith, solar_azimuth, dni_extra, airmass,
)  # fmt: skip
poa_components = pvlib.irradiance.poa_components(
    aoi, clearsky["dni"], sky_diffuse, poa_ground_diffuse=0
```

AdamRJensen (approved):
Just minor details

pvlib/shading.py
echedey-ls: > A better approximation is obtained by taking into account the shaded blocks. A ‘‘block’’ is here defined as a group of cells protected by one bypass diode. A block is shaded when at least one of its cells is shaded. A first possibility is to consider that the power of a block is fully cancelled when the block is shaded.

The paper does not specify it, but from the original tests spreadsheet I confirm they are considered shaded if a minimal amount of shade impacts on them.

echedey-ls: I've added *partially* to the docstring, thou maybe I shouldn't.

AdamRJensen: I like it! Especially if you've confirmed with the spreadsheet

pvlib/shading.py
echedey-ls: I've changed the verb to "zeroes":

`     This experimental model zeroes a fraction of the direct and circumsolar`

I may have not translated that correctly. I've seen "to zero" as a verb. Should I reword it to something along `This experimental model decreases a fraction of the direct and circumsolar with each shaded block`

pvlib/shading.py
```diff
@@ -553,3 +553,141 @@ def shaded_fraction1d(
...
+    shaded_fraction,
+    shaded_blocks,
+    total_blocks,
+):
+    r"""
+    A shading loss power factor for non-monolithic silicon
+    modules and arrays with an arbitrary number of bypass diodes.
+
+    This experimental model zeroes a fraction of the direct and circumsolar
+    irradiance based on the number of *blocks* affected by the shadow.
```
cwhanse, line 570:
```suggestion
    This experimental model reduces the direct and circumsolar
    irradiance reaching the module's cells based on the number of *blocks*
    affected by the shadow.
```

pvlib/shading.py
AdamRJensen: Let's use the word reduces

> This experimental model reduces a fraction of the direct and circumsolar with each shaded block

kandersolar:
Great contribution @echedey-ls!

Issue #2068, opened 2024-05-26
Google Summer of Code: Floating PV


IoannisSifnaios:
This issue gives an overview of the plan and progress of my [GSoC ](https://summerofcode.withgoogle.com/) project.

The title of my project is "Floating Solar - Enabling modeling of floating PV with pvlib", where I will be adding floating PV temperature models in the pvlib python’s library. The official project description can be found [here](https://summerofcode.withgoogle.com/programs/2024/projects/Z32UN1Qx), and my mentors during this project will be @AdamRJensen and @kandersolar. My detailed plan is the following:

## Calculation of ambient parameters for floating PV:

- Albedo for water bodies (non-ocean) as described [here](https://ttu-ir.tdl.org/server/api/core/bitstreams/7a06ecbe-c254-44c7-9ee2-d49dbb3bd9f3/content)
- Is relevant to #1227 

## Module temperature for floating PV:

- Gallery example of the expansion of the Faiman expression for floating PV (using the methodologies from [here ](https://www.mdpi.com/1996-1073/16/20/7153)and [here](https://www.sciencedirect.com/science/article/pii/S0038092X21002085))

- Addition of the Lindholm model (described [here](https://onlinelibrary.wiley.com/doi/abs/10.1002/pip.3451))

- Addition of the Rahaman thermal model (described [here](https://www.sciencedirect.com/science/article/pii/S0960148123003713))

As the project develops, I plan to link these tasks to individual issues and pull requests.

---

Please don't hesitate to give any feedback! Since this is the first time I am contributing to pvlib (or any other open-source software, tbh), I am really looking forward to learning as much as possible from all of you!

IoannisSifnaios:
It looks like there is already a plan to add a `floating.py` module (as mentioned in #1937).

IoannisSifnaios:
My GSoC is completed, thus this issue is closed. Thanks to everyone for their assistance!

kandersolar:
Thanks @IoannisSifnaios for the great contributions!

PR #2075, opened 2024-06-04
Lindholm temperature model for floating PV

IoannisSifnaios:
- [X] Part of #2068
 - [X] Tests added

### DIFF
--- a/docs/sphinx/source/reference/pv_modeling/temperature.rst
+++ b/docs/sphinx/source/reference/pv_modeling/temperature.rst
@@ -18,6 +18,7 @@ PV temperature models
    temperature.noct_sam
    temperature.prilliman
    pvsystem.PVSystem.get_cell_temperature
+   temperature.lindholm_floating
    temperature.generic_linear
    temperature.GenericLinearModel
 
--- a/pvlib/temperature.py
+++ b/pvlib/temperature.py
@@ -1092,6 +1092,208 @@ def prilliman(temp_cell, wind_speed, unit_mass=11.1, coefficients=None):
     return smoothed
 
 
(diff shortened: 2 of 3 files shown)


kandersolar:
Taking a closer look at the references now, I'm having trouble understanding the calculations.  I think my main source of confusion is that the references use `temp_module_front` and `temp_module_back` as inputs to the calculations, i.e. they must be known ahead of time.  Is that true?  How is a user supposed to come up with those values -- measured with thermocouples on the module surfaces, or can they be modeled somehow?  I would have expected front and rear temperature to be _outputs_ of a temperature model, not _inputs_ to one.

Surely if the front and rear temperatures are known, the cell temperature could be estimated much more simply (`tcell = (tfront + tback) / 2`??) than with such a complex model as this...  What am I misunderstanding?

IoannisSifnaios:
You are correct, @kandersolar! The model takes the `temp_module_front` and `temp_module_back` as input to calculate the radiative heat loss toward the sky and water, respectively. In the paper, the authors measured these temperatures using thermocouples on some panels.

From a modeling perspective, this definitely makes things harder, as it is not easy to know these temperatures in advance. I am not sure what a solution could be; maybe use some other model to estimate these temperatures (e.g., sapm_module)? I am not an expert on this, so maybe someone else has a better idea?

At least for the case study they validated their model with, the cell temperature was not the average of the front and back module temperatures (though I guess that could be module-dependent...).

image

IoannisSifnaios:
After communicating with the model's author, the model could be used to estimate the module temperature using an iterative procedure. However, since this is not documented in any peer-reviewed journal, it was decided not to include this model in pvlib.

PR #2079, opened 2024-06-06
Add albedo function for inland water bodies

IoannisSifnaios:
- [X] Part of #2068
 - [X] Tests added

This PR adds a function for calculating the albedo of inland water bodies as described in the paper: Dvoracek M.J., Hannabas B. (1990). "Prediction of albedo for use in evapotranspiration and irrigation scheduling." IN: Visions of the Future, American Society of Agricultural Engineers 04-90: 692-699.

You can contact me if you would like me to send you the paper.


docs/sphinx/source/whatsnew/v0.11.0.rst
```diff
@@ -23,7 +23,9 @@ Enhancements
   shade perpendicular to ``axis_azimuth``. The function is applicable to both
   fixed-tilt and one-axis tracking systems.
   (:issue:`1689`, :pull:`1725`, :pull:`1962`)
-
+* Add function :py:func:`pvlib.albedo.albedo_water`, to calculate the
```
echedey-ls, line 26: I find the name a bit repetitive. If it's going to end in the `pvlib.albedo` namespace, I suggest renaming the function to `inland_water`. `albedo_water` does not exclude the ocean, and that may be a nice clarification.

pvlib/albedo.py
```diff
@@ -0,0 +1,103 @@
...
+
+    References
+    ----------
+    .. [1] Dvoracek M.J., Hannabas B. (1990). "Prediction of albedo for use in
+       evapotranspiration and irrigation scheduling." IN: Visions of the Future
+       American Society of Agricultural Engineers 04-90: 692-699.
+    """  # noqa: E501
+
+    if surface_condition is not None:
+        color_coeff = WATER_COLOR_COEFFS[surface_condition]
```
echedey-ls, line 99: Can you implement any error handling and custom error message here? In case an invalid key is passed.

pvlib/albedo.py
```diff
@@ -0,0 +1,103 @@
...
+        Sun elevation angle. [degrees]
+
+    color_coeff : float, optional
+        Water color coefficient. [-]
+
+    wave_roughness_coeff : float, optional
+        Water wave roughness coefficient. [-]
+
+    surface_condition : string, optional
+        If supplied, overrides ``color_coeff`` and ``wave_roughness_coeff``.
```
echedey-ls, line 47: Hmmm, I think it should raise an error or emit a warning, instead of silently omitting parameters (unintended behaviour of an user)?

pvlib/albedo.py
```diff
@@ -0,0 +1,103 @@
...
+        ``surface_condition`` can be one of the following:
+        'clear_water_no_waves', 'clear_water_ripples_up_to_2.5cm',
+        'clear_water_ripples_larger_than_2.5cm_occasional_whitecaps',
+        'clear_water_frequent_whitecaps', 'green_water_ripples_up_to_2.5cm',
+        'muddy_water_no_waves'.
```
echedey-ls, line 52:
```suggestion
        ``'clear_water_no_waves'``, ``'clear_water_ripples_up_to_2.5cm'``,
        ``'clear_water_ripples_larger_than_2.5cm_occasional_whitecaps'``,
        ``'clear_water_frequent_whitecaps'``, ``'green_water_ripples_up_to_2.5cm'``,
        ``'muddy_water_no_waves'``.
```

Maybe in a list it could be read easier 🤔

echedey-ls (commented):
An initial review. I like the docs redaction.

docs/sphinx/source/whatsnew/v0.11.0.rst
kandersolar: I think even more specificity is needed.  This inland water albedo model is one of many that pvlib could potentially implement, so I think we should leave room for them.  Assuming the model itself doesn't have a name (I did not see one in the reference or in any PV papers citing it), how about `inland_water_dvoracek`?  I don't like it, but at least it is specific...

pvlib/albedo.py
```diff
@@ -0,0 +1,103 @@
...
+    'clear_water_ripples_larger_than_2.5cm_occasional_whitecaps': 0.23,
+    'clear_water_frequent_whitecaps': 0.3,
+    'green_water_ripples_up_to_2.5cm': 0.22,
+    'muddy_water_no_waves': 0.19
+}
+
+WATER_ROUGHNESS_COEFFS = {
+    'clear_water_no_waves': 0.29,
+    'clear_water_ripples_up_to_2.5cm': 0.7,
+    'clear_water_ripples_larger_than_2.5cm_occasional_whitecaps': 1.23,
```
kandersolar, line 19:
```suggestion
    'clear_water_ripples_larger_than_2.5cm_occasional_whitecaps': 1.25,
```

pvlib/albedo.py
```diff
@@ -0,0 +1,103 @@
...
+    'clear_water_frequent_whitecaps': 2,
+    'green_water_ripples_up_to_2.5cm': 0.7,
+    'muddy_water_no_waves': 0.29
+}
+
+
+def albedo_water(solar_elevation, color_coeff=None, wave_roughness_coeff=None,
+                 surface_condition=None):
```
kandersolar, line 27: I suggest placing `surface_condition` before `color_coeff` and `wave_roughness_coeff`.  I think most users of this function will be using the pre-determined coefficients, so we should not "hide" the relevant parameter at the end of the list.

pvlib/albedo.py
kandersolar: a bulleted list like this may be a good option: https://github.com/pvlib/pvlib-python/blob/main/pvlib/iam.py#L667-L675

pvlib/albedo.py
```diff
@@ -0,0 +1,103 @@
...
+    | Clear water ripples larger than 2.5 cm (occasional whitecaps) | 0.23                          | 1.23                                   |
+    +---------------------------------------------------------------+-------------------------------+----------------------------------------+
+    | Clear water frequent whitecaps                                | 0.30                          | 2.00                                   |
+    +---------------------------------------------------------------+-------------------------------+----------------------------------------+
+    | Green water ripples up to 2.5cm                               | 0.22                          | 0.70                                   |
+    +---------------------------------------------------------------+-------------------------------+----------------------------------------+
+    | Muddy water no waves                                          | 0.19                          | 0.29                                   |
+    +---------------------------------------------------------------+-------------------------------+----------------------------------------+
```
kandersolar, line 89:
```suggestion
    +------------------------+-------------------+-------------------------+
    | Surface and condition  | Color coefficient | Wave roughness          |
    |                        | (:math:`c`)       | coefficient (:math:`r`) |
    +========================+===================+=========================+
    | Clear water, no waves  | 0.13              | 0.29                    |
    +------------------------+-------------------+-------------------------+
    | Clear water, ripples   | 0.16              | 0.70                    |
    | up to 2.5 cm           |                   |                         |
    +------------------------+-------------------+-------------------------+
    | Clear water, ripples   | 0.23              | 1.25                    |
    | larger than 2.5 cm     |                   |                         |
    | (occasional whitecaps) |                   |                         |
    +------------------------+-------------------+-------------------------+
    | Clear water,           | 0.30              | 2.00                    |
    | frequent whitecaps     |                   |                         |
    +------------------------+-------------------+-------------------------+
    | Green water, ripples   | 0.22              | 0.70                    |
    | up to 2.5cm            |                   |                         |
    +------------------------+-------------------+-------------------------+
    | Muddy water, no waves  | 0.19              | 0.29                    |
    +------------------------+-------------------+-------------------------+
```

kandersolar: This alternative form should render the same in the HTML (assuming I have not made a mistake), but is much nicer to look at in IDEs and such.

pvlib/albedo.py
```diff
@@ -0,0 +1,103 @@
...
+    .. [1] Dvoracek M.J., Hannabas B. (1990). "Prediction of albedo for use in
+       evapotranspiration and irrigation scheduling." IN: Visions of the Future
+       American Society of Agricultural Engineers 04-90: 692-699.
+    """  # noqa: E501
+
+    if surface_condition is not None:
+        color_coeff = WATER_COLOR_COEFFS[surface_condition]
+        wave_roughness_coeff = WATER_ROUGHNESS_COEFFS[surface_condition]
+
+    albedo = color_coeff ** (wave_roughness_coeff * sind(solar_elevation) + 1)
```
kandersolar, line 102: What to do about negative solar elevation?  Clip elevation to zero before this calculation?  Set albedo to nan at night?

kandersolar (commented):
A few comments.  Nice and clean code!

pvlib/albedo.py
AdamRJensen: I guess you can do a check that if `surface_condition` is not None, then the other two parameters should be none, and vice versa.

pvlib/albedo.py
AdamRJensen: I like the bulleted list. Remember that you need the vertical blank space for the list to render correctly.

pvlib/albedo.py
```diff
@@ -0,0 +1,110 @@
...
+        \rho = c^{(r sin(\alpha) + 1)}
+
+    Inputs to the model are the water color coefficient :math:`c` [-], the
+    water wave roughness coefficient :math:`r` [-] and the solar elevation
+    :math:`\alpha` [degrees]. Parameters are provided in [1]_ , and are coded
+    for convenience in :data:`~pvlib.albedo.WATER_COLOR_COEFFS` and
+    :data:`~pvlib.albedo.WATER_ROUGHNESS_COEFFS`. The values of these
+    coefficients are experimentally determined.
+
+    +---------------------------------------------------------------+-------------------------------+----------------------------------------+
```
AdamRJensen, line 75:
```suggestion
```
I think this is left over from the early formatting of the table?

pvlib/albedo.py
AdamRJensen: If an incorrect key is specified, the dictionary will raise a KeyError. I think that should be helpful enough and the user can consult the list of `surface_conditions` in the documentation. So I vote not adding any custom error message for invalid keys.

pvlib/albedo.py
AdamRJensen: I think we should do something along those lines. However, it's going to be a bit of data wrangling with the various input types (ints, floats, series). Maybe something along the lines of:

    if len(solar_elevation) > 1:
        if solar_elevation<0:
            solar_elevation = 0
    else:
        solar_elevation[solar_elevation<0] = 0

cwhanse (approved):
I have no comments beyond those made by other reviewers. Approving pending resolution of other comments.

pvlib/albedo.py
IoannisSifnaios: Solar elevation was clipped at zero.

pvlib/albedo.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def inland_water_dvoracek(solar_elevation, surface_condition=None,
+                          color_coeff=None, wave_roughness_coeff=None):
+    r"""
+    Estimation of albedo for inland water bodies.
+
+    The available surface conditions are for inland water bodies, e.g., lakes
+    and ponds. For ocean/open sea, an albedo value of 0.06 is recommended.
+    See :const:`pvlib.irradiance.SURFACE_ALBEDOS`.
```
AdamRJensen, line 34:
```suggestion
    and ponds. For ocean/open sea, see
    :const:`pvlib.irradiance.SURFACE_ALBEDOS`.
```
In case the 0.06 value changes at some point, it's probably preferred to just give the reference.

pvlib/albedo.py
```diff
@@ -0,0 +1,120 @@
...
+    color_coeff : float, optional
+        Water color coefficient. [-]
+
+    wave_roughness_coeff : float, optional
+        Water wave roughness coefficient. [-]
+
+    Returns
+    -------
+    albedo : numeric
+        Albedo for inland water bodies.
```
AdamRJensen, line 61:
```suggestion
        Albedo for inland water bodies. [-]
```

pvlib/albedo.py
```diff
@@ -0,0 +1,120 @@
...
+
+    Notes
+    -----
+    The equation for calculating the albedo :math:`\rho` is given by
+
+    .. math::
+       :label: albedo
+
+        \rho = c^{(r sin(\alpha) + 1)}
```
AdamRJensen, line 70:
```suggestion
        \rho = c^{(r \cdot \sin(\alpha) + 1)}
```
\sin will make "sin" non-italic as it's not a variable

pvlib/albedo.py
```diff
@@ -0,0 +1,120 @@
...
+    .. [1] Dvoracek M.J., Hannabas B. (1990). "Prediction of albedo for use in
+       evapotranspiration and irrigation scheduling." IN: Visions of the Future
+       American Society of Agricultural Engineers 04-90: 692-699.
+    """
+
+    if surface_condition is None:
+        if (color_coeff is None) or (wave_roughness_coeff is None):
+            raise ValueError('Either a surface_condition has to be chosen or a'
+                             'combination of both color and wave roughness'
+                             'coefficients.')
```
AdamRJensen, line 112:
```suggestion
            raise ValueError('Either a `surface_condition` has to be chosen or a'
                             'combination of `color_coeff` and'
                             '`wave_roughness_coeffs`.')
```

pvlib/tests/test_albedo.py
```diff
@@ -0,0 +1,56 @@
...
+    result = albedo.inland_water_dvoracek(solar_elevation=solar_elevs,
+                                          color_coeff=color_coeffs,
+                                          wave_roughness_coeff=roughness_coeffs)   # noqa: E501
+    expected = pd.Series([0.1, 0.1, 0.12875, 0.06278, 0.064], index=times)
+    assert_series_equal(expected, result, atol=1e-5)
+
+
+def test_albedo_water_invalid():
+    with pytest.raises(ValueError):
+        albedo.inland_water_dvoracek(45)
```
AdamRJensen, line 56: I'd probably make this test a bit more explicit. E.g., test that a ValueError is raised when passing 'not_a_surface_condition' to `surface_conditions`.

Also, test that a ValueError is raised when only the `color_coeff` is specified and `surface_roughness_coeff` is not. These tests can be under the same function (test_albedo_water_invalid) but with more `with pytests..` statements after each other.

AdamRJensen (approved):

pvlib/albedo.py
```diff
@@ -0,0 +1,120 @@
...
+    .. [1] Dvoracek M.J., Hannabas B. (1990). "Prediction of albedo for use in
+       evapotranspiration and irrigation scheduling." IN: Visions of the Future
+       American Society of Agricultural Engineers 04-90: 692-699.
+    """
+
+    if surface_condition is None:
+        if (color_coeff is None) or (wave_roughness_coeff is None):
+            raise ValueError('Either a `surface_condition` has to be chosen or'
+                             ' a combination of `color_coeff` and'
+                             '`wave_roughness_coeff`.')
```
kandersolar, line 112:
```suggestion
            raise ValueError('Either a `surface_condition` has to be chosen or'
                             ' a combination of `color_coeff` and'
                             ' `wave_roughness_coeff`.')
```

pvlib/tests/test_albedo.py
```diff
@@ -0,0 +1,65 @@
...
+    result = albedo.inland_water_dvoracek(solar_elevation=solar_elevs,
+                                          color_coeff=color_coeffs,
+                                          wave_roughness_coeff=roughness_coeffs)   # noqa: E501
+    expected = pd.Series([0.1, 0.1, 0.12875, 0.06278, 0.064], index=times)
+    assert_series_equal(expected, result, atol=1e-5)
+
+
+def test_albedo_water_invalid():
+    with pytest.raises(ValueError):  # no surface info given
+        albedo.inland_water_dvoracek(solar_elevation=45)
```
kandersolar, line 56: Can we use `match=` in each of these `pytest.raises` calls to ensure that the correct error message is being raised (in addition to the correct error _type_)?  For example, this one could be ``` with pytest.raises(ValueError, match="'Either a `surface_condition`"): ```

pvlib/albedo.py
```diff
@@ -0,0 +1,120 @@
...
+    surface_condition : string, optional
+        If supplied, overrides ``color_coeff`` and ``wave_roughness_coeff``.
+        ``surface_condition`` can be one of the following:
+
+        * 'clear_water_no_waves'
+        * 'clear_water_ripples_up_to_2.5cm'
+        * 'clear_water_ripples_larger_than_2.5cm_occasional_whitecaps'
+        * 'clear_water_frequent_whitecaps'
+        * 'green_water_ripples_up_to_2.5cm'
+        * 'muddy_water_no_waves'.
```
echedey-ls, line 50:
```suggestion
        * ``'clear_water_no_waves'``
        * ``'clear_water_ripples_up_to_2.5cm'``
        * ``'clear_water_ripples_larger_than_2.5cm_occasional_whitecaps'``
        * ``'clear_water_frequent_whitecaps'``
        * ``'green_water_ripples_up_to_2.5cm'``
        * ``'muddy_water_no_waves'``
```
I prefer code formatting for code constants

pvlib/albedo.py
```diff
@@ -0,0 +1,120 @@
...
+       evapotranspiration and irrigation scheduling." IN: Visions of the Future
+       American Society of Agricultural Engineers 04-90: 692-699.
+    """
+
+    if surface_condition is None:
+        if (color_coeff is None) or (wave_roughness_coeff is None):
+            raise ValueError('Either a `surface_condition` has to be chosen or'
+                             ' a combination of `color_coeff` and'
+                             '`wave_roughness_coeff`.')
+    else:
```
echedey-ls, line 113: It may be and edge-case, but what should happen if all the optional parameters are provided? I prefer an exception (or a warning) instead of any parameter being ignored.
Bonus case: what should happen if all optional parameters are not passed a value?

I would change the if blocks to something along the following snippet, IDK what the other reviewers think about it @kandersolar @AdamRJensen 

```python
    if surface_condition is not None and (
        color_coeff is None and wave_roughness_coeff is None
    ):
        # use surface_condition
        pass
    elif surface_condition is None and not (
        color_coeff is None or wave_roughness_coeff is None
    ):
        # use color_coeff and wave_roughness_coeff
        pass
    else:
        raise ValueError(
            "Either a `surface_condition` has to be chosen or"
            " a combination of `color_coeff` and"
            "`wave_roughness_coeff`."
        )
```

PD: sorry for this suggestion at the current progress of the PR

pvlib/albedo.py
```diff
@@ -0,0 +1,120 @@
...
+        color_coeff = WATER_COLOR_COEFFS[surface_condition]
+        wave_roughness_coeff = WATER_ROUGHNESS_COEFFS[surface_condition]
+
+    solar_elevation = np.where(solar_elevation < 0, 0, solar_elevation)
+
+    albedo = color_coeff ** (wave_roughness_coeff * sind(solar_elevation) + 1)
+    return albedo
```
kandersolar, line 120: It's possible to get an array out despite putting a Series in:

```python
In [399]: inland_water_dvoracek(pd.Series([45]), "muddy_water_no_waves")
Out[399]: array([0.13516185])
```

This is not an uncommon problem when using numpy tricks like `np.where`.  Our usual workaround is to fix up the types at the end of the function like this:

```python
    if isinstance(solar_elevation, pd.Series):
        albedo = pd.Series(albedo, index=solar_elevation.index)

    return albedo
```

A little inelegant, but it gets the job done.

kandersolar (commented):
A few last comments :)

echedey-ls (commented):
Minor comments. Merging feels close 🚀

pvlib/albedo.py
AdamRJensen: @echedey-ls looks sleek to me!

pvlib/albedo.py
AdamRJensen: Wouldn't the type check of int/floats I suggested earlier do the job?

kandersolar: Type checking has some tricky edge cases (e.g. `isinstance(np.array([45.])[0], (int, float))` is `False` because it is an `np.float64`, not a `float`).  A test based on `np.isscalar` might be ok.  

But I think it is generally better to fix up types at the end instead of following different calculation paths based on type.  Less chance of accidentally calculating different output values for different types.

pvlib/albedo.py
IoannisSifnaios: The way it is written now if you only give the solar_elevation and omit the rest, you get the ValueError

pvlib/tests/test_albedo.py
```diff
@@ -0,0 +1,84 @@
+import numpy as np
+import pandas as pd
+import pytest
+from pvlib import albedo
+
+from .conftest import assert_series_equal
+from numpy.testing import assert_allclose
+
+
+def test_albedo_water_default():
```
echedey-ls, line 10:
```suggestion
def test_inland_water_dvoracek_default():
```
And all the remaining

pvlib/albedo.py
IoannisSifnaios: @kandersolar the problem is that `np.where` returns an array (even if you pass a pd.Series for the solar_elevation) - so the if statement is never true. I did a fix, but it might not be the prettiest one... can you approve it (or suggest an alternative)?

pvlib/albedo.py
```diff
@@ -0,0 +1,137 @@
...
+    :const:`pvlib.irradiance.SURFACE_ALBEDOS`.
+
+    Parameters
+    ----------
+    solar_elevation : numeric
+        Sun elevation angle. [degrees]
+
+    surface_condition : string, optional
+        If supplied, overrides ``color_coeff`` and ``wave_roughness_coeff``.
+        ``surface_condition`` can be one of the following:
```
echedey-ls, line 45:
```suggestion
        One of the following:
```

pvlib/albedo.py
```diff
@@ -0,0 +1,137 @@
...
+        Water color coefficient. [-]
+
+    wave_roughness_coeff : float, optional
+        Water wave roughness coefficient. [-]
+
+    Returns
+    -------
+    albedo : numeric
+        Albedo for inland water bodies. [-]
+
```
echedey-ls, line 64:
```suggestion

    Raises
    ------
    ValueError
        If neither of ``surface_condition`` nor the pair ``color_coeff`` and ``wave_roughness_coeff`` are given.
        If ``surface_condition`` and any of ``color_coeff`` or ``wave_roughness_coeff`` are given.
        These parameters are mutually exclusive.

    KeyError
        If ``surface_condition`` is invalid.

```
Feel free to use a more natural language.
Maybe you can note the exclusiveness of the optional parameters elsewhere too.

echedey-ls (approved):
LGTM with the changes down below 🚀

kandersolar (approved):
LGTM.  Thanks @IoannisSifnaios and everyone that reviewed!

PR #2104, opened 2024-06-22
stream_temperature_function

IoannisSifnaios:
- [X] Related to #2068
 - [X] Tests added

Addition of the Stefan and Preud'homme equation for estimating the mean daily stream water temperature using the ambient air temperature as input. The paper can be found [here](https://doi.org/10.1111/j.1752-1688.1993.tb01502.x).

Although this equation was created for estimating stream temperature, there are a number of publications that have used it for estimating the lakewater and seawater temperature (not sure if this is a good approach, though). From a quick search, some indicative publications that used it for estimating water temperature for floating PV are: [1](https://doi.org/10.1016/j.renene.2023.03.076), [2](https://doi.org/10.1016/j.renene.2021.04.082), [3](https://doi.org/10.1016/j.heliyon.2022.e11896), [4](https://ieeexplore.ieee.org/document/9600926), [5](https://ieeexplore.ieee.org/abstract/document/9725955).

[Doc rendering](https://pvlib-python--2104.org.readthedocs.build/en/2104/reference/generated/pvlib.floating.stream_temperature_stefan.html)


IoannisSifnaios:
This function was not tested for very low mean daily ambient temperatures. Thus, if air temperatures lower than -6.6 degC are provided, the water temperature from this function will be negative. This can be true in some cases (e.g. if there is supercooling), but under normal conditions, it is not possible. So the question is: Should we limit the water surface temperature to 0 or return np.nan when low ambient air temperatures are used?

pvlib/tests/test_floating.py
```diff
@@ -0,0 +1,33 @@
...
+    air_temps = np.array([-5, 2.5, 20, 30, 40])
+    result = floating.daily_stream_temperature_stefan(temp_air=air_temps)
+    expected = np.array([1.25, 6.875, 20, 27.5, 35])
+    assert_allclose(expected, result, atol=1e-3)
+
+
+def test_daily_stream_temperature_stefan_series():
+    times = pd.date_range(start="2015-01-01 00:00", end="2015-01-05 00:00",
+                          freq="1d")
+    air_temps = pd.Series([-5, 2.5, 20, 30, 40], index=times)
```
echedey-ls, line 29: Both these lines plus the `expected = ...` down below and lines 20-21:

```python
    air_temps = np.array([-5, 2.5, 20, 30, 40])
    result = floating.daily_stream_temperature_stefan(temp_air=air_temps)
```

repeat the same values. You can try to use a fixture (a pytest.fixture). There are plenty of examples in the repo, for example this one:
https://github.com/pvlib/pvlib-python/blob/55e0a36fac1e3931b92f5f11508f6c09d903e0cc/pvlib/tests/test_irradiance.py#L68C1-L70C74

It would be great if you return both the input and the expected output from the function. [pytest.fixture docs](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html).

Then, for the numpy types testing, use the numpy array view of the series, something along `pd.Series().values` or `.to_numpy()` (from the top of my head, if it gives you an error you won't have trouble finding what to use in the documentation).

Also, instead of testing the positive clipping of the output independently, you can add that to this data.

echedey-ls (commented):
I can't access the paper so this is a bit of a blind review, but here are some initial ideas.

As always, feel free to make (or not) the changes with your own style.

IoannisSifnaios:
Thanks @echedey-ls and @AdamRJensen for all the help with the `pytest.fixture`.

Is there a way to use `fixture` values as input to a `pytest.mark.parametrize` test?

cwhanse:
Looks good so far @IoannisSifnaios

pvlib/floating.py
```diff
@@ -0,0 +1,66 @@
...
+"""
+
+import numpy as np
+import pandas as pd
+
+
+def stream_temperature_stefan(temp_air):
+    r"""
+    Estimation of daily stream water temperature based on daily ambient
+    temperature.
```
cwhanse, line 13:
```suggestion
    Estimate daily stream water temperature from daily ambient air temperature.
```

pvlib/floating.py
```diff
@@ -0,0 +1,66 @@
...
+    Returns
+    -------
+    water_temperature : numeric
+        Daily average stream water temperature. [°C]
+
+    Notes
+    -----
+    The equation for calculating the daily average stream water temperature
+    :math:`T_w` from the daily ambient air temperature :math:`T_{air}` is
+    provided in [1]_ and given by:
```
cwhanse, line 29:
```suggestion
    Daily average stream water temperature
    :math:`T_w` is calculated from daily ambient air temperature :math:`T_{air}` as
    provided in [1]_:
```

pvlib/floating.py
```diff
@@ -0,0 +1,66 @@
...
+
+    .. math::
+       :label: stream
+
+        T_w = 5 + 0.75 \cdot T_{air}
+
+    The predicted daily stream water temperatrues of this equation had a
+    standard deviation of 2.7 °C compared to measurements. Small, shallow
```
cwhanse, line 37:
```suggestion
    The predicted daily stream water temperatures had a
    standard deviation of 2.7 °C compared to measurements. Small, shallow
```

pvlib/floating.py
```diff
@@ -0,0 +1,66 @@
...
+    streams had smaller deviations than large, deep rivers.
+
+    It should be noted that this equation is limited to streams, i.e., water
+    bodies that are well mixed in vertical and transverse direction of a cross
```
cwhanse, line 41:
```suggestion
    bodies that are well mixed in the vertical and transverse directions of a cross
```

pvlib/floating.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Warning
+    -------
+    The expression has been developed for inland streams and is thus not
+    suitable for estimating ocean temperature.
+
+    References
+    ----------
+    .. [1] Stefan H. G., Preud'homme E. B. (1993). "Stream temperature
+       estimation from air temperature." IN: Journal of the American Water
```
cwhanse, line 54:
```suggestion
       estimation from air temperature." Journal of the American Water
```

kandersolar:
> Although this equation was created for estimating stream temperature, there are a number of publications that have used it for estimating the lakewater and seawater temperature (not sure if this is a good approach, though).

Perhaps it is worth noting this in the function's docstring?

pvlib/floating.py
```diff
@@ -0,0 +1,70 @@
...
+    It should be noted that this equation is limited to streams, i.e., water
+    bodies that are well mixed in the vertical and transverse directions of a
+    cross section. Also, it is only tested on ice-free streams. Consequently,
+    when the mean ambient air temperature is lower than -6 °C, the surface
+    stream water temperature is returned as NaN.
+
+    Also, although this equation was created for estimating stream temperature,
+    there are a number of publications that have used it for estimating the
+    lakewater and seawater temperature for floating PV, without having any
+    proof that this is a legitimate approach.
```
AdamRJensen, line 48:
```suggestion
    While this equation was created for estimating stream temperature,
    there are a number of publications that have used it for estimating the
    lakewater and seawater temperature for floating PV, although without
    providing any proof of the formula's validity for such purposes.
```

AdamRJensen (approved):

pvlib/floating.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Warning
+    -------
+    The expression has been developed for inland streams and is thus not
+    suitable for estimating ocean temperature.
+
```
adriesse, line 54:
```suggestion
    
```
This is covered by the notes.

pvlib/floating.py
```diff
@@ -0,0 +1,70 @@
...
+        T_w = 5 + 0.75 \cdot T_{air}
+
+    The predicted daily stream water temperatures had a
+    standard deviation of 2.7 °C compared to measurements. Small, shallow
+    streams had smaller deviations than large, deep rivers.
+
+    It should be noted that this equation is limited to streams, i.e., water
+    bodies that are well mixed in the vertical and transverse directions of a
+    cross section. Also, it is only tested on ice-free streams. Consequently,
+    when the mean ambient air temperature is lower than -6 °C, the surface
```
adriesse, line 42:
```suggestion
    when the mean ambient air temperature is lower than -6.66 °C, the surface
```

pvlib/floating.py
```diff
@@ -0,0 +1,70 @@
...
+       Resources Association 29-1: 27-45.
+       :doi:`10.1111/j.1752-1688.1993.tb01502.x`
+    """
+
+    temp_stream = 5 + 0.75 * temp_air
+
+    temp_stream = np.where(temp_stream < 0, np.nan, temp_stream)
+
+    if isinstance(temp_air, pd.Series):
+        temp_stream = pd.Series(temp_stream, index=temp_air.index)
```
adriesse, line 68:
```suggestion
    temp_stream[temp_stream < 0] = np.nan

```
Pretty sure this will work.

pvlib/floating.py
```diff
@@ -0,0 +1,70 @@
...
+    -----
+    Daily average stream water temperature
+    :math:`T_w` is calculated from daily ambient air temperature
+    :math:`T_{air}` as provided in [1]_:
+
+    .. math::
+       :label: stream
+
+        T_w = 5 + 0.75 \cdot T_{air}
+
```
adriesse, line 34: I like the notes on validity.

adriesse (changes requested):

adriesse:
Thinking about this again, it seems unlikely that floating PV will be built on the type of water body modeled here, so are we encouraging the possible misuse for other types of bodies of water?  I would suspect that a more suitable model for lakes and larger bodies exists somewhere.

pvlib/floating.py
```diff
@@ -0,0 +1,70 @@
...
+    The predicted daily stream water temperatures had a
+    standard deviation of 2.7 °C compared to measurements. Small, shallow
+    streams had smaller deviations than large, deep rivers.
+
+    It should be noted that this equation is limited to streams, i.e., water
+    bodies that are well mixed in the vertical and transverse directions of a
+    cross section. Also, it is only tested on ice-free streams. Consequently,
+    when the mean ambient air temperature is lower than -6.66 °C, the surface
+    stream water temperature is returned as NaN.
```
RDaxini, line 43:
```suggestion
    stream water temperature is returned as ``np.nan``.
```

pvlib/floating.py
```diff
@@ -0,0 +1,70 @@
...
+
+    While this equation was created for estimating stream temperature,
+    there are a number of publications that have used it for estimating the
+    lakewater and seawater temperature for floating PV, although without
+    providing any proof of the formula's validity for such purposes.
```
RDaxini, line 48: Nothing really wrong with the original, but I was just wondering whether we could make this sentence more concise. This is not a 'strong' suggestion though, feel free to ignore it. Please keep your version if you prefer.
```suggestion
    Although this model was developed to estimate stream temperature,
    a number of publications have used it to estimate lakewater and
    seawater temperatures for floating PV without evidence of validity
    for such applications.
```

pvlib/floating.py
IoannisSifnaios: The use of np.where was done such that the function also works with floats, whereas the ``temp_stream[temp_stream<0]`` wouldn't work with floats.

pvlib/floating.py
```diff
@@ -0,0 +1,66 @@
...
+        Daily average stream water temperature. [°C]
+
+    Notes
+    -----
+    Daily average stream water temperature
+    :math:`T_w` is calculated from daily ambient air temperature
+    :math:`T_{air}` as provided in [1]_:
+
+    .. math::
+       :label: stream
```
echedey-ls, line 31:
```suggestion
```

Unnecessary unless you want to reference it, or need automatic equation numbering - does not seem to be the case.

echedey-ls (approved):
I can't access the paper, but this proposal looks perfectly fine. Nice job!

Non-blocking comments down below.

IoannisSifnaios:
> Thinking about this again, it seems unlikely that floating PV will be built on the type of water body modeled here, so are we encouraging the possible misuse for other types of bodies of water? I would suspect that a more suitable model for lakes and larger bodies exists somewhere.

@adriesse I also thought about this (and asked some oceanography researchers regarding models for sea temperature), but there are no simple models that can estimate this, which I think is the reason why people used this model in the first place (although it might not be very appropriate). Thus, since we have mentioned very explicitly the limitations of the model, I still think it makes sense to add it in lack of a more appropriate model.

adriesse:
> > Thinking about this again, it seems unlikely that floating PV will be built on the type of water body modeled here, so are we encouraging the possible misuse for other types of bodies of water? I would suspect that a more suitable model for lakes and larger bodies exists somewhere.
> 
> @adriesse I also thought about this (and asked some oceanography researchers regarding models for sea temperature), but there are no simple models that can estimate the sea surface temperature, which I think is the reason why people used this model in the first place (although it might not be very appropriate). Thus, since we have mentioned the limitations of the model very explicitly, I still think it makes sense to add it in lack of a more appropriate model. Hopefully, in the future, some simple model might be developed, and then it could be directly compared to this one in order to estimate how "off" the results are...

I haven't read these papers, but it sounds pretty questionable to take a model out of one context and use it in another context without some kind of checks or validation.  If we put that in pvlib it's like giving someone a hammer when you know they need a screwdriver.

I suggest putting in an empty `lake_temperature` function that raises a `NotImplementedError` instead.  Maybe that will encourage people to look into it!!

cwhanse:
@adriesse @IoannisSifnaios it's been a while - can you recap where we are with this PR and its reviews?

IoannisSifnaios:
> @adriesse @IoannisSifnaios it's been a while - can you recap where we are with this PR and its reviews?

I have implemented all the suggested review comments. However, we haven't reached an agreement regarding whether this function should be included in pvlib. 

The main argument for not including it is that it calculates the temperature of moving water bodies (i.e., streams), which is an unlikely choice for installing a floating PV system. As Anton mentioned, some lake or sea temperature models would be more appropriate.

The main argument for including it is that I could not find functions that estimate lake or sea temperature using ambient measurements and that this expression has been used in some publications (see description of the PR). So if we decide to include it in pvlib, it is with a warning about its usage and limitations.

I think we should maybe put it to a vote and possibly close this PR if we agree not to include this function.

docs/sphinx/source/whatsnew/v0.11.1.rst
```diff
@@ -10,6 +10,12 @@ Deprecations
 
 Enhancements
 ~~~~~~~~~~~~
+* Add function :py:func:`pvlib.floating.stream_temperature_stefan`, to calculate the
+  water temperature for inland water bodies.
```
cwhanse, line 14:
```suggestion
  water temperature for streams and rivers.
```

pvlib/floating.py
```diff
@@ -0,0 +1,64 @@
...
+    standard deviation of 2.7 °C compared to measurements. Small, shallow
+    streams had smaller deviations than large, deep rivers.
+
+    It should be noted that this equation is limited to streams, i.e., water
+    bodies that are well mixed in the vertical and transverse directions of a
+    cross section. Also, it is only tested on ice-free streams. Consequently,
+    when the mean ambient air temperature is lower than -6.66 °C, the surface
+    stream water temperature is returned as ``np.nan``.
+
+    Although this model was developed to estimate stream temperature,
```
cwhanse, line 44:
```suggestion
..warning::
    This model was developed to estimate stream temperature. A number of
```

cwhanse (approved):
I think the function has place in the library with the appropriate caveats (a warning about application).

RDaxini:
Since there's no concern about the technical soundness of the model itself and it is a model that is used by the PV modelling community, then as a library of models for the modelling community I think it makes sense to include it.

As for concerns about its application, I think the function is clearly labelled for what it is so users are adequately informed of the relevant caveats concerning its application.

cwhanse:
@IoannisSifnaios I see a number of papers referring lake surface water temperature (LWST) models, or sea surface temperature (SST) models. Most models appear to be dynamic; basically, future temperature is a function of past temperature and exogenous variables. Are these models more complex that your application needs?

This discussion (about water surface temperature models for non-flowing water bodies) probably needs its own issue.

RDaxini (approved):

IoannisSifnaios:
> @IoannisSifnaios I see a number of papers referring lake surface water temperature (LWST) models, or sea surface temperature (SST) models. Most models appear to be dynamic; basically, future temperature is a function of past temperature and exogenous variables. Are these models more complex that your application needs?
> 
> This discussion (about water surface temperature models for non-flowing water bodies) probably needs its own issue.

@cwhanse SST models are quite complicated and usually require solving turbulence equations etc. (I actually had a discussion with an oceanographer about this...). Regarding lakes, if you found a paper with a suitable model, please mention it in an issue, and I can take it from there! ;-)

adriesse:
I'm not yet convinced that this is a good idea. Whether we like it or not, inclusion in pvlib imparts a certain credibility and also increases visibility.  This means that we have to be careful about what we include, whether we like it or not.

Building PV over running water seems to be a niche application, so strictly speaking this model function would benefit few people.  On the other hand there will be lots of people building PV over larger water bodies, and if this is the only model they find, they will likely use it despite it not having been validated for their application.  I don't see that as a benefit to anyone. (The situation is similar with the wind speed height model that was recently merged.)

From the abstract, it appears that the empirical model was derived from measurements on 11 streams in one geographic area only, and ambient temperatures were measured between 0 and 144 miles from those streams.  If someone has used it for a lake in India (one of the references), does that bolster the case _for_ inclusion?  Or does it rather bolster the case _against_ inclusion?

One problem that is hinted at in the abstract is the thermal mass of the water body.  From personal experience living near a large lake I can attest to the moderating effect on air temperature in early summer and early winter.  I doubt that the size of the water body and the location of the temperature measurement in relation to it can be entirely ignored for modeling floating PV systems in general.

Finally, there is no hurry.  Even if we don't know of a suitable alternative right now, there is a steady stream(!) of papers on floating PV and surely more suitable models will show up.  Or perhaps someone will validate this one more broadly and put my concerns to rest.

AdamRJensen:
The discussion above certainly brings up a lot of good points, e.g., about the validation of the algorithm and whether its usage is justifiable. I'm leaning toward closing this PR and seeing how the modeling of floating PV develops.

For floating PV on the sea, I think using reanalysis data for water temperature from ERA5 or similar could be a good option.

IoannisSifnaios:
I think I will close this for now. Thank you for the interesting discussions!

PR #2138, opened 2024-07-19
coerce and rotate pvgis TMY data to desired tz and year

### LINKED ISSUE #2139: add boilerplate to fix pvgis tmy to a timezone and year
pvgis is the easiest weather data to retrieve b/c no credentials required, but it always returns data in UTC, which then needs to be adjusted to local timezone, causing the year to not start at midnight.

This is some boilerplate that I use everytime I pull pvgis, so rather than clogging up my codebase, I thought it would be easier if it were upstream. Basically:
1. if february is a leap year, change it to 1990
2. rotate the dataframe by the desired utc offset (note, don't shift the index)
3. convert the timezone
4. change indices to desired year

Tested in this Google Colab notebook [`test_coerce_pvgis.ipynb`](https://colab.research.google.com/drive/1pibVTd02QUEJ45oShWuaqzywqviaxWFo#scrollTo=FcNx2GTQKdmQ)

keep doing this manually in my code.

**Additional context**
This was discussed here https://github.com/pvlib/pvlib-python/discussions/1528#discussioncomment-9177537

mikofski:
- add private function `_coerce_and_roll_pvgis_tmy()`
- add `roll_utc_offset` and `coerce_year` params to docstring for `get_pvgis_tmy`
- call private function if `roll_utc_offset` is not `None` or `coerce_year` is not `None`

discussed in https://github.com/pvlib/pvlib-python/discussions/1528#discussioncomment-9177537

 - [x] Closes #2139 
 - [x] Tests added


mikofski:
Tested in this Colab notebook: [`test_coerce_pvgis.ipynb`](https://colab.research.google.com/drive/1pibVTd02QUEJ45oShWuaqzywqviaxWFo?usp=sharing).

cwhanse:
Showing my ignorance of PVGIS: does it always return 8760 timestamps? 

If PVGIS does return 8760, what should I expect if a user coerces to a leap year? I assume they would get a DatetimeIndex with Feb 29 missing.

Second question: if I rotate to UTC-10 (`utc_offset=-10`), what should I expect as the result? An index of length 8750, or an index of length 8760 but with the last 10 hours in the following year? If its 8760, what data are in the last 10 hours?

mikofski:
- `get_pvgis_tmy()` always returns 8760 in utc
- yes, you are correct. If coerced to a leap year then Feb 29 is skipped, this matches other functions in `iotools` with this argument
- if you rotate to utc offset of -10 then the output will be a 8760 starting at midnight of the forced year, the first 10 rows are “rotated” to the bottom, because in utc-10 they are from 2am to 11pm on December 31st. The tests check the output to see that the first timestamp is January 1st at midnight, and the last timestamp is December 31st at 11pm

Where is utc-10 is that Hawaii?

cwhanse:
> Where is utc-10 is that Hawaii?

Dunno I just picked a number for illustration.

cwhanse:
> if you rotate to utc offset of -10 then the output will be a 8760 starting at midnight of the forced year, the first 10 rows are “rotated” to the bottom, because in utc-10 they are from 2am to 11pm on December 31st. 

Ah. Basically we assume a TMY year is periodic, thus every hour's data is defined in the rotated index.

pvlib/iotools/pvgis.py
```diff
@@ -390,9 +391,32 @@ def read_pvgis_hourly(filename, pvgis_format=None, map_variables=True):
     raise ValueError(err_msg)
 
 
+def _coerce_and_rotate_pvgis(pvgis_data, tz, year):
```
adriesse, line 394:
```suggestion
def _coerce_and_rotate_tmy(tmy_data, tz):
```
Should only be used for tmy, right?  Source probably doesn't matter.

Does pvgis put out Feb 29?  I hope not.  I think it would be ok to just use 1990.

pvlib/iotools/pvgis.py
```diff
@@ -424,6 +448,12 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
         base url of PVGIS API, append ``tmy`` to get TMY endpoint
     timeout : int, default 30
         time in seconds to wait for server response before timeout
+    utc_offset: int, default 0
+        Use to specify a timezone other than the default UTC zero. Will force
+        year to ``coerce_year`` if not zero.
```
adriesse, line 453:
```suggestion
        year to 1990 if not zero.
```

pvlib/iotools/pvgis.py
```diff
@@ -424,6 +448,12 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
...
+    coerce_year: int, default 1990
```
adriesse, line 454:
```suggestion
    coerce_year: boolean
```

pvlib/iotools/pvgis.py
```diff
@@ -424,6 +448,12 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
...
+        Use to force indices to desired year. Ignored if ``utc_offset`` is
+        zero.
```
adriesse, line 456:
```suggestion
        Used to force indices to 1990. Set to True if ``utc_offset`` is
        non-zero.
```

pvlib/iotools/pvgis.py
```diff
@@ -390,9 +391,32 @@ def read_pvgis_hourly(filename, pvgis_format=None, map_variables=True):
...
+    """
+    # assert tz//1 == tz
+    tzname = f'Etc/GMT{-tz:+d}'
+    # check if February is leap year
+    feb1 = pvgis_data[pvgis_data.index.month == 2].index.year[0]
+    if calendar.isleap(feb1):
+        # replace Feb year with a non-leap year
+        pvgis_data.index = [
+            ts.replace(year=1990) if ts.month == 2 else ts
+            for ts in pvgis_data.index]
```
adriesse, line 407: Can skip all this if we reduce the options.

pvlib/iotools/pvgis.py
```diff
@@ -390,9 +391,32 @@ def read_pvgis_hourly(filename, pvgis_format=None, map_variables=True):
...
+    pvgis_data = pd.concat(
+        [pvgis_data.iloc[-tz:], pvgis_data.iloc[:-tz]],
+        axis=0).tz_convert(tzname)
```
adriesse, line 410: I kind of liked the `np.roll()` solution but in python there are always _so_ many ways to achieve the same result!

pvlib/iotools/pvgis.py
mikofski: Yes pvgis leaves out Feb 29 always even if month is 2012, a leap year

pvlib/iotools/pvgis.py
mikofski: I don’t recall why I chose this approach but I can test if `np.roll` works better

pvlib/iotools/pvgis.py
adriesse: If you always coerce the year to 1990 there will be no exception.

mikofski: i think the order of operations has to be
1. Sanitize February
2. Rotate hours
3. Concert time zone
4. replace years

If I try to coerce years first then when I convert time zone I get two years.

i can try to experiment with the order but this was what seemed to work best. Let me play around with it some more and maybe I can get rid of step 1. Although I do really like the option to coerce to any year not just 1990.

mikofski: I found a way to make it work without sanitizing February and with `np.roll` by replacing the year and time zone at the same time. Any other way always returns two years.

pvlib/iotools/pvgis.py
mikofski: it works!

pvlib/iotools/pvgis.py
mikofski: this doesn't follow the usage in other functions, and I think that pvgis main issue is the `utc_offset` so my preference is to keep both args and allow user to coerce years like other `iotools` do

pvlib/iotools/pvgis.py
mikofski: changed name of private function to `_coerce_and_rotate_pvgis_tmy`

pvlib/iotools/pvgis.py
mikofski: I prefer to keep function to coerce years, matches other functions in iotools

pvlib/iotools/pvgis.py
mikofski: doesn't match coerce year in other iotools which lets user force any year

pvlib/tests/iotools/test_pvgis.py
```diff
@@ -435,6 +435,22 @@ def _compare_pvgis_tmy_basic(expected, meta_expected, pvgis_data):
...
+    pvgis_data = get_pvgis_tmy(45, 8, utc_offset=2)  # Turin
+    jan1_midnight1990 = pd.Timestamp('1990-01-01 00:00:00', tz='Etc/GMT-2')
+    dec31_midnight1990 = pd.Timestamp('1990-12-31 23:00:00', tz='Etc/GMT-2')
+    assert pvgis_data[0].index[0] == jan1_midnight1990
+    assert pvgis_data[0].index[-1] == dec31_midnight1990
+    pvgis_data = get_pvgis_tmy(45, 8, utc_offset=2, coerce_year=2021)  # Turin
+    jan1_midnight2021 = pd.Timestamp('2021-01-01 00:00:00', tz='Etc/GMT-2')
+    dec31_midnight2021 = pd.Timestamp('2021-12-31 23:00:00', tz='Etc/GMT-2')
+    assert pvgis_data[0].index[0] == jan1_midnight2021
+    assert pvgis_data[0].index[-1] == dec31_midnight2021
```
cwhanse, line 451: Can we also check that a few values are aligned with the index correctly?

pvlib/iotools/pvgis.py
adriesse: No problem, I hadn't looked at the other functions.

pvlib/tests/iotools/test_pvgis.py
mikofski: I checked noon for every month with default and coerced year. Note I conflated a few other updates to the test:
- check that UTC time zone is unchanged if using defaults
- update Turin is actually CET which is UTC+1, not +2, oops! I got confused because currently it is CEST, or summertime which _is_ +2, 😆 ha-ha
- use some variables to be DRY to replace some WET & hardcoded args like the coerced year used 3x, time zone used 4x, and the "Etc/GMT" tz name also used 4x
- I didn't test the combo of `utc_offset` is zero, and `coerce_year` is set - should do nothing - do we need this test?

cwhanse (approved):

pvlib/iotools/pvgis.py
```diff
@@ -424,6 +441,12 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
         base url of PVGIS API, append ``tmy`` to get TMY endpoint
     timeout : int, default 30
         time in seconds to wait for server response before timeout
+    utc_offset: int, default 0
+        Use to specify a timezone other than the default UTC zero. Will force
+        year to ``coerce_year`` if not zero.
```
kandersolar, line 446: This parameter names and description don't make any mention of rolling the data.  I think at least the parameter description should explain the rolling.  Ideally the parameter name would somehow reflect it as well, although I'm having trouble coming up with a good name.  Somehow `roll_according_to_utc_offset` doesn't quite roll off the tongue...

kandersolar (commented):
FYI, the PSM3 TMY API does this "roll over" also: https://assessingsolar.github.io/unofficial-psm3-userguide/pages/rollover.html

adriesse:
> FYI, the PSM3 TMY API does this "roll over" also: https://assessingsolar.github.io/unofficial-psm3-userguide/pages/rollover.html

TMY methods have ways of merging between months from different years.  Ideally this would be done here.

pvlib/iotools/pvgis.py
adriesse: I was going to leave good enough alone, but since we have more suggestions: I would roll automatically based on longitude.

pvlib/iotools/pvgis.py
```diff
@@ -510,6 +533,9 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
     if map_variables:
         data = data.rename(columns=VARIABLE_MAP)
 
+    if utc_offset != 0:
+        data = _coerce_and_roll_pvgis_tmy(data, utc_offset, coerce_year)
```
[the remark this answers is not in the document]
mikofski: I did recognize that forcing year could be useful by itself (eg when utc is zero)  but figured users could more easily replace the index with time series of their choice since no rolling is required. Eg:
```python
data.index = pd.date_range(start='year-01-01 01:00', freq='T', …)
```
I didn’t want to make `utc_offset` be either boolean or integer, because I’m afraid it’s a risky design that later will cause unexpected problems, eg zero is also false. I also didn’t want to add a 3rd new argument, so that’s why I left it like this.

pvlib/iotools/pvgis.py
mikofski: I can add explanation of rolling and maybe somewhere we should add description that TMY is always expected to start and end at midnight on Jan 1/December 31?

mikofski: >roll automatically based on longitude.

I don’t want to make rolling mandatory. Using longitude is clever, but not all time zones are strictly longitudinal, and I’ve had issues using that approach.

adriesse: No problem.

pvlib/iotools/pvgis.py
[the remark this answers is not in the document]
kandersolar: Would something like the following work?

```python
def get_pvgis_tmy(..., roll_offset=0, coerce_year=1990):
    """
    Parameters
    ----------
    ...
    roll_offset : int, default 0
        A UTC offset in hours.  If provided, the dataframe will be "rolled"
        so that data begins at midnight, blah blah
        It is likely desirable to use this parameter in combination with
        ``coerce_year``.
    coerce_year : int, default 1990
        blah blah
    """

    if roll_offset != 0:
        tz = pytz.timezone(f'Etc/GMT{-roll_offset:+d}')
        data = data.tz_convert(tz)
        data = pd.DataFrame(np.roll(data, roll_offset, axis=0),
                            columns=data.columns,
                            index=np.roll(data.index, roll_offset))

    if coerce_year:
        data.index = [x.replace(year=1990) for x in data.index]
```

pvlib/iotools/pvgis.py
```diff
@@ -424,6 +441,13 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
         base url of PVGIS API, append ``tmy`` to get TMY endpoint
     timeout : int, default 30
         time in seconds to wait for server response before timeout
+    utc_offset: int, default 0
+        Use to specify a time zone other than the default UTC zero and roll
+        dataframe by ``utc_offset`` so that it starts at midnight on January
+        1st. If not zero, will also force year to ``coerce_year``.
+    coerce_year: int, default 1990
```
kandersolar, line 448: Let's do `coerce_year=None` instead, for consistency with the other pvlib functions that have `coerce_year` (`read_epw`, `read_tmy3`).  

If we do have a default year, that would be a change in behavior that could warrant a deprecation period...

pvlib/iotools/pvgis.py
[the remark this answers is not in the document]
mikofski: I don’t think this will work. When I tried this variant I got an error that February 29 is not a date. The `tz_convert()` changed the hour which made March 1st at midnight become 1am and made midnight now February 29th. I don’t want to have to sanitize the February year PVGIS chooses before converting. Look back at 31800e9 and d0bd0a0.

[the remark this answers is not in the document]
kandersolar: Argh, good point.  Shame on me for not actually trying it.  

I still think it's strange for `coerce_year` to only work when `utc_offset` is nonzero.  Does your code still work if it runs `_coerce_and_roll_pvgis_tmy` even when `utc_offset` is zero?  Seems like it should...

pvlib/iotools/pvgis.py
mikofski: I think if we make both defaults `None`I can check identity, add a comment to avoid implicit booleaness, and then make it possible to coerce the year even for utc zero. I will send an update later today.

I prefer `utc_offset` over `roll_offset` because this function converts the time zone to a new utc offset, which is not explained in “roll offset” while utc offset is a well known concept. How about `roll_utc_offset` which unfortunately sounds like a verb, but makes some sense if I let the default be `None`?

pvlib/iotools/pvgis.py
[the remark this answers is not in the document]
mikofski: I can make it work - give me a second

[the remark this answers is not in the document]
adriesse: > Argh, good point. Shame on me for not actually trying it.
> 
> I still think it's strange for `coerce_year` to only work when `utc_offset` is nonzero. Does your code still work if it runs `_coerce_and_roll_pvgis_tmy` even when `utc_offset` is zero? Seems like it should...

`align_to_local_noon` might be less cryptic.

pvlib/iotools/pvgis.py
adriesse: > Somehow `roll_according_to_utc_offset` doesn't quite roll off the tongue...

How about `align_to_local_noon`?

mikofski: @adriesse  I changed it to `roll_utc_offset` is that okay? It's both a verb `roll` and a number `utc_offset`. I like your suggestion, and thought my tests should probably have check transit times, which is the real test of whether it's in the right time zone, but that was a bridge too far.

pvlib/iotools/pvgis.py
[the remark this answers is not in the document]
mikofski: @kandersolar ready for review -- see my comment (updated) and my last 2 commits.

pvlib/iotools/pvgis.py
adriesse: It's all good.  No strong feelings about any of this.

pvlib/iotools/pvgis.py
```diff
@@ -424,6 +443,13 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
         base url of PVGIS API, append ``tmy`` to get TMY endpoint
     timeout : int, default 30
         time in seconds to wait for server response before timeout
+    roll_utc_offset: int, default None
+        Use to specify a time zone other than the default UTC zero and roll
+        dataframe by ``roll_utc_offset`` so it starts at midnight on January
+        1st. Ignored if ``None``, otherwise will force year to ``coerce_year``.
+    coerce_year: int, default None
+        Use to force indices to desired year. Will default to 1990 if
+        ``coerce_year`` is ``None``, but ``roll_utc_offset`` is not ``None``.
```
kandersolar, line 452:
```suggestion
    roll_utc_offset: int, optional
        Use to specify a time zone other than the default UTC zero and roll
        dataframe by ``roll_utc_offset`` so it starts at midnight on January
        1st. Ignored if ``None``, otherwise will force year to ``coerce_year``.
    coerce_year: int, optional
        Use to force indices to desired year. Will default to 1990 if
        ``coerce_year`` is not specified, but ``roll_utc_offset`` is specified.
```

For #1574

docs/sphinx/source/whatsnew/v0.11.1.rst
```diff
@@ -16,6 +16,10 @@ Enhancements
 * Add new parameters for min/max absolute air mass to
   :py:func:`pvlib.spectrum.spectral_factor_firstsolar`.
   (:issue:`2086`, :pull:`2100`)
+* Add ``utc_offset`` and ``coerce_year`` arguments to
```
kandersolar, line 19:
```suggestion
* Add ``roll_utc_offset`` and ``coerce_year`` arguments to
```

kandersolar (approved):
LGTM.  We may want to add similar functionality for other TMY datasets in the future, so I think it's worth @AdamRJensen taking a look too if he has the time.

AdamRJensen:
> > LGTM. We may want to add similar functionality for other TMY datasets in the future,
> 
> LGTM2
> 
> My perception is that this is fixing a flaw in the PVGIS service. Hopefully other TMY sources align on local time.

The PVGIS team is currently reworking the entire process, so this is a good time to make requests about features and provide feedback. Tagging @NikosAlexandris

pvlib/iotools/pvgis.py
```diff
@@ -510,6 +536,16 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
     if map_variables:
         data = data.rename(columns=VARIABLE_MAP)
 
+    if roll_utc_offset is not None or coerce_year is not None:
+        # XXX: be explicit, test for identity not implicit booleaness
+        # roll_utc_offset None but coerce_year isn't, set year with utc zero
+        if roll_utc_offset is None:  # XXX: None and zero are both False
+            roll_utc_offset = 0
```
AdamRJensen, line 543: Is this really a desired behavior? Alternatively would be to raise a ValueError if roll_utc_offset is missing

pvlib/iotools/pvgis.py
```diff
@@ -510,6 +536,16 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
     if map_variables:
         data = data.rename(columns=VARIABLE_MAP)
 
+    if roll_utc_offset is not None or coerce_year is not None:
+        # XXX: be explicit, test for identity not implicit booleaness
```
AdamRJensen, line 540: what does these "xxx" mean?

AdamRJensen (approved):
I tested it and it seems to work great 🥳 

One thing I noticed was that the index name is maintained as "time(UTC)" when utc_roll_offset is None and is removed when utc_roll_offset is specified. Should this be tested? Maybe it doesn't matter

pvlib/iotools/pvgis.py
```diff
@@ -390,9 +390,28 @@ def read_pvgis_hourly(filename, pvgis_format=None, map_variables=True):
     raise ValueError(err_msg)
 
 
+def _coerce_and_roll_pvgis_tmy(pvgis_data, tz, year):
```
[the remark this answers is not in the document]
mikofski: Agreed, removed "pvgis" from name in 91c169e.

pvlib/iotools/pvgis.py
mikofski: I use it as a warning to beware or pay attention or to flag problematic code. I used it here because I was worried about implicit booleaness, especially that `None` and 0 are both false, which could be a problem because zero is a valid UTC offset, but evaluates the same as `None` even though we want the opposite outcome.

Various places define it differently, but I disagree that's it's synonymous with or related to `FIXME:` or `TODO:`
- [wikipedia > computer programming comments > tags](https://en.wikipedia.org/wiki/Comment_(computer_programming)#Tags)
- [SO what does `XXX` mean?](https://softwareengineering.stackexchange.com/questions/65467/what-does-xxx-mean-in-a-comment)
- [SO what is the meaning of `XXX`?](https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments)
- [SO best practices rule 9](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/)

pvlib/iotools/pvgis.py
mikofski: I think it's important because it allows user to coerce year, but leave TZ alone - especially useful in GMT. @kandersolar requested this in https://github.com/pvlib/pvlib-python/pull/2138#discussion_r1696869191. I reworked `_coerce_and_roll_tmy()` to accept either `None` or zero as UTC, so I removed this from `get_pvgis_tmy()` for simplicity.

pvlib/iotools/pvgis.py
mikofski: I removed the warning comments now that I handle `tz=None` in `_coerce_and_roll_tmy()` instead of here.

pvlib/iotools/pvgis.py
```diff
@@ -424,6 +443,13 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
         base url of PVGIS API, append ``tmy`` to get TMY endpoint
     timeout : int, default 30
         time in seconds to wait for server response before timeout
+    roll_utc_offset: int, optional
+        Use to specify a time zone other than the default UTC zero and roll
+        dataframe by ``roll_utc_offset`` so it starts at midnight on January
+        1st. Ignored if ``None``, otherwise will force year to ``coerce_year``.
+    coerce_year: int, optional
+        Use to force indices to desired year. Will default to 1990 if
+        ``coerce_year`` is not specified, but ``roll_utc_offset`` is specified.
```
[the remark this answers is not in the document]
mikofski: Understandable. I have a slight preference for the default behavior. I believe folks familiar with PVsyst will assume 1990. Maybe we should leave this PR open for a few days to invite more feedback?

mikofski:
> One thing I noticed was that the index name is maintained as "time(UTC)" when utc_roll_offset is None and is removed when utc_roll_offset is specified. Should this be tested? Maybe it doesn't matter

I renamed the index and added tests in https://github.com/pvlib/pvlib-python/pull/2138/commits/1e9ee64186eb74457f1b2769093b406dc39af7a7

pvlib/iotools/pvgis.py
mikofski: Done in fb2813d.

pvlib/iotools/pvgis.py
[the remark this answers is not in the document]
mikofski: Since there's no further feedback, should we merge this PR?

pvlib/iotools/pvgis.py
AdamRJensen: I guess a user could pass in a non-utc localized index (assuming the timezone would get picked up by the function) but then it would get overridden if the timezone wasn't separately given as an input?

mikofski: `_coerce_and_roll_tmy()` is a private function in the `pvgis.py` module that only gets called by `get_pvgis_tmy()`, and PVGIS TMY are always UTC. This is the purpose of #2139 which is to localize TMY dataset from PVGIS and roll it so it starts at midnight.

IMHO private functions are undocumented, so responsibility is the at the user’s own risk if `_coerce_and_roll_tmy()` is passed a localized index.

AdamRJensen: Right right, I'm not requesting changes just wanted to clarify my understanding.

AdamRJensen:
Thanks @mikofski Great to see this new feature in the iotools!

NikosAlexandris:
> > My perception is that this is fixing a flaw in the PVGIS service. Hopefully other TMY sources align on local time.
> 
> The PVGIS team is currently reworking the entire process, so this is a good time to make requests about features and provide feedback. Tagging @NikosAlexandris

Thank you for pinging me here. Will look into this as time permits.

NikosAlexandris:
> Tested in this Colab notebook: [`test_coerce_pvgis.ipynb`](https://colab.research.google.com/drive/1pibVTd02QUEJ45oShWuaqzywqviaxWFo?usp=sharing).

@mikofski Would you have some time to help me better understand the "issue" described in this Notebook ?  Thank you.

mikofski:
Hi @NikosAlexandris , in past versions of ‘iotools.get_pvgis_tmy()` the data-frame returned was always in UTC time. In my experience TMY files have traditionally been in local time, starting & ending at midnight. Therefore, the “issue” tested in the notebook is if the additional arguments roll utc offset and coerce year will rotate the data frame to coincide with local time, starting at midnight in the desired year

PR #2136, opened 2024-07-18
restructure pvlib/spectrum

RDaxini:
- [X] Closes #2125
 - [X] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [X] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

Context and discussion in #2125

Issue #2125, opened 2024-07-11
Proposal: split up mismatch.py


RDaxini:
**Is your feature request related to a problem? Please describe.**
I've been working with the spectrum unit of pvlib a fair bit recently for my GSoC project and I find that the [`pvlib/spectrum/mismatch.py`](https://github.com/pvlib/pvlib-python/blob/main/pvlib/spectrum/mismatch.py) file currently contains a range of functions, several of which are not directly related to spectral mismatch.

Current functions in `pvlib/spectrum/mismatch.py`:

[pvlib.spectrum.get_example_spectral_response](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.get_example_spectral_response.html)
[pvlib.spectrum.get_am15g](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.get_am15g.html) (deprecated v0.11, removed 0.12)
[pvlib.spectrum.get_reference_spectra](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.get_reference_spectra.html)
[pvlib.spectrum.calc_spectral_mismatch_field](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.calc_spectral_mismatch_field.html)
[pvlib.spectrum.spectral_factor_caballero](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.spectral_factor_caballero.html)
[pvlib.spectrum.spectral_factor_firstsolar](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.spectral_factor_firstsolar.html)
[pvlib.spectrum.spectral_factor_sapm](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.spectral_factor_sapm.html)
[pvlib.spectrum.spectral_factor_pvspec](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.spectral_factor_pvspec.html)
[pvlib.spectrum.spectral_factor_jrc](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.spectral_factor_jrc.html)
[pvlib.spectrum.sr_to_qe](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.sr_to_qe.html)
[...]

PR #2151, opened 2024-08-05
Split test_spectrum.py

RDaxini:
Restructure tests in line with `mismatch.py` restructuring. Prompted by final comments in #2136.

 - [X] Relates to #2125 and #2136


RDaxini:
Should I add to the same whatsnew entry as #2136 once that is merged, or create a new entry?

kandersolar:
> Should I add to the same whatsnew entry as https://github.com/pvlib/pvlib-python/pull/2136 once that is merged, or create a new entry?

I'd say this PR is implicit in the whatsnew entry for #2136, so no need for another one here IMHO

pvlib/tests/spectrum/test_mismatch.py
```diff
@@ -1,11 +1,10 @@
...
+from numpy.testing import assert_allclose, assert_approx_equal
 import pandas as pd
 import numpy as np
 from pvlib import spectrum
-from pvlib._deprecation import pvlibDeprecationWarning
 
-from .conftest import DATA_DIR, assert_series_equal, fail_on_pvlib_version
+from ..conftest import DATA_DIR, assert_series_equal
 
 SPECTRL2_TEST_DATA = DATA_DIR / 'spectrl2_example_spectra.csv'
```
kandersolar, line 9: This variable, and the `spectrl2_data` fixture, can be removed from this file.  They're only needed in the new `test_spectrl2.py`.

kandersolar (approved):

pvlib/tests/spectrum/test_mismatch.py
RDaxini: These parts are still required in `test_mismatch.py` because the first test for `calc_mismatch_field` uses spectlr2 data

kandersolar: Indeed so!  My mistake.

kandersolar: In that case, can we extract that fixture out into a `conftest.py` file inside this new `pvlib/tests/spectrum` directory?  See https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files

Then we do not have to have redundant definitions in these two files.

RDaxini: Nice idea. I had been wondering if this was possible somehow but was not really aware of how `conftest.py` works to be able to suggest this. Commit [3729860](https://github.com/pvlib/pvlib-python/pull/2151/commits/3729860d35f8da81b18be92c5eefb8b7b86a3d5f) should resolve this.

kandersolar:
Thanks @RDaxini!

PR #2212, opened 2024-09-16
WIP: Convert between CEC and PVsyst single diode models

### LINKED ISSUE #1935: Add converters between single diode models
pvlib provides three single diode models: CEC, PVSyst and De Soto, with the first two being of more interest. pvlib copies and distributes the SAM file containing parameters for the CEC model.

It seems reasonable that some would like to convert from CEC parameters to a set of PVSyst model parameters, such that the conversion produces equivalent power and energy simulations. Something like `pvlib.ivtools.sdm.convert` modeled after `pvlib.iam.convert`.

Interested in reactions to this idea:
- Would you use it? 
- What criteria should be used to judge that two diode models are "equivalent?"

cwhanse:
- [x] Closes #1935
 - [x] Tests added

Primary reference is in review. Feedback on function naming and signature is welcome.

### DIFF
--- a/pvlib/ivtools/sdm.py
+++ b/pvlib/ivtools/sdm.py
@@ -7,12 +7,14 @@ Function names should follow the pattern "fit_" + name of model + "_" +
 """
 
 import numpy as np
+import pandas as pd
 
 from scipy import constants
 from scipy import optimize
 from scipy.special import lambertw
 
-from pvlib.pvsystem import calcparams_pvsyst, singlediode, v_from_i
+from pvlib.pvsystem import (calcparams_pvsyst, calcparams_cec, singlediode,
+                            v_from_i)
 from pvlib.singlediode import bishop88_mpp
 
 from pvlib.ivtools.utils import rectify_iv_curve, _numdiff
@@ -22,6 +24,17 @@ from pvlib.tools import _first_order_centered_difference
 
 
 CONSTANTS = {'E0': 1000.0, 'T0': 25.0, 'k': constants.k, 'q': constants.e}
+
+
+IEC61853 = pd.DataFrame(
+    columns=['effective_irradiance', 'temp_cell'],
+    data = np.array(
+        [[100, 100, 100, 100, 200, 200, 200, 200, 400, 400, 400, 400,
+          600, 600, 600, 600, 800, 800, 800, 800, 1000, 1000, 1000, 1000,
+          1100, 1100, 1100, 1100],
+         [15, 25, 50, 75, 15, 25, 50, 75, 15, 25, 50, 75, 15, 25, 50, 75,
+          15, 25, 50, 75, 15, 25, 50, 75, 15, 25, 50, 75]]).T,
+    dtype=np.float64)
 
 
 def fit_cec_sam(celltype, v_mp, i_mp, v_oc, i_sc, alpha_sc, beta_voc,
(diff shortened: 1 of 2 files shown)


pvlib/ivtools/sdm.py
```diff
@@ -24,6 +26,17 @@
...
+    columns=['effective_irradiance', 'temp_cell'],
+    data = np.array(
+        [[100, 100, 100, 100, 200, 200, 200, 200, 400, 400, 400, 400,
+          600, 600, 600, 600, 800, 800, 800, 800, 1000, 1000, 1000, 1000,
+          1100, 1100, 1100, 1100],
+         [15, 25, 50, 75, 15, 25, 50, 75, 15, 25, 50, 75, 15, 25, 50, 75,
+          15, 25, 50, 75, 15, 25, 50, 75, 15, 25, 50, 75]]).T,
+    dtype=np.float64)
+
+
```
mikofski, line 39: [Mesh grid](https://numpy.org/doc/stable/reference/generated/numpy.meshgrid.html)?

adriesse:
> Feedback on function naming and signature is welcome.

I guess one might discuss whether you are converting models or sets of parameters.

cwhanse:
Test failure for `ubuntu-latest, 3.8, conda, -min` is because that environment installs scipy=1.6.0, which doesn't allow for constraints when using Nelder-Mead, so the optimizer gets a different, unsatisfactory, solution.

echedey-ls:
> I guess one might discuss whether you are converting models or sets of parameters.

Do you have any proposal in mind? For example, `cec_params_to_pvsyst`, `params[_from]_pvsyst_to_cec` [`params_pvsyst2cec`]. I don't really have a strong preference on wording.

> scipy=1.6.0, which doesn't allow for constraints when using Nelder-Mead

scipy=1.7.0 is the first version to support it, and is also supported in Py3.9 (taking into account upcoming drop of Py3.8 support). I would edit this requirement for future >=Py3.9 release (#2213).

Source of scipy docs on nelder-mead; scipy 1.6.3 bumps to 1.7.0:
- 1.6.3 https://docs.scipy.org/doc/scipy-1.6.3/reference/optimize.minimize-neldermead.html?highlight=nelder%20mead
- 1.7.0 https://docs.scipy.org/doc/scipy-1.7.0/reference/reference/optimize.minimize-neldermead.html

mikofski:
Is the simplex solver required? Why not try Powell instead?

cwhanse:
> Is the simplex solver required? Why not try Powell instead?

@leliadeville

leliadeville:
> Is the simplex solver required? Why not try Powell instead?

@mikofski I used Nelder-Mead to develop and check the algorithm, but I think the optimization method would be good to have as a parameter.

mikofski:
Any particular reason for choosing Simplex? Good idea to make `method` a parameter. I would be curious to see how other methods like Powell compare. My experience is that the system of equations is stiff and requires some transformation to improve the condition. An excellent primer for selecting methods is https://gael-varoquaux.info/scipy-lecture-notes/advanced/mathematical_optimization/index.html#choosing-a-method

cwhanse:
For context, here's the submitted paper. The main point is determining that optimization is preferred over two other techniques (functions currently in pvlib, and an approach that equates quantities that should have the same values at the reference condition. The optimization could certainly be tuned and improved.

[sdm_translation_submitted.pdf](https://github.com/user-attachments/files/17084394/sdm_translation_submitted.pdf)

mikofski:
>The main point is determining that optimization is preferred …

I agree 💯. I wrote a [blog on using implicit methods to solve SDM system of equations](https://breakingbytes.github.io/examples-of-implicit-versus-explicit-solutions-for-photovoltaic-solar-energy-systems.html#examples-of-implicit-versus-explicit-solutions-for-photovoltaic-solar-energy-systems) with several [gists](https://gist.github.com/mikofski/df318d1f892767ac7c762e732fecaa7f#file-pv-system-implicit-solver-ipynb) comparing
using convex optimization after transforming the variables to improve the condition of the Jacobian to explicit methods using Bishop’s method. Because the equations are differential I used the SciPy wrapper around the MINPACK FORTRAN HYBRDJ solver but other solvers might be more efficient.

mikofski:
I understand now, after reading the paper. This is a method of directly obtaining parameters without generating an IEC-61853 matrix as the go between first. Very clever! So the minimization method of the objective function is secondary, and there are no variables to transform. Thanks for patiently explaining. Great job @leliadeville

cwhanse:
> This is a method of directly obtaining parameters without generating an IEC-61853 matrix as the go between first. 

I may be misunderstanding, but generating the IEC-61853 matrix data for the source model is the first step. The target model's parameters are found by fitting the target model to the IEC-61853 matrix data.

mikofski:
You are right. Sorry I misread. Looking again, this does solve each model independently from the matrix. Honestly I’m not sure what I was thinking before. 

However, I think it would be interesting to see by comparison a method that does use the Jacobian, similar to what I discussed in the gist linked above or in [PVMismatch `gen_coeffs()`](https://github.com/SunPower/PVMismatch/blob/master/pvmismatch/contrib/gen_coeffs/__init__.py), the two approaches differ by single and two diode model, but both transform the variables into either parabolic or log space and then use a gradient search method to solve the convex optimization problem. The method may suffer from poor initial guess but typically converges faster than simplex. Anyway, there is no publication, so I apologize for ranting. Thanks.

adriesse:
> > I guess one might discuss whether you are converting models or sets of parameters.
> 
> @adriesse , do you have any proposal in mind? For example, `cec_params_to_pvsyst`, `params[_from]_pvsyst_to_cec` [`params_pvsyst2cec`]. I don't really have a strong preference on wording.

I may have just been referring to the title of the PR.  No need to overload the function names.

pvlib/ivtools/sdm.py
cwhanse: I think the explicit construction is better, since it omits certain combinations e.g. 100 W/m2 and 75C. The original test construction overlooked this point.

cwhanse:
@pvlib/pvlib-maintainer I am perplexed by the test failure. Test passes when I run it locally, and the results compare favorably when I look at the numbers.

Is there some pvlib setting that may be affecting the `isclose` function?

echedey-ls:
> I am perplexed by the test failure. Test passes when I run it locally, and the results compare favorably when I look at the numbers.
> 
> Is there some pvlib setting that may be affecting the `isclose` function?

@cwhanse, I can reproduce it on Windows x64, Python 3.11.8, in a venv, and Numpy version:

```
>>> from importlib.metadata import version
>>> version("numpy")
'2.1.3'
```
Hope that helps with your setup. Sry for not getting a deeper look at it, not enough time lately to FOSS.

cwhanse:
@echedey-ls thanks, after a little more investigation I figured it was configuration dependent. There's one parameter being estimated that's very sensitive to what must be round-off level differences.

cwhanse:
Closing due to a decision that this algorithm is not sufficiently mature. We found that convergence is sensitive to initial condition and haven't found a method to overcome the sensitive dependence.

Issue #1253, opened 2021-07-02
variables_style_rules.csv could use an update


kanderso-nrel:
The [table of parameter names](https://pvlib-python.readthedocs.io/en/latest/variables_style_rules.html) in the docs is a bit dusty and doesn't perfectly reflect today's pvlib.  To make it easier to figure out where the gaps are, here's a script that scans the package, finds the parameter names currently in use, and compares with that table:

```python
import pvlib
import pandas as pd
from inspect import getmembers, isfunction, ismodule, signature
from collections import defaultdict

skip_modules = ['version', 'forecast', 'iotools', 'location', 'modelchain', 'tools']


def get_parameter_map():
    """
    Find all parameter names used in public pvlib functions
    """
    parameter_map = defaultdict(set)  # map parameter names to functions that use them

    for module_name, module in getmembers(pvlib, ismodule):
        if module_name in skip_modules or module_name.startswith("_"):
            continue
        for function_name, function in getmembers(module, isfunction):
            if function_name.startswith("_"):
                continue
            full_function_name = module_name + "." + function_name
            parameter_names = signature(function).parameters.keys()
            for parameter_name in parameter_names:
                parameter_map[parameter_name].add(full_function_name)

    return parameter_map


parameter_map = get_parameter_map()

# read the docs variable name table
url = 'https://raw.githubusercontent.com/pvlib/pvlib-python/master/pvlib/data/variables_style_rules.csv'
df = pd.read_csv(url, sep=';')
listed_names = set(df['variable'])
unlisted_names = set(parameter_map.keys()) - listed_names
print(unlisted_names)
```

`unlisted_names` has over 300 elements.  I think it makes sense to leave most of them out of the table, for example function-specific names like `upper_line_length` and `u0`, but a few of the unlisted ones are common enough to warrant inclusion, e.g. `albedo`.  It also shows a few inconsistencies like `delta_t` vs `deltaT` and `poa_irradiance` vs `poa_global`.

An idea, maybe a bad one:  once we've figured out which of the unlisted parameter names we don't want to include in the table, put them in an "ignore list" and create a CI check that performs the above scan to check for new parameter names that aren't in the table but maybe should be.

echedey-ls:
Add `fill_factor` to the description list, via #2046.

AdamRJensen:
[...]

PR #2234, opened 2024-09-29
Change Variables and Symbols page into a Glossary of terms

### LINKED ISSUE #1421: Docstring parameter descriptions are somewhat redundant and inconsistent
Conventions like pvlib's definition of `surface_tilt` and `surface_azimuth` are documented somewhat haphazardly in docstring parameter descriptions.  This is objectionable in that it is redundant (same description appears in many places) and inconsistent (not all descriptions are the same for a particular parameter).  For example many functions in `pvlib.irradiance` have parameter descriptions like this:

https://github.com/pvlib/pvlib-python/blob/df1c56e92d91701319adad4fe5cf98837edad0fb/pvlib/irradiance.py#L553-L555

while some others have this:

https://github.com/pvlib/pvlib-python/blob/df1c56e92d91701319adad4fe5cf98837edad0fb/pvlib/irradiance.py#L329

I'm proposing we convert the "Variables and Symbols" page into a sphinx [glossary](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#glossary).  Much like how `:py:func:` links to function definitions, sphinx glossaries allow you to link terms to their definitions using ``` :term:`surface_tilt` ```.  So a glossary like this:

```
.. glossary::

   surface_tilt
      Collector surface tilt angle, measured upwards from horizontal.
      Zero tilt means a horizontal surface (facing directly upwards).
      See also :term:`surface_azimuth`.

   ...
```
would let all `surface_tilt` docstring parameter descriptions reduce to something like this:

```
surface_tilt : numeric
    See :term:`surface_tilt` [degrees]
```

The downside I see to extracting and centralizing this information is that the gallery definitions are only immediately accessible in the built sphinx docs, so `help(pvlib.irradiance.klucher)` and `pvlib.irradiance.klucher??` will be less useful (or, depending on your viewpoint, less cluttered).  So there is an argument to be made that centralizing this information is not a net gain and we should continue including it in the docstrings themselves.

**Additional context**
#1253 is vaguely relevant

RDaxini:
- [X] Partially addresses #1421 ([comment](https://github.com/pvlib/pvlib-python/issues/1421#issuecomment-2344049224))
 - [ ] ~Tests added~

The glossary would replace the [variables and symbols page](https://pvlib-python--2234.org.readthedocs.build/en/2234/user_guide/variables_style_rules.html). This is a limited example. Links:
[Glossary](https://pvlib-python--2234.org.readthedocs.build/en/2234/user_guide/glossary.html)
[Function](https://pvlib-python--2234.org.readthedocs.build/en/2234/reference/generated/pvlib.irradiance.haydavies.html) (see parameters `surface_tilt` and `dni`, and second paragraph in docstring)

Points to discuss: [see this comment](https://github.com/pvlib/pvlib-python/issues/1421#issuecomment-2381425653)

---
Note:
* Took the liberty of alphabetizing the list since I did not see a logical order in the original
* Added/updated `dni` and `surface_tilt` definition to show example enhancements (linked to the [example function](https://pvlib-python--2234.org.readthedocs.build/en/2234/reference/generated/pvlib.irradiance.haydavies.html))

Generally, I think adding/updating definitions (description, units, etc.) and adding/removing new/duplicate variables (#1421, #1253) probably requires a case-by-case discussion so I have not gone further here save for the two examples


RDaxini:
The current variables and symbols page defines function parameters as they appear in the function, e.g. lowercase `dni`. In the  function docstring, we also have 'normal text', e.g. DNI or Direct Normal Irradiance.

Questions:
1) Do we want to link both instances --- parameter definition and docstring --- to the glossary page?
2) Do we want the glossary to contain the parameter terms, e.g. dni, or something else e.g. an expanded form "Direct Normal Irradiance"

If we want the glossary to contain the parameter terms with a definition, and also link terms from the docstring to the same glossary, we can write something like:

https://github.com/pvlib/pvlib-python/blob/a7ba8eaeb620a17c8a241fe5d86830638ea0dab8/pvlib/irradiance.py#L738-L739

to generate custom text in the docstring and either the same, or:

https://github.com/pvlib/pvlib-python/blob/a7ba8eaeb620a17c8a241fe5d86830638ea0dab8/pvlib/irradiance.py#L763-L764

in the parameter definitions.

My view:
> Do we want to link both instances --- parameter definition and docstring --- to the glossary page?

I think yes, but this would require more effort from authors.

> Do we want the glossary to contain the parameter terms, e.g. dni, or something else e.g. an expanded form "Direct Normal Irradiance"

Undecided...
We need to define parameters as they appear in the function somewhere, glossary is one option. The function docstring is another option, in which case the glossary terms could be the 'normal text' versions.

Final point:
It was commented in the linked issue that glossary definitions should not replace definitions on the functions page. Not all parameters have a full definition (e.g. `dni` in the example function). If a full definition is not present, linking to the glossary term in the parameter list is important. If there is already the same definition in the parameter list as there is in the glossary, maybe linking to the glossary there is less important but linking to the glossary in the 'normal text' docstring is more important. This could impact the answer to the two questions above.

pvlib/irradiance.py
```diff
@@ -745,6 +744,10 @@ def haydavies(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
         defined as degrees from horizontal (e.g. surface facing up = 0,
         surface facing horizon = 90)
 
+    surface_tilt : numeric
+        Panel tilt from the horizontal, in decimal degrees, see
+        :term:`surface_tilt`.
```
RDaxini, line 749: add `see :term:`surface_tilt` with the full definition, or shorten definition (like this example) and add it?

echedey-ls: I'd rather have short definitions to get rid of small inconsistencies in the redactions. And that it's kind of simple it's definition, no need to clutter it.

AdamRJensen:
Irradiation is energy (e.g., Wh/m^2) whereas irradiance is power (e.g., W/m^2). I think it's better to use irradiance

pvlib/irradiance.py
echedey-ls: We are in agreement with that; if I understood correctly, Dax was referring to the full definition:
```
surface_tilt : numeric
        Surface tilt angles in decimal degrees. The tilt angle is
        defined as degrees from horizontal (e.g. surface facing up = 0,
        surface facing horizon = 90)
```
It does not appear on the diff by default, just gotta click on those expand arrows.

RDaxini:
Thoughts on retaining any of the following text from `variables_style_rules.rst` before deleting the file? I haven't reviewed each link to verify it is an optimal source of information, but each link is still working and, overall, I think it's nice to include general educational content via external links. I am in favour of copying this over to the new page.

https://github.com/pvlib/pvlib-python/blob/cbe4cc5a8bb3bb529f2dd885aebcd527fedbbc63/docs/sphinx/source/user_guide/variables_style_rules.rst?plain=1#L15-L25

cwhanse:
I think its worth keeping the SodaPro references. We could drop the IEC 61724 links.

echedey-ls:
A random thought, without any intentions: maybe instead of splitting and creating a whole new page, both of them could be put into the same one.

kandersolar:
If the old page is being removed, then these references need to be updated:

```bash
$ git grep -n -F "variables_style_rules"
docs/sphinx/source/contributing/style_guide.rst:16::ref:`variables_style_rules`. We could be better about consistency.
docs/sphinx/source/index.rst:29:There is a :ref:`variable naming convention <variables_style_rules>` to
docs/sphinx/source/user_guide/index.rst:18:   variables_style_rules
docs/sphinx/source/user_guide/variables_style_rules.rst:1:.. _variables_style_rules:
docs/sphinx/source/user_guide/variables_style_rules.rst:9:   :file: ../../../../pvlib/data/variables_style_rules.csv
docs/sphinx/source/user_guide/weather_data.rst:86:into standard pvlib names (see :ref:`variables_style_rules`).
docs/sphinx/source/whatsnew/v0.3.0.txt:50:    * :ref:`variables_style_rules` (:issue:`102`)
pvlib/iotools/midc.py:196:    :ref:`variables_style_rules`.
```

RDaxini:
> If the old page is being removed, then these references need to be updated:
> 
> ```shell
> $ git grep -n -F "variables_style_rules"
> docs/sphinx/source/contributing/style_guide.rst:16::ref:`variables_style_rules`. We could be better about consistency.
> docs/sphinx/source/index.rst:29:There is a :ref:`variable naming convention <variables_style_rules>` to
> docs/sphinx/source/user_guide/index.rst:18:   variables_style_rules
> docs/sphinx/source/user_guide/variables_style_rules.rst:1:.. _variables_style_rules:
> docs/sphinx/source/user_guide/variables_style_rules.rst:9:   :file: ../../../../pvlib/data/variables_style_rules.csv
> docs/sphinx/source/user_guide/weather_data.rst:86:into standard pvlib names (see :ref:`variables_style_rules`).
> docs/sphinx/source/whatsnew/v0.3.0.txt:50:    * :ref:`variables_style_rules` (:issue:`102`)
> pvlib/iotools/midc.py:196:    :ref:`variables_style_rules`.
> ```

@kandersolar done, except for the 0.3.0 whatsnew entry --- do I need to do anything there? Doesn't make sense to change the reference to "glossary", but should the :ref:`` be removed since the link won't work anyway (a warning is probably generated somewhere), or some kind of clarification be added to say the page was removed in 11.2, or nothing?

docs/sphinx/source/user_guide/glossary.rst
```diff
@@ -0,0 +1,195 @@
...
+    wind_speed
+        Wind speed
+
+
+For further explanation of the variables, common symbols, and
+units, refer to the following sources from `SoDa Service <http://www.soda-pro.com/home>`_:
+
+   * `Acronyms, Terminology and Units <https://www.soda-pro.com/help/general/acronyms-terminology-and-units>`_
+   * `Plane orientations and radiation components <https://www.soda-pro.com/help/general/plane-orientations-and-radiation-components>`_
+   * `Time references <https://www.soda-pro.com/help/general/time-references>`_
```
echedey-ls, line 192:
```suggestion
* `Acronyms, Terminology and Units <https://www.soda-pro.com/help/general/acronyms-terminology-and-units>`_
* `Plane orientations and radiation components <https://www.soda-pro.com/help/general/plane-orientations-and-radiation-components>`_
* `Time references <https://www.soda-pro.com/help/general/time-references>`_
```

It's rendering as citation, changes it to normal bulleted-list.

docs/sphinx/source/user_guide/index.rst
```diff
@@ -15,6 +15,6 @@ User Guide
    clearsky
    bifacial
    weather_data
-   variables_style_rules
+   glossary
```
echedey-ls, line 18: Can we move this line one below? I find super anti-intuitive the next topic is about SDE, then the FAQ:

```
   weather_data
   singlediode
   glossary
   faq
```

echedey-ls (approved):
@RDaxini minor comments down below, feel free to discard any of them.

> [v0.3.0 whatsnew entry] should the :ref:`` be removed since the link won't work anyway 

I would strive to not have any warnings in the docs - that's noise when reviewing. Plain code formatting is good for me.

LGTM, loving these improvements on documentation 😋

docs/sphinx/source/user_guide/glossary.rst
RDaxini: I don't understand this one, the original has a shaded background and so does this? It doesn't look like a citation(?)

echedey-ls: You are completely right. Block quote is the right term of what it does now, sorry for the confusion.

https://bashtage.github.io/sphinx-material/rst-cheatsheet/rst-cheatsheet.html

RDaxini: No problem, I appreciate you going through the docs carefully to try find these kinda things. I implemented your other suggestion to change the order of the pages btw

echedey-ls: Nah man, reviewing is the easy part and more entertaining than wasting time xD

docs/sphinx/source/user_guide/glossary.rst
```diff
@@ -0,0 +1,195 @@
...
+
+    surface_azimuth
+        Azimuth angle of the surface
+
+    surface_tilt
+        Tilt angle of the surface
+
+    surface_tilt
+        Panel tilt from horizontal [°]. For example, a surface facing up = 0°,
+        surface facing horizon = 90°.
```
kandersolar, line 153:
```suggestion
    surface_tilt
        Panel tilt from horizontal [°]. For example, a surface facing up = 0°,
        surface facing horizon = 90°.
```

`surface_tilt` is duplicated.  I think the longer one is better.

kandersolar (approved):

AdamRJensen:
The terms Nomenclature or Terminology seem to be more correct than Glossary, as the list includes both terms and symbols/variables. Glossary I believe should only contain terms/definitions.

Terminology seems preferred to me as it's more easily understandable to non-native speakers.

RDaxini:
> The terms Nomenclature or Terminology seem to be more correct than Glossary, as the list includes both terms and symbols/variables. Glossary I believe should only contain terms/definitions.
> 
> Terminology seems preferred to me as it's more easily understandable to non-native speakers.

I'm not opposed to changing the name. Just another thought: how about going back to the old name then, "Variables and Symbols"? I don't have a strong preference

docs/sphinx/source/contributing/style_guide.rst
```diff
@@ -13,7 +13,7 @@ pvlib python generally follows the `PEP 8 -- Style Guide for Python Code
 is 79 characters.
 
 pvlib python uses a mix of full and abbreviated variable names. See
-:ref:`variables_style_rules`. We could be better about consistency.
+:ref:`glossary`. We could be better about consistency.
```
cwhanse, line 16:
```suggestion
:ref:`glossary`. 
```
I don't think that comment helps a new contributor.

cwhanse (approved):

docs/sphinx/source/user_guide/glossary.rst
```diff
@@ -0,0 +1,192 @@
...
+    bhi
+        Beam/direct horizontal irradiance
+
+    dhi
+        Diffuse horizontal irradiance
+
+    dni
+        Direct normal irradiance [Wm⁻²]. Irradiance received per unit area by a
+        surface perpendicular (normal) to the sun's rays that propagate in a
+        straight line from the sun.
```
IoannisSifnaios, line 45: Is there a reason why DNI has an extensive explanation of what it is (including units) while the other irradiances don't? I kind of like the idea of reporting the unit that each term is (usually) measured in pvlib.

docs/sphinx/source/user_guide/glossary.rst
```diff
@@ -0,0 +1,192 @@
...
+        vector pointing towards the sun’s center
+    
+    aoi_projection
+        cos(aoi)
+
+    ape
+        Average photon energy
+
+    apparent_zenith
+        Refraction-corrected solar zenith angle in degrees
```
IoannisSifnaios, line 34: Here the unit is mentioned in text. Should it be [°] instead?

docs/sphinx/source/user_guide/glossary.rst
```diff
@@ -0,0 +1,192 @@
...
+        Nominal inverter efficiency
+
+    eta_inv_ref
+        Reference inverter efficiency
+
+    g_poa_effective
+        Broadband plane of array effective irradiance
+
+    gamma_pdc
+        Module temperature coefficient. Typically in units of 1/C.
```
IoannisSifnaios, line 69: Again maybe try to be consistent on whether we report the units and how

IoannisSifnaios (commented):
Looks good overall! I left one or two comments. Regarding the name of the section, I also agree that Nomenclature (assigning of a word or phrase to a particular object, event, or property) would be a more suitable term than Glossary (terms that are either newly introduced, uncommon, or specialized)

AdamRJensen:
> I don't mind "glossary", but I'm not opposed to changing the name. Just another thought: how about going back to the old name then, "Variables and Symbols"? I don't have a strong preference

I would be ok with either "Terminology" or "Variables and Symbols".

echedey-ls:
I don't really like "Variables and Symbols"; to be precise it's more about "Common Parameters" - but I agree "Terminology" is good if it gets expanded to other terms in the future.

An small side-note, the supposed link in https://pvlib-python--2234.org.readthedocs.build/en/2234/contributing/style_guide.html (first section "Code Style", second paragraph, second sentence) to the glossary is pointing to CPython's glossary: https://docs.python.org/3/glossary.html#glossary . It may be due to the crossreference at the top of the file missing an underscore at the beginning.

RDaxini:
@IoannisSifnaios thanks for reviewing. I was thinking it'd be better to keep the scope of this PR focused on creating the glossary. DNI was expanded as an example (definition+link in function docs). I agree the others need updating, but I think that deciding how best to define each term / groups of terms might be better addressed in a separate PR. 
Happy to revisit this if you or anyone else thinks otherwise

RDaxini:
I created #2284 to keep track of additional suggestions and address the comments from @IoannisSifnaios and @AdamRJensen (thanks for the suggestions)
Is deciding the name the only task remaining? From @IoannisSifnaios's definition, nomenclature sounds like the most technically accurate. Terminology makes me think more of words/phrases, but we have abbreviations and initialisms. That might just be my own inaccurate subjective impression though. I guess the four options now are:
Nomenclature, Variables and Symbols, Glossary, Terminology
How do we decide?

cwhanse:
> Nomenclature, Variables and Symbols, Glossary, Terminology

I'm OK with any of these.

> How do we decide?

Sometimes the PR originator just chooses what color to paint the bikeshed.

RDaxini:
> Will merge later today if no objections are raised by then

I'm still happy to make changes to any aspect if necessary, anyone please feel free to leave a comment 👍🏽

kandersolar:
Thanks @RDaxini and all other participants!  It's fun to see this idea finally implemented :)

Issue #1544, opened 2022-09-06
Migrate efficiency models from pvpltools-python


adriesse:
**Describe the solution you'd like**
A number of different models exist to estimate efficiency (or power) as a function of irradiance and temperature.  I coded and published many of them in my own repository [pvpltools-python](https://github.com/adriesse/pvpltools-python) as it was quicker to do than a pvlib PR, but I think they would fit nicely into pvlib-python.  

https://github.com/adriesse/pvpltools-python/blob/master/pvpltools/module_efficiency.py

**Additional context**
Oh yes, I almost forgot: I developed such a model myself also and it's in there.

adriesse:
What do you think, could the module go in pretty much as-is, or should it be organized differently?

cwhanse:
I'm not in favor of adding the functions to `pvsystem.py` but that's the only current code module that makes any sense. Most functions return module efficiency `eta` (or help do so). There's a gap to span between having `eta` and getting DC power, which could be a new function in `pvsystem.py`. So for this base layer, I'd favor a new module e.g. `pvlib.dc_efficiency.py` and I immediately dislike the name.

adriesse:
> So for this base layer, I'd favor a new module e.g. `pvlib.dc_efficiency.py` and I immediately dislike the name.

I dislike that name too.  :)

adriesse:
I would welcome more opinions on code organization before I continue.
  
Another option would be to just submit the ADR model and ADR fitting function.  In that case, what would be the preferred place(s) to put them?

kanderso-nrel:
#1572 seems OK to me so far.  `dc_efficiency` seems too broad to me.

adriesse:
Looking forward to some reviews...

adriesse:
I am having some second thoughts about where to put this code.  In particular, I think that ```pvefficiency``` may be too narrow for the module.  Steve has already said that LFM does more than efficiency, and I think my model could do more than that as well.  Which brings up my doubts about using the bare ```adr``` as a function name here also.

Could we create a new module called ```pvmodule``` or perhaps even better ```pvarray```?  Initially LFM core functions and ADR efficiency and fitting could go in there there and eventually more stuff could be moved from ```pvsystem``` to reduce its size.

If this is too complex an issue we could discuss it under #1581, but I would really like to get this ADR code into the next release.

cwhanse:
[...]

Issue #1350, opened 2021-12-07
account for major pvwatts changes


wholmgren:
The [SAM Version 2021.12.02 release notes](https://nrel.github.io/SAM/doc/releasenotes.html) describe major changes to the PVWatts model:

> With this release, we have updated SAM's implementation to PVWatts V8 (pvwattsv8 in SSC) with new features listed below. The online PVWatts® Calculator continues to use pvwattsv5 and will be updated later. All versions of PVWatts are available via the SAM Software Development Kit and PySAM. (729)
> 
>     Reorganize user interface to group advanced inputs into a collapsible panel.
>     Add snow loss model when weather file includes snow depth data in cm.
>     Support for bifacial modules.
>     Internal module and module thermal models based on same model as Detailed PV model.
>     Internal inverter model based on same model as Detailed PV model.
>     Inputs for monthly soiling as DC loss.

We should add a note or a warning to `pvsystem.pvwatts_dc` and `inverter.pvwatts`. And `ModelChain.with_pvwatts` is also now ambiguous.

kanderso-nrel:
Append `v5` to the function names?  So `ModelChain.with_pvwattsv5`, etc, and deprecate the originals

cwhanse:
After thumbing up, second thought: append `v5` to base layer functions but add a parameter to `ModelChain.with_pvwatts`.

adriesse:
The release note itself seems ambiguous, since the "Detailed PV Model" has several module and inverter options, and incomplete, since the parameters for those options aren't identified.  Regardless, I think it's a good idea to create and maintain ways in which pvlib users can run a version of their choice and document what that choice implies technically.

adriesse:
The following note from the SAM website may also be relevant:

> Note: The PVWatts website uses a different version numbering system than the PVWatts model. For example, the website "Version 6" uses the PVWatts Version 5 model.

adriesse:
Is the overall idea to offer a separate set of functions for each (or at least more than one) pvwatts version?

kanderso-nrel:
I think there are two independent goals here.  For one, to resolve the ambiguity of what models our current "pvwatts_dc" et al. functions implement.  For two, to make space for emulating other PVWatts versions, which yes is something I'd like to do for at least v8.

FYI I suspect much of PVWatts v8 is already implemented in pvlib (although called something else, e.g. `sam_noct` and the CEC models), but I'm waiting until they publish an official description before taking a closer look.

[...]

PR #2235, opened 2024-09-29
Change g_poa_effective to effective_irradiance

AdamRJensen:
- [x] Related to #1253
 - [x] Tests added

`g_poa_effective` is an alternative name for `effective_irradiance`. The latter is the standard in pvlib (at least the most widely used).

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.11.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.11.2.rst
@@ -6,6 +6,8 @@ v0.11.2 (Anticipated December, 2024)
 
 Deprecations
 ~~~~~~~~~~~~
+* Changed the `g_poa_effective` parameter name to
+  `effective_irradiance` in :py:func:`~pvlib.pvsystem.PVSystem.pvwatts_dc`.
 
 
 Enhancements
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -839,7 +839,7 @@ class PVSystem:
         )
 
     @_unwrap_single_value
-    def pvwatts_dc(self, g_poa_effective, temp_cell):
+    def pvwatts_dc(self, effective_irradiance, temp_cell):
         """
         Calculates DC power according to the PVWatts model using
         :py:func:`pvlib.pvsystem.pvwatts_dc`, `self.module_parameters['pdc0']`,
@@ -847,15 +847,15 @@ class PVSystem:
 
         See :py:func:`pvlib.pvsystem.pvwatts_dc` for details.
         """
-        g_poa_effective = self._validate_per_array(g_poa_effective)
+        effective_irradiance = self._validate_per_array(effective_irradiance)
         temp_cell = self._validate_per_array(temp_cell)
         return tuple(
-            pvwatts_dc(g_poa_effective, temp_cell,
+            pvwatts_dc(effective_irradiance, temp_cell,
                        array.module_parameters['pdc0'],
                        array.module_parameters['gamma_pdc'],
                        **_build_kwargs(['temp_ref'], array.module_parameters))
-            for array, g_poa_effective, temp_cell
-            in zip(self.arrays, g_poa_effective, temp_cell)
+            for array, effective_irradiance, temp_cell
+            in zip(self.arrays, effective_irradiance, temp_cell)
         )
 
     def pvwatts_losses(self):
@@ -2764,7 +2764,7 @@ def scale_voltage_current_power(data, voltage=1, current=1):
     return df_sorted
 
 
-def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
+def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
     r"""
     Implements NREL's PVWatts DC power model. The PVWatts DC model [1]_ is:
 
@@ -2780,7 +2780,7 @@ def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
 
     Parameters
     ----------
-    g_poa_effective: numeric
+    effective_irradiance: numeric
         Irradiance transmitted to the PV cells. To be
         fully consistent with PVWatts, the user must have already
         applied angle of incidence losses, but not soiling, spectral,
@@ -2808,7 +2808,7 @@ def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
            (2014).
     """  # noqa: E501
 
-    pdc = (g_poa_effective * 0.001 * pdc0 *
+    pdc = (effective_irradiance * 0.001 * pdc0 *
            (1 + gamma_pdc * (temp_cell - temp_ref)))
 
     return pdc
--- a/docs/sphinx/source/user_guide/pvsystem.rst
+++ b/docs/sphinx/source/user_guide/pvsystem.rst
@@ -75,7 +75,7 @@ data irradiance and temperature.
 
 .. ipython:: python
 
-    pdc = system.pvwatts_dc(g_poa_effective=1000, temp_cell=30)
+    pdc = system.pvwatts_dc(effective_irradiance=1000, temp_cell=30)
     print(pdc)
 
 Methods attached to a PVSystem object wrap the corresponding functions in
@@ -84,8 +84,8 @@ using data stored in the PVSystem attributes. Compare the
 :py:meth:`~pvlib.pvsystem.PVSystem.pvwatts_dc` method signature to the
 :py:func:`~pvlib.pvsystem.pvwatts_dc` function signature:
 
-    * :py:meth:`PVSystem.pvwatts_dc(g_poa_effective, temp_cell) <pvlib.pvsystem.PVSystem.pvwatts_dc>`
-    * :py:func:`pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.) <pvlib.pvsystem.pvwatts_dc>`
+    * :py:meth:`PVSystem.pvwatts_dc(effective_irradiance, temp_cell) <pvlib.pvsystem.PVSystem.pvwatts_dc>`
+    * :py:func:`pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.) <pvlib.pvsystem.pvwatts_dc>`
 
 How does this work? The :py:meth:`~pvlib.pvsystem.PVSystem.pvwatts_dc`
 method looks in `PVSystem.module_parameters` for the `pdc0`, and


echedey-ls:
Should a deprecation-warning period be established? I'm not aware of the internal methodology you guys make use of, but I've seen that [changing params have been deprecated before in this project](https://github.com/pvlib/pvlib-python/blob/4cfda4a14366217ba9bb3b15d2531c61b7507e69/docs/sphinx/source/whatsnew/v0.7.0.rst?plain=1#L109C1-L117C57). See #773.

RDaxini:
> The previous definition was also not correct. IMO it doesn't you can't use the term "broadband" about effective irradiance which has potentially been modified for spectral effects.
> 
> > g_poa_effective;broadband plane of array effective irradiance.

If you apply a dimensionless spectral factor correction to a value of broadband irradiance in Wm⁻², it's still a broadband value in Wm⁻² but just a fraction of the original broadband value. Broadband just means its the irradiance between a (wide) wavelength range, no?

docs/sphinx/source/whatsnew/v0.11.2.rst
```diff
@@ -6,7 +6,9 @@ v0.11.2 (Anticipated December, 2024)
 
 Deprecations
 ~~~~~~~~~~~~
-
+* Changed the `g_poa_effective` parameter name to
+  `effective_irradiance` in :py:func:`~pvlib.pvsystem.PVSystem.pvwatts_dc`.
+  (:pull:`2235`)
```
RDaxini, line 11: Maybe it's still worth linking the related issue?  I understand they aren't directly connected though. Just a thought
```suggestion
  (:issue:`1253`, :pull:`2235`)
```

AdamRJensen:
> Should a deprecation-warning period be established? I'm not aware of the internal methodology you guys make use of, but I've seen that [changing params have been deprecated before in this project](https://github.com/pvlib/pvlib-python/blob/4cfda4a14366217ba9bb3b15d2531c61b7507e69/docs/sphinx/source/whatsnew/v0.7.0.rst?plain=1#L109C1-L117C57). See #773.

This would be ideal. But when it's just s rename I don't see how it can be deprecated nicely.

cwhanse:
I don't think I agree with this change of name `g_poa_effective` to `effective_irradiance`. The current description is of a quantity that is not `poa_global` but also is not `effective_irradiance` which is supposed to be after accounting for reflections, soiling, and spectrum. I agree that the description can be improved.

Just a comment about names: sometimes variable names were chosen to more closely connect the code with the text in the reference. Didn't happen here (the reference uses :math:`I_{tr}` so I'm guessing the submitter of this code came up with `g_poa_effective` as a compromise :math:`G_{POA}` being common in literature, and adding "_effective" because reflections have been removed.

```
    g_poa_effective: numeric
    effective_irradiance: numeric
        Irradiance transmitted to the PV cells. To be
        fully consistent with PVWatts, the user must have already
        applied angle of incidence losses, but not soiling, spectral,
        etc. [W/m^2]
```

echedey-ls:
> This would be ideal. But when it's just a rename I don't see how it can be deprecated nicely.

@AdamRJensen , we can work it out, similarly to #773 with some little variations. Hint: signature will need to be:

```python
def pvwatts_dc(self, effective_irradiance, temp_cell, *, g_poa_effective=None):
```

and the warning when it's used:

```python
if g_poa_effective:
    warnings.warn("Use 'effective_irradiance' to suppress this warning", pvlibDeprecationWarning)
```

then the warning test.

> because reflections have been removed

If I understand it well, then `[post_]iam_irradiance`, `iam_modified_irradiance`, `irradiance_after_iam`, `poa_global_after_iam`, `poa_g_after_iam` may be options that fit the description better. What do you think about it @cwhanse ?

I'm +1 to changing the parameter name and description, +1 to using a self-explanatory parameter name, +1 for `poa_global_after_iam` if I had to choose, not a strong opinion.

RDaxini:
> The current description is of a quantity that is not `poa_global` but also is not `effective_irradiance` which is supposed to be after accounting for reflections, soiling, and spectrum. I agree that the description can be improved.

We already have a definition of `effective_irradiance`, which includes both reflection and spectral effects:
https://github.com/pvlib/pvlib-python/blob/4cfda4a14366217ba9bb3b15d2531c61b7507e69/pvlib/pvsystem.py#L2309-L2311

So the definition proposed here (AOI correction only) would conflict with the existing definition above.

I was just reading through the pvwatts manual now --- page 9 eqn 8 --- perhaps we could just just use the same "transmitted" terminology here instead, i.e. `transmitted_irradiance`?

cwhanse:
> "transmitted" terminology here instead, i.e. transmitted_irradiance?

The De Soto model paper uses the term "absorbed irradiance" for this quantity

kandersolar:
If this parameter name changes, we should consider doing it in combination with broader changes to that function (name and module, see #1350 and https://github.com/pvlib/pvlib-python/issues/1544#issuecomment-1358068972).  

PVWatts v5 doesn't account for spectrum (so you could say implicitly $M=1$). Soiling is applied as a generic loss factor after DC power is modeled.  I agree that `effective_irradiance` is not quite right for the quantity in question, but that bothers me less than inventing a new term would.  I suggest renaming to `effective_irradiance` and listing the caveats in the parameter description.

AdamRJensen:
> If this parameter name changes, we should consider doing it in combination with broader changes to that function (name and module, see #1350 and [#1544 (comment)](https://github.com/pvlib/pvlib-python/issues/1544#issuecomment-1358068972)).
> 
> PVWatts v5 doesn't account for spectrum (so you could say implicitly M = 1 ). Soiling is applied as a generic loss factor after DC power is modeled. I agree that `effective_irradiance` is not quite right for the quantity in question, but that bothers me less than inventing/adopting a new term would. I suggest renaming to `effective_irradiance` and listing the caveats in the parameter description.

+1

cwhanse:
> I suggest renaming to effective_irradiance and listing the caveats in the parameter description.

The distinction between `poa_global` and `effective_irradiance` is probably not important to typical users of `pvwatts_dc`. If we are choosing to rename `g_poa_effective` to one of the already-used terms, I would prefer `poa_global` rather than `effective_irradiance`. 

Although I'll point out the irony of this conversation at the same time we are discussing various instances of inconsistent parameter names.

adriesse:
I would vote for `effective_irradiance` because that's consistent with the other DC module models/functions.

RDaxini:
(not sure if my earlier comment was holding anything up so, just in case, a follow up comment...)

In general I am not keen on reusing an existing variable with a different definition ([comment](https://github.com/pvlib/pvlib-python/pull/2235#issuecomment-2381632496)), but in this specific case I think @kandersolar's [comment](https://github.com/pvlib/pvlib-python/pull/2235#issuecomment-2383641130)

> (so you could say implicitly M = 1 )

addresses the overlap to a sufficient extent so I'll leave my vote neutral (not opposed) on `effective_irradiance` / other suggestions:)

docs/sphinx/source/whatsnew/v0.11.3.rst
```diff
@@ -41,6 +41,10 @@ Bug fixes
   returns an IANA string. (:issue:`2340`, :pull:`2341`)
 * :py:func:`~pvlib.iotools.get_pvgis_tmy` with ``outputformat='csv'`` now
   works with the updated data format returned by PVGIS. (:issue:`2344`, :pull:`2395`)
+* Changed the `g_poa_effective` parameter name to
+  `effective_irradiance` in :py:func:`~pvlib.pvsystem.PVSystem.pvwatts_dc`.
```
echedey-ls, line 45:
```suggestion
* Rename parameter name ``g_poa_effective`` to ``effective_irradiance`` in
  :py:func:`~pvlib.pvsystem.PVSystem.pvwatts_dc` and :py:func:`~pvlib.pvsystem.pvwatts_dc`.
```

pvlib/pvsystem.py
```diff
@@ -2799,7 +2799,7 @@ def scale_voltage_current_power(data, voltage=1, current=1):
     return df_sorted
 
 
-def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
+def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
```
echedey-ls, line 2802:
```suggestion
from pvlib._deprecation import renamed_kwarg_warning

@renamed_kwarg_warning("0.12.0", "g_poa_effective", "effective_irradiance", "0.13.0")
def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
```
What do you think about deprecating the parameter names with the decorator introduced in #2237?

pvlib/pvsystem.py
```diff
@@ -846,23 +846,23 @@ def scale_voltage_current_power(self, data):
         )
 
     @_unwrap_single_value
-    def pvwatts_dc(self, g_poa_effective, temp_cell):
+    def pvwatts_dc(self, effective_irradiance, temp_cell):
```
echedey-ls, line 849:
```
@renamed_kwarg_warning("0.12.0", "g_poa_effective", "effective_irradiance", "0.13.0")
@_unwrap_single_value
def pvwatts_dc(self, effective_irradiance, temp_cell):
```

echedey-ls (commented):
Some nitpicking on the readme and proposal of using the may-be-keyword argument decorator to give some graceful warnings. Seems a good change to me.

pvlib/pvsystem.py
AdamRJensen: It's obviously a very good idea :)

AdamRJensen:
@echedey-ls I think I fixed it as per your suggestions - any chance you could take another look? 😄

kandersolar:
Whatever change we make here, it would be nice (though not required, since we're just deprecating, not removing) to make it in the .0 release this week.

> If we are choosing to rename `g_poa_effective` to one of the already-used terms, I would prefer `poa_global` rather than `effective_irradiance`.

@cwhanse any chance this preference has changed since September?  I think yours is the only voice of opposition to `effective_irradiance`.

FWIW I still think the reasoning in https://github.com/pvlib/pvlib-python/pull/2235#issuecomment-2383641130 is sound; PVWatts v5 accounts for soiling elsewhere, and spectrum not at all, so IMHO the irradiance quantity in question is `effective_irradiance` in spirit as far as PVWatts v5 is concerned.  I don't think calling it `poa_global` is suitable given that it's supposed to account for IAM.

wholmgren:
FWIW the original PVWatts PR does have some discussion about parameter names. https://github.com/pvlib/pvlib-python/pull/195#issuecomment-227856824

I suspect that many users of `pvwatts_dc` are applying soiling or spectral losses before calling this function. So my vote in 2025 is to call it `effective_irradiance`, note the discrepancy with the literature, and move on. I don't think we want to introduce something like the originally proposed `irradiance_transmitted` or variations.

cwhanse:
> any chance this preference has changed since September? I think yours is the only voice of opposition to effective_irradiance.

I'm ok with `effective_irradiance`.

I concur with the reasoning in https://github.com/pvlib/pvlib-python/pull/2235#issuecomment-2383641130 which, in pvlib terms, is equivalent to `pvwatts_dc` setting `spectrum_model='no_loss'`. From this perspective the input is `effective_irradiance`.

In fact, that leads to a possible definition of `effective_irradiance`: broadband plane-of-array irradiance reduced by reflections, soiling and adjusted for spectrum.

That's how `effective_irradiance` is usually computed for PV modeling. In the context of model calibration from outdoor measurements, `effective_irradiance` is different: the irradiance estimated from measured `Isc`.

pvlib/pvsystem.py
```diff
@@ -843,24 +843,26 @@ def scale_voltage_current_power(self, data):
             for array, data in zip(self.arrays, data)
         )
 
+    @renamed_kwarg_warning(
+        "0.13.0", "g_poa_effective", "effective_irradiance", "0.14.0")
```
kandersolar:
```suggestion
        "0.13.0", "g_poa_effective", "effective_irradiance")
```

pvlib/pvsystem.py
```diff
@@ -2856,7 +2858,9 @@ def scale_voltage_current_power(data, voltage=1, current=1):
     return df_sorted
 
 
-def pvwatts_dc(g_poa_effective, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
+@renamed_kwarg_warning(
+    "0.13.0", "g_poa_effective", "effective_irradiance", "0.14.0")
```
kandersolar:
```suggestion
    "0.13.0", "g_poa_effective", "effective_irradiance")
```

kandersolar (commented):
@AdamRJensen there's a `g_poa_effective` in the new agriPV example too

adriesse:
> In fact, that leads to a possible definition of `effective_irradiance`: broadband plane-of-array irradiance reduced by reflections, soiling and adjusted for spectrum.

I'm still in favour of using this variable name here, but not convinced about the definition.

Issue #2231, opened 2024-09-28
Input parameter naming convention - time vs times


AdamRJensen:
Several pvlib functions take an index as one of the input parameters, however, it is inconsistent whether this parameter is called `times` or `time`.

I think we should document the preferred naming, add it to the symbols and variables list, and make an effort to standardize.

## See a few examples below of `TIME`:

https://github.com/pvlib/pvlib-python/blob/4cfda4a14366217ba9bb3b15d2531c61b7507e69/pvlib/solarposition.py#L30
https://github.com/pvlib/pvlib-python/blob/4cfda4a14366217ba9bb3b15d2531c61b7507e69/pvlib/solarposition.py#L128
https://github.com/pvlib/pvlib-python/blob/4cfda4a14366217ba9bb3b15d2531c61b7507e69/pvlib/tools.py#L120

## Here's a few examples of `TIMES`:
https://github.com/pvlib/pvlib-python/blob/4cfda4a14366217ba9bb3b15d2531c61b7507e69/pvlib/tools.py#L424
https://github.com/pvlib/pvlib-python/blob/4cfda4a14366217ba9bb3b15d2531c61b7507e69/pvlib/solarposition.py#L392
https://github.com/pvlib/pvlib-python/blob/4cfda4a14366217ba9bb3b15d2531c61b7507e69/pvlib/solarposition.py#L1343

RDaxini:
Related: #1253
In particular [this comment](https://github.com/pvlib/pvlib-python/issues/1253#issuecomment-2344790740) (btw I have updated the table with time/times now thanks to this issue)

adriesse:
I vote for `time`.

AdamRJensen:
I personally prefer times 😋

kandersolar:
A good guiding principle may be to use plural words only when multiple values are necessary.  For example, I think `times` is an appropriate name for the `_get_sample_intervals(times, win_length)` function, since it actually requires multiple timestamps.   For other functions (e.g. `hour_angle`) that work fine with single timestamps, would we say that the function returns `hour_angles`, and do the rise/set functions return `sunrises` and `sunsets`?  Here I think singular makes sense.

An argument against `time` in general is that it has the same name as a python standard library.

cwhanse:
> An argument against time in general is that it has the same name as a python standard library.

Good point. I overlooked this when agreeing with "time" with the singular/plural grammar in mind. I retract my vote; pvlib should not shadow any standard python names.

[...]

PR #2237, opened 2024-10-02
[MAINT]: Decorator to warn about changed parameter names & allow for deprecation period

echedey-ls:
- [x] Eases addressing #2231 (partially covered here) and #2235
 - [x] Tests added
    - do we want this in the public API (I'd say no)?

@AdamRJensen is threatening my mail inbox with issues reports from the parameter renaming in #2231 and #2235 (just kidding). So maybe instead of making breaking changes we can have some deprecation period.

Maybe over-engineering a pattern in each function for each rename is not the best choice ([my own idea](https://github.com/pvlib/pvlib-python/pull/2235#issuecomment-2381632318)), but we can work out some decorator to do so.

### Here is that decorator, instructions and recommendations on how to use it:
- Introduced and tested a decorator that emits a warning when the old parameter name is used
  - It does NOT modify docs
  - It does NOT check in which version of pvlib to fail automatically, that's left to the test suite
  - For these non-features, I can try to do them if you are sure of what you want.
- Example of edits required in `pvlib.solarposition.hour_angle`, docs down below.

### Wiki?

In addition, I propose adding instructions on how to proceed with renamings in the wiki.

### Docs
- `pvlib.solarposition.hour_angle`:
  - Main latest: https://pvlib-python.readthedocs.io/en/latest/reference/generated/pvlib.solarposition.hour_angle.html
  - This PR: https://pvlib-python--2237.org.readthedocs.build/en/2237/reference/generated/pvlib.solarposition.hour_angle.html
- `pvlib.solarposition.sun_rise_set_transit_spa`:
  - Main latest: https://pvlib-python.readthedocs.io/en/latest/reference/generated/pvlib.solarposition.sun_rise_set_transit_spa.html
  - This PR: https://pvlib-python--2237.org.readthedocs.build/en/2237/reference/generated/pvlib.solarposition.sun_rise_set_transit_spa.html#pvlib-solarposition-sun-rise-set-transit-spa

### DIFF
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -410,7 +410,8 @@ def sun_rise_set_transit_spa(time, latitude, longitude, how='numpy',
         Must be localized to the timezone for ``latitude`` and ``longitude``.
 
         .. versionchanged:: 0.11.2
-            Renamed from ``times`` to ``time``. Removal in ``v0.12.0``.
+            The ``times` parameter has been renamed ``time``. The deprecated
+            ``times`` parameter will be removed in ``v0.12.0``.
 
     latitude : float
         Latitude in degrees, positive north of equator, negative to south
--- a/pvlib/solarposition.py
+++ b/pvlib/solarposition.py
@@ -23,7 +23,7 @@ import pandas as pd
 import scipy.optimize as so
 import warnings
 
-from pvlib import atmosphere
+from pvlib import atmosphere, tools
 from pvlib.tools import datetime_to_djd, djd_to_datetime
 
 
@@ -199,11 +199,7 @@ def spa_c(time, latitude, longitude, pressure=101325, altitude=0,
         raise ImportError('Could not import built-in SPA calculator. ' +
                           'You may need to recompile the SPA code.')
 
-    # if localized, convert to UTC. otherwise, assume UTC.
-    try:
-        time_utc = time.tz_convert('UTC')
-    except TypeError:
-        time_utc = time
+    time_utc = tools._pandas_to_utc(time)
 
     spa_out = []
 
@@ -312,13 +308,11 @@ def spa_python(time, latitude, longitude,
         avg. yearly air pressure in Pascals.
     temperature : int or float, optional, default 12
         avg. yearly air temperature in degrees C.
-    delta_t : float, optional, default 67.0
+    delta_t : float or array, optional, default 67.0
         Difference between terrestrial time and UT1.
         If delta_t is None, uses spa.calculate_deltat
         using time.year and time.month from pandas.DatetimeIndex.
         For most simulations the default delta_t is sufficient.
-        *Note: delta_t = None will break code using nrel_numba,
-        this will be fixed in a future version.*
         The USNO has historical and forecasted delta_t [3]_.
     atmos_refrac : float, optional
         The approximate atmospheric refraction (in degrees)
@@ -378,7 +372,9 @@ def spa_python(time, latitude, longitude,
 
     spa = _spa_python_import(how)
 
-    delta_t = delta_t or spa.calculate_deltat(time.year, time.month)
+    if delta_t is None:
+        time_utc = tools._pandas_to_utc(time)
+        delta_t = spa.calculate_deltat(time_utc.year, time_utc.month)
 
     app_zenith, zenith, app_elevation, elevation, azimuth, eot = \
         spa.solar_position(unixtime, lat, lon, elev, pressure, temperature,
@@ -418,13 +414,11 @@ def sun_rise_set_transit_spa(times, latitude, longitude, how='numpy',
         Options are 'numpy' or 'numba'. If numba >= 0.17.0
         is installed, how='numba' will compile the spa functions
         to machine code and run them multithreaded.
-    delta_t : float, optional, default 67.0
+    delta_t : float or array, optional, default 67.0
         Difference between terrestrial time and UT1.
         If delta_t is None, uses spa.calculate_deltat
         using times.year and times.month from pandas.DatetimeIndex.
         For most simulations the default delta_t is sufficient.
-        *Note: delta_t = None will break code using nrel_numba,
-        this will be fixed in a future version.*
     numthreads : int, optional, default 4
         Number of threads to use if how == 'numba'.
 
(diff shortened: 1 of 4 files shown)


echedey-ls:
First thing is to consider if such a feature is needed/wanted for pvlib, feel free to weight in :)

cwhanse:
Seems elegant and useful, and easy to remove if it proves troublesome.

pvlib/tests/test__deprecation.py
kandersolar, line 1: Some tests in `test_conftest.py` can/should probably be migrated here.  Can be a follow-up PR.

pvlib/tests/test__deprecation.py
```diff
@@ -0,0 +1,39 @@
...
+    # assert no warning is raised when using the new kwarg
+    assert renamed_kwarg_func(new_kwarg=1) == 1  # as keyword argument
+    assert renamed_kwarg_func(1) == 1  # as positional argument
+
+    # assert a warning is raised when using the old kwarg
+    with pytest.warns(Warning, match="Parameter 'old_kwarg' has been renamed"):
+        assert renamed_kwarg_func(old_kwarg=1) == 1
+
+    # assert an error is raised when using both the old and new kwarg
+    with pytest.raises(ValueError, match="they refer to the same parameter."):
```
kandersolar, line 32: Maybe worth testing that `func.__name__` and such get interpolated correctly?

pvlib/solarposition.py
```diff
@@ -1340,15 +1341,20 @@ def solar_zenith_analytical(latitude, hourangle, declination):
...
 
     Parameters
     ----------
-    times : :class:`pandas.DatetimeIndex`
+    time : :class:`pandas.DatetimeIndex`
         Corresponding timestamps, must be localized to the timezone for the
         ``longitude``.
+
+        .. versionchanged:: 0.11.2
+            Renamed from ``times`` to ``time``.
```
kandersolar, line 1356: It would be nice if the docs somehow indicated that `times` is still allowed, for now.  Maybe we should we use `deprecated` instead of `versionchanged` during the deprecation period, and `versionchanged` afterwards?

kandersolar (commented):
I am cautiously on board with the decorator :)

pvlib/solarposition.py
AdamRJensen: I don't see a point to state that it's still allowed. The less work the better so I vote for version changed only.  Deprecating is already cumbersome and I'm hesitant to touch that require it

pvlib/tests/test__deprecation.py
```diff
@@ -0,0 +1,39 @@
...
+    def func(new_kwarg):
+        return new_kwarg
+
+    return func
+
+
+def test_renamed_kwarg_warning(renamed_kwarg_func):
+    # assert no warning is raised when using the new kwarg
+    assert renamed_kwarg_func(new_kwarg=1) == 1  # as keyword argument
+    assert renamed_kwarg_func(1) == 1  # as positional argument
```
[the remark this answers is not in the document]
echedey-ls: Oh damn, I overlooked that :(

pvlib/tests/test__deprecation.py
echedey-ls: Will be a follow-up PR, I'm out for a few days.

pvlib/solarposition.py
echedey-ls: My line of thought was exactly the same of @AdamRJensen 's. However, I did hesitate about the message. It can be changed to something like:

```rst
.. versionchanged:: 0.12.0
    Renamed from ``times`` to ``time``. ``times`` was deprecated in ``0.11.2``.
```
or
```rst
.. versionchanged:: 0.11.2
    Renamed from ``times`` to ``time``. ``times`` will be removed in ``0.12.0``.
```
which explains both things but clutters the docs IMO.

I don't mind changing it thou.

echedey-ls: But it's true that in v0.12.0 a follow-up PR to clean up the test that I have pushed now in https://github.com/pvlib/pvlib-python/pull/2237/commits/65e144b200b35d1c1c757158f6d1786c4066ac9b is required

echedey-ls:
I'd say this is now ready to choose a normalized deprecation/versionchanged directive format and discuss how to proceed in the future removal, or keep the admonitions.

I have a slight preference for what's been done at `sun_rise_set_transit_spa`, and just remove the parameter `deprecated` directive in v0.12.

pvlib/solarposition.py
```diff
@@ -404,8 +406,12 @@ def sun_rise_set_transit_spa(times, latitude, longitude, how='numpy',
 
     Parameters
     ----------
-    times : pandas.DatetimeIndex
+    date : pandas.DatetimeIndex
```
cwhanse, line 409: How did this name change creep into this PR?

cwhanse: https://github.com/pvlib/pvlib-python/issues/2231#issuecomment-2389814884

echedey-ls: I changed my mind about waiting for feedback in the issue just to show another example, finally choose between each admonition type and test whether any decorator argument persisted between different uses.

I can revert back that change, sorry.

I considered this one should be renamed to `date` rather than `time`. In fact, the inner function being wrapped uses `dates` as the parameter name.

echedey-ls: Resolved, feel free to leave a review :D

echedey-ls:
Let me know if new actions are desired in this PR, or if you prefer to split add decorator / address issues in different PRs.

pvlib/solarposition.py
```diff
@@ -404,8 +406,12 @@ def sun_rise_set_transit_spa(times, latitude, longitude, how='numpy',
 
     Parameters
     ----------
-    times : pandas.DatetimeIndex
+    time : pandas.DatetimeIndex
         Must be localized to the timezone for ``latitude`` and ``longitude``.
+
+        .. versionchanged:: 0.11.2
+            Renamed from ``times`` to ``time``. Removal in ``v0.12.0``.
```
AdamRJensen, line 413:
```suggestion
            The ``times` parameter has been renamed ``time``. The deprecated
            ``times`` parameter will be removed in ``v0.12.0``.
```
I think this is a clearer

AdamRJensen (commented):
I think this is awesome!

kandersolar (commented):
@echedey-ls I hate to ask for more work, but I think we should limit the scope of this PR to only creating the decorator (not applying it to real pvlib functions).  Would you mind splitting out the `solarposition.py` parameter renames into new PR(s)?  That will give them better visibility.  

It was helpful to see the decorator "in action" on real examples though :)

echedey-ls:
No problem. It doesn't make much sense to open the other PR now without the decorator on main, so I'll leave that up to you. These branch's commits already have the necessary changes to cherrypick them or just start from scratch, they aren't that many lines.

kandersolar (approved):
Thanks @echedey-ls!

PR #2277, opened 2024-10-22
[MAINT]: Remove tests from wheel distro

### LINKED ISSUE #2271: Exclude unnecssary data from pvlib packaging
Most of the data in the data folder are only used for testing and not useful for any examples. However, they add up a considerable amount of space (MBs) relative to the pvlib package itself. For applications such as WASM (code running in the browser), all this clutter slows down loading significantly. The data folder is currently 57 MB out of a total of 63 MB.

Of course, there are data files that we want to keep such as the Linke turbidity and CEC database. However, a rough estimation shows that removing unused data files can reduce the pvlib package but half!

Excluding/including data files can be specified in the [pyproject.toml file](https://setuptools.pypa.io/en/latest/userguide/datafiles.html), for example:

    [tool.setuptools.packages.find]
    where = ["src"]  # list of folders that contain the packages (["."] by default)
    include = ["my_package*"]  # package names should match these glob patterns (["*"] by default)
    exclude = ["my_package.tests*"]  # exclude packages matching these glob patterns (empty by default)

This can also be done in the **MANIFEST.in** file, which is pvlib's current approach, and seems to have finer control.

The pvlib/tests folder can also be excluded, and save 2.5 MB.

echedey-ls:
- [x] Closes #2271
 - [  ] Tests added
 - [  ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.
 - [  ] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.

My biggest fear is that `setuptools` is caching things so please make a fresh local test. So far everything looks great to me:

- Tests and test data included in sdist
- Neither tests nor test data included in bdist

Partially addresses #1056 by moving all data files exclusively used for testing into `pvlib/tests/data`

Whatsnew already updated with quantified sizes from wheel both zipped and once extracted compared against `v0.11.1`.


echedey-ls:
I haven't checked whether `conda-forge` will complain because I don't have a clue on how it tests packages, from sdist, bdist, ... Maybe anyone more knowledgable can foretell about that.

kandersolar:
I think there shouldn't be any problem with conda-forge.  It doesn't run the tests, nor does it rely on the presence of the test data files.  Additionally, the process is entirely driven by the sdist (not wheel) files on PyPI anyway, so if we only change the wheel files, I think there can't be an issue.

For reference, here's the conda-forge recipe for pvlib-python: https://github.com/conda-forge/pvlib-python-feedstock/blob/main/recipe/meta.yaml

docs/sphinx/source/whatsnew/v0.11.2.rst
```diff
@@ -10,7 +10,10 @@ Deprecations
 
 Enhancements
 ~~~~~~~~~~~~
-
+* Reduced space requirements by removing tests and test files from wheel.
```
AdamRJensen, line 13:
```suggestion
* Reduced space requirements by excluding tests and test files from wheel.
```

AdamRJensen (approved):
A brief review from my phone. But I like it!

echedey-ls:
As I posted in #2271, I'm +1 to making the flat layout. Feel free to react to this message if you [dis]agree.

echedey-ls:
That's it. Flat structure serious proposal. One thing, the `/scripts` folder is getting into the `sdist`. May as well ignore it in the `MANIFEST.in`. Not doing it now cause it may be scope creep. Happy new year btw.

kandersolar:
@echedey-ls any idea why some data files are showing as having their contents changed?

image

echedey-ls:
I was utterly wrong - they were committed as CRLF, now they will stay like that. In any case, 100% recommend normalization of the line endings 🌞

codecov.yml
```diff
@@ -14,14 +14,15 @@ flags:
   core:
     paths:
       - pvlib/
+      - tests/
```
kandersolar, line 17: I think we do not want `tests` listed here

tests/iotools/test_solcast.py
kandersolar, line 1: Looks like this file still has the CR/LF issue :)

kandersolar (approved):
I think the time has come to move forward with this PR.  @echedey-ls there's a merge conflict, and two last comments below.

codecov.yml
echedey-ls: I took me a while to grasp what was desired here, but I got to this for the following reasons:
1. It previously was
    ```yaml
    paths:
      - pvlib/
      - '!pvlib/iotools/'
      - '!pvlib/tests/iotools/'
    ```
    I understand that in version `pvlib/tests` was included, right?
2. `tests/` also has code to test, for example:
   * [conftest.py:fail_on_pvlib_version(...)](https://github.com/pvlib/pvlib-python/blob/8bc594ca3587d8b76466062399b64fa9920c1b18/pvlib/tests/conftest.py#L21)
     Tested in
      https://github.com/pvlib/pvlib-python/blob/8bc594ca3587d8b76466062399b64fa9920c1b18/pvlib/tests/test__deprecation.py#L13C1-L28C20
  I did suppose `codecov` is smart enough not expect coverage of the test functions. I would like to verify it but the new GitHub experience ✨  does not let me see the coverage result file.

kandersolar: Ok, I agree now.  I retract my objection.  Good catch!!

echedey-ls:
Thanks, Kevin 😃

PR #2296, opened 2024-11-18
fix: missing subscript

FelixKoTU:
Subscript "s" is missing for the number of cells in series (N_s) in latex math expressions.

### DIFF
--- a/docs/sphinx/source/user_guide/singlediode.rst
+++ b/docs/sphinx/source/user_guide/singlediode.rst
@@ -24,7 +24,7 @@ diode model equation.
 
 .. math::
 
-   I = I_L - I_0 \left(\exp \left(\frac{V + I R_s}{n Ns V_{th}} \right) - 1 \right)
+   I = I_L - I_0 \left(\exp \left(\frac{V + I R_s}{n N_s V_{th}} \right) - 1 \right)
        - \frac{V + I R_s}{R_{sh}}
 
 Lambert W-function is the inverse of the function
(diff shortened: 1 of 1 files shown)


docs/sphinx/source/user_guide/singlediode.rst
```diff
@@ -111,4 +111,4 @@ Clifford W. Hansen, Sandia `Report SAND2015-2065
 
 [4] "Computer simulation of the effects of electrical mismatches in
 photovoltaic cell interconnection circuits" JW Bishop, Solar Cell (1988)
-:doi:`10.1016/0379-6787(88)90059-2`
\ No newline at end of file
+:doi:`10.1016/0379-6787(88)90059-2`
```
RDaxini, line 114: What changed here? It looks like nothing happened, but it shows as a change

RDaxini (commented):
Thanks for spotting and fixing this @FelixKoTU.
Shall we expand the scope a little bit? We could link the references properly in line 20: `by Jain and Kapoor _[1] _[2], and Hansen _[3].` We could also standardize the formatting of the reference list. See [here](https://pvlib-python.readthedocs.io/en/latest/contributing/style_guide.html#references) for further info

docs/sphinx/source/user_guide/singlediode.rst
FelixKoTU: This seems to be an oversight on my part, I did not mean to alter that line.

cwhanse: > We could link the references properly in line 20: by Jain and Kapoor _[1] _[2], and Hansen _[3]. 
> We could also standardize the formatting of the references

Only if @FelixKoTU wants to, in my view.

This page really doesn't fit in a User Guide, it's more of detailed background we ought to have somewhere but didn't know where else to put it. My point is that we ought to consider rewriting this page and editing the references can be done in that revision.

RDaxini: > Only if @FelixKoTU wants to, in my view.
> 
> This page really doesn't fit in a User Guide, it's more of detailed background we ought to have somewhere but didn't know where else to put it. My point is that we ought to consider rewriting this page and editing the references can be done in that revision.

Fair point!

RDaxini (approved):

kandersolar:
> What changed here? It looks like nothing happened, but it shows as a change

I think it's whether the line ends in `\n` or not: https://github.com/FelixKoTU/pvlib-python/commit/5359efe3cd564807792cd99f6a80647928736859.patch

Doesn't really matter IMHO.  Thanks @FelixKoTU!

RDaxini:
> Doesn't really matter IMHO

for sure no big deal I was kinda just curious 😅

PR #2274, opened 2024-10-21
Adopt `dni_clear` variable name for clearsky DNI

RDaxini:
- [X] Related: #2272
 - [X] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [x] Tests ~added~ modified
 - [ ] ~Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.~
 - [ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [X] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

`dni_clear` is already in the variables and symbols list
Also related: #1253, #2306

Issue #2272, opened 2024-10-20
Standardise variable name for clearsky GHI and DNI, add to glossary


RDaxini:
**Is your feature request related to a problem? Please describe.**
I have counted three variations of clear sky GHI in the docs:

`clearsky` (in `clearsky.detect_clearsky`)
`clearsky_ghi` (in `irradiance.clearsky_index`)
`ghi_clearsky` (in `irradiance.dirindex`)
and no entry in the [variables and symbols page](https://pvlib-python.readthedocs.io/en/stable/user_guide/variables_style_rules.html) (soon to become a glossary #2234) 

for DNI, we have `dni_clear` in the variables and symbols page but also `clearsky_dni` (in `irradiance.dni`) and `dni_clearsky` (in `irradiance.dirindex`) in the docs.

**Describe the solution you'd like**
* Agree a single format
* Add/modify an entry in the variables and symbols page

**Additional context**
Related: #1253

RDaxini:
All, save for just `clearsky`, seem reasonable to me. I probably prefer either `clearsky_xxx` (written as you'd say it) or `xxx_clear` (shorter)

cwhanse:
I prefer `ghi_clear` but it's a mild preference. Parallels the few terms that are a conditional irradiance, i.e., `dni_clear`, `ghi_extra`, `poa_global`, etc.

RDaxini:
I think we're set on `ghi_clear` (also `dni_clear` for where it isn't already the case)

Would updating these variables be classed as breaking changes and require some kind of deprecation? I am thinking about #2237
Is a separate PR required for each function that is changed, or is one PR for all the functions involving the changing variable appropriate? (e.g. #2273)

adriesse:
Did you consider having just one PR for these two names since the changes are so closely related?

adriesse:
> > Did you consider having just one PR for these two names since the changes are so closely related?
> 
> @adriesse I asked about PR structure [here](https://github.com/pvlib/pvlib-python/issues/2272#issuecomment-2427750656) but thought each variable had to be separate, no response so I was not really sure. I confess I think I underestimated the overlap. Is it possible to combine PRs after creation? We should do that if it would help with merging later.

Sorry I didn't speak up earlier.  I'm not the best person to advise on the PR machinations.  :(

echedey-ls:
[...]

PR #2306, opened 2024-11-25
Adopt ghi_clear variable name for clearsky GHI

RDaxini:
- [X] Relates to #2272
 - [x] Tests added and modified

Also related: #1253, #2274


pvlib/irradiance.py
```diff
@@ -1614,6 +1614,11 @@ def ghi_from_poa_driesse_2023(surface_tilt, surface_azimuth,
         return ghi
 
 
+@renamed_kwarg_warning(
+    since='11.2',
+    old_param_name='clearsky_ghi',
+    new_param_name='ghi_clear',
+    removal="12.0")
 def clearsky_index(ghi, clearsky_ghi, max_clearsky_index=2.0):
```
echedey-ls, line 1622:
```suggestion
def clearsky_index(ghi, ghi_clear, max_clearsky_index=2.0):
```

pvlib/irradiance.py
```diff
@@ -2151,6 +2156,11 @@ def _dirint_bins(times, kt_prime, zenith, w, delta_kt_prime):
     return kt_prime_bin, zenith_bin, w_bin, delta_kt_prime_bin
 
 
+@renamed_kwarg_warning(
+    since='11.2',
+    old_param_name='ghi_clearsky',
+    new_param_name='ghi_clear',
+    removal="12.0")
 def dirindex(ghi, ghi_clearsky, dni_clearsky, zenith, times, pressure=101325.,
```
echedey-ls, line 2164:
```suggestion
def dirindex(ghi, ghi_clear, dni_clearsky, zenith, times, pressure=101325.,
```

echedey-ls (commented):
> I am pretty lost here:(

Kinda my fault for not documenting what I thought was obvious, sorry 😬 
You have to rename the variables in the signature, see suggestions.

pvlib/tests/test_irradiance.py
```diff
@@ -1235,6 +1235,12 @@ def test_clearsky_index():
     assert_series_equal(out, expected)
 
 
+def test_clearsky_index_renaming():
+    with pytest.warns(pvlibDeprecationWarning, match='ghi_clear'):
+        ghi, clearsky_ghi = 200, 300
+        irradiance.clearsky_index(ghi, clearsky_ghi)
```
echedey-ls, line 1241:
```suggestion
        irradiance.clearsky_index(ghi, clearsky_ghi=clearsky_ghi)
```
As positional arguments, it's all good to continue using them since the parameter hasn't changed position. But it's when using keyword arguments when things can go wrong. If a user had `clearsky_index(ghi=..., clearsky_ghi=...)`, now that the signature has changed, it would break and needs to be updated to the same, but using `ghi_clear` instead of the former should make it work.

The decorator is a shorthand to rename to the new param name if the old name was provided as the keyword param.

RDaxini: ah of course that makes sense thanks @echedey-ls !!

RDaxini: Thanks! I think part of the problem here was just my python knowledge and understanding of how to use decorators, in particular for deprecations. 
If you do develop something, perhaps you could use this PR as a base for the example once it's ready.
Now I know how to do this, I will complete #2274

pvlib/tests/test_irradiance.py
```diff
@@ -1235,6 +1257,12 @@ def test_clearsky_index():
     assert_series_equal(out, expected)
 
 
+def test_clearsky_index_clearsky_ghi_deprecation():
```
echedey-ls, line 1260:
```suggestion
@fail_on_pvlib_version("0.12")
def test_clearsky_index_clearsky_ghi_deprecation():
```
At some point during the decorator PR (#2237) I proposed adding this to the test, so removing the decorator in the future is not forgotten. But it didn't draw any attention. What do maintainers think about it? @kandersolar

Sry Dax for the review in dribs and drabs.

RDaxini: > Sry Dax for the review in dribs and drabs.

it's totally fine! I'll commit this change for now then if others disagree we could always revert the change

echedey-ls: > Question though, v0.12? is it not 12.0?

Yes, see the other uses of `@fail_on_pvlib_version`. It is due to its implementation. Intuitively the leftmost part is more important than the last numbers. `major.minor.patch` is the usual way to state them. You wouldn't want something to fail both on pvlib `0.12.0` and `1.12.0`.

pvlib/irradiance.py
```diff
@@ -2151,15 +2156,20 @@ def _dirint_bins(times, kt_prime, zenith, w, delta_kt_prime):
     return kt_prime_bin, zenith_bin, w_bin, delta_kt_prime_bin
 
 
-def dirindex(ghi, ghi_clearsky, dni_clearsky, zenith, times, pressure=101325.,
+@renamed_kwarg_warning(
+    since='11.2',
+    old_param_name='ghi_clearsky',
+    new_param_name='ghi_clear',
+    removal="12.0")
```
kandersolar, line 2163: Any pvlib version string (currently) should be `0.X.Y`.  Note that version strings have a specific format and technical meaning: https://semver.org/.  Additionally, they aren't just for humans to read; they identify versions on e.g. PyPI (see `0.11.1`, not `11.1`, as the latest version here: https://pypi.org/project/pvlib/) so that `pip install pvlib==0.11.1` does the right thing.  So the `0.` is important and cannot be omitted.

AdamRJensen:
I recommend that you hold on working concurrently on this PR (#2306) and #2274 as they concern many of the same functions and will create a headache to merge.

pvlib/tests/test_irradiance.py
```diff
@@ -1095,6 +1095,29 @@ def test_dirindex(times):
...
+            np.array([939.95469881]),
+            index=times
+        )
+        zenith = pd.Series(
+            np.array([10.56413562]),
+            index=times
+        )
+        pressure, tdew = 93193, 10
+        irradiance.dirindex(
+            ghi=ghi, ghi_clearsky=ghi_clearsky, dni_clearsky=dni_clearsky,
```
echedey-ls, line 1117:
```suggestion
            ghi=ghi, ghi_clearsky=ghi_clearsky, dni_clear=dni_clearsky,
```
Nitpicking about the warnings. Not problematic as of now, both warnings appear and the `ghi_clear` is matched, but since there already is a dedicated test for deprecated `dni_clearsky` (`test_dirindex_dni_clearsky_deprecation`, Line 1153) it is more atomic to limit the one the test is for.

echedey-ls (approved):
Ignore the comment below, PR nailed, @RDaxini ! Well done using that random, undocumented decorator.

pvlib/tests/test_irradiance.py
```diff
@@ -1095,6 +1095,29 @@ def test_dirindex(times):
...
 
+@fail_on_pvlib_version("0.13")
+def test_dirindex_ghi_clearsky_deprecation():
+    with pytest.warns(pvlibDeprecationWarning, match='ghi_clear'):
+        times = pd.DatetimeIndex(['2014-06-24T18-1200'])
+        ghi = pd.Series([1038.62], index=times)
+        ghi_clearsky = pd.Series(
+            np.array([1042.48031487]),
+            index=times
+        )
```
kandersolar, line 1106: Nitpick: it's best practice to include only the warning-emitting code inside the `with pytest.warns(...):` block.  Reasons: it makes the intent of the test clearer (future readers don't have to figure out which line of code is supposed to be emitting the warning), and it theoretically makes the test more robust (technically, it is possible that `pandas` could start emitting `pvlibDeprecationWarnings` talking about `ghi_clear`...).

So let's move all this variable initialization code outside the `pytest.warns` block.

kandersolar: Also I'm guessing much of this was just copy/paste from other tests, but no need for the `np.array`s here; `pd.Series([1042.4...], index=times)` should work fine.

pvlib/irradiance.py
```diff
@@ -2175,14 +2188,17 @@ def dirindex(ghi, ghi_clearsky, dni_clear, zenith, times, pressure=101325.,
     ghi : array-like
         Global horizontal irradiance. [Wm⁻²]
 
-    ghi_clearsky : array-like
+    ghi_clear : array-like
         Global horizontal irradiance from clear sky model. [Wm⁻²]
 
+    .. versionchanged:: 0.11.2
+        Renamed from ``ghi_clearsky`` to ``ghi_clear``.
```
kandersolar, line 2195:
```suggestion
        .. versionchanged:: 0.11.2
            Renamed from ``ghi_clearsky`` to ``ghi_clear``.
```

Sorry Dax, I meant the `..` needed to be aligned with the "Global".  Right now sphinx seems to be interpreting `versionchanged` as another parameter.  I think this suggested indentation should work (and apply the same to the others, of course), but let's check the RTD build to be sure.

kandersolar: Looks like it worked.  Now the others should be changed to match

image

RDaxini:
I think we're good now

PR #1997, opened 2024-03-23
Updated pd.date_range-API in plot_sunpath_diagrams.py

PhilBrk8:
freq='H' is depricated.
the parameter 'closed' was changed to 'inclusive'

 - [ ] Closes #xxxx
 - [ ] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [ ] Tests added
 - [ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.
 - [ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [ ] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [ ] Pull request is nearly complete and ready for detailed review.
 - [ ] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

Issue #1926, opened 2023-12-03
Replacing pandas with polars


langestefan:
**Is your feature request related to a problem? Please describe.**
Pandas is used extensively in pvlib all over the place. However after using it for many years I think it is now outdated and needs to be replaced. It is greedy, slow, inconsistent and the API is awkward and difficult to use. 

**Describe the solution you'd like**
I would like to propose replacing pandas with [polars](https://github.com/pola-rs/polars). Polars is a DataFrame interface built in Rust with many, many performance benefits. Too many to list them here, so I suggest visiting their page to learn more. 

I know this is a very big change, but I believe it will lead to some serious performance and usability benefits. So there is no real plan yet, I would like to just start the discussion and see what others think. I think a good starting point would be to benchmark some performance-critical pvlib code in pandas and polars so we can quantify what the performance benefits for pvlib would be.

mikofski:
>It is greedy, slow, inconsistent and the API is awkward and difficult to use.

I ❤️ this! 🤣

jules-ch:
Like @kevinsa5

I'd say it's way to early to support polars instead of pandas. The API is still unstable, version 0.X.
I'm a big user of polars on personal projects because of performance lazyloading, async collection, integration with the Apache Arrow Ecosystem.

But I really don't see any improvement in using polars here, apart from using the new shiny library, an effort is being made to use numpy array as inputs in functions where pandas were used before see #1455.
Performance-wise, benchmarks need to be made first and since we're not processing GB of data, we probably won't see any difference in performance, especially since we can use the new interface with the Arrow ecosystem with pandas 2.0.

Even when processing 20 years of weather data for a single location it's not even close to several millions of rows.
IMO I don't think it will make a significant gain in perf but I'm eager to see some benchmarks.

Moving forward using Numpy types instead of Dataframe or Series will make integrating with Polars or any Dataframe library easy.

We can probably use polars already if that's what you want with a bunch of PVLIB API.

markcampanelli:
I think it would be more appropriate to simply make pvlib "agnostic" to if one is using `pandas` vs. `polars` (vs. just plain old `numpy` float vectors and a Python `list` of timestamps). Not sure if this could be done with some "timestamped data provider" interface abstraction to lighten the load on users. Given the current evolution of pvlib, this would most likely be painful and disruptive to the `pandas` lovers. Also, sorting all this out would be quite the lift, I fear.

langestefan:
[...]

PR #2338, opened 2024-12-22
updated imports

PhilBrk8:


cwhanse:
@PhilBrk8 in my view, library-wide reformatting should be a deliberate decision connected with a set of linting rules, and tied to a major release, i.e., whenever we decide to move to v1.0. Although the edits here are improvements, I am discouraged from accepting them without already having discussed the linter and rules in the CI and having concurrence from maintainers about those rules.

Can you explain the motivation to update just the solarposition.py file? It appears that you started there then branched out to the full directory tree.

echedey-ls:
Wow. These are some pretty big changes @PhilBrk8 !! I was halfway redacting when @cwhanse weighed in, but his comment sums up what I was about to say.

Just to clarify, many times some specific style has been adopted to make equations more readable (in general some parts of the code) so that also needs to be taken into account by using comment macros to turn off the formatter.

The `imp` package removal seems good to me, and narrowing the `ImportError` to `ModuleNotFoundError`, just to clean up some legacy code and avoid the highly-unlikely problems derived from the imports.

In any case, and only if you are motivated to do so, I encourage you to address any of the [issues tagged as "good first issue"](https://github.com/pvlib/pvlib-python/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22), they really make for good improvements in this repo!! Feel free to reach out if you need any assistance 😉

PhilBrk8:
Hi @cwhanse & @echedey-ls 
quite a long time ago @kandersolar told me he would generally be up for it if I want to make it a project for myself to bring pvlib up to a newer / more modern python version.

I initially just wanted to implement type hints into the functions, but then saw some other things along the way.

Regarding the formatting:
as @cwhanse noticed correctly, I started with solarposition.py
Basically just because I have set formatOnSave=true in VS-Code.

I naively thought the PR is only applied for the commits up to its time of creation, but when I realized the newer ones are included automatically, I changed the PR-Name and left the commits for the whole tree in, to see what you think of it.

-----------------------------------------------------------------------------------------------------------------------------------------
But yes, reviewing / refactoring / modernizing all files is way out of scope for it to be done this way.

I then thought initially I just do everything on my fork, because I got curious about a modern & polars version of pvlib and thought it might be interesting to benchmark a "modern"-pvlib against the original at some point.
But if chenges turn out to be useful improvements, then this is also not the way it should be done I guess, but with a more thought through branching strategy. 
------------------------------------------
So then I came up with the following idea / proposal:
Are you generally up to create a "development" / "modern" branch in the original repo?
I would then keep the changes in respective brances like:
- feature/reformatting
- feature/type_hints
- feature/polars
- feature/refactor


Or should I rather create a discussion for that?

kandersolar:
> quite a long time ago @kandersolar told me he would generally be up for it if I want to make it a project for myself to bring pvlib up to a newer / more modern python version.

I'm not sure what "newer / more modern" means here -- python itself (e.g. python 3.13?), or dependencies like polars?  If the latter, I think there was likely some misunderstanding.  For example I thought the proposal here was more related to functionality/models than python versions: https://github.com/pvlib/pvlib-python/pull/1997#issuecomment-2066547388

In any case, it is best to open an issue proposing the change before putting in the work for a PR.  Sometimes there is a better way to accomplish the goal, or maybe the change is not desirable for reasons that could be explained in the issue.  Having that discussion first helps prevent wasted effort if the PR has to be changed or closed after the discussion.

> So then I came up with the following idea / proposal:

Several of these topics already have discussions, e.g. #1926 and #2062.  If you want to pursue these ideas, I think the first step is to get consensus that the idea is beneficial for pvlib.  None of the mentioned ideas (bulk reformatting, type hints, polars) have achieved that consensus yet.  So engaging in those discussions with new and persuasive arguments in favor of the ideas is the next step.  Only once the community is convinced does it make sense to talk about how specifically it should be implemented, e.g. branching strategy and such.

kandersolar:
Closing, as the above comments show that someone needs to champion a discussion before changes like these can be merged.  And if that consensus is reached, I think it would be easier to start fresh instead of resolving all the conflicts in this branch.

PR #2378, opened 2025-02-06
Add NSRDB GOES v4 to iotools

### LINKED ISSUE #2326: Add NSRDB GOES V4
NSRDB's latest "V4" dataset appears to be available. It would be nice to have this available in iotools.

See https://github.com/NREL/SAM/issues/1920 for some additional info. Quoting @cpaulgilman:

>NSRDB V4.0.0 data (1998-2023) is now available from new NREL Developer API endpoints:
>
>https://developer.nrel.gov/docs/solar/nsrdb/nsrdb-GOES-tmy-v4-0-0-download/
>
>https://developer.nrel.gov/docs/solar/nsrdb/nsrdb-GOES-aggregated-v4-0-0-download/
>
>SAM's Advanced Downloads option can access these endpoints, but the direct download from the Location and Resource page still accesses the older V3.2.2 (1998-2022) API endpoint.
>
>Note there is an "aggregated" and "conus" endpoint. The latter seems to be a subset of the former, but need to determine the correct endpoint to use.

Also note that there was an issue with DNI that has been fixed (also in that SAM issue's comments).

williamhobbs:
- [x] Closes #2326
 - [x] Tests added


williamhobbs:
What's the best practice for lines like this one that end up too long? Do I simply add `# noqa` or `# noqa: F401`? (I don't know what F401 is for, but I've seen it used here...)

```python
GOES_URL = NSRDB_API_BASE + "/api/nsrdb/v2/solar/nsrdb-GOES-aggregated-v4-0-0-download.csv"
```

https://github.com/williamhobbs/pvlib-python/blob/4af81a76b4d8bb86b992b5a87ca81dbec6ecb8d5/pvlib/iotools/goes4.py#L15

mikofski:
You could use parentheses 

```python
GOES_URL = NSRDB_API_BASE + ("/api/nsrdb/v2/solar/"
"nsrdb-GOES-aggregated-v4-0-0-download.csv")
```

or just break the string smaller and add them

```python
API_STUB = "/api/nsrdb/v2/solar/"
ENDPOINT = "nsrdb-GOES-aggregated-v4-0-0-download.csv"
GOES_URL = (NSRDB_API_BASE
    + API_STUB + ENDPOINT)
```

imo avoid ignoring pep8 warnings or else they become pointless.

williamhobbs:
Thanks, @mikofski!

docs/sphinx/source/whatsnew/v0.11.3.rst
```diff
@@ -12,6 +12,7 @@ Enhancements
 ~~~~~~~~~~~~
 * Fix a bug in :py:func:`pvlib.bifacial.get_irradiance_poa` which may have yielded non-zero
   ground irradiance when the sun was below the horizon. (:issue:`2245`, :pull:`2359`)
+* Added NREL NSRDB GOES v4 to `pvlib.iotools` (:issue:`2326`, :pull:`2378`)
```
echedey-ls, line 15:
```suggestion
* Add NREL NSRDB GOES v4 API client to :py:mod:`pvlib.iotools`. See :py:func:`~pvlib.iotools.get_goes4`, :py:func:`~pvlib.iotools.read_goes4` and :py:func:`~pvlib.iotools.parse_goes4`. (:issue:`2326`, :pull:`2378`)
```
I couldn't help but review this. The remaining code looks good, I will review in depth in the future :)

williamhobbs:
I'm not 100% sure that I set up the tests in `test_goes4.py` correctly. I copied what was done in `test_psm3.py` and manually downloaded and reformatted as needed the test data csv files. 

See `test_read_goes4_map_variables()` in particular. `columns_mapped` includes variables like `'Ozone'` that are part of a manual download but are not addressed in pvlib. 

I think it's correct, but I could be misunderstanding the goals of some of the tests.

williamhobbs:
Ok, made the switch from "goes4" to "psm4". 

I haven't done anything with @kandersolar's suggestion for different `get_` functions for different endpoints (e.g., `get_nsrdb_goes4_aggregated`, `get_nsrdb_goes4_conus`, and `get_nsrdb_goes4_tmy`). I'll wait for some consensus on that one.

williamhobbs:
Ok, I made an effort to create separate `get_` functions in a new branch: https://github.com/williamhobbs/pvlib-python/tree/nsrdb_psm4_dedicated_get_functions.

I'll hold off on merging that with the branch in this PR (https://github.com/williamhobbs/pvlib-python/tree/nsrdb_goes_v4) until someone suggests that I do it - I don't want to make a mess of things too quickly.

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,388 @@
+"""
+Get NSRDB PSM4
+see
+https://developer.nrel.gov/docs/solar/nsrdb/nsrdb-GOES-conus-v4-0-0-download/
```
AdamRJensen, line 4: Personally, I think the URL should be in the function references and not in the module description.

My suggestion is to have the module comment be just a single one-liner, e.g. "Functions for reading and retrieving data from NSRDB PSM4."

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,388 @@
...
+    'Dew Point': 'temp_dew',
+    'Relative Humidity': 'relative_humidity',
+    'Pressure': 'pressure',
+    'Wind Speed': 'wind_speed',
+    'Wind Direction': 'wind_direction',
+    'Surface Albedo': 'albedo',
+    'Precipitable Water': 'precipitable_water',
+    'AOD': 'aod',
+    'Alpha': 'alpha',
+    'Asymmetry': 'asymmetry',
```
AdamRJensen, line 45: I don't know what these are and they aren't part of the pvlib standard variables, so it's unusual for us to map them.

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+def get_psm4(latitude, longitude, api_key, email, names='tmy', interval=60,
+             attributes=ATTRIBUTES, leap_day=True, full_name=PVLIB_PYTHON,
+             affiliation=PVLIB_PYTHON, map_variables=True, url=None,
+             timeout=30):
+    """
+    Retrieve NSRDB PSM4 timeseries weather data from the PSM4 API. The NSRDB
+    is described in [1]_ and the PSM4 API is described in [2]_, [3]_, and
+    [4]_.
```
AdamRJensen, line 80:
```suggestion
    Retrieve NSRDB PSM4 timeseries weather data from the PSM4 API.
    
    The NSRDB is described in [1]_ and the PSM4 API is described in [2]_,
    [3]_, and [4]_.
```

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,388 @@
...
+        NREL API uses this to automatically communicate messages back
+        to the user only if necessary
+    names : str, default 'tmy'
+        PSM4 API parameter specifing year (e.g. ``2020``) or TMY variant to
+        download (e.g. ``'tmy'`` or ``'tgy-2022'``).  The allowed values
+        update periodically, so consult the NSRDB references below for the
+        current set of options.
+    interval : int, {60, 5, 15, 30}
+        interval size in minutes, must be 5, 15, 30 or 60. Must be 60 for
+        typical year requests (i.e., tmy/tgy/tdy).
```
AdamRJensen, line 100: I think we should call this ``time_period`` which is used by other iotools functions. I believe that @kandersolar regretted naming it ``interval`` (which was taken from NSRDBs naming convention).

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,388 @@
...
+    attributes : list of str, optional
```
AdamRJensen, line 101: Again I think this is better named something like ``parameters``.

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,388 @@
...
+        except JSONDecodeError:
+            errors = response.content.decode('utf-8')
+        raise requests.HTTPError(errors, response=response)
+    # the CSV is in the response content as a UTF-8 bytestring
+    # to use pandas we need to create a file buffer from the response
+    fbuf = io.StringIO(response.content.decode('utf-8'))
+    return parse_psm4(fbuf, map_variables)
+
+
+def parse_psm4(fbuf, map_variables=True):
```
AdamRJensen, line 221: I personally think we should get rid of the parse functions as it creates a LOT of duplicate documentation and significantly more functions than needed. It's unnecessary clutter and it takes very little to modify _read_ functions to also take filebuffers.

Alternatively, we could consider making the _parse_ functions private.

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,388 @@
...
+    """
+    Parse an NSRDB PSM4 weather file (formatted as SAM CSV). The NSRDB
+    is described in [1]_ and the SAM CSV format is described in [2]_.
```
AdamRJensen, line 224:
```suggestion
    Parse an NSRDB PSM4 weather file (formatted as SAM CSV).
    
    The NSRDB is described in [1]_ and the SAM CSV format is described in [2]_.
```

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,388 @@
...
+        metadata['altitude'] = metadata.pop('Elevation')
+
+    return data, metadata
+
+
+def read_psm4(filename, map_variables=True):
+    """
+    Read an NSRDB PSM4 weather file (formatted as SAM CSV). The NSRDB
+    is described in [1]_ and the SAM CSV format is described in [2]_.
+
```
AdamRJensen, line 358:
```suggestion
    Read an NSRDB PSM4 weather file (formatted as SAM CSV).
    
    The NSRDB is described in [1]_ and the SAM CSV format is described in [2]_.

```

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,388 @@
...
+    Parameters
+    ----------
+    filename: str
```
AdamRJensen, line 361: I think it should be str or path-like?

pvlib/tests/iotools/test_psm4.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    try:
+        demo_key = os.environ["NREL_API_KEY"]
+    except KeyError:
+        warnings.warn(
+            "WARNING: NREL API KEY environment variable not set! "
+            "Using DEMO_KEY instead. Unexpected failures may occur."
+        )
+        demo_key = 'DEMO_KEY'
+    return demo_key
```
AdamRJensen, line 46: If it's the same key as for the PSM3, then I suggest importing it from that test module (should be possible?) to avoid duplicating code.

AdamRJensen (commented):
A few initial comments

AdamRJensen:
Given the unanimous vote and the three weeks passed (see https://github.com/pvlib/pvlib-python/issues/2326#issuecomment-2698878568), you can go ahead with splitting the functions into individual ones.

Please still consider the comments above 😄

williamhobbs:
Ok @AdamRJensen, I implemented many of your comments into https://github.com/williamhobbs/pvlib-python/tree/nsrdb_psm4_dedicated_get_functions. I may be at the limit of my git/GitHub knowledge, though. What do I do next to get those changes (along with the dedicated functions) from that branch into this PR? I can follow up offline if that's best...

pvlib/iotools/psm4.py
kandersolar: I am generally +1 to ditching `parse_` and having `read_` work with buffers as well as filenames.  But let's do it as a follow-up PR.

pvlib/iotools/psm4.py
williamhobbs: removed

pvlib/iotools/psm4.py
williamhobbs: done

pvlib/iotools/psm4.py
williamhobbs: done

pvlib/iotools/psm4.py
williamhobbs: done

pvlib/iotools/psm4.py
williamhobbs: done

pvlib/iotools/psm4.py
williamhobbs: done

pvlib/iotools/psm4.py
williamhobbs: done

pvlib/iotools/psm4.py
williamhobbs: done

pvlib/tests/iotools/test_psm4.py
williamhobbs: holding off on this, per a conversation with @kandersolar

williamhobbs:
@AdamRJensen - I think I addressed all of your comments in one way or another. Thanks for the review - I think I forgot to say that earlier.

pvlib/tests/iotools/test_psm4.py
kandersolar: I moved the fixture to conftest and imported it in both psmN test modules

tests/conftest.py
```diff
@@ -84,6 +84,26 @@ def assert_frame_equal(left, right, **kwargs):
                                   reason='does not run on windows')
 
 
+@pytest.fixture(scope="module")
+def nrel_api_key():
+    """Supplies pvlib-python's NREL Developer Network API key.
+
+    Azure Pipelines CI utilizes a secret variable set to NREL_API_KEY
```
wholmgren, line 91:
```suggestion
    pvlib's CI utilizes a secret variable set to NREL_API_KEY
```

wholmgren (approved):

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,822 @@
...
+        in decimal degrees, between -180 and 180, east is positive
+    api_key : str
+        NREL Developer Network API key
+    email : str
+        NREL API uses this to automatically communicate messages back
+        to the user only if necessary
+    names : str, default '2023'
+        PSM4 API parameter specifing year (e.g. ``2023``) to download. The
+        allowed values update periodically, so consult the NSRDB reference
+        below for the current set of options.
```
AdamRJensen, line 103: No need to follow the strange naming convention of PSM. I suggest renaming this to ``"year"``.

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,822 @@
...
+    api_key : str
+        NREL Developer Network API key
+    email : str
+        NREL API uses this to automatically communicate messages back
+        to the user only if necessary
+    names : str, default 'tmy'
+        PSM4 API parameter specifing TMY variant to download (e.g. ``'tmy'``
+        or ``'tgy-2022'``).  The allowed values update periodically, so
+        consult the NSRDB references below for the current set of options.
+    time_period : int, {60}
```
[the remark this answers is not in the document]
kandersolar: @AdamRJensen could you clarify how this relates to your previous comment https://github.com/pvlib/pvlib-python/pull/2378#discussion_r2009195497

pvlib/iotools/psm4.py
williamhobbs: I think `year` makes sense for aggregated, conus, and full disk datasets. What do you think about the TMY dataset? Should we call it `year`, even though it has to include `tmy` or similar?

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,822 @@
...
+        When true, renames columns of the Dataframe to pvlib variable names
+        where applicable. See variable :const:`VARIABLE_MAP`.
+    url : str, optional
+        Full API endpoint URL. If not specified, the PSM4 GOES Aggregated v4
+        URL is used.
+    timeout : int, default 30
+        time in seconds to wait for server response before timeout
+    utc: bool, default : False
+        retrieve data with timestamps converted to UTC. False returns
+        timestamps in local standard time of the selected location
```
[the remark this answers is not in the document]
williamhobbs: Done.

pvlib/iotools/psm4.py
AdamRJensen: I agree that it's not ideal, but I think it's the best option. I'm open to other alternatives.

pvlib/iotools/psm4.py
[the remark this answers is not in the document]
AdamRJensen: My previous comment was rubbish.. I apologize.

pvlib/iotools/psm4.py
williamhobbs: Ok. I made the change to `year`.

pvlib/iotools/psm4.py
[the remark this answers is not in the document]
williamhobbs: Done. I replaced `time_period` with `time_step`.

docs/sphinx/source/reference/iotools.rst
```diff
@@ -26,6 +26,12 @@ of sources and file formats relevant to solar energy modeling.
    iotools.read_crn
    iotools.read_solrad
    iotools.get_solrad
+   iotools.get_nsrdb_psm4_aggregated
+   iotools.get_nsrdb_psm4_tmy
+   iotools.get_nsrdb_psm4_conus
+   iotools.get_nsrdb_psm4_full_disc
+   iotools.read_nsrdb_psm4
+   iotools.parse_nsrdb_psm4
```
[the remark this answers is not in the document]
williamhobbs: I vote for the other Will H.'s recommendation.

kandersolar:
@williamhobbs looks like some updates to the tests are needed: https://github.com/pvlib/pvlib-python/actions/runs/14574837205/job/40878680163?pr=2378#step:6:327

(the "remote data" tests only run when a maintainer adds the `remote-data` label to a PR, so that's why these test failures did not appear previously)

kandersolar:
Thanks @williamhobbs for iterating with the tests :) Many of the remaining code coverage issues are for API error handling conditions that we might not be able to easily induce, so I'd say let's ignore those.  However, it does look like `get_nsrdb_psm4_full_disc` isn't currently tested.  Can you add another test function for that?

williamhobbs:
@kandersolar, I think I finally addressed all the tests. Not sure about the failing codecov checks, though.

One note/question: I manually removed the Alpha and Asymmetry columns from the test psm4 csv files (after removing them from the variable map based on https://github.com/pvlib/pvlib-python/pull/2378/#discussion_r2009194528). Was simply deleting those columns from the csv files the right thing to do? I honestly can't remember how I created those files originally.

williamhobbs:
> ...it does look like `get_nsrdb_psm4_full_disc` isn't currently tested. Can you add another test function for that?

@kandersolar, yes, I'll work on that.

kandersolar:
> One note/question: I manually removed the Alpha and Asymmetry columns from the test psm4 csv files (after removing them from the variable map based on https://github.com/pvlib/pvlib-python/pull/2378). Was simply deleting those columns from the csv files the right thing to do?

It looks like no tests make use of those columns (in the PSM3 tests either).  So maybe ideally it would be good to include them, but since there's no special functionality related to those columns that we need to test, I think it's okay to leave them out.

williamhobbs:
> > One note/question: I manually removed the Alpha and Asymmetry columns from the test psm4 csv files (after removing them from the variable map based on #2378). Was simply deleting those columns from the csv files the right thing to do?
> 
> It looks like no tests make use of those columns (in the PSM3 tests either). So maybe ideally it would be good to include them, but since there's no special functionality related to those columns that we need to test, I think it's okay to leave them out.

I think leaving Alpha and Asymmetry columns in was causing `test_read_nsrdb_psm4_map_variables` to fail, which is why I removed them.

williamhobbs:
@kandersolar I think it's ready now.

pvlib/iotools/psm4.py
```diff
@@ -0,0 +1,819 @@
...
+    'wind_speed': 'wind_speed',
+    'wind_direction': 'wind_direction',
+    'albedo': 'surface_albedo',
+    'precipitable_water': 'total_precipitable_water',
+    'aod': 'aod',
+}
+
+
+def get_nsrdb_psm4_aggregated(latitude, longitude, api_key, email,
+                              year='2023', time_step=60,
```
kandersolar, line 74: I just noticed the default value `year='2023'` in this function and some others.  Is there a good reason for it?  I don't see much convenience in having a default year since presumably most users will override it for the year they actually need anyway.  And if it's possible to leave the parameter unspecified, it might funnel careless users towards requesting the wrong year unintentionally.

I think we should change it so that `year` does not have a default value for the calendar year (i.e. non-TMY) functions. @williamhobbs @adamrjensen any thoughts?

AdamRJensen: I think that's a mighty fine idea!

cwhanse: Does the NSRDB return an informative error if a year is requested for which there is no data? That might be why a default was included.

williamhobbs: Sounds good to me. 

@cwhanse, I get this when making a request to NSRDB without including a year or if I pick a year outside the allowed range (e.g., 1997):

`"errors":["The required 'years' or 'names' parameter must be a single number between 1998 and 2023. The comma delimited option is only available when using the JSON variant of this endpoint."]`

PR #2380, opened 2025-02-07
Include solarfactors in python 3.12 tests

kandersolar:
With the new solarfactors v1.6.0 release (and the shapely 2.0 compatibility it brings), it can be successfully installed and used on python 3.12 and above.  That means we don't need to exclude it from the python 3.12 test environment anymore.

### DIFF
--- a/ci/requirements-py3.12.yml
+++ b/ci/requirements-py3.12.yml
@@ -25,4 +25,4 @@ dependencies:
     - statsmodels
     - pip:
         - nrel-pysam>=2.0
-        # - solarfactors  # required shapely<2 isn't available for 3.12
+        - solarfactors


echedey-ls (approved):
Tests are successful ✓: https://github.com/pvlib/pvlib-python/actions/runs/13205591970/job/36867915441?pr=2380

Quick question, is there any reason not to test optional dependencies in the bare environment setup?
https://github.com/pvlib/pvlib-python/blob/2ca40aa377d0871905f3c12f7ab0de0d437e2e21/.github/workflows/pytest.yml#L72C1-L76C21

kandersolar:
Two reasons, I am guessing:
1) Historical: a long time ago, in a pvlib version far far away, I think it was difficult/impossible to install some of the optional dependencies using only pip.  And nobody has changed the test configuration since then.
2) Still relevant today: if all of the tests are run with `[optional]` installed, then we don't have any check that pvlib works without `[optional]`.

PR #2381, opened 2025-02-07
Replace zenith with solar_zenith in variable map for pvlib:psm3

### LINKED ISSUE #2377: zenith vs solar_zenith in psm3 variable maps
Should `'zenith': 'solar_zenith_angle', ` in `REQUEST_VARIABLE_MAP` instead be `'solar_zenith': 'solar_zenith_angle', `? In `VARIABLE_MAP` it is `solar_zenith`.

https://github.com/pvlib/pvlib-python/blob/2ca40aa377d0871905f3c12f7ab0de0d437e2e21/pvlib/iotools/psm3.py#L45-L61

ayushjariyal:
- [ ] Closes #2377

### DIFF
--- a/pvlib/iotools/psm3.py
+++ b/pvlib/iotools/psm3.py
@@ -49,7 +49,7 @@ REQUEST_VARIABLE_MAP = {
     'ghi_clear': 'clearsky_ghi',
     'dhi_clear': 'clearsky_dhi',
     'dni_clear': 'clearsky_dni',
-    'zenith': 'solar_zenith_angle',
+    'solar_zenith': 'solar_zenith_angle',
     'temp_air': 'air_temperature',
     'temp_dew': 'dew_point',
     'relative_humidity': 'relative_humidity',
--- a/pvlib/tests/iotools/test_psm3.py
+++ b/pvlib/tests/iotools/test_psm3.py
@@ -171,11 +171,11 @@ def test_read_psm3_map_variables():
     """test read_psm3 map_variables=True"""
     data, metadata = psm3.read_psm3(MANUAL_TEST_DATA, map_variables=True)
     columns_mapped = ['Year', 'Month', 'Day', 'Hour', 'Minute', 'dhi', 'ghi',
-                    'dni', 'ghi_clear', 'dhi_clear', 'dni_clear',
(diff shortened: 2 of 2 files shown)


cwhanse:
@ayushjariyal I feel like this bug should have been caught in the tests. Are you willing to add a check to [this test](https://github.com/pvlib/pvlib-python/blob/2ca40aa377d0871905f3c12f7ab0de0d437e2e21/pvlib/tests/iotools/test_psm3.py#L170) to verify that all intended columns have been mapped?

williamhobbs:
@cwhanse, does this test also need to be updated to check for the intended columns?

https://github.com/pvlib/pvlib-python/blob/2ca40aa377d0871905f3c12f7ab0de0d437e2e21/pvlib/tests/iotools/test_psm3.py#L185

This test already includes `'solar_zenith'`, but maybe I'm misunderstanding the test workflow:

https://github.com/pvlib/pvlib-python/blob/2ca40aa377d0871905f3c12f7ab0de0d437e2e21/pvlib/tests/iotools/test_psm3.py#L170

cwhanse:
> @cwhanse, does this test also need to be updated to check for the intended columns?

In both cases I think the better answer is no. 

The error (`zenith` in place of `solar_zenith`) is in a map from pvlib names to PSM3 names. The results of that mapping are not passed to other pvlib functions; they are only checked for a correct index. Tthe test structure (for all of iotools, I think, and to a large extent all of pvlib) does not enforce pvlib's parameter nomenclature except when calling another pvlib function. That kind of enforcement strikes me as complicated, probably undesirable, and certainly, well beyond the scope of this PR.

I vote to only fix the name and thank the contributor.

williamhobbs:
> I vote to only fix the name and thank the contributor.

@cwhanse, I agree. Thanks for the clarification.

cwhanse:
@ayushjariyal I don't think we need a note in the whatsnew for this fix, and you're already in the list of contributors. So this is done I think.

cwhanse (approved):

RDaxini:
Can we change the title of this issue to something like (suggestion) "Change `zenith` to `solar_zenith` in psm3 variable map"? Might make it easier to find if needed for future reference

kandersolar:
Thanks @ayushjariyal!

Issue #2330, opened 2024-12-11
Add new PVsyst temperature model parameter sets


cwhanse:
PVsyst has added a set of temperature model parameters for roof mounted systems that allow air flow beneath the modules, termed "semi-integration".

**Describe the solution you'd like**
Add this parameter set to pvlib\temperature\pvsyst_cell, and as a value for PVSystem's `racking_model` parameter.

#2293 should be merged first.

Muhammad-Rebaal:
Hi @cwhanse !
I would like to solve this issue and do the required enhancement so could you please assign this issue to me?
The solution I'd like to propose is : 

`PVsyst's "semi-integration" model to pvlib.temperature.pvsyst_cell and as a racking_model option in PVSystem for improved accuracy in roof-mounted system modeling.`

Thank you!

cwhanse:
@Muhammad-Rebaal sure thing.

Details of the PVsyst model are [here](https://www.pvsyst.com/help-pvsyst7/thermal_loss.htm).

kandersolar:
Closed by #2415

PR #2415, opened 2025-03-21
Enhanced the Software with Semi-Integration

Muhammad-Rebaal:
Hi @cwhanse !

Hope you are doing well !

I've open a PR for the enhancement asked in the issue [#2330](https://github.com/pvlib/pvlib-python/issues/2330) for semi_integration model for roof top mounted systems and also write the test for it . So, Could you please take a look at it & let me if anything needs to change ?

Thank You !

### DIFF
--- a/tests/test_semi_integration.py
+++ b/tests/test_semi_integration.py
@@ -1,0 +1,27 @@
+import numpy as np
+from pvlib import temperature
+from pvlib import pvsystem
+
+# Get the pvsyst_cell function directly
+from pvlib.temperature import pvsyst_cell
+
+
+def test_pvsyst_cell_semi_integrated():
+    """Test semi-integrated PVsyst cell temperature model."""
+    poa = 1000  # W/m^2
+    temp_air = 25  # °C
+    wind_speed = 1  # m/s
+    
+    # Test with default parameters for semi_integrated
+    result = pvsyst_cell(poa, temp_air, wind_speed, u_c=20.0, u_v=0.0)
+    expected = temp_air + ((poa * 0.9 * (1 - 0.1)) / (20.0 + 0.0 * wind_speed))
+    assert np.isclose(result, expected)
+    
+    # Test with temperature.pvsyst_cell wrapper
+    temp_params = {'u_c': 20.0, 'u_v': 0.0}
+    mount = pvsystem.FixedMount(surface_tilt=30, surface_azimuth=180,
+                                racking_model='semi_integrated')
+    array = pvsystem.Array(mount=mount, module_type='glass_glass',
+                          temperature_model_parameters=temp_params)
+    result = array.get_cell_temperature(poa, temp_air, wind_speed, 'pvsyst')
+    assert np.isclose(result, expected)


cwhanse:
Hi @Muhammad-Rebaal there's a lot that needs to be done differently here.

1. The parameters for the semi-integrated case need to be added to the dict [here](https://github.com/pvlib/pvlib-python/blob/0deb93f86d4c1c8d04b35edde42f9eb33ec3deae/pvlib/temperature.py#L16)
2. The docstring of `temperature.pvsyst_cell` needs to have the semi-integrated case added.
3. The docstring and methods of pvsystem/PVSystem need to have semi-integrated added as an option for `racking_model`. You can locate where by searching for one of the existing parameter names for the PVsyst temperature model, e.g. `freestanding`
4. The unit test should only test that the correct parameters are inferred when `racking_model` is semi-integrated. That test goes [here](https://github.com/pvlib/pvlib-python/blob/0deb93f86d4c1c8d04b35edde42f9eb33ec3deae/tests/test_pvsystem.py#L683). The correct calculation by the `pvsyst_cell` function is already tested.

Muhammad-Rebaal:
> Hi @Muhammad-Rebaal there's a lot that needs to be done differently here.
> 
> 1. The parameters for the semi-integrated case need to be added to the dict [here](https://github.com/pvlib/pvlib-python/blob/0deb93f86d4c1c8d04b35edde42f9eb33ec3deae/pvlib/temperature.py#L16)
> 2. The docstring of `temperature.pvsyst_cell` needs to have the semi-integrated case added.
> 3. The docstring and methods of pvsystem/PVSystem need to have semi-integrated added as an option for `racking_model`. You can locate where by searching for one of the existing parameter names for the PVsyst temperature model, e.g. `freestanding`
> 4. The unit test should only test that the correct parameters are inferred when `racking_model` is semi-integrated. That test goes [here](https://github.com/pvlib/pvlib-python/blob/0deb93f86d4c1c8d04b35edde42f9eb33ec3deae/tests/test_pvsystem.py#L683). The correct calculation by the `pvsyst_cell` function is already tested.

Oh my bad ! Actually, I changed those stuff but it wasn't pushed by me 😬

pvlib/temperature.py
```diff
@@ -393,19 +383,21 @@ def pvsyst_cell(poa_global, temp_air, wind_speed=1.0, u_c=29.0, u_v=0.0,
     air temperature :math:`T_{a}` (C) and wind speed :math:`WS` (m/s). Model
     output is cell temperature :math:`T_{C}`. Model parameters depend both on
     the module construction and its mounting. Parameters are provided in
-    [1]_ for open (freestanding) and close (insulated) mounting configurations,
-    , and are coded for convenience in
+    [1]_ for open (freestanding), close (insulated), and intermediate
+    (semi_integrate) mounting configurations, and are coded for convenience in
```
cwhanse:
```suggestion
    (semi_integrated) mounting configurations, and are coded for convenience in
```

cwhanse (approved):

pvlib/temperature.py
kandersolar: This typo is already fixed in the latest commits

Muhammad-Rebaal:
> Thanks @Muhammad-Rebaal!

My pleasure 🙌🏻

PR #2416, opened 2025-03-21
Remove pvgis_tmy outputformat='basic'

AdamRJensen:
 - [x] Tests added

PVGIS offers multiple output formats. For hourly data the file formats ['csv', 'json', 'basic'] are available, whereas for TMY data file format of ['csv', 'json', 'epw', 'basic'] are available.

I suggest that we drop support for the ``'basic'`` output format as there is no point to it; the file type is the same as the csv option but without any header or metadata. For this reason, the file option is also not available on the pvgis web interface and is already not supported by pvlib's pvgis_hourly functions.

The motivation is to reduce the maintenance burden, as dropping support for this file format reduces the code complexity significantly without any drawbacks for users. I'm currently planning more changes to the pvgis functions, but would like this change to be merged first. The planned changes mostly involve changing the outputs to be (data,meta) to be consistent with the rest of the iotools.

### DIFF
--- a/pvlib/iotools/pvgis.py
+++ b/pvlib/iotools/pvgis.py
@@ -429,7 +429,7 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
     longitude : float
         Longitude in degrees east
     outputformat : str, default 'json'
-        Must be in ``['csv', 'basic', 'epw', 'json']``. See PVGIS TMY tool
+        Must be in ``['csv', 'epw', 'json']``. See PVGIS TMY tool
         documentation [2]_ for more info.
     usehorizon : bool, default True
         include effects of horizon
@@ -461,11 +461,11 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
     data : pandas.DataFrame
         the weather data
     months_selected : list
-        TMY year for each month, ``None`` for basic and EPW
+        TMY year for each month, ``None`` for EPW
     inputs : dict
-        the inputs, ``None`` for basic and EPW
+        the inputs, ``None`` for EPW
     metadata : list or dict
-        file metadata, ``None`` for basic
+        file metadata
 
     Raises
     ------
@@ -516,9 +516,6 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
     elif outputformat == 'csv':
         with io.BytesIO(res.content) as src:
             data, months_selected, inputs, meta = _parse_pvgis_tmy_csv(src)
-    elif outputformat == 'basic':
-        with io.BytesIO(res.content) as src:
-            data, months_selected, inputs, meta = _parse_pvgis_tmy_basic(src)
     elif outputformat == 'epw':
         with io.StringIO(res.content.decode('utf-8')) as src:
             data, meta = parse_epw(src)
@@ -590,13 +587,6 @@ def _parse_pvgis_tmy_csv(src):
     return data, months_selected, inputs, meta
 
 
-def _parse_pvgis_tmy_basic(src):
-    data = pd.read_csv(src)
-    data.index = pd.to_datetime(
-        data['time(UTC)'], format='%Y%m%d:%H%M', utc=True)
-    data = data.drop('time(UTC)', axis=1)
-    return data, None, None, None
-
 
 def read_pvgis_tmy(filename, pvgis_format=None, map_variables=True):
     """
@@ -610,11 +600,9 @@ def read_pvgis_tmy(filename, pvgis_format=None, map_variables=True):
         Format of PVGIS file or buffer. Equivalent to the ``outputformat``
         parameter in the PVGIS TMY API. If ``filename`` is a file and
         ``pvgis_format`` is not specified then the file extension will be used
-        to determine the PVGIS format to parse. For PVGIS files from the API
-        with ``outputformat='basic'``, please set ``pvgis_format`` to
-        ``'basic'``.
+        to determine the PVGIS format to parse.
         If ``filename`` is a buffer, then ``pvgis_format`` is required and must
-        be in ``['csv', 'epw', 'json', 'basic']``.
+        be in ``['csv', 'epw', 'json']``.
     map_variables: bool, default True
         When true, renames columns of the Dataframe to pvlib variable names
         where applicable. See variable :const:`VARIABLE_MAP`.
@@ -625,18 +613,18 @@ def read_pvgis_tmy(filename, pvgis_format=None, map_variables=True):
     data : pandas.DataFrame
         the weather data
     months_selected : list
-        TMY year for each month, ``None`` for basic and EPW
+        TMY year for each month, ``None`` for EPW
     inputs : dict
-        the inputs, ``None`` for basic and EPW
+        the inputs, ``None`` for EPW
     metadata : list or dict
-        file metadata, ``None`` for basic
+        file metadata
 
     Raises
     ------
     ValueError
         if ``pvgis_format`` is not specified and the file extension is neither
         ``.csv``, ``.json``, nor ``.epw``, or if ``pvgis_format`` is provided
-        as input but isn't in ``['csv', 'epw', 'json', 'basic']``
+        as input but isn't in ``['csv', 'epw', 'json']``
     TypeError
         if ``pvgis_format`` is not specified and ``filename`` is a buffer
 
@@ -652,8 +640,7 @@ def read_pvgis_tmy(filename, pvgis_format=None, map_variables=True):
         outputformat = Path(filename).suffix[1:].lower()
     else:
         outputformat = pvgis_format
-    # parse the pvgis file based on the output format, either 'epw', 'json',
-    # 'csv', or 'basic'
+    # parse the pvgis file based on the output format, either 'epw', 'json', or 'csv'
 
     # EPW: use the EPW parser from the pvlib.iotools epw.py module
     if outputformat == 'epw':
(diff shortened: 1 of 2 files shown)


pvlib/iotools/pvgis.py
```diff
@@ -677,24 +665,20 @@ def read_pvgis_tmy(filename, pvgis_format=None, map_variables=True):
...
+                    _parse_pvgis_tmy_csv(fbuf)
 
     else:
-        # raise exception if pvgis format isn't in ['csv','basic','epw','json']
+        # raise exception if pvgis format isn't in ['csv','epw','json']
         err_msg = (
-            "pvgis format '{:s}' was unknown, must be either 'epw', 'json', "
-            "'csv', or 'basic'").format(outputformat)
+            "pvgis format '{:s}' was unknown, must be either 'json', 'csv',"
+            "or 'epw'").format(outputformat)
```
kandersolar, line 681:
```suggestion
            " or 'epw'").format(outputformat)
```

tests/iotools/test_pvgis.py
```diff
@@ -630,23 +622,6 @@ def test_read_pvgis_tmy_csv(expected, month_year_expected, inputs_expected,
                                meta_expected, csv_meta, pvgis_data)
 
 
-def test_read_pvgis_tmy_basic(expected, meta_expected):
-    fn = TESTS_DATA_DIR / 'tmy_45.000_8.000_2005_2023.txt'
```
kandersolar, line 634: Can we ditch this `tmy_45.000_8.000_2005_2023.txt` file now?

pvlib/iotools/pvgis.py
```diff
@@ -516,17 +516,16 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
     elif outputformat == 'csv':
         with io.BytesIO(res.content) as src:
             data, months_selected, inputs, meta = _parse_pvgis_tmy_csv(src)
-    elif outputformat == 'basic':
-        with io.BytesIO(res.content) as src:
-            data, months_selected, inputs, meta = _parse_pvgis_tmy_basic(src)
```
kandersolar, line 521: What do you think about raising a nicer error message here?  For example `outputformat='basic' is no longer supported, please use outputformat='csv' instead`.

kandersolar (commented):
I think there is no technical barrier to deprecating `outputformat='basic'` for a while before removing it altogether.  However, I bet `basic` is hardly used, so for the sake of expediency and not taxing @AdamRJensen's patience, I am okay with no deprecation period here, although I grumble nonetheless.

I did not do a thorough review, but these changes seem okay to me (assuming nobody objects to removing `basic` in the first place).  A couple small notes below.

tests/iotools/test_pvgis.py
AdamRJensen: Yes, good catch!

pvlib/iotools/pvgis.py
AdamRJensen: Yes, it's been implemented.

AdamRJensen:
> I think there is no technical barrier to deprecating `outputformat='basic'` for a while before removing it altogether. However, I bet `basic` is hardly used, so for the sake of expediency and not taxing @AdamRJensen's patience, I am okay with no deprecation period here, although I grumble nonetheless.

I am planning other breaking changes (consistent output of (data,meta)) which cannot have a deprecation period. So since these functions will be affected by a breaking change in the next release, it's better to include this breaking change too; which would minimize the number of times users would potentially need to update their code.

@mikofski It would be great if you review this since you're the original author!

AdamRJensen:
It would be nice to get this merged soon, so I can move on to the next changes of the PVGIS functions (standardizing the output of ``data,meta``

@mikofski it would be great if you could take a brief look!

pvlib/iotools/pvgis.py
```diff
@@ -373,7 +373,7 @@ def read_pvgis_hourly(filename, pvgis_format=None, map_variables=True):
         return _parse_pvgis_hourly_json(src, map_variables=map_variables)
 
     # CSV: use _parse_pvgis_hourly_csv()
-    if outputformat == 'csv':
+    elif outputformat == 'csv':
```
mikofski, line 376: unnecessary, this code can't be reached if `outputformat` is `'json'`

pvlib/iotools/pvgis.py
```diff
@@ -383,11 +383,17 @@ def read_pvgis_hourly(filename, pvgis_format=None, map_variables=True):
                     fbuf, map_variables=map_variables)
         return pvgis_data
 
-    # raise exception if pvgis format isn't in ['csv', 'json']
-    err_msg = (
-        "pvgis format '{:s}' was unknown, must be either 'json' or 'csv'")\
-        .format(outputformat)
-    raise ValueError(err_msg)
+    elif outputformat == 'basic':
```
mikofski, line 386: I think this condition and these changes should be moved to `get_pvgis_tmy()` instead of here in `read_pvgis_hourly()` which does not accept `basic` as input to the `pvgis_format` argument

pvlib/iotools/pvgis.py
mikofski: I believe it was implemented it in `read_pvgis_hourly()` instead of here in `get_pvgis_tmy()`

pvlib/iotools/pvgis.py
```diff
@@ -516,17 +522,16 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
...
         with io.BytesIO(res.content) as src:
             data, months_selected, inputs, meta = _parse_pvgis_tmy_csv(src)
-    elif outputformat == 'basic':
-        with io.BytesIO(res.content) as src:
-            data, months_selected, inputs, meta = _parse_pvgis_tmy_basic(src)
     elif outputformat == 'epw':
         with io.StringIO(res.content.decode('utf-8')) as src:
             data, meta = parse_epw(src)
             months_selected, inputs = None, None
+    # raise exception if pvgis format isn't in ['csv', 'json', 'epw']
```
mikofski, line 529: unnecessary, this line will never be reached because pvgis will return 400 BAD REQUEST if the `outputformat` isn't one of the options specified by the API

pvlib/iotools/pvgis.py
```diff
@@ -516,17 +522,16 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
...
     else:
-        # this line is never reached because if outputformat is not valid then
-        # the response is HTTP/1.1 400 BAD REQUEST which is handled earlier
-        pass
+        err_msg = (
+            "pvgis format '{:s}' was unknown, must be either 'json', 'csv', or"
+            " 'epw'.").format(outputformat)
+        raise ValueError(err_msg)
```
mikofski, line 534: move the condition for `'basic'` and changes from `read_pvgis_hourly()` to here:
```
elif outputformat == 'basic':
    err_msg = ("outputformat='basic' is no longer supported, please use ",
                      "outputformat='csv' instead.")
    raise ValueError(err_msg)

File: pvlib/iotools/pvgis.py [hunk 8]
```diff
@@ -383,11 +383,17 @@ def read_pvgis_hourly(filename, pvgis_format=None, map_variables=True):
...
-        "pvgis format '{:s}' was unknown, must be either 'json' or 'csv'")\
-        .format(outputformat)
-    raise ValueError(err_msg)
+    elif outputformat == 'basic':
+        err_msg = "outputformat='basic' is no longer supported, please use "\
+            "outputformat='csv' instead."
+        raise ValueError(err_msg)
+
+    else:
+        # raise exception if pvgis format isn't in ['csv', 'json']
```
mikofski, line 392: move this up to the top where `outputformat` is defined to raise an exception asap.

tests/iotools/test_pvgis.py
```diff
@@ -630,23 +622,6 @@ def test_read_pvgis_tmy_csv(expected, month_year_expected, inputs_expected,
...
-    # XXX: can't infer outputformat from file extensions for basic
-    with pytest.raises(ValueError, match="pvgis format 'txt' was unknown"):
```
mikofski, line 636: there are not tests now for raising an exception if either the file extension or `pvgis_format` was not `'json'` or `'csv'`. Don't need the basic test file, can create a named temp file instead.

mikofski (changes requested):
I think the "nicer" error message @kandersolar requested for `get_pvgis_tmy()` was accidently added to `read_pvgis_hourly()` which never accepted `'basic'` as an input, but `get_pvgis_tmy()` doesn't raise this "nicer" message. This reveals that `read_pvgis_hourly()` didn't have a test condition (until now) to handle if `outputformat` was not either `'json'` or `'csv'`. Removal of the "basic" test data removed a test if the file extension was not either `'json'`, '`csv'`, or `'epw'`

pvlib/iotools/pvgis.py
AdamRJensen: Moving it would duplicate code as far as I see it, so I think it's better to leave it where it is.

pvlib/iotools/pvgis.py
AdamRJensen: Changing it to ``elif`` just makes it read a bit nicer, but doesn't change anything functionally.

pvlib/iotools/pvgis.py
AdamRJensen: Good catch!

tests/iotools/test_pvgis.py
AdamRJensen: There is now a dedicated test that checks if an error is raised when a filename with a non-recognized file format is passed.

pvlib/iotools/pvgis.py
```diff
@@ -533,17 +534,15 @@ def get_pvgis_tmy(latitude, longitude, outputformat='json', usehorizon=True,
...
-            data, months_selected, inputs, meta = _parse_pvgis_tmy_basic(src)
     elif outputformat == 'epw':
         with io.StringIO(res.content.decode('utf-8')) as src:
             data, meta = parse_epw(src)
             months_selected, inputs = None, None
-    else:
-        # this line is never reached because if outputformat is not valid then
-        # the response is HTTP/1.1 400 BAD REQUEST which is handled earlier
-        pass
+    # raise exception if pvgis format isn't in ['csv', 'json', 'epw']
```
AdamRJensen, line 541:
```suggestion
```

PR #2413, opened 2025-03-19
Finalize v0.12.0

kandersolar:
- [x] Closes #2407
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing/index.html)
 - ~[ ] Tests added~
 - ~[ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.~
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - ~[ ] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.~
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

PR #2418, opened 2025-03-22
Add example whatsnew file

AdamRJensen:
Part of pvlib's [release procedures](https://github.com/pvlib/pvlib-python/wiki/Release-procedures) is to add a new empty Whatsnew file after the release of a new version.

I believe the current practice is to take the previous whatsnew file and delete the content. This is tedious. More of a problem is that not all releases include all types of contributions and thus the added whatsnew file is often missing sections. While, these sections typically get added, they tend to get added in a random order (maybe not a big deal).

In any case, to make the release procedure smoother I suggest having a Whatsnew template that we use as the bases for the new  Whatsnew file that is added after a release. The example Whatsnew file has all possible sections (at least the ones we've used historically). Please comment if you think more should be added, if some are obsolete, or if you think that they should be ordered differently.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.X.Y-example.rst
+++ b/docs/sphinx/source/whatsnew/v0.X.Y-example.rst
@@ -1,0 +1,45 @@
+.. _whatsnew_0XXYY:
+
+
+v0.X.Y (Month XX, 20YY)
+-----------------------
+
+Breaking Changes
+~~~~~~~~~~~~~~~~
+
+
+Deprecations
+~~~~~~~~~~~~
+
+
+Bug fixes
+~~~~~~~~~
+
+
+Enhancements
+~~~~~~~~~~~~
+
+
+Documentation
+~~~~~~~~~~~~~
+
+
+Testing
+~~~~~~~
+
+
+Benchmarking
+~~~~~~~~~~~~
+
+
+Requirements
+~~~~~~~~~~~~
+
+
+Maintenance
+~~~~~~~~~~~
+
+
+Contributors
+~~~~~~~~~~~~
+


docs/sphinx/source/whatsnew/v0.X.Y-example.rst
```diff
@@ -0,0 +1,45 @@
+.. _whatsnew_0XXYY:
```
echedey-ls, line 1:
```suggestion
.. _whatsnew_0_X_Y:
```
This is just personal taste, IK it's not the legacy form, but easier to read.

echedey-ls (approved):
Good catch, this is a nice improvement IMO.

I think we could merge Testing and Benchmarking into Maintenance, (maybe rename to Development or a broader concept), since they don't really impact users (benchmarking means changes to asv benchmarks, right? I assume it's not about speed improvements in algorithms, that should probably go into Enhancements).

And on the other hand, Requirements section is similar to Breaking Changes. For example, bumping python or numpy versions.

AdamRJensen:
> I think we could merge Testing and Benchmarking into Maintenance, (maybe rename to Development or a broader concept), since they don't really impact users (benchmarking means changes to asv benchmarks, right? I assume it's not about speed improvements in algorithms, that should probably go into Enhancements).

Benchmarking indeed does refer to ASV. I would be fine with merging benchmarking and testing and naming it "Development".

> And on the other hand, Requirements section is similar to Breaking Changes. For example, bumping python or numpy versions.
I think it's nice to keep a separate Requirements section as in my view a breaking change is our code that is changing, whereas changes in requirements are very different and most often not a big deal to most people.

Still open to other ideas!

docs/sphinx/source/whatsnew/v0.X.Y-example.rst
AdamRJensen: I also prefer underscores. But what is the purpose of this tag and where does it show up (it at all)?

echedey-ls: This is the hyperlink target, you would use it in rST like ```:ref:`whatsnew_0_X_Y` ```

cwhanse:
I disagree with lumping Testing and Benchmarking into Development.  Testing is a useful category of change on its own. Development is a broad term that I think of as adding content.

RDaxini:
Could we revisit [this idea](https://github.com/pvlib/pvlib-python/pull/2413#discussion_r2004305184)? Adding `.. module:: pvlib` or `currentmodule`.
Would be worth double checking how it works and testing it on an example. This PR might be the right place to do that.

AdamRJensen:
> Could we revisit [this idea](https://github.com/pvlib/pvlib-python/pull/2413#discussion_r2004305184)? Adding `.. module:: pvlib` or `currentmodule`. Would be worth double checking how it works and testing it on an example. This PR might be the right place to do that.

@RDaxini Could you add a link to where I can read more about what this would do?

echedey-ls:
@AdamRJensen @RDaxini this is the `currentmodule` reference: https://www.sphinx-doc.org/en/master/usage/domains/python.html#directive-py-currentmodule

In any case, I'm slightly opposed to using it, cause I find it clearer to state whether the references are about something inside pvlib or outside (remember there is intersphinx enabled for packages like numpy, scipy or pandas)

AdamRJensen:
GIven the lack of consensus then I suggest that we hold of on the changes proposed by @RDaxini and discuss that elsewhere.

@kandersolar would you give your thoughts on this PR given that you have been acting release manager for the past many releases?

kandersolar (approved):
Fine with me.  This reminds me that I think I neglected to add the next whatsnew after the last release.

A side note: I expected sphinx to emit a warning along the lines of "v0.X.Y-example.rst not included in any toctree", but didn't see one in the log.  I'd be interested to learn why not if anyone knows why...

echedey-ls:
> A side note: I expected sphinx to emit a warning along the lines of "v0.X.Y-example.rst not included in any toctree", but didn't see one in the log. I'd be interested to learn why not if anyone knows why...

I did wonder the same recently, but found the wise magic spell is written here:

https://github.com/pvlib/pvlib-python/blame/main/docs/sphinx/source/conf.py#L126

docs/sphinx/source/whatsnew/v0.X.Y-example.rst
echedey-ls: Can we revisit this? Even thou it changes the current naming scheme of past files, I don't see any downsides.

AdamRJensen: Certainly, I'm for it.

kandersolar:
Thanks @AdamRJensen!

PR #2436, opened 2025-04-11
[DOC] update scipy intersphinx url

kandersolar:
Trivial change to address the following docs build warning:

> intersphinx inventory has moved: https://docs.scipy.org/doc/scipy/reference/objects.inv -> https://docs.scipy.org/doc/scipy/objects.inv


echedey-ls (approved):

kandersolar:
Thanks @echedey-ls!

PR #2486, opened 2025-06-17
Add 2025 coefficients for Huld model

### LINKED ISSUE #2461: Consider updating `k` coefficients in `pvlib.pvarray.huld`
The EU JRC has published updated coefficients for what we call the `huld` model: https://doi.org/10.1002/pip.3926

We could update the coefficient value currently available in `pvlib.pvarray.huld`, or add these as new ones.

The paper also briefly describes the fitting procedure they used, although I'm not sure it's detailed enough to change the situation with #1979.

OmarBahamida:
Enhance Huld model with EU JRC coefficients

- Introduced a new function to infer updated coefficients for the Huld model based on EU JRC research.
- Added a parameter to the existing Huld function to optionally use these updated coefficients.
- Updated documentation to reflect the new functionality and included references to the EU JRC paper.
- Added tests to verify the implementation and ensure compatibility with existing functionality.

Closes #2461

### DIFF
--- a/pvlib/pvarray.py
+++ b/pvlib/pvarray.py
@@ -236,7 +236,7 @@ def _infer_k_huld(cell_type, pdc0, k_version):
     pdc0 : numeric
         Power of the modules at reference conditions [W]
     k_version : str
-        Either '2011' or '2025'.
+        Either 'pvgis5' or 'pvgis6'.
 
     Returns
     -------
@@ -263,7 +263,7 @@ def _infer_k_huld(cell_type, pdc0, k_version):
                                -0.000043, 0.0),
                        'cis': (-0.011001, -0.029734, -0.002887, 0.000217,
                                -0.000163, 0.0),
-                       'cdte': (-0.020644, -0.035316, -0.003406, 0.000073,
+                       'cdte': (-0.020644, -0.035136, -0.003406, 0.000073,
                                 -0.000141, 0.000002)}
     else:
         raise ValueError(f'Invalid k_version={k_version}: must be either '


cwhanse:
I'm -1 on the new kwarg `use_eu_jrc`. The way I read the 2025 reference, the intent is to replace the older set of coefficients, to reflect "modern" modules. If we need to retain the older coefficients for continuity, we can add a table to the docstring.

Comments for my fellow maintainers?

echedey-ls:
I'm between @cwhanse and @OmarBahamida ; I prefer a `version` keyword only parameter that can be either `2025` or `2011`, and defaults to the latest one. I think it's valuable to ease the use of older parameters, given their lifetime is long and post analysis of systems is great for diagnosis.

pvlib/pvarray.py
```diff
@@ -332,10 +373,15 @@ def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
            E. Dunlop. A power-rating model for crystalline silicon PV modules.
            Solar Energy Materials and Solar Cells 95, (2011), pp. 3359-3369.
            :doi:`10.1016/j.solmat.2011.07.026`.
+    .. [2] EU JRC paper, "Updated coefficients for the Huld model",
+           https://doi.org/10.1002/pip.3926
```
echedey-ls: Looks like you didn't change this to fit the IEEE citation format nor the `:doi:` directive (see the citation just above).

I recommend you to read https://pvlib-python.readthedocs.io/en/latest/contributing/style_guide.html

tests/test_pvarray.py
```diff
@@ -69,3 +69,23 @@ def test_huld():
     with pytest.raises(ValueError,
                        match='Either k or cell_type must be specified'):
         res = pvarray.huld(1000, 25, 100)
+
+
+def test_huld_eu_jrc():
```
echedey-ls: Can you check for some specific values, up to 4-5 decimals? It would be better if you calculate them outside of your Python implementation.

AdamRJensen:
> I'm between @cwhanse and @OmarBahamida ; I prefer a `version` keyword only parameter that can be either `2025` or `2011`, and defaults to the latest one. I think it's valuable to ease the use of older parameters, given their lifetime is long and post analysis of systems is great for diagnosis.

I fully support this implementation. It would be a shame to remove existing features. This way we can also have a deprecation cycle, i.e., a period where a warning is raised if the ``version`` parameter is not specified. Eventually, we can make the breaking change of switching the default version to the most recent version of the coefficients.

cwhanse:
Any set of coefficients can be provided directly to the existing function, so kwargs like `version` aren't strictly necessary.

But if we lean towards providing built-in parameters, as a convenience, or as a means to repeat previous analyses (where parameters were not explicit, which is a different problem), then I think the user should be required to affirmatively select them: `version=None` would have my vote.

adriesse:
I think I lean towards @AdamRJensen 's suggestion, which maintains the default behaviour for now.

cwhanse:
@OmarBahamida any plan to move this forward? If not, I will do so. Thanks for starting this change.

OmarBahamida:
Hi all,
I’ve just pushed changes to fix the Flake8 linter errors (trailing whitespace and long lines).
I also verified that all new code is fully covered by tests (100% coverage for pvlib/pvarray.py).
Sorry for my late response, I haven’t had enough time recently, but I appreciate your patience. 
Thank you for your feedback and support.

cwhanse:
@OmarBahamida we'd like to get this change into the next release. I took the liberty of restructuring the new kwarg and the test that looks up the coefficients.

OmarBahamida:
@cwhanse Thank you very much for your help and for restructuring the code and tests! I’m glad to hear this will be included in the next release. I appreciate your feedback and support throughout this process.

cwhanse:
@pvlib/pvlib-maintainer this is ready for reviews

pvlib/pvarray.py
```diff
@@ -225,24 +225,54 @@ def adr_wrapper(xdata, *params):
...
+    else:
+        raise ValueError(f'Invalid k_version={k_version}: must be either '
+                         '"2011" or "2025" as a string')
     k = tuple([x*pdc0 for x in huld_params[cell_type.lower()]])
     return k
 
 
-def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
+def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None,
+         k_version=None):
```
adriesse:
```suggestion
         k_version='2011'):
```
Why not make it explicit here?

pvlib/pvarray.py
```diff
@@ -328,14 +362,21 @@ def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
...
+           Photovoltaic Module Technologies,"
+           Progress in Photovoltaics: Research and Applications 33,
+           no. 8 (2025): 905–917, :doi:`10.1002/pip.3926`.
     """
     if k is None:
         if cell_type is not None:
-            k = _infer_k_huld(cell_type, pdc0)
+            if k_version is None:
+                k_version = '2011'
+            k = _infer_k_huld(cell_type, pdc0, k_version)
```
adriesse:
```suggestion
            k = _infer_k_huld(cell_type, pdc0, k_version)
```

adriesse (commented):
Looks good to me.

pvlib/pvarray.py
cwhanse: Yes, of course. String is immutable.

pvlib/pvarray.py
```diff
@@ -274,6 +304,10 @@ def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
     cell_type : str, optional
         If provided, must be one of ``'cSi'``, ``'CIS'``, or ``'CdTe'``.
         Used to look up default values for ``k`` if ``k`` is not specified.
+    k_version : str, optional
+        Either `'2011'` (default) or `'2025'`. Used to select default values
+        for ``k`` if ``k`` is not specified. If `'2011'`, values from [1]_
+        are used; if `'2025'` values are from [2]_.
```
echedey-ls:
```suggestion
    k_version : str, default ``"2011"``
        Either ``'2011'`` (default) or ``'2025'``. Used to select default values
        for ``k`` if ``k`` is not specified. If ``'2011'``, values from [1]_
        are used; if ``'2025'`` values are from [2]_.
```
Code formatting constants. They currently are in italics.
https://pvlib-python--2486.org.readthedocs.build/en/2486/reference/generated/pvlib.pvarray.huld.html#pvlib.pvarray.huld

echedey-ls (approved):
Absolute nice tests! LGTM :)

echedey-ls:
Well, I guess at some point pvlib will have to change the default constants year, but that can be a problem for the future pvlib.

pvlib/pvarray.py
```diff
@@ -274,6 +304,10 @@ def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
     cell_type : str, optional
         If provided, must be one of ``'cSi'``, ``'CIS'``, or ``'CdTe'``.
         Used to look up default values for ``k`` if ``k`` is not specified.
+    k_version : str, default ``"2011"``
+        Either ``'2011'`` (default) or ``'2025'``. Used to select default values
```
cwhanse:
```suggestion
        Either ``'2011'`` (default) or ``'2025'``. Selects the default value
```

pvlib/pvarray.py
```diff
@@ -274,6 +304,10 @@ def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
     cell_type : str, optional
         If provided, must be one of ``'cSi'``, ``'CIS'``, or ``'CdTe'``.
         Used to look up default values for ``k`` if ``k`` is not specified.
+    k_version : str, default ``"2011"``
+        Either ``'2011'`` (default) or ``'2025'``. Selects the default value
```
cwhanse:
```suggestion
        Either ``'2011'`` (default) or ``'2025'``. Selects the values
```

pvlib/pvarray.py
```diff
@@ -225,24 +225,54 @@ def adr_wrapper(xdata, *params):
...
+                       'cdte': (-0.046689, -0.072844, -0.002262, 0.000276,
+                                0.000159, -0.000006)}
+    elif k_version == '2025':
+        # Updated coefficients from EU JRC paper
+        huld_params = {'csi': (-0.017162, -0.040289, -0.004681, 0.000148,
+                               0.000169, 0.000005),
+                       'cis': (-0.005521, -0.038576, -0.003711, -0.000901,
+                               -0.001251, 0.000001),
+                       'cdte': (-0.046477, -0.072509, -0.002252, 0.000275,
+                                0.000158, -0.000006)}
```
kandersolar: I can't find where the numbers for 2025 came from.  What am I missing?  I expected them to come from Table 3, but those are completely different:

image

kandersolar: After looking into it a bit more, I remain confused.  Some of the "2025" numbers in this PR seem to be present in this 2010 paper: https://doi.org/10.1016/j.solener.2009.12.002

cwhanse: Well...I simply copied from the submitted code and didn't check the "updated" values against the 2025 paper. I glanced at the first few of the 2010 values and thought, that's good. Note to self for next time.

Now that you point to it: the 2010 numbers aren't in the 2010 paper, they are taken from the PVGIS online documentation.

pvlib/pvarray.py
```diff
@@ -225,24 +225,54 @@ def adr_wrapper(xdata, *params):
...
     # equation that has factored Pdc0 out of the polynomial:
     #  P = G/1000 * Pdc0 * (1 + k1 log(Geff) + ...) so these parameters are
     # multiplied by pdc0
-    huld_params = {'csi': (-0.017237, -0.040465, -0.004702, 0.000149,
-                           0.000170, 0.000005),
-                   'cis': (-0.005554, -0.038724, -0.003723, -0.000905,
-                           -0.001256, 0.000001),
-                   'cdte': (-0.046689, -0.072844, -0.002262, 0.000276,
-                            0.000159, -0.000006)}
+    if k_version == '2011':
```
kandersolar: I don't think `'2011'` is a good choice of identifier for the older coefficients.  To me it suggests the 2011 paper is the source of those coefficients, which isn't the case.  

I think something like `PVGIS5.3` (the current version) and `PVGIS{something}` could be a good option, but I don't know what version will have the new coefficients.  I've emailed Nikos Alexandris at JRC asking about this.

kandersolar: Nikos responded that PVGIS 6 will use the new coefficients by default.  Maybe `PVGIS5` and `PVGIS6` are good identifiers, then.

pvlib/pvarray.py
```diff
@@ -274,6 +305,11 @@ def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None):
     cell_type : str, optional
         If provided, must be one of ``'cSi'``, ``'CIS'``, or ``'CdTe'``.
         Used to look up default values for ``k`` if ``k`` is not specified.
+    k_version : str, optional
+        Either ``'pvgis5'`` (default) or ``'pvgis6'``. Selects values
+        for ``k`` if ``k`` is not specified. If ``'pvgis5'``, values are
+        from PVGIS documentation and are labeled in [2]_ as "current".
+        If ``'pvgis6'`` values are from [2]_ labeled as "updated".
```
adriesse, line 312:
```suggestion
        Selects a set of values for ``k`` if ``k`` is not specified: either ``'pvgis5'`` (default)
        or ``'pvgis6'``.  For ``'pvgis5'``, ``k`` values are from PVGIS documentation, 
        which are also identified as "current" in [2]_ . For ``'pvgis6'``, values are 
        from [2]_ where they are labeled as "updated".
```

adriesse (approved):

pvlib/pvarray.py
cwhanse: I'll not take this suggestion as the phrasing implies that a value of `'pvgis5'` is assigned to `k` rather than being a value of `k_version`.

pvlib/pvarray.py
```diff
@@ -225,24 +225,55 @@ def adr_wrapper(xdata, *params):
...
+                               -0.001256, 0.000001),
+                       'cdte': (-0.046689, -0.072844, -0.002262, 0.000276,
+                                0.000159, -0.000006)}
+    elif k_version.lower() == 'pvgis6':
+        # Coefficients from EU JRC paper
+        huld_params = {'csi': (-0.0067560, -0.016444, -0.003015, -0.000045,
+                               -0.000043, 0.0),
+                       'cis': (-0.011001, -0.029734, -0.002887, 0.000217,
+                               -0.000163, 0.0),
+                       'cdte': (-0.020644, -0.035316, -0.003406, 0.000073,
```
kandersolar:
```suggestion
                       'cdte': (-0.020644, -0.035136, -0.003406, 0.000073,
```

pvlib/pvarray.py
```diff
@@ -225,24 +225,55 @@ def adr_wrapper(xdata, *params):
...
+    Get the EU JRC updated coefficients for the Huld model.
+
+    Parameters
+    ----------
+    cell_type : str
+        Must be one of 'csi', 'cis', or 'cdte'
+    pdc0 : numeric
+        Power of the modules at reference conditions [W]
+    k_version : str
+        Either '2011' or '2025'.
```
kandersolar:
```suggestion
        Either 'pvgis5' or 'pvgis6'.
```

kandersolar:
Thanks @OmarBahamida and all other contributors here!

PR #2492, opened 2025-07-01
Fix typos in ModelChain error messages

RDaxini:
- [X] Closes #2490
 - [X] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing/index.html)
 - [ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [X] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [X] Pull request is nearly complete and ready for detailed review.
 - [X] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

PR #2495, opened 2025-07-02
Deprecate `modelchain.get_orientation`

RDaxini:
- [X] Related discussion #2492
 - [X] Tests added
 - [X] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

~See discussion in related #2492~

~Original test `test_bad_get_orientation():` only checked whether a value error was raised, the new test (with what I think is a clearer name) checks not only whether a value error is raised but also whether the error string matches~

### DIFF
--- a/docs/sphinx/source/reference/modelchain.rst
+++ b/docs/sphinx/source/reference/modelchain.rst
@@ -115,12 +115,3 @@ on the information in the associated :py:class:`~pvsystem.PVSystem` object.
    modelchain.ModelChain.infer_temperature_model
    modelchain.ModelChain.infer_losses_model
 
-Functions
----------
-
-Functions for power modeling.
-
-.. autosummary::
-   :toctree: generated/
-
-   modelchain.get_orientation
--- a/docs/sphinx/source/reference/modelchain.rst
+++ b/docs/sphinx/source/reference/modelchain.rst
@@ -115,3 +115,12 @@ on the information in the associated :py:class:`~pvsystem.PVSystem` object.
    modelchain.ModelChain.infer_temperature_model
    modelchain.ModelChain.infer_losses_model
 
+Functions
+---------
+
+Functions for power modeling.
+
+.. autosummary::
+   :toctree: generated/
+
+   modelchain.get_orientation
--- a/docs/sphinx/source/whatsnew/v0.13.1.rst
+++ b/docs/sphinx/source/whatsnew/v0.13.1.rst
@@ -10,6 +10,8 @@ Breaking Changes
 
 Deprecations
 ~~~~~~~~~~~~
+* Deprecate :py:func:`~pvlib.modelchain.get_orientation`. Removal scheduled for
+  0.14.0. (:pull:`2691`)
 
 
 Bug fixes
@@ -26,9 +28,7 @@ Documentation
 
 Testing
 ~~~~~~~
-* Update test for :py:func:`~pvlib.modelchain.get_orientation` to check
-  whether the correct error message is raised when an invalid strategy is
-  provided. (:issue:`2492`, :pull:`2691`)
+
 
 Benchmarking
 ~~~~~~~~~~~~
@@ -46,3 +46,4 @@ Maintenance
 Contributors
 ~~~~~~~~~~~~
 * Elijah Passmore (:ghuser:`eljpsm`)
+* Rajiv Daxini (:ghuser:`RDaxini`)
--- a/docs/sphinx/source/whatsnew/v0.13.1.rst
+++ b/docs/sphinx/source/whatsnew/v0.13.1.rst
@@ -11,7 +11,7 @@ Breaking Changes
 Deprecations
 ~~~~~~~~~~~~
 * Deprecate :py:func:`~pvlib.modelchain.get_orientation`. Removal scheduled for
-  0.14.0. (:pull:`2691`)
+  ``v0.14.0``. (:pull:`2691`)
 
 
 Bug fixes
--- a/docs/sphinx/source/whatsnew/v0.13.1.rst
+++ b/docs/sphinx/source/whatsnew/v0.13.1.rst
@@ -26,7 +26,9 @@ Documentation
 
 Testing
 ~~~~~~~
-
+* Update test for :py:func:`~pvlib.modelchain.get_orientation` to check
+  whether the correct error message is raised when an invalid strategy is
+  provided. (:issue:`2492`, :pull:`2691`)
 
 Benchmarking
 ~~~~~~~~~~~~
--- a/docs/sphinx/source/whatsnew/v0.13.1.rst
+++ b/docs/sphinx/source/whatsnew/v0.13.1.rst
@@ -10,8 +10,7 @@ Breaking Changes
 
 Deprecations
 ~~~~~~~~~~~~
-* Deprecate :py:func:`~pvlib.modelchain.get_orientation`. Removal scheduled for
-  ``v0.14.0``. (:pull:`2691`)
+* Deprecate :py:func:`~pvlib.modelchain.get_orientation`. (:pull:`2691`)
 * Rename parameter name ``aparent_azimuth`` to ``solar_azimuth`` in :py:func:`~pvlib.tracking.singleaxis`.
   (:issue:`2479`, :pull:`2480`)
 
--- a/docs/sphinx/source/whatsnew/v0.13.1.rst
+++ b/docs/sphinx/source/whatsnew/v0.13.1.rst
@@ -10,7 +10,7 @@ Breaking Changes
 
 Deprecations
 ~~~~~~~~~~~~
-* Deprecate :py:func:`~pvlib.modelchain.get_orientation`. (:pull:`2691`)
+* Deprecate :py:func:`~pvlib.modelchain.get_orientation`. (:pull:`2495`)
 * Rename parameter name ``aparent_azimuth`` to ``solar_azimuth`` in :py:func:`~pvlib.tracking.singleaxis`.
   (:issue:`2479`, :pull:`2480`)
 
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -62,8 +62,8 @@ SAPM_CONFIG = dict(
 
 
 @deprecated(
-    since="0.13",
-    removal="0.14",
+    since="0.13.1",
+    removal="the next version",
     name="pvlib.modelchain.get_orientation",
     alternative=None,
     addendum=None,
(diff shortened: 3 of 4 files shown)


RDaxini:
I could also add tests to check whether the correct values are returned according to the strategy provided, is this necessary?
For example:
```
def test_get_orientation_flat():
    surface_tilt, surface_azimuth = modelchain.get_orientation('flat')
    assert surface_tilt == 0
    assert surface_azimuth == 180
```

cwhanse:
> tests to check whether the correct values are returned

Optional IMO. I'm actually wondering why modelchain.py has this function. It was added in 2016 in the initial modelchain build and hasn't been used since in the `ModelChain` methods. There's a `PVSystem.get_orientation` method but that looks like just casting to a dict.

RDaxini:
@cwhanse fair observation, are you proposing we deprecate and remove the function entirely?

wholmgren:
Yes get rid of it!

docs/sphinx/source/reference/modelchain.rst
```diff
@@ -115,12 +115,3 @@ on the information in the associated :py:class:`~pvsystem.PVSystem` object.
...
 
-Functions
----------
-
-Functions for power modeling.
-
-.. autosummary::
-   :toctree: generated/
-
-   modelchain.get_orientation
```
echedey-ls: Strictly following deprecations as a concept, this shouldn't be removed till the removal - a just in case clarification. Anyway, we all are sure this function hasn't found it's way into many codebases so I agree with removing it here.

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -10,6 +10,8 @@ Breaking Changes
 
 Deprecations
 ~~~~~~~~~~~~
+* Deprecate :py:func:`~pvlib.modelchain.get_orientation`. Removal scheduled for
+  0.14.0. (:pull:`2691`)
```
echedey-ls:
```suggestion
  ``v0.14.0``. (:pull:`2691`)
```

echedey-ls (approved):
LGTM :)
Just a comment and a formatting preference down below.

docs/sphinx/source/reference/modelchain.rst
RDaxini: Thanks for explaining that @echedey-ls

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -10,6 +10,8 @@ Breaking Changes
 
 Deprecations
 ~~~~~~~~~~~~
+* Deprecate :py:func:`~pvlib.modelchain.get_orientation`. Removal scheduled for
+  ``v0.14.0``. (:pull:`2691`)
```
kandersolar:
```suggestion
* Deprecate :py:func:`~pvlib.modelchain.get_orientation`.  (:pull:`2691`)
```
In case v0.14.0 turns out to be in the next few releases, I think it is better to not mention a version number.

However, I notice that generally I seem to be the only one advocating for not mentioning specific versions regarding deprecation removals.  Maybe it's time to discuss and adopt an actual policy here?  In the meantime, feel free to overrule me!

pvlib/modelchain.py
```diff
@@ -59,6 +61,13 @@
 )
 
 
+@deprecated(
+    since="0.13",
+    removal="0.14",
```
kandersolar:
```suggestion
```

pvlib/modelchain.py
```diff
@@ -59,6 +61,13 @@
 )
 
 
+@deprecated(
+    since="0.13",
```
kandersolar:
```suggestion
    since="0.13.1",
```

tests/test_modelchain.py
```diff
@@ -1785,11 +1785,6 @@ def test_invalid_models(model, sapm_dc_snl_ac_system, location):
         ModelChain(sapm_dc_snl_ac_system, location, **kwargs)
 
 
-def test_bad_get_orientation():
-    with pytest.raises(ValueError):
-        modelchain.get_orientation('bad value')
-
-
```
kandersolar: To add to @echedey-ls's point above, the usual practice would be to wait to remove the tests until we remove the functionality.   For the deprecation period, we would usually retain the tests, just slightly modified so that the function calls are inside `with pytest.warns(pvlibDeprecationWarning, match="..."):` blocks to (1) ensure the deprecation warning is working as expected and (2) prevent that warning from escaping into the test suite's logging output.

I suggest just for the sake of best practice we should do that here too.

docs/sphinx/source/whatsnew/v0.13.1.rst
RDaxini: > In case v0.14.0 turns out to be in the next few releases, I think it is better to not mention a version number.

Hmm that's a fair point, I confess I had not thought much about that. I think I am on board with your suggestion. If no-one objects in the next day or two, I'll commit your suggestion.

> Maybe it's time to discuss and adopt an actual policy here?

+1

echedey-ls: I like setting a version more for us, so it does not get forgotten. I understand Kevin's point is to make it more agile.
So I propose that instead of an specific semver, we set it to `removal="next version"`. That's agile, and concrete for users and us.

```python
>>> import pvlib.modelchain as mc
>>> mc.get_orientation('flat')
<stdin>:1: pvlibDeprecationWarning: The pvlib.modelchain.get_orientation function was deprecated in pvlib 0.13 and will be removed in next version.
(0, 180)
>>>
```

And let's set the patch number too, so it's deprecated in 0.13.x and next version clearly is 0.13.(x+1).

RDaxini:
@echedey-ls @kandersolar thanks for the feedback about the deprecation scheduling. Please review, let me know if the current version is in line with what you guys have in mind.
I agree with keeping something there for maintainers not to lose track of the issue, and retaining some flexibility in the removal. @echedey-ls let's update the deprecations tracker issue with this PR too.

docs/sphinx/source/whatsnew/v0.13.1.rst
kandersolar: I think many people (me included) would interpret "next version" to include patches, so the next version after 0.13.1 would be 0.13.2 (or 0.14.0, if there is no 0.13.2).  So I don't think "next version" solves the problem in my comment above :(

> I like setting a version more for us, so it does not get forgotten.

I agree, but I think the deprecation tracker is the best tool for that :)

RDaxini: I had the same thought so I followed up with an off-github conversation with Echedey to clarify, I thought I might be the only one to have misunderstood 😅 I think the idea was to say *something* so there is a message, but without committing to something specific? @echedey-ls may be able to clarify further.

> I agree, but I think the deprecation tracker is the best tool for that :)

Makes sense, so is the suggestion here just to stick with no removal statement (whatever the default is) and then make sure we all stay on top of the deprecation tracker? (Thanks for creating that @echedey-ls!) --what is the consensus?

kandersolar: > stick with no removal statement (whatever the default is) and then make sure we all stay on top of the deprecation tracker

This has my vote!

FYI the default is "soon":

https://github.com/pvlib/pvlib-python/blob/908d3da1f898c8caf50954a356ed4310e4a57062/pvlib/_deprecation.py#L141-L142

echedey-ls: > I think many people (me included) would interpret "next version" to include patches, so the next version after 0.13.1 would be 0.13.2 (or 0.14.0, if there is no 0.13.2). So I don't think "next version" solves the problem in my comment above :(

I think that does address what I interpret the problem is, that is not declaring we are going to be agile in the removal that will happen in whatever next version is. You are right that for ensuring removals get done we have the deprecation tracker, my bad for a shitty argument.

My reasons to still support "next" do not make for a strong opinion:
1. "soon" is not precise in the same way "next" is. And precedences of "soon" removals meant 2+ minor versions. I think we should strive for super clear declaration of intentions. And being so does not look harmful in this situation.
2. removals in patch versions are against pvlib-flavoured semver which usually happens at minor versions. I would just heavily differentiate this edge-deprecation and removal from any other, to not mislead (new) users with the usual procedures involved in pvlib. Not all users may be aware of the deprecations tracker.

I'm in favour of removing this one in a patch version, thou. Unless anybody else objects, please go ahead with the current choice @RDaxini , I don't want to block such a small change.

RDaxini: I've pushed @kandersolar's suggestion in the last commit. Please double check.
Sounds like this subject requires further (separate) discussion for future though.

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -48,4 +50,4 @@ Contributors
 ~~~~~~~~~~~~
 * Elijah Passmore (:ghuser:`eljpsm`)
 * Rajiv Daxini (:ghuser:`RDaxini`)
-* Omar Bahamida (:ghuser:`OmarBahamida`)
\ No newline at end of file
+* Omar Bahamida (:ghuser:`OmarBahamida`)
```
RDaxini: Not sure why this shows as me having changed something
Might have arisen somehow from when I resolved conflicts with the whatsnew file

kandersolar: Trailing newline, presumably.  Don't worry about it

kandersolar:
Sorry, one more thing @RDaxini: can you also catch the warning in the tests using `with pytest.warns(pvlibDeprecationWarning, match='...'):`?

See https://github.com/pvlib/pvlib-python/actions/runs/16630932041/job/47059848104?pr=2495#step:9:84

tests/test_modelchain.py
```diff
@@ -1786,8 +1788,9 @@ def test_invalid_models(model, sapm_dc_snl_ac_system, location):
 
 
 def test_bad_get_orientation():
-    with pytest.raises(ValueError):
-        modelchain.get_orientation('bad value')
+    with pytest.warns(pvlibDeprecationWarning, match='will be removed soon'):
+        with pytest.raises(ValueError):
+            modelchain.get_orientation('bad value')
```
RDaxini, line 1793: is including both in the same test okay? passed locally

echedey-ls: I think so. I don't know if there is a pytest method to assert raises and warns at the same time. I doubt there is.

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -10,6 +10,7 @@ Breaking Changes
 
 Deprecations
 ~~~~~~~~~~~~
+* Deprecate :py:func:`~pvlib.modelchain.get_orientation`. (:pull:`2691`)
```
kandersolar:
```suggestion
* Deprecate :py:func:`~pvlib.modelchain.get_orientation`. (:pull:`2495`)
```

Issue #2497, opened 2025-07-08
Use scipy's new `optimize.elementwise` functions


kandersolar:
scipy v1.15.0 (Jan 2025) offers two new element-wise optimization functions: [scipy.optimize.elementwise.find_root](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.elementwise.find_root.html) and [scipy.optimize.elementwise.find_minimum](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.elementwise.find_minimum.html).  (See previous brief discussion of this scipy functionality in https://github.com/pvlib/pvlib-python/issues/974).

These functions can likely provide significant speed-ups for single-diode equation calculations.  For example, rerunning the benchmark in https://github.com/pvlib/pvlib-python/issues/974#issuecomment-1719578738 using the new `find_root` function gives this timing comparison for `bishop88_i_from_v` ("chandrupatla" is the new scipy function):

<img width="481" height="279" alt="Image" src="https://github.com/user-attachments/assets/a30a35d9-81ba-4a59-9955-279986e666db" />

The `find_minimum` function could likely speed up the lambertw approach as well, although I have not benchmarked it against our `_golden_sect_DataFrame` function.

The question is: when can we start using these in pvlib?  The required scipy version is quite new, and not available for python 3.9.  Must we wait a couple years until our minimum scipy is v1.15?  Or should we add the functionality now, and just raise an error if the installed version of scipy is insufficient?

cwhanse:
> Or should we add the functionality now, and just raise an error if the installed version of scipy is insufficient?

I'm in favor, if the error message suggests that the fix is to select a different method _or_ to update scipy.

kandersolar:
To-do list for this issue:

- [x] Add `method='chandrupatla'` to `pvlib.pvsystem` functions: `singlediode`, `i_from_v`, `v_from_i`, `max_power_point`
- [x] Add `method='chandrupatla'` to `pvlib.singlediode` functions: `bishop88_mpp`, `bishop88_v_from_i`, `bishop88_i_from_v`
- [x] Replace `_golden_sect_DataFrame` with `find_minimum`

PR #948, opened 2020-03-31
add capability for reverse bias

cwhanse:
- [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing.html)
 - [x] Tests added
 - [x] Updates entries to [`docs/sphinx/source/api.rst`](https://github.com/pvlib/pvlib-python/blob/master/docs/sphinx/source/api.rst) for API changes.
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/master/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [x] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [x] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels and Milestone are assigned to the Pull Request and linked Issue.

Adds reverse bias capability to single diode functions. The Lambert W technique cannot solve the single diode equation with the term for current at reverse bias. The Bishop '88 methods can solve the equation. Tested with `method='newton'`, not clear if `method='brentq'` is reliable for this application.

PR #2498, opened 2025-07-08
Add `scipy.optimize.elementwise.find_root` (`method='chandrupatla'`) to bishop88 functions

kandersolar:
- [x] Partway towards #2497 
 - [x] Tests added

This PR adds `method='chandrupatla'` to the bishop88 functions.  I decided to leave `lambertw` for a separate PR, since it's not clear how to opt into using the new scipy function instead of our golden section search.

Some basic timing comparisons for the wrapper functions in `pvlib.pvsystem` (all times in milliseconds):

| function      | inputs                 | lambertw | brentq | newton | chandrupatla |
|---------------|------------------------|----------|--------|--------|--------------|
| `singlediode` | 8760 simulation        | 156      | 6370   | 32.2   | 48           |
| `v_from_i`    | I-V curve, 1000 points | 0.685    | 136    | 3.39   | 5.62         |
| `i_from_v`    | I-V curve, 1000 points | 0.664    | 87.5   | 1.06   | 3.64         |

Still need to figure out a clean way to skip the tests on python 3.9, where this new functionality is not available.

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -2722,7 +2722,8 @@ def v_from_i(current, photocurrent, saturation_current, resistance_series,
 
     method : str
         Method to use: ``'lambertw'``, ``'newton'``, ``'brentq'``, or
-        ``'chandrupatla'``. *Note*: ``'brentq'`` is limited to non-negative current.
+        ``'chandrupatla'``. *Note*: ``'brentq'`` is limited to
+        non-negative current.
 
         .. note::
            ``'chandrupatla'`` requires scipy 1.15 or greater.
@@ -2808,7 +2809,8 @@ def i_from_v(voltage, photocurrent, saturation_current, resistance_series,
 
     method : str
         Method to use: ``'lambertw'``, ``'newton'``, ``'brentq'``, or
-        ``'chandrupatla'``. *Note*: ``'brentq'`` is limited to non-negative current.
+        ``'chandrupatla'``. *Note*: ``'brentq'`` is limited to
+        non-negative current.
 
         .. note::
            ``'chandrupatla'`` requires scipy 1.15 or greater.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,8 +3,8 @@ import platform
 import warnings
 
 import pandas as pd
+import scipy
 import os
-import sys
 from packaging.version import Version
 import pytest
 from functools import wraps
@@ -197,10 +197,9 @@ requires_pandas_2_0 = pytest.mark.skipif(not has_pandas_2_0,
 
 # single-diode equation functions have method=='chandrupatla', which relies
 # on scipy.optimize.elementwise.find_root, which is only available in
-# scipy>=1.15.  That is only available for python 3.10 and above, so
-# we need to skip those tests on python 3.9.
-# TODO remove this when we drop support for python 3.9.
-chandrupatla_available = sys.version_info >= (3, 10)
+# scipy>=1.15.
+# TODO remove this when our minimum scipy is >=1.15
+chandrupatla_available = Version(scipy.__version__) >= Version("1.15.0")
 chandrupatla = pytest.param(
     "chandrupatla", marks=pytest.mark.skipif(not chandrupatla_available,
                                              reason="needs scipy 1.15")


pvlib/singlediode.py
```diff
@@ -162,12 +162,11 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
     # calculate temporary values to simplify calculations
     v_star = diode_voltage / nNsVth  # non-dimensional diode voltage
     g_sh = 1.0 / resistance_shunt  # conductance
-    if breakdown_factor > 0:  # reverse bias is considered
```
kandersolar, line 165: Had to get rid of the `if` for vectorized evaluation in `find_root`.  The math works out the same either way, it's just a bit of unnecessary computation when `breakdown_factor` is zero.

(see also #1821)

pvlib/singlediode.py
```diff
@@ -177,18 +176,14 @@ def bishop88(diode_voltage, photocurrent, saturation_current,
         grad_i_recomb = np.where(is_recomb, i_recomb / v_recomb, 0)
         grad_2i_recomb = np.where(is_recomb, 2 * grad_i_recomb / v_recomb, 0)
         g_diode = saturation_current * np.exp(v_star) / nNsVth  # conductance
-        if breakdown_factor > 0:  # reverse bias is considered
```
kandersolar, line 180: Same here

mikofski:
I think Chandrupatla is a great alternative to Brent Q or Toms-748, neither which are vectorized. The Cythonized version of Brent Q is problematic to deploy in pvlib, and Newton can fail if guesses are too far from the roots. Another option to consider is Toms-748 with Numba

echedey-ls:
> Another option to consider is Toms-748 with Numba

Feel free to open an issue regarding that! It's a bit out-of-scope for this PR IMO (not the issue, as long as it's timing gets compared there). It would make for a great contribution thou.

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -18,7 +18,14 @@ Bug fixes
 
 Enhancements
 ~~~~~~~~~~~~
-
+* Added ``method='chandrupatla'`` to :py:func:`pvlib.pvsystem.singlediode`,
```
echedey-ls: Any comment on speed improvements, for context? For example, `see issue#... for a brief comparison of the different solvers` or `this is faster than brentq but slower than newton, but convergence is guaranteed because it is bounded` (I don't really know if it really qualifies as bounded just by having an initial bracket)

pvlib/pvsystem.py
```diff
@@ -2498,7 +2498,8 @@ def singlediode(photocurrent, saturation_current, resistance_series,
 
     method : str, default 'lambertw'
         Determines the method used to calculate points on the IV curve. The
-        options are ``'lambertw'``, ``'newton'``, or ``'brentq'``.
+        options are ``'lambertw'``, ``'newton'``, ``'brentq'``, or
+        ``'chandrupatla'`` (requires scipy 1.15 or greater).
```
echedey-ls: Personal mild preference, feel free to ignore.


```suggestion
        ``'chandrupatla'``.
        
        .. note::
           ``'chandrupatla'`` requires scipy 1.15 or greater.
```

pvlib/singlediode.py
```diff
@@ -247,12 +242,14 @@ def bishop88_i_from_v(voltage, photocurrent, saturation_current,
...
+       Either ``'newton'``, ``'brentq'``, or ``'chandrupatla'`` (requires
+       scipy 1.15 or greater). ''method'' must be ``'newton'``
        if ``breakdown_factor`` is not 0.
     method_kwargs : dict, optional
         Keyword arguments passed to root finder method. See
-        :py:func:`scipy:scipy.optimize.brentq` and
-        :py:func:`scipy:scipy.optimize.newton` parameters.
+        :py:func:`scipy:scipy.optimize.brentq`,
+        :py:func:`scipy:scipy.optimize.newton`, and
+        :py:func:`scipy:scipy.optimize.elementwise.find_root` for parameters.
```
echedey-ls:
```suggestion
        :py:func:`scipy:scipy.optimize.elementwise.find_root` (when ``method='chandrupatla'``) for parameters.
```
As long as there is only one mismatch in the method name and the scipy function, it's not important, but wondering if it's better to be extra clear.

pvlib/singlediode.py
```diff
@@ -388,7 +409,8 @@ def bishop88_v_from_i(current, photocurrent, saturation_current,
     breakdown_exp : float, default 3.28
         avalanche breakdown exponent :math:`m` [unitless]
     method : str, default 'newton'
-       Either ``'newton'`` or ``'brentq'``. ''method'' must be ``'newton'``
+       Either ``'newton'``, ``'brentq'``, or ``'chandrupatla'`` (requires
+       scipy 1.15 or greater). ''method'' must be ``'newton'``
```
echedey-ls: At this point I would send users to ```See methods allowed by :py:func:`pvlib.singlediode.bishop88_mpp` ``` or equivalent.

pvlib/singlediode.py
```diff
@@ -388,7 +409,8 @@ def bishop88_v_from_i(current, photocurrent, saturation_current,
...
        if ``breakdown_factor`` is not 0.
     method_kwargs : dict, optional
```
echedey-ls, line 423: May also need updating;

```
method_kwargs : dict, optional
        Keyword arguments passed to root finder function chose by ``method``. 
        See:

          * ``method='brentq'``: :py:func:`scipy:scipy.optimize.brentq`
          * ``method='newton'``: :py:func:`scipy:scipy.optimize.newton`
          * ``method='chandrupatla'``: :py:func:`scipy:scipy.optimize.elementwise.find_root`

        ``'full_output': True`` is allowed, and ``optimizer_output`` would be
        returned. See examples section.
```

pvlib/singlediode.py
```diff
@@ -474,6 +496,29 @@ def vd_from_brent(voc, i, iph, isat, rs, rsh, gamma, d2mutau, NsVbi,
...
+                f"({str(e)})"
+            )
+            raise ImportError(msg)
+
+        shape = _shape_of_max_size(current, voc_est)
+        vlo = np.zeros(shape)
+        vhi = np.full(shape, voc_est)
+        bounds = (vlo, vhi)
+        kwargs_trimmed = method_kwargs.copy()
+        kwargs_trimmed.pop("full_output", None)  # not valid for find_root
```
echedey-ls, line 527: Great detail. I would raise an exception thou, and remove the claim that `full_output=True` is supported in general (my suggestion above is mildly incorrect, but just in case you want to ignore either that, this or both)

pvlib/singlediode.py
```diff
@@ -474,6 +496,29 @@ def vd_from_brent(voc, i, iph, isat, rs, rsh, gamma, d2mutau, NsVbi,
...
+        try:
+            from scipy.optimize.elementwise import find_root
+        except ModuleNotFoundError as e:
+            # TODO remove this when our minimum scipy version is >=1.15
+            msg = (
+                "method='chandrupatla' requires scipy v1.15 or greater. "
+                "Select another method, or update your version of scipy. "
+                f"({str(e)})"
+            )
+            raise ImportError(msg)
```
echedey-ls:
```suggestion
            msg = (
                "method='chandrupatla' requires scipy v1.15 or greater. "
                "Select another method, or update your version of scipy. "
            )
            raise ImportError(msg) from e
```

or 


```suggestion
            msg = (
                "method='chandrupatla' requires scipy v1.15 or greater. "
                "Select another method, or update your version of scipy. "
                f"({str(e)})"
            )
            raise ImportError(msg) from None
```
https://stackoverflow.com/questions/24752395/python-raise-from-usage

The first one will show that the raised exception is a direct cause of the `ModuleNotFoundError`. The second one will omit any reference to the `ModuleNotFoundError`.

As of now, it will show that another exception happened while handling it, which can be made more clear. Let me know if you want real-pvlib examples.

pvlib/singlediode.py
```diff
@@ -474,6 +496,29 @@ def vd_from_brent(voc, i, iph, isat, rs, rsh, gamma, d2mutau, NsVbi,
         vd = newton(func=lambda x, *a: fi(x, current, *a), x0=x0,
                     fprime=lambda x, *a: bishop88(x, *a, gradients=True)[3],
                     args=args, **method_kwargs)
+    elif method == 'chandrupatla':
+        try:
+            from scipy.optimize.elementwise import find_root
+        except ModuleNotFoundError as e:
+            # TODO remove this when our minimum scipy version is >=1.15
+            msg = (
+                "method='chandrupatla' requires scipy v1.15 or greater. "
```
echedey-ls:
```suggestion
            msg = (
                "method='chandrupatla' requires scipy v1.15 or greater (available for Python3.10+). "
```

pvlib/singlediode.py
kandersolar: Even though `full_output` isn't an option for `find_root`, I think it makes sense to keep it.  `find_root` returns a rich object always, so the pvlib code returns that if `full_output` is requested.  I guess I am thinking of it more as a pvlib parameter than a kwarg that gets passed through to scipy.

pvlib/singlediode.py
kandersolar: I intentionally avoided `raise ... from ...` because (if I remember correctly) it used to not work well in some ipython consoles.  However, I just checked and it seems to work fine now.  I'll use the first one.  Thanks for the suggestion!

pvlib/singlediode.py
echedey-ls: That's a convincing argument :)
I wasn't aware of the always-rich output of `find_root`.

kandersolar:
Coverage failure should be ignored.  It is failing because coverage is calculated based on the python 3.9 jobs, and these tests are skipped on python 3.9 because of the required scipy version.

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -18,7 +18,16 @@ Bug fixes
 
 Enhancements
 ~~~~~~~~~~~~
-
+* Added ``method='chandrupatla'`` (faster than ``brentq`` and slower than ``newton``,
```
cwhanse:
```suggestion
* Add ``method='chandrupatla'`` (faster than ``brentq`` and slower than ``newton``,
```

Nit: we should prefer present tense for what's new notes.

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -18,7 +18,16 @@ Bug fixes
...
+  but convergence is guaranteed) to
```
cwhanse:
```suggestion
  but convergence is guaranteed) as an option for
```

pvlib/pvsystem.py
```diff
@@ -2713,8 +2721,13 @@ def v_from_i(current, photocurrent, saturation_current, resistance_series,
         0 < nNsVth
 
     method : str
-        Method to use: ``'lambertw'``, ``'newton'``, or ``'brentq'``. *Note*:
-        ``'brentq'`` is limited to 1st quadrant only.
+        Method to use: ``'lambertw'``, ``'newton'``, ``'brentq'``, or
+        ``'chandrupatla'``. *Note*: ``'brentq'`` is limited to 1st quadrant
+        only.
```
```suggestion
        ``'chandrupatla'``. *Note*: ``'brentq'`` is limited to non-negative current.
```

pvlib/pvsystem.py
```diff
@@ -2795,8 +2808,13 @@ def i_from_v(voltage, photocurrent, saturation_current, resistance_series,
...
```
```suggestion
        ``'chandrupatla'``. *Note*: ``'brentq'`` is limited to non-negative current.
```

pvlib/singlediode.py
cwhanse: Not sure now why that was there, maybe thinking to save few microseconds.

pvlib/singlediode.py
```diff
@@ -247,12 +242,19 @@ def bishop88_i_from_v(voltage, photocurrent, saturation_current,
     breakdown_exp : float, default 3.28
         avalanche breakdown exponent :math:`m` [unitless]
     method : str, default 'newton'
-       Either ``'newton'`` or ``'brentq'``. ''method'' must be ``'newton'``
-       if ``breakdown_factor`` is not 0.
+        Either ``'newton'``, ``'brentq'``, or ``'chandrupatla'``.
+        ''method'' must be ``'newton'`` if ``breakdown_factor`` is not 0.
```
cwhanse, line 246: Why this requirement for `'newton'`? The arguments passed when `method='brentq'` or `method='chandrupatla'` include all the reverse bias parameters.

tests/test_pvsystem.py
```diff
@@ -1400,44 +1407,43 @@ def test_PVSystem_i_from_v(mocker):
...
-def test_i_from_v_size():
-    with pytest.raises(ValueError):
-        pvsystem.i_from_v([7.5] * 3, 7., 6e-7, [0.1] * 2, 20, 0.5)
-    with pytest.raises(ValueError):
-        pvsystem.i_from_v([7.5] * 3, 7., 6e-7, [0.1] * 2, 20, 0.5,
-                          method='brentq')
+@pytest.mark.parametrize('method', ['lambertw', 'brentq', 'newton',
+                                    chandrupatla])
+def test_i_from_v_size(method):
+    if method == 'newton':
```
cwhanse, line 1413: Is this why the previous restriction to `method='newton'`?

pvlib/singlediode.py
kandersolar: I do not know the answer to this.  I see in #948 you wrote this:

> The Lambert W technique cannot solve the single diode equation with the term for current at reverse bias. The Bishop '88 methods can solve the equation. Tested with method='newton', not clear if method='brentq' is reliable for this application.

cwhanse: My apologies for asking you to explain something I added 5 years ago :). I didn't pick up that all you did was reformat a bit of documentation.

I think that comment was added because the tests didn't confirm that `'brentq'` works. I can't think of a reason why it wouldn't work. I suspect there was a vectorization issue that I didn't see how to fix at the time. It may be gone with removal of the `if breakdown > 0:` statement.

tests/test_pvsystem.py
mikofski: I think the reason for this requirement was because Brent is a bounded search, and was searching between zero and `v_oc_est`, but I guess if you know the reverse bias breakdown voltage `v_rbd` then you can use that instead of zero? Perhaps the concern was that voltage asymptote and never intersects with `v_rbd` while we know that voltage increases greater than `v_oc_est` but is that a necessary requirement for Brent?

mikofski: I’m okay with removing Brent without deprecation if that’s in scope

kandersolar: I don't think removing brentq should be considered until our minimum scipy version is recent enough to have chandrupatla available.  Guessing that won't be for be a year or two.

pvlib/singlediode.py
cwhanse: I think we could remove that constraint requiring `method='newton'` with `breakdown_factor>0`. But in a follow-on PR.

tests/test_pvsystem.py
```diff
@@ -1400,44 +1407,43 @@ def test_PVSystem_i_from_v(mocker):
...
 
 
-def test_i_from_v_size():
-    with pytest.raises(ValueError):
-        pvsystem.i_from_v([7.5] * 3, 7., 6e-7, [0.1] * 2, 20, 0.5)
-    with pytest.raises(ValueError):
-        pvsystem.i_from_v([7.5] * 3, 7., 6e-7, [0.1] * 2, 20, 0.5,
-                          method='brentq')
+@pytest.mark.parametrize('method', ['lambertw', 'brentq', 'newton',
+                                    chandrupatla])
```
cwhanse, line 1411: quotes needed `'chandrupatla'`?

tests/test_pvsystem.py
```diff
@@ -1400,44 +1407,43 @@ def test_PVSystem_i_from_v(mocker):
...
     assert isinstance(out, dict)
     for k, v in out.items():
         assert np.isclose(v, expected[k])
-    out = pvsystem.max_power_point(IL, I0, Rs, Rsh, nNsVth, method='newton')
-    for k, v in out.items():
-        assert np.isclose(v, expected[k])
 
 
-def test_mpp_recombination():
+@pytest.mark.parametrize('method', ['brentq', 'newton', chandrupatla])
```
cwhanse, line 1445:
```suggestion
@pytest.mark.parametrize('method', ['brentq', 'newton', 'chandrupatla'])
```

There are several more instances of `chandrupatla` instead of `'chandrupatla'`

tests/test_pvsystem.py
kandersolar: quotes not needed; chandrupatla is a pytest param that evaluates to the string `'chandrupatla'`, but skips when the necessary scipy version isn't installed.  See its definition here: https://github.com/pvlib/pvlib-python/pull/2498/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R203-R206

PR #2499, opened 2025-07-15
Add iotools functions for Meteonorm 

AdamRJensen:
 - [x] Tests added

This PR adds functions for retrieving irradiance and weather data from Meteonorm. This is the last function I plan on contributing to the commercial iotools, as pvlib then will support what I consider the major providers of satellite-derived irradiance data.

Documentation:
- [get_meteonorm](https://pvlib-python--2499.org.readthedocs.build/en/2499/reference/generated/pvlib.iotools.get_meteonorm.html)
- [get_meteonorm_tmy](https://pvlib-python--2499.org.readthedocs.build/en/2499/reference/generated/pvlib.iotools.get_meteonorm_tmy.html)

For anyone willing to review this PR, I can provide a temporary API key.

### DIFF
--- a/tests/iotools/test_meteonorm.py
+++ b/tests/iotools/test_meteonorm.py
@@ -51,7 +51,7 @@ def expected_meta():
 @pytest.fixture
 def expected_meteonorm_index():
     expected_meteonorm_index = \
-        pd.date_range('2023-01-01', '2024-12-31 23:59', freq='1h', tz='UTC') \
+        pd.date_range('2023-01-01', '2023-12-31 23:59', freq='1h', tz='UTC') \
         + pd.Timedelta(minutes=30)
     expected_meteonorm_index.freq = None
     return expected_meteonorm_index
@@ -114,7 +114,7 @@ def test_get_meteonorm_training(
         expected_meteonorm_data):
     data, meta = pvlib.iotools.get_meteonorm_observation_training(
         latitude=50, longitude=10,
-        start='2023-01-01', end='2025-01-01',
+        start='2023-01-01', end='2024-01-01',
         api_key=demo_api_key,
         parameters=['ghi', 'global_horizontal_irradiance_with_shading'],
         time_step='1h',
@@ -159,8 +159,8 @@ def test_get_meteonorm_realtime(demo_api_key, demo_url, expected_columns_all):
 def test_get_meteonorm_forecast_basic(demo_api_key, demo_url):
     data, meta = pvlib.iotools.get_meteonorm_forecast_basic(
         latitude=50, longitude=10,
-        start=pd.Timestamp.now(tz='UTC'),
-        end=pd.Timestamp.now(tz='UTC') + pd.Timedelta(hours=5),
+        start='+1hours',
+        end=pd.Timestamp.now(tz='UTC') + pd.Timedelta(hours=6),
         api_key=demo_api_key,
         parameters='ghi',
         url=demo_url)


pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+    surface_azimuth : float, optional
+        Orientation (azimuth angle) of the (fixed) plane. Clockwise from north
+        (north=0, east=90, south=180, west=270). The default is 180.
+    time_step: {'1min', '1h'}, optional
+        Frequency of the time series. The default is '1h'.
+    horizon: optional
+        Specification of the hoirzon line. Can be either 'flat' or 'auto', or
+        specified as a list of 360 horizon elevation angles. The default is
+        'auto'.
+    terrain: string, optional
```
IoannisSifnaios:
```suggestion
    terrain: str, optional
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+        default is 0.2.
+    turbidity: list or 'auto', optional
+        List of 12 monthly mean atmospheric Linke turbidity values. The default
+        is 'auto'.
+    random_seed: int, optional
+        Random seed to be used for stochastic processes. Two identical requests
+        with the same random seed will yield identical results.
+    clear_sky_radiation_model : {'esra', 'solis'}
+        Which clearsky model to use. The default is 'esra'.
+    data_version : string, optional
```
IoannisSifnaios:
```suggestion
    data_version : str, optional
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+        Version of Meteonorm climatological data to be used. The default is
+        'latest'.
+    future_scenario: string, optional
```
IoannisSifnaios:
```suggestion
    future_scenario: str, optional
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+        Future climate scenario.
+    future_year : integer, optional
```
IoannisSifnaios:
```suggestion
    future_year : int, optional
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+    interval_index: bool, optional
+        Whether the index of the returned data object is of the type
+        pd.DatetimeIndex or pd.IntervalIndex. This is an experimental feature
+        which may be removed without warning. The default is False.
+    map_variables: bool, optional
+        When true, renames columns of the Dataframe to pvlib variable names
+        where applicable. See variable :const:`VARIABLE_MAP`. The default is
+        True.
+    url: str, optional.
+        Base URL of the Meteonorm API. 'climate/tmy'` is
```
IoannisSifnaios:
```suggestion
        Base URL of the Meteonorm API. `'climate/tmy'` is
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+
+    Parameters
+    ----------
+    latitude: float
+        In decimal degrees, north is positive (ISO 19115).
+    longitude: float
+        In decimal degrees, east is positive (ISO 19115).
+    api_key: str
+        Meteonorm API key.
+    parameters: list, optional
```
IoannisSifnaios: Maybe this should be list or 'all' (like the turbidity)?

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+        List of parameters to request or 'all' to get all parameters. The
+        default is 'all'.
+    surface_tilt: float, optional
+        Tilt angle from horizontal plane. The default is 0.
+    surface_azimuth : float, optional
+        Orientation (azimuth angle) of the (fixed) plane. Clockwise from north
+        (north=0, east=90, south=180, west=270). The default is 180.
+    time_step: {'1min', '1h'}, optional
+        Frequency of the time series. The default is '1h'.
+    horizon: optional
```
IoannisSifnaios:
```suggestion
    horizon: str, optional
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+        default is 'all'.
+    surface_tilt: float, optional
+        Tilt angle from horizontal plane. The default is 0.
+    surface_azimuth: float, optional
+        Orientation (azimuth angle) of the (fixed) plane. Clockwise from north
+        (north=0, east=90, south=180, west=270). The default is 180.
+    time_step : {'1min', '15min', '1h'}, optional
+        Frequency of the time series. The parameter is ignored when requesting
+        forcasting data. The default is '15min'.
+    horizon : optional
```
IoannisSifnaios:
```suggestion
    horizon : str, optional
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+        Specification of the horizon line. Can be either a 'flat', 'auto', or
```
maschwanden:
```suggestion
        Specification of the horizon line. Can be either 'flat', 'auto', or
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,337 @@
...
+        Specification of the hoirzon line. Can be either 'flat' or 'auto', or
+        specified as a list of 360 horizon elevation angles. The default is
+        'auto'.
+    terrain: string, optional
+        Local terrain situation. Must be one of: ['open', 'depression',
+        'cold_air_lake', 'sea_lake', 'city', 'slope_south',
+        'slope_west_east']. The default is 'open'.
+    albedo: float, optional
+        Ground albedo. Albedo changes due to snow fall are modelled. The
+        default is 0.2.
```
maschwanden:
```suggestion
    albedo: float, optional
        Fixed ground albedo. If not set, a default albedo of 0.2 is used and albedo changes due to snow fall are modelled.
```
Mostly copy-pasted from our API docs. 

Important: When the **fixed albedo** is set, albedo changes due to snow fall are **not** modelled! Thus, this setting does not set a "baseline albedo" which is altered by snow, but rather a fixed albedo which **ignores** changes due to snow.

docs/sphinx/source/reference/iotools.rst
```diff
@@ -92,6 +92,16 @@ Commercial datasets
 Accessing these APIs typically requires payment.
 Datasets provide near-global coverage.
 
+Meteonorm
+*********
+
+.. autosummary::
+   :toctree: generated/
+
+   iotools.get_meteonorm
```
maschwanden: For Solcast there are specialized functions for retrieving TMY, historical and forecast data. For Meteonorm there is currently only a TMY and a "everthing else" getter function (for the endpoints `/observation/training`, `/observation/realtime`, `/forecast/basic`, and `/forecast/precision`). What's the reasoning behind implementing `get_meteonorm` instead of `get_meteonorm_observation_training`, `get_meteonorm_observation_realtime`, `get_meteonorm_forecast_basic`, and `get_meteonorm_forecast_precision`? There are pros and cons for both variants, but i would argue, that for the end user it's easier if there's a specialized function for every endpoint.

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,349 @@
...
+        specified, UTC is assumed. Relative datetime strings are supported.
+    api_key : str
+        Meteonorm API key.
+    endpoint : str
+        API endpoint, see [3]_. Must be one of:
+
+        * ``'observation/training'`` - historical data with a 7-day delay
+        * ``'observation/realtime'`` - near-real time (past 7-days)
+        * ``'forecast/basic'`` - forecast with hourly resolution
+        * ``'forecast/precision'`` - forecast with 15-min resolution
```
maschwanden:
```suggestion
        * ``'forecast/precision'`` - forecast with hourly, 15-min, or 1-min resolution
```
The resolution is set by `time_step` (see [here](https://docs.meteonorm.com/api/forecast#forecast-precision)).

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,349 @@
...
+        'surface_azimuth': surface_azimuth,
+        'horizon': horizon,
+    }
+
+    # Allow specifying single parameters as string
+    if isinstance(parameters, str):
+        parameter_list = \
+            list(VARIABLE_MAP.keys()) + list(VARIABLE_MAP.values())
+        if parameters in parameter_list:
+            parameters = [parameters]
```
maschwanden:
```suggestion
        parameters = [parameters]
```
I do not understand why only parameters corresponding to either a key or value of `VARIABLE_MAP` are "cast" to a list. IMHO the code is easier to read if the cast is done always (the mapping of pvlib variable names to Meteonorm variable names is either way done in the next if block).

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,349 @@
...
+        # allow the use of pvlib parameter names
+        parameter_dict = {v: k for k, v in VARIABLE_MAP.items()}
+        parameters = [parameter_dict.get(p, p) for p in parameters]
+        params['parameters'] = ','.join(parameters)
+
+    if not isinstance(horizon, str):
+        params['horizon'] = ','.join(map(str, horizon))
+
+    if 'forecast' not in endpoint.lower():
+        params['frequency'] = TIME_STEP_MAP.get(time_step, time_step)
```
maschwanden: `/forecast/precision` does have a parameter `frequency`. Allowing the user to select a (higher) frequency is one of the advantages of this endpoint over `/forecast/basic`.

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,349 @@
...
+
+    headers = {"Authorization": f"Bearer {api_key}"}
+
+    response = requests.get(
+        urljoin(url, endpoint.lstrip('/')), headers=headers, params=params)
+
+    if not response.ok:
+        # response.raise_for_status() does not give a useful error message
+        raise requests.HTTPError(response.json())
```
maschwanden: If a more concise error message is desired, we could extract the error message here (the structure of the error response is defined in the OpenAPI and thus will not change in `v1`).
 ```suggestion
        raise requests.HTTPError("Meteonorm API returned an error: " 
        + response.json()['error']['message'])
```

File: pvlib/iotools/meteonorm.py [hunk 16]
```diff
@@ -0,0 +1,349 @@
...
+
+    return data, meta
+
+
+TMY_ENDPOINT = 'climate/tmy'
+
+
+def get_meteonorm_tmy(latitude, longitude, api_key,
+                      parameters='all', *, surface_tilt=0,
+                      surface_azimuth=180, time_step='15min', horizon='auto',
```
maschwanden: Meteonorm `/climate/tmy` currently only supports hourly and minute frequency. Maybe we add 15-min frequency in the future but that's not a high priority feature for us.
```suggestion
                      surface_azimuth=180, time_step='1h', horizon='auto',
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,349 @@
...
+                      terrain='open', albedo=0.2, turbidity='auto',
```
maschwanden: `terrain` could be confused with `horizon`, e.g. Solcast calls it's horizon parameter "terrain_shading". We therefore decided to name the parameter "situation" (which is also far from optimal).

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,349 @@
...
+        'future_scenario': future_scenario,
+        'future_year': future_year,
+    }
+
+    # Allow specifying single parameters as string
+    if isinstance(parameters, str):
+        parameter_list = \
+            list(VARIABLE_MAP.keys()) + list(VARIABLE_MAP.values())
+        if parameters in parameter_list:
+            parameters = [parameters]
```
maschwanden: Same as above.

tests/iotools/test_meteonorm.py
```diff
@@ -0,0 +1,323 @@
...
+@pytest.mark.remote_data
+@pytest.mark.flaky(reruns=RERUNS, reruns_delay=RERUNS_DELAY)
+def test_get_meteonorm_tmy_HTTPError(demo_api_key, demo_url):
+    with pytest.raises(
+            HTTPError, match='parameter "surface_azimuth"'):
+        _ = pvlib.iotools.get_meteonorm_tmy(
+            latitude=50, longitude=10,
+            api_key=demo_api_key,
+            parameters='dhi',
+            # Infeasible surface_titl
```
maschwanden:
```suggestion
            # Infeasible surface_tilt
```

pvlib/iotools/meteonorm.py
AdamRJensen: I have changed it to defaulting to meteonorm's default (no value specified). I decided to call it "constant albedo" instead of "fixed albedo". Thanks for the correction.

pvlib/iotools/meteonorm.py
AdamRJensen: That is a good point, I agree!

pvlib/iotools/meteonorm.py
AdamRJensen: What about calling it `terrain_situation` here? Just calling it "situation" could refer to a lot of situations..

@maschwanden could you elaborate a bit on what this parameter influences?

docs/sphinx/source/reference/iotools.rst
AdamRJensen: I might be convinced to have a separate function for ``get_meteonorm_observation`` and ``get_meteonorm_forecast`` with a keyword parameter to switch between the subcases. The main motivation is to reduce the duplicate docstring and reduce the maintenance burden.

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,348 @@
...
+from urllib.parse import urljoin
+
+URL = 'https://api.meteonorm.com/v1/'
+
+VARIABLE_MAP = {
+    'global_horizontal_irradiance': 'ghi',
+    'diffuse_horizontal_irradiance': 'dhi',
+    'direct_normal_irradiance': 'dni',
+    'direct_horizontal_irradiance': 'bhi',
+    'global_clear_sky_irradiance': 'ghi_clear',
```
kandersolar: `dhi_clear`, `dni_clear`, `bhi_clear` can be listed as well

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,348 @@
...
+    Raises
+    ------
+    requests.HTTPError
+        Raises an error when an incorrect request is made.
+
+    Returns
+    -------
+    data : pd.DataFrame
+        Time series data. The index corresponds to the start (left) of the
+        interval unless ``interval_index`` is set to False.
```
kandersolar: Should this say `True`?

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,348 @@
...
+    References
+    ----------
+    .. [1] `Meteonorm
+       <https://meteonorm.com/>`_
+    .. [2] `Meteonorm API
+       <https://docs.meteonorm.com/docs/getting-started>`_
+    .. [3] `Meteonorm API reference
+       <https://docs.meteonorm.com/api>`_
+    """
+    # Relative date strings are not yet supported
```
kandersolar: What does "relative" mean here?

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,348 @@
...
+        'horizon': horizon,
+        'response_format': 'json',
+    }
+
+    # Allow specifying single parameters as string
+    if isinstance(parameters, str):
+        parameters = [parameters]
+
+    # convert list to string with values separated by commas
+    if not isinstance(parameters, (str, type(None))):
```
kandersolar: I don't get this condition.  `parameters` cannot be a `str` at this point (due to the list-wrapping line above).  And is `None` supported?  If so, it's not documented.

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,348 @@
...
+        Specification of the horizon line. Can be either 'flat' or 'auto', or
+        specified as a list of 360 integer horizon elevation angles.
+        'auto'.
+    terrain : str, default : 'open'
+        Local terrain situation. Must be one of: ['open', 'depression',
+        'cold_air_lake', 'sea_lake', 'city', 'slope_south',
+        'slope_west_east'].
+    albedo : float, optional
+        Constant ground albedo. If no value is specified a baseline albedo of
+        0.2 is used and albedo cahnges due to snow fall is modeled. If a value
```
kandersolar:
```suggestion
        0.2 is used and albedo changes due to snow fall are modeled. If a value
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,348 @@
...
+    Raises
+    ------
+    requests.HTTPError
+        Raises an error when an incorrect request is made.
+
+    Returns
+    -------
+    data : pd.DataFrame
+        Time series data. The index corresponds to the start (left) of the
+        interval unless ``interval_index`` is set to False.
```
kandersolar: Same here

pvlib/iotools/meteonorm.py
kandersolar: Ah, maybe the "+12hours" format for the forecast API?

docs/sphinx/source/reference/iotools.rst
kandersolar: I'm not sure what the right answer here is.  Maybe @AdamRJensen's proposal is the best balance.

pvlib/iotools/meteonorm.py
AdamRJensen: I think this is a relic from previous iterations. The if statement has been removed.

docs/sphinx/source/reference/iotools.rst
maschwanden: Yes, I see that this is a maintenance nightmare with doc-strings of this size.
I assume messing around with the `__doc__` field to share a part of the documentation is not really an option; this becomes unreadable very fast.

Despite that, every request to the API costs the user real money and thus I'm critical about using parameters - instead of separate functions - to decide which endpoint to call.

> [...] keyword parameter to switch between the subcases.

@AdamRJensen How would this look like? One option would be to use the exact same signature but the value to the parameter `endpoint` has to be `"basic"`/`"precision"` for `get_meteonorm_forecast` and `"realtime"`/`"training"` for `get_meteonorm_observation`.
Note that there is no parameter `frequency` for the `/forecast/basic` endpoint, thus it would be a good idea to raise a warning or exception if the user uses `endpoint="basic", time_step="1min"`.

pvlib/iotools/meteonorm.py
maschwanden: Agree. I like `terrain_situation`, if it wouldn't be a breaking change i would even consider renaming it in the API :).

This parameter can increase the TMY quality by providing information about the climatic/geographic/topographic situation of the location to Meteonorm. If set, monthly variations/corrections are applied to the parameters **temperature** and **wind_speed**.

Some examples:
- `terrain_situation="summit"` :  wind speeds are increased
- `terrain_situation="city"` : the temperatures are increased
- `terrain_situation="cold_air_lake"` : the temperatures are decreased

pvlib/iotools/meteonorm.py
maschwanden: Perfect :+1: "constant albedo" sounds good.

pvlib/iotools/meteonorm.py
AdamRJensen: I have changed the parameter name to ``terrain_situation``.

docs/sphinx/source/reference/iotools.rst
AdamRJensen: > @AdamRJensen How would this look like? One option would be to use the exact same signature but the value to the parameter `endpoint` has to be `"basic"`/`"precision"` for `get_meteonorm_forecast` and `"realtime"`/`"training"` for `get_meteonorm_observation`. Note that there is no parameter `frequency` for the `/forecast/basic` endpoint, thus it would be a good idea to raise a warning or exception if the user uses `endpoint="basic", time_step="1min"`.

@maschwanden If we are to split it into separate forecast and observation functions, then this is exactly how I envisioned it. I'm curious what other members of @pvlib/pvlib-maintainer have to say about this.

Also, I've added an error message if the basic forecast is requested with a frequency other than '1_hour'.

AdamRJensen: @maschwanden Check out the changes in the last commit. Would this approach be ok with you?

wholmgren: If I understand correctly, @maschwanden proposed `meteonorm.get_meteonorm_observation_training` is similar to the existing `solcast.get_solcast_historic` function and the  `meteonorm.get_meteonorm_observation_realtime` is similar to the `solcast.get_solcast_live` function. I would favor API parallelism over concerns about docstrings.

pvlib/iotools/meteonorm.py
```diff
@@ -31,18 +31,19 @@
 }
 
 
-def get_meteonorm(latitude, longitude, start, end, api_key, endpoint,
-                  parameters='all', *, surface_tilt=0, surface_azimuth=180,
-                  time_step='15min', horizon='auto', interval_index=False,
-                  map_variables=True, url=URL):
+def get_meteonorm_observation(
+        latitude, longitude, start, end, api_key, endpoint='training',
```
maschwanden: I think `'realtime'` would be the better default, it's much cheaper (thus an accidental request will not cost you too much) and `/observation/realtime` is used on a regular basis, whereas `/observation/training` is probably only called once per location to train a ML model.

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,466 @@
...
+
+    if not isinstance(horizon, str):
+        params['horizon'] = ','.join(map(str, horizon))
+
+    if 'basic' not in endpoint:
+        params['frequency'] = TIME_STEP_MAP.get(time_step, time_step)
+    else:
+        if time_step not in ['1h', '1_hour']:
+            raise ValueError("The 'forecast/basic' api endpoint only "
+                             "supports ``time_step='1h'``.")
```
maschwanden: I would rather put this code into `get_meteonorm_forecast`, since it's forecast specific.

Thus: 
```suggestion
    params['frequency'] = TIME_STEP_MAP.get(time_step, time_step)
```

And in `get_meteonorm_forecast` add:

```

    if endpoint == 'basic' and time_step not in ['1h', '1_hour']:
        raise ValueError("The 'forecast/basic' api endpoint only "
                         "supports ``time_step='1h'``.")
```

docs/sphinx/source/reference/iotools.rst
maschwanden: Yes :+1:, this looks good to me!

@wholmgren : Yes, that's what I proposed.

pvlib/iotools/meteonorm.py
AdamRJensen: I have split the functions into four separate functions, so now that the forecast_basic has it's own function this is no longer needed (as there is no time_step parameter for this function).

pvlib/iotools/meteonorm.py
AdamRJensen: No longer applicable, as all end points have separate functions now.

docs/sphinx/source/reference/iotools.rst
AdamRJensen: I've revised it such that all end points now have their own functions.

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,723 @@
...
+    parameters,
+    surface_tilt,
+    surface_azimuth,
+    time_step,
+    horizon,
+    interval_index,
+    map_variables,
+    url,
+    endpoint,
+):
```
kandersolar: We don't use this formatting anywhere else in pvlib, do we?  I'm not a fan of it personally.

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -20,6 +20,11 @@ Bug fixes
 
 Enhancements
 ~~~~~~~~~~~~
+* Add iotools functions to retrieve irradiance and weather data from Meteonorm:
+  :py:func:`~pvlib.iotools.get_meteonorm_forecast_basic`, :py:func:`~pvlib.iotools.get_meteonorm_forecast_precision`,
+  `~pvlib.iotools.get_meteonorm_observation_realtime`, `~pvlib.iotools.get_meteonorm_observation_training`,
```
kandersolar:
```suggestion
  :py:func:`~pvlib.iotools.get_meteonorm_observation_realtime`, :py:func:`~pvlib.iotools.get_meteonorm_observation_training`,
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,723 @@
...
+        Metadata.
+
+    Examples
+    --------
+    >>> # Retrieve historical time series data
+    >>> df, meta = pvlib.iotools.get_meteonorm_observation_training(  # doctest: +SKIP
+    ...     latitude=50, longitude=10,  # doctest: +SKIP
+    ...     start='2023-01-01', end='2025-01-01',  # doctest: +SKIP
+    ...     api_key='redacted',  # doctest: +SKIP
+    ...     endpoint='training')  # doctest: +SKIP
```
kandersolar:
```suggestion
    ...     api_key='redacted')  # doctest: +SKIP
```

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,723 @@
...
+    References
+    ----------
+    .. [1] `Meteonorm
+       <https://meteonorm.com/>`_
+    .. [2] `Meteonorm API
+       <https://docs.meteonorm.com/docs/getting-started>`_
+    .. [3] `Meteonorm API reference
+       <https://docs.meteonorm.com/api>`_
+    """
+    params = {
```
kandersolar: Seems like `_get_meteonorm` could be made flexible enough to accommodate the TMY endpoint as well, but okay with me to not pursue that in this PR.

tests/iotools/test_meteonorm.py
```diff
@@ -0,0 +1,316 @@
...
+@pytest.fixture
+def expected_meteonorm_index():
+    expected_meteonorm_index = \
+        pd.date_range('2023-01-01', '2024-12-31 23:59', freq='1h', tz='UTC')
+    expected_meteonorm_index.freq = None
+    return expected_meteonorm_index
+
+
+@pytest.fixture
+def expected_metenorm_data():
```
kandersolar:
```suggestion
def expected_meteonorm_data():
```
nitpick

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,723 @@
...
+    data : pd.DataFrame
+        Time series data. The index corresponds to the start (left) of the
+        interval unless ``interval_index`` is set to True.
+    meta : dict
+        Metadata.
+
+    See Also
+    --------
+    pvlib.iotools.get_meteonorm_observation,
+    pvlib.iotools.get_meteonorm_forecast
```
kandersolar: Update

pvlib/iotools/meteonorm.py
```diff
@@ -0,0 +1,723 @@
...
+    -------
+    data : pd.DataFrame
+        Time series data. The index corresponds to the start (left) of the
+        interval unless ``interval_index`` is set to True.
+    meta : dict
+        Metadata.
+
+    See Also
+    --------
+    pvlib.iotools.get_meteonorm_observation,
```
kandersolar: Update

kandersolar (approved):
Few minor things to clean up, otherwise OK with me.

pvlib/iotools/meteonorm.py
AdamRJensen: Done!

docs/sphinx/source/reference/iotools.rst
maschwanden: Looks good!

AdamRJensen:
I will merge this PR on Thursday the 14th unless anyone objects.

PR #2500, opened 2025-07-15
Add NASA POWER to iotools

IoannisSifnaios:
- ~~[ ] Closes #xxxx~~
 - [ ] Tests added

As agreed with @AdamRJensen, a function to get data from NASA POWER.

### DIFF
--- a/pvlib/iotools/nasa_power.py
+++ b/pvlib/iotools/nasa_power.py
@@ -1,0 +1,145 @@
+"""Function for reading and retrieving data from NASA POWER."""
+
+import pandas as pd
+import requests
+
+
+URL = 'https://power.larc.nasa.gov/api/temporal/hourly/point'
+
+VARIABLE_MAP = {
+    'ALLSKY_SFC_SW_DWN': 'ghi',
+    'ALLSKY_SFC_SW_DIFF': 'dhi',
+    'ALLSKY_SFC_SW_DNI': 'dni',
+    'CLRSKY_SFC_SW_DWN': 'ghi_clear',
+    'T2M': 'temp_air_2m',
+    'WS2M': 'wind_speed_2m',
+    'WS10M': 'wind_speed_10m',
+}
+
+
+def get_nasa_power(latitude, longitude, start_date, end_date,
+                   parameters, time_standard='utc', community='re',
+                   site_elevation=None, wind_elevation=None,
+                   wind_surface=None, map_variables=True):
+    """
+    Retrieve irradiance and weather data from NASA POWER.
+
+    A general description of NASA POWER is given in [1]_ and the API is
+    described in [2]_. A detailed list of the available parameters can be
+    found in [3]_.
+
+    Parameters
+    ----------
+    latitude: float
+        In decimal degrees, north is positive (ISO 19115).
+    longitude: float
+        In decimal degrees, east is positive (ISO 19115).
+    start_date: datetime like
+        First timestamp of the requested period. If a timezone is not
+        specified, UTC is assumed.
+    end_date: datetime like
+        Last timestamp of the requested period. If a timezone is not
+        specified, UTC is assumed.
+    parameters: str, list
+        List of parameters. Some common parameters are mentioned below; for the
+        full list see [3]_:
+            * ALLSKY_SFC_SW_DWN: Global Horizontal Irradiance (GHI) [Wm⁻²]
+            * ALLSKY_SFC_SW_DIFF: Diffuse Horizonatl Irradiance (DHI) [Wm⁻²]
+            * ALLSKY_SFC_SW_DNI: Direct Normal Irradiance (DNI) [Wm⁻²]
+            * CLRSKY_SFC_SW_DWN: Clear-sky GHI [Wm⁻²]
+            * T2M : Air temperature at 2 m [C]
+            * WS2M: Wind speed at 2 m [m/s]
+            * WS10M: Wind speed at 10 m [m/s]
+    community: str, default: 're'
+        Can be one of the following depending on which parameters are of
+        interest. Note that in many cases this choice might affect the units
+        of the parameter:
+            * 're': renewable energy
+            * 'sb': sustainable buildings
+            * 'ag': agroclimatology
+    time_standard: str, default:'utc'
+        Can be either:
+            * Universal Time Coordinated (utc): is the standard time measure
+            used by the world.
+            * Local Solar Time (lst): A 15 Degrees swath that represents solar
+            noon at the middle longitude of the swath.
+    site_elevation: float, optional
+        The custom site elevation in meters to produce the corrected
+        atmospheric pressure adjusted for elevation.
+    wind_elevation: float, optional
+        The custom wind elevation in meters to produce the wind speed adjusted
+        for elevation. Has to be between 10 and 300 m; see [4]_.
+    wind_surface: str, optional
+        The definable surface type to adjust the wind speed. For a list of the
+        surface types see [4]_.
+    map_variables: bool, default: True
+        When true, renames columns of the Dataframe to pvlib variable names
+        where applicable. The default is True. See variable
+        :const:`VARIABLE_MAP`.
+
+    Raises
+    ------
+    requests.HTTPError
+        Raises an error when an incorrect request is made.
+    ValueError
+        Raises an error if more than 15 parameters are requested in one call.
+
+    Returns
+    -------
+    data : pd.DataFrame
+        Time series data. The index corresponds to the start (left) of the
+        interval.
+
+    References
+    ----------
+    .. [1] `NASA Prediction Of Worldwide Energy Resources (POWER)
+       <https://power.larc.nasa.gov/>`_
+    .. [2] `NASA POWER API
+       <https://power.larc.nasa.gov/api/pages/>`_
+    .. [3] `NASA POWER API parameters
+       <https://power.larc.nasa.gov/parameters/>`_
+    .. [4] `NASA POWER corrected wind speed parameters
+       <https://power.larc.nasa.gov/docs/methodology/meteorology/wind/>_
+    """
+    if len(parameters) > 15:
+        raise ValueError("A maximum of 15 parameters can currently be "
+                         "requested in one submission.")
+
+    start = pd.Timestamp(start_date)
+    end = pd.Timestamp(end_date)
+    start = start.tz_localize('UTC') if start.tzinfo is None else start
+    end = end.tz_localize('UTC') if end.tzinfo is None else end
+
+    params = {
+        'latitude': latitude,
+        'longitude': longitude,
+        'start': start.strftime('%Y%m%d'),
+        'end': end.strftime('%Y%m%d'),
+        'community': community,
+        'parameters': ','.join(parameters),  # make parameters in a string
+        'format': 'json',
+        'user': None,
+        'header': True,
+        'time-standard': time_standard,
+        'site-elevation': site_elevation,
+        'wind-elevation': wind_elevation,
+        'wind-surface': wind_surface,
+    }
+
+    response = requests.get(URL, params=params)
+    if not response.ok:
+        # response.raise_for_status() does not give a useful error message
+        raise requests.HTTPError(response.json())
+
+    # Parse the data to dataframe
+    data = response.json()
+    hourly_data = data['properties']['parameter']
+    df = pd.DataFrame(hourly_data)
+    df.index = pd.to_datetime(df.index, format='%Y%m%d%H')
+    df = df.sort_index()
+
+    # Rename according to pvlib convention
+    if map_variables:
+        df = df.rename(columns=VARIABLE_MAP)
+
+    return df


IoannisSifnaios:
@AdamRJensen could you give some first feedback on this?

Also, should there also be a `read_nasa_power` function or not?

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,144 @@
...
+        Can be either 'utc' or 'lst':
+            * Universal Time Coordinated (utc)
+            * Local Solar Time (lst): A 15 Degrees swath that represents solar
+              noon at the middle longitude of the swath
+    site_elevation: float, optional
+        The custom site elevation in meters to produce the corrected
+        atmospheric pressure adjusted for elevation.
+    wind_elevation: float, optional
+        The custom wind elevation in meters to produce the wind speed adjusted
+        for elevation. Has to be between 10 and 300 m; see [4]_.
```
RDaxini: Just a small comment/question.
Is this the same as wind height, as defined here for example: [`pvlib.temperature.fuentes`](https://pvlib-python.readthedocs.io/en/v0.11.1/reference/generated/pvlib.temperature.fuentes.html)
If so, maybe we could standardize these and choose one of `wind_height` or `wind_elevation`

cwhanse: The way I read [4], `wind_elevation` here is the height assumed for the wind speed and direction returned by the source. In `fuentes` `wind_height` is the height associated with the input wind speed, and that is used to scale wind to `module_height`. To pass the NASA data into `fuentes`, `wind_height`=`wind_elevation` is correct.

I agree with @RDaxini suggestion to tie these quantities together with the same parameter name.

IoannisSifnaios: Done!

AdamRJensen:
> Also, should there also be a `read_nasa_power` function or not?

I suggest just adding a ``get_nasa_power`` function for now. If there is a need for a read function, we can always add that later.

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,150 @@
...
+        * ``'re'``: renewable energy
+        * ``'sb'``: sustainable buildings
+        * ``'ag'``: agroclimatology
+
+    time_standard: str, default: ``'utc'``
+        Can be either ``'utc'`` or ``'lst'``:
+
+        * Universal Time Coordinated (utc)
+        * Local Solar Time (lst): A 15 Degrees swath that represents solar
+          noon at the middle longitude of the swath
```
AdamRJensen:
```suggestion
    time_standard: str, optional
        Must be one of the below options. The default is ``'utc'``.

        * ``'utc'`` universal time coordinated
        * ``'lst'``: local solar time.  15 degree swath that represents solar
          noon at the middle longitude of the swath
```

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,150 @@
...
+
+    elevation: float, optional
+        The custom site elevation in meters to produce the corrected
+        atmospheric pressure adjusted for elevation.
+    wind_height: float, optional
+        The custom wind height in meters to produce the wind speed adjusted
+        for height. Has to be between 10 and 300 m; see [4]_.
+    wind_surface: str, optional
+        The definable surface type to adjust the wind speed. For a list of the
+        surface types see [4]_.
```
AdamRJensen: Do these have defaults?

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,150 @@
...
+    map_variables: bool, optional, default: True
+        When true, renames columns of the Dataframe to pvlib variable names
+        where applicable. The default is True. See variable
+        :const:`VARIABLE_MAP`.
```
AdamRJensen:
```suggestion
    map_variables: bool, optional
        When true, renames columns of the Dataframe to pvlib variable names
        where applicable. See variable :const:`VARIABLE_MAP`.
        The default is True.
```

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,150 @@
...
+    response = requests.get(URL, params=params)
+    if not response.ok:
+        # response.raise_for_status() does not give a useful error message
+        raise requests.HTTPError(response.json())
+
+    # Parse the data to dataframe
+    data = response.json()
+    hourly_data = data['properties']['parameter']
+    df = pd.DataFrame(hourly_data)
+    df.index = pd.to_datetime(df.index, format='%Y%m%d%H').tz_localize('UTC')
```
AdamRJensen, line 135: Are the returned timestamps always UTC? What about when ``time_standard`` is 'lst'?

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,150 @@
...
+    latitude: float
+        In decimal degrees, north is positive (ISO 19115).
+    longitude: float
+        In decimal degrees, east is positive (ISO 19115).
+    start: datetime like
+        First timestamp of the requested period. If a timezone is not
+        specified, UTC is assumed.
+    end: datetime like
+        Last timestamp of the requested period. If a timezone is not
+        specified, UTC is assumed.
```
AdamRJensen:
```suggestion
    start: datetime like
        First timestamp of the requested period.
    end: datetime like
        Last timestamp of the requested period.
```
No need to convert to UTC when you can only request full days.

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,150 @@
...
+    parameters: str, list
```
AdamRJensen, line 45: To make this easier to use, I suggest adding a list of default parameters. These could include GHI, DHI, DNI, temperature, and wind speed.

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,150 @@
...
+       <https://power.larc.nasa.gov/api/pages/>`_
+    .. [3] `NASA POWER API parameters
+       <https://power.larc.nasa.gov/parameters/>`_
+    .. [4] `NASA POWER corrected wind speed parameters
+       <https://power.larc.nasa.gov/docs/methodology/meteorology/wind/>`_
+    """
+    start = pd.Timestamp(start)
+    end = pd.Timestamp(end)
+    start = start.tz_localize('UTC') if start.tzinfo is None else start
+    end = end.tz_localize('UTC') if end.tzinfo is None else end
```
AdamRJensen:
```suggestion
```
No need for this, since you can only request full days.

pvlib/iotools/nasa_power.py
AdamRJensen: I suggest adding some of the following information
> Wind Elevation values in Meters are required to be between 10m and 300m.
> If you provide a correct wind surface alias please include a surface elevation with the request.

AdamRJensen: Also, add the units
> Site Elevation site-elevation values in Meters.

pvlib/iotools/nasa_power.py
AdamRJensen: I would just remove the option and let users calculate lst themselves. Then mention in the outputs that the timestamps are in UTC.

AdamRJensen (commented):
Initial review

pvlib/iotools/nasa_power.py
IoannisSifnaios: No, they don't have defaults.

pvlib/iotools/nasa_power.py
IoannisSifnaios: There are [336 parameters](https://power.larc.nasa.gov/parameters/) available... Also, temperature and wind speed are given at different heights (2m, 10m, 50m). I guess it could be easy for people to miss the fact that there are many more parameters available if we give some default ones. Isn't the list in the description sufficient?

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -19,7 +19,7 @@ Bug fixes
 
 Enhancements
 ~~~~~~~~~~~~
-
+* Added function :py:func:`pvlib.iotools.get_nasa_power` to retrieve data from NASA POWER. (:pull:`2500`)
```
AdamRJensen:
```suggestion
* Added function :py:func:`~pvlib.iotools.get_nasa_power` to retrieve data from NASA POWER.
  (:pull:`2500`)
```

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,142 @@
+"""Function for reading and retrieving data from NASA POWER."""
+
+import pandas as pd
+import requests
+
+
```
AdamRJensen, line 6:
```suggestion
```

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,142 @@
+"""Function for reading and retrieving data from NASA POWER."""
```
AdamRJensen:
```suggestion
"""Functions for reading and retrieving data from NASA POWER."""
```

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,142 @@
...
+    'ALLSKY_SFC_SW_DNI': 'dni',
+    'CLRSKY_SFC_SW_DWN': 'ghi_clear',
+    'T2M': 'temp_air_2m',
+    'WS2M': 'wind_speed_2m',
+    'WS10M': 'wind_speed_10m',
+}
+
+
+def get_nasa_power(latitude, longitude, start, end, parameters,
+                   time_standard='utc', community='re', elevation=None,
```
AdamRJensen:
```suggestion
                   community='re', elevation=None,
```

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,142 @@
...
+    .. [2] `NASA POWER API
+       <https://power.larc.nasa.gov/api/pages/>`_
+    .. [3] `NASA POWER API parameters
+       <https://power.larc.nasa.gov/parameters/>`_
+    .. [4] `NASA POWER corrected wind speed parameters
+       <https://power.larc.nasa.gov/docs/methodology/meteorology/wind/>`_
+    """
+    start = pd.Timestamp(start)
+    end = pd.Timestamp(end)
+
```
AdamRJensen, line 105:
```suggestion

    # allow the use of pvlib parameter names
    parameter_dict = {v: k for k, v in VARIABLE_MAP.items()}
    parameters = [parameter_dict.get(p, p) for p in parameters]

```

IoannisSifnaios: If we do this, I guess the users could ask for GHI instead of the weird name NASA POWER has. But would it then make sense to give the parameter names in the docstring?

AdamRJensen: I think it would be fine to leave the original names in the docstring. But it would be nice to add a comment to the docstring that the standard pvlib parameter names can also be used.

pvlib/iotools/nasa_power.py
AdamRJensen: Note, there are only 36 parameters available with the API and an hourly frequency.

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,141 @@
...
+import requests
+
+URL = 'https://power.larc.nasa.gov/api/temporal/hourly/point'
+
+VARIABLE_MAP = {
+    'ALLSKY_SFC_SW_DWN': 'ghi',
+    'ALLSKY_SFC_SW_DIFF': 'dhi',
+    'ALLSKY_SFC_SW_DNI': 'dni',
+    'CLRSKY_SFC_SW_DWN': 'ghi_clear',
+    'T2M': 'temp_air_2m',
```
AdamRJensen:
```suggestion
    'T2M': 'temp_air',
```
2 m air temperature is relatively standard, so this should be mapped to ``temp_air``. If users for some reason want a 10 m air temperature, then that's still possible with the T10M parameter.

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,141 @@
...
+    'WS2M': 'wind_speed_2m',
+    'WS10M': 'wind_speed_10m',
+}
+
+
+def get_nasa_power(latitude, longitude, start, end, parameters,
+                   community='re', elevation=None,
+                   wind_height=None, wind_surface=None, map_variables=True):
```
AdamRJensen: The URL should be added as a parameter ``url=URL``

pvlib/iotools/nasa_power.py
AdamRJensen: Anyhow, most pvlib users surely would want the three irradiance components and air temperature and wind speed (10m).

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # Parse the data to dataframe
+    data = response.json()
+    hourly_data = data['properties']['parameter']
+    df = pd.DataFrame(hourly_data)
+    df.index = pd.to_datetime(df.index, format='%Y%m%d%H').tz_localize('UTC')
+
+    # Make metadata dictionary
+    meta = {key: data[key] for key in ['header', 'messages', 'parameters',
+                                       'times', 'type']}
```
AdamRJensen:
```suggestion
    # Create metadata dictionary
    meta = data['header']
    meta['times'] = data['times']
    meta['parameters'] = data['parameters']
```
A bit cleaner

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,153 @@
...
+
+        * ``ALLSKY_SFC_SW_DWN``: Global Horizontal Irradiance (GHI) [Wm⁻²]
+        * ``ALLSKY_SFC_SW_DIFF``: Diffuse Horizontal Irradiance (DHI) [Wm⁻²]
+        * ``ALLSKY_SFC_SW_DNI``: Direct Normal Irradiance (DNI) [Wm⁻²]
+        * ``T2M``: Air temperature at 2 m [C]
+        * ``WS10M``: Wind speed at 10 m [m/s]
+
+    community: str
+        Can be one of the following depending on which parameters are of
+        interest. The default is ``'re'``. Note that in many cases this choice
```
kandersolar:
```suggestion
    community: str, default 're'
        Can be one of the following depending on which parameters are of
        interest. Note that in many cases this choice
```

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,153 @@
...
+        The custom wind height in meters to produce the wind speed adjusted
+        for height. Has to be between 10 and 300 m; see [4]_.
+    wind_surface: str, optional
+        The definable surface type to adjust the wind speed. For a list of the
+        surface types see [4]_. If you provide a wind surface alias please
+        include a site elevation with the request.
+    map_variables: bool, optional
+        When true, renames columns of the Dataframe to pvlib variable names
+        where applicable. See variable :const:`VARIABLE_MAP`.
+        The default is `True`.
```
kandersolar:
```suggestion
    map_variables: bool, default True
        When true, renames columns of the Dataframe to pvlib variable names
        where applicable. See variable :const:`VARIABLE_MAP`.
```

tests/iotools/test_nasa_power.py
```diff
@@ -0,0 +1,88 @@
...
+@pytest.fixture
+def ghi_series(data_index):
+    ghi = [
+        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 50.25, 184.2, 281.55, 368.3, 406.48,
+        386.45, 316.05, 210.1, 109.05, 12.9, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0
+    ]
+    return pd.Series(data=ghi, index=data_index, name='ghi')
+
+
+def test_get_nasa_power(data_index, ghi_series):
```
kandersolar:
```suggestion

@pytest.mark.remote_data
@pytest.mark.flaky(reruns=RERUNS, reruns_delay=RERUNS_DELAY)
def test_get_nasa_power(data_index, ghi_series):
```

All test functions that access internet APIs should have these two decorators :)

And add `from tests.conftest import RERUNS, RERUNS_DELAY` at the top of the file.

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,153 @@
...
+    parameters: str, list
+        List of parameters. The default parameters are mentioned below; for the
+        full list see [3]_. Note that the pvlib naming conventions can also be
+        used.
+
+        * ``ALLSKY_SFC_SW_DWN``: Global Horizontal Irradiance (GHI) [Wm⁻²]
+        * ``ALLSKY_SFC_SW_DIFF``: Diffuse Horizontal Irradiance (DHI) [Wm⁻²]
+        * ``ALLSKY_SFC_SW_DNI``: Direct Normal Irradiance (DNI) [Wm⁻²]
+        * ``T2M``: Air temperature at 2 m [C]
+        * ``WS10M``: Wind speed at 10 m [m/s]
```
kandersolar: I suggest using pvlib names here and in the default parameter list.  As a user, I don't want to have to mentally translate `ALLSKY_SFC_SW_DWN` into `ghi` when reading these docs.

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,153 @@
...
+    if not response.ok:
+        # response.raise_for_status() does not give a useful error message
+        raise requests.HTTPError(response.json())
+
+    # Parse the data to dataframe
+    data = response.json()
+    hourly_data = data['properties']['parameter']
+    df = pd.DataFrame(hourly_data)
+    df.index = pd.to_datetime(df.index, format='%Y%m%d%H').tz_localize('UTC')
+    df = df.replace(-999, np.nan)
```
kandersolar: perhaps worth using `meta['fill_value']` instead of hardcoding -999, in case they change it someday?

kandersolar (commented):
A few comments, overall looks pretty good to me!  Seems like the API is a little slow but not terrible: 16 seconds for a year of data.

pvlib/iotools/nasa_power.py
cwhanse: @kandersolar to clarify: are you suggesting that the pvlib function accept values like `'ghi'` in place of `'ALLSKY_SFC_SW_DIFF'`?

kandersolar: The pvlib function currently accepts both of those and I am not suggesting changing that.  The suggestion is for the default values to use pvlib names instead of NASA POWER names, to (IMHO) make the docstring a little clearer.

IoannisSifnaios:
When running pytest locally, it says that the first test is skipped. Any ideas why this is (and also, is it a problem)?

echedey-ls:
> When running pytest locally, it says that the first test is skipped. Any ideas why this is (and also, is it a problem)?

@IoannisSifnaios that's due to the flags that make the iotools tests only run when explicitly requested. The pytest command you must run is in [`workflows/pytest-remote-data.yml`](https://github.com/pvlib/pvlib-python/blob/main/.github/workflows/pytest-remote-data.yml):

`        run: pytest tests/iotools --cov=./ --cov-report=xml --remote-data`

cov parameters are just for the coverage file, you don't need em.

IoannisSifnaios:
> > When running pytest locally, it says that the first test is skipped. Any ideas why this is (and also, is it a problem)?
> 
> @IoannisSifnaios that's due to the flags that make the iotools tests only run when explicitly requested. The pytest command you must run is in [`workflows/pytest-remote-data.yml`](https://github.com/pvlib/pvlib-python/blob/main/.github/workflows/pytest-remote-data.yml):
> 
> ` run: pytest tests/iotools --cov=./ --cov-report=xml --remote-data`
> 
> cov parameters are just for the coverage file, you don't need em.

I guess based on the workflow description, there should also be a `remote_data` label added, right?

echedey-ls:
> there should also be a remote_data label added, right?

Yup, that's right!

Edit: in fact, see it's one of the instructions (the last one) on the PR template

tests/iotools/test_nasa_power.py
```diff
@@ -0,0 +1,92 @@
...
+    assert meta['time_standard'] == 'UTC'
+    assert meta['title'] == 'NASA/POWER Source Native Resolution Hourly Data'
+    # Assert that the index is parsed correctly
+    pd.testing.assert_index_equal(data.index, data_index)
+    # Test one column
+    pd.testing.assert_series_equal(data['ALLSKY_SFC_SW_DWN'], ghi_series,
+                                   check_freq=False, check_names=False)
+
+
+def test_get_nasa_power_pvlib_params_naming(data_index, ghi_series):
```
kandersolar, line 50:
```suggestion
@pytest.mark.remote_data
@pytest.mark.flaky(reruns=RERUNS, reruns_delay=RERUNS_DELAY)
def test_get_nasa_power_pvlib_params_naming(data_index, ghi_series):
```

tests/iotools/test_nasa_power.py
```diff
@@ -0,0 +1,92 @@
...
+                                              end=data_index[-1],
+                                              parameters=['ghi'])
+    # Assert that the index is parsed correctly
+    pd.testing.assert_index_equal(data.index, data_index)
+    # Test one column
+    pd.testing.assert_series_equal(data['ghi'], ghi_series,
+                                   check_freq=False)
+
+
+def test_get_nasa_power_map_variables(data_index):
```
kandersolar, line 65:
```suggestion
@pytest.mark.remote_data
@pytest.mark.flaky(reruns=RERUNS, reruns_delay=RERUNS_DELAY)
def test_get_nasa_power_map_variables(data_index):
```

tests/iotools/test_nasa_power.py
```diff
@@ -0,0 +1,92 @@
...
+                                              end=data_index[-1])
+    mapped_column_names = ['ghi', 'dni', 'dhi', 'temp_air', 'wind_speed']
+    for c in mapped_column_names:
+        assert c in data.columns
+    assert meta['latitude'] == 44.76
+    assert meta['longitude'] == 7.64
+    assert meta['altitude'] == 705.88
+
+
+def test_get_nasa_power_wrong_parameter_name(data_index):
```
kandersolar, line 81:
```suggestion
@pytest.mark.remote_data
@pytest.mark.flaky(reruns=RERUNS, reruns_delay=RERUNS_DELAY)
def test_get_nasa_power_wrong_parameter_name(data_index):
```

tests/iotools/test_nasa_power.py
```diff
@@ -0,0 +1,92 @@
...
+    # Test if HTTPError is raised if a wrong parameter name is asked
+    with pytest.raises(HTTPError, match=r"ALLSKY_SFC_SW_DLN"):
+        pvlib.iotools.get_nasa_power(latitude=44.76,
+                                     longitude=7.64,
+                                     start=data_index[0],
+                                     end=data_index[-1],
+                                     parameters=['ALLSKY_SFC_SW_DLN'])
+
+
+def test_get_nasa_power_duplicate_parameter_name(data_index):
```
kandersolar, line 93:
```suggestion
@pytest.mark.remote_data
@pytest.mark.flaky(reruns=RERUNS, reruns_delay=RERUNS_DELAY)
def test_get_nasa_power_duplicate_parameter_name(data_index):
```

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,153 @@
...
+        * Wind speed at 10 m [m/s]
+
+    community: str, default 're'
+        Can be one of the following depending on which parameters are of
+        interest. Note that in many cases this choice
+        might affect the units of the parameter.
+
+        * ``'re'``: renewable energy
+        * ``'sb'``: sustainable buildings
+        * ``'ag'``: agroclimatology
```
kandersolar, line 63: Does this parameter do anything other than change some units?  If so, it might be better to keep the function signature simple and just hardcode `re`.  Just a thought!

kandersolar (approved):

pvlib/iotools/nasa_power.py
IoannisSifnaios: It gives you access to some other parameters in some cases. For example, irradiance is in all 3 of them but some climate parameters (not necessarily relevant to pvlib though) might be in only one of them.

pvlib/iotools/nasa_power.py
```diff
@@ -0,0 +1,153 @@
...
+    'ALLSKY_SFC_SW_DNI': 'dni',
+    'CLRSKY_SFC_SW_DWN': 'ghi_clear',
+    'T2M': 'temp_air',
+    'WS2M': 'wind_speed_2m',
+    'WS10M': 'wind_speed',
+}
+
+
+def get_nasa_power(latitude, longitude, start, end,
+                   parameters=DEFAULT_PARAMETERS, community='re', url=URL,
```
AdamRJensen:
```suggestion
                   parameters=DEFAULT_PARAMETERS, *, community='re', url=URL,
```
I suggest adding an asterisk which forces the remaining parameters to be specified as keywords only (not positional). The motivation is that this makes it easier to add and deprecate parameters.
@kandersolar let me hear your opinion on this.

AdamRJensen: The url should be the last parameter (standard iotools practice)

docs/sphinx/source/whatsnew/v0.13.1.rst
```diff
@@ -19,7 +19,8 @@ Bug fixes
 
 Enhancements
 ~~~~~~~~~~~~
-
+* Added function :py:func:`~pvlib.iotools.get_nasa_power` to retrieve data from NASA POWER.
```
echedey-ls:
```suggestion
* Add :py:func:`pvlib.iotools.get_nasa_power` to retrieve data from NASA POWER free API.
```
Use imperative mood for the changelogs (latest trend in pvlib, e.g. recommended by https://common-changelog.org/)
I suggest keeping the iotools namespace visible, I think it is valuable in the context of new functionality. And stating it's a free API - for me, that guarantees it's worth considering 🤑

echedey-ls (approved):
Minor nitpick in Whatsnew. LGTM 💯

pvlib/iotools/nasa_power.py
kandersolar: I have been reluctant to introduce `*` (and `/`) into function signatures:
1. I have not encountered it much in role model packages like scipy (though perhaps it is there and I do not notice) 
2. It doesn't seem to me that we should force a code style opinion onto users
3. It does not come with any obvious benefit to pvlib's maintenance

However, item 3 is partly rebutted by the point about deprecation, which is a good one that I had not considered.

I don't feel strongly about it.  To me it seems reasonable to use `*` in this case, and if others think it's a good idea to start using it more in pvlib then I won't object, although I think in some cases there could be disagreement about what position in the parameters list it should take.

AdamRJensen: Alright let's try it out here. It's easy remove but adding it later would be a breaking change. I'm partly advocating for it because I'm toying with the idea of adding a parameter for allowing setting the index to be a time interval/period index.

kandersolar:
Thanks @IoannisSifnaios!

Issue #1684, opened 2023-03-06
is there still a need for irradiance functions to return OrderDict since Python-3.7 is minimum version?


mikofski:
**Is your feature request related to a problem? Please describe.**
Python-3.7 is the minimum python. I believe since Python-3ish, all dicts are ordered by default. So do we still need to use `OrderedDict` for returns in `irradiance.py`?

**Describe the solution you'd like**
If no longer necessary, remove import of `OrderedDict` and replace all calls with simple dictionary.

**Describe alternatives you've considered**
Leave it as is. It seems to function now, and there may be an added benefit b/c an `OrderedDict` can be compared to another mapping that doesn't respect order?

**Additional context**
related to #1455 and raised in https://github.com/pvlib/pvlib-python/pull/1179#discussion_r1121847498

AdamRJensen:
This [article](https://realpython.com/python-ordereddict/) describes a few reasons why one might want to continue using OrderedDicts:

> **1 - Intent signaling**: If you use OrderedDict over dict, then your code makes it clear that the order of items in the dictionary is important. You’re clearly communicating that your code needs or relies on the order of items in the underlying dictionary.
> **2 - Control over the order of items**: If you need to rearrange or reorder the items in a dictionary, then you can use [.move_to_end()](https://docs.python.org/3/library/collections.html#collections.OrderedDict.move_to_end) and also the enhanced variation of [.popitem()](https://docs.python.org/3/library/collections.html#collections.OrderedDict.popitem).
> **3 - Equality test behavior**: If your code compares dictionaries for equality, and the order of items is important in that comparison, then OrderedDict is the right choice.
> **Backward compatibility** Relying on regular dict objects to preserve the order of it

Though I still think that I favor the simplicity of getting rid of OrderedDicts.

markcampanelli:
I’d like to point out that one should never rely on the _implementation_ of `dict` being ordered (as it is in newer CPython versions, IIRC for better performance). Thus, if a dictionary must be ordered for some good reason, then keep it declared as an `OrderedDict`.

kandersolar:
As I understand it, dicts being ordered according to insertion was a CPython implementation detail in 3.6 that got upgraded to an official language feature in 3.7 onward, so it should be reliable for our purposes :)

markcampanelli:
[...]

Issue #434, opened 2018-02-22
Return sky diffuse and components as dataframe in irradiance.perez


cwhanse:
`irradiance.perez` can return isotropic, circumsolar and horizon components of sky diffuse irradiance if `return_components = True`.  Currently, if `return_components` then results are returned as a tuple `(sky_diffuse, diffuse_components)` with the elements being a pandas.Series and pandas.Dataframe, respectively.

I propose that we always return a Dataframe with column `sky_diffuse` and optionally columns for `isotropic, circumsolar, horizon`.  This avoids having the output of the perez model of a different type (tuple instead of pandas object) and having to unpack the tuple when diffuse_components are requested.

wholmgren:
This is an interesting idea. I'm concerned about maintaining API compatibility across all of the transposition models. I think the default behavior of all of the models should be the same, if possible. So, I'm reluctant to default to returning a one-column DataFrame from ``perez`` while the others return Series. 

We could make similar changes to all of the models. I don't know if that would be useful or busy work. What would the keys look like for each function? For example

``isotropic`` returns ``sky_diffuse, isotropic``?
``haydavies`` returns ``sky_diffuse, isotropic, circumsolar``?
``perez`` returns ``sky_diffuse, isotropic, circumsolar, horizon``?

cwhanse:
Good point about compatibility.  We could return a Series if `return_components = False` rather than a one column data frame.  The hiccup for me is dealing with the tuple instead of a pandas object.

We could propagate the component capability into the other sky diffuse models, although perez is likely to be the most frequently used for this purpose.

wholmgren:
Sure, that sounds like a quick fix then. 

For reference, the ``return_components`` feature was added in #259. It does not appear that we discussed the idea of returning a single DataFrame with the total and the components.

wholmgren:
@cwhanse I suggest that we limit the scope of this particular issue to ``irradiance.perez`` and create new issues if want to expand this feature to new transposition models.

PR #2527, opened 2025-08-03
Adapts irradiance.isotropic to return_components framework.

### LINKED ISSUE #1553: Consider allowing `return_components` in all transposition model functions
Access to sky diffuse irradiance in component form (typically, isotropic + circumsolar + horizon) is needed for diffuse IAM and shading/view factor loss modeling.  `pvlib.irradiance.perez` has the ability to return separate in-plane sky diffuse components in addition to the combined quantity `sky_diffuse`, but currently `perez` is the only transposition function with that ability. Adding this `return_components` option to more models was briefly discussed in #434 but it was decided to limit it to `perez` at that time.  I'd like to pick up that discussion in the context of pvlib's current and future expanded capabilities regarding diffuse IAM and shade modeling.

As I understand them, the `isotropic`, `haydavies`, `klucher`, and `reindl` models can all be considered as fitting naturally into the same `isotropic`/`circumsolar`/`horizon` organization as `perez` (though the simpler models might have one or two of those components equal to zero).  I'm not sure about the `king` model.  So, with the possible exception of `king`, it seems that there is no conceptual barrier to extending the other models with a `return_components` option to return the same components as `perez` does.

In addition to making the function level of pvlib more useful on its own, a uniform interface for diffuse components across all (or most) transposition functions would enable improvements to the higher-level constructs `ModelChain` and `infinite_sheds`:
- `infinite_sheds` could be generalized to make use of anisotropic transposition models (#1551)
- `ModelChain` could apply the same IAM to the `circumsolar` component that it does to `poa_direct`, as well as make use of other component-specific IAMs from e.g. `marion_diffuse`

One concern about extending `return_components` to all models in #434 was the API (returning a DataFrame, e.g.).  I think I'm okay with what is currently used in `perez`, but bring it up for discussion here.

ramaroesilva:
- [ ] Closes #1553
 - [ ] Tests added

### DIFF
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -587,7 +587,7 @@ def get_ground_diffuse(surface_tilt, ghi, albedo=.25, surface_type=None):
     return diffuse_irrad
 
 
-def isotropic(surface_tilt, dhi):
+def isotropic(surface_tilt, dhi, return_components=False):
     r'''
     Determine diffuse irradiance from the sky on a tilted surface using
     the isotropic sky model.
@@ -613,10 +613,23 @@ def isotropic(surface_tilt, dhi):
     dhi : numeric
         Diffuse horizontal irradiance. [Wm⁻²] DHI must be >=0.
 
+    return_components : bool, default False
+        Flag used to decide whether to return the calculated diffuse components
+        or not. If `False`, ``sky_diffuse`` is returned. If `True`,
+        ``diffuse_components`` is returned.
+
     Returns
     -------
-    diffuse : numeric
-        The sky diffuse component of the solar radiation.
+    poa_sky_diffuse : numeric
+        The sky diffuse component of the solar radiation on a tilted
+        surface.
+
+    diffuse_components : OrderedDict (array input) or DataFrame (Series input)
+        Keys/columns are:
+            * sky_diffuse: Total sky diffuse
+            * isotropic
+            * circumsolar
+            * horizon
 
     References
     ----------
@@ -630,9 +643,20 @@ def isotropic(surface_tilt, dhi):
        Energy vol. 201. pp. 8-12
        :doi:`10.1016/j.solener.2020.02.067`
     '''
-    sky_diffuse = dhi * (1 + tools.cosd(surface_tilt)) * 0.5
+    poa_sky_diffuse = dhi * (1 + tools.cosd(surface_tilt)) * 0.5
 
-    return sky_diffuse
+    if return_components:
+        diffuse_components = OrderedDict()
+        diffuse_components['poa_sky_diffuse'] = poa_sky_diffuse
+
+        # Calculate the different components
+        diffuse_components['poa_isotropic'] = poa_sky_diffuse
+        diffuse_components['poa_circumsolar'] = 0
+        diffuse_components['poa_horizon'] = 0
+
+        return diffuse_components
+    else:
+        return poa_sky_diffuse
 
 
 def klucher(surface_tilt, surface_azimuth, dhi, ghi, solar_zenith,
@@ -782,8 +806,9 @@ def haydavies(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
         or supply ``projection_ratio``.
 
     return_components : bool, default `False`
-        If `False`, ``sky_diffuse`` is returned.
-        If `True`, ``diffuse_components`` is returned.
+       Flag used to decide whether to return the calculated diffuse components
+       or not. If `False`, ``sky_diffuse`` is returned. If `True`,
+       ``diffuse_components`` is returned.
 
     Returns
     --------
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -615,27 +615,17 @@ def isotropic(surface_tilt, dhi, return_components=False):
 
     return_components : bool, default False
         Flag used to decide whether to return the calculated diffuse components
-        or not. If `False`, ``poa_sky_diffuse`` is returned. If `True`,
-        ``diffuse_components`` is returned.
+        or not. If `True`, the resulting dataframe will include three more
+        columns. If not, only ``poa_sky_diffuse`` column is returned.
 
     Returns
     -------
-    poa_sky_diffuse : numeric
+    poa_sky_diffuse : DataFrame
         The sky diffuse component of the solar radiation on a tilted
-        surface.
-
-    diffuse_components : dict (array input) or DataFrame (Series input)
-        Keys/columns are:
-            * sky_diffuse (the sum of the components below)
-            * isotropic
-            * circumsolar
-            * horizon
-            * poa_sky_diffuse (the sum of the components below)
-            * poa_isotropic
-            * poa_circumsolar
-            * poa_horizon
-        The first four elements will be deprecated in v0.14.0 and are kept
-        to avoit breaking changes.
+        surface. Columns provided controlled by ``return_components`` argument.
+          If `False`, ``total`` is returned.
+          If `True`, ``isotropic``, ``circumsolar`` and ``horizon``.
+        Index is either inherited from dhi or default.
 
     References
     ----------
@@ -650,30 +640,29 @@ def isotropic(surface_tilt, dhi, return_components=False):
        :doi:`10.1016/j.solener.2020.02.067`
     '''
 
-    poa_sky_diffuse = dhi * (1 + tools.cosd(surface_tilt)) * 0.5    
-        
-    if return_components:
-        is_pandas = isinstance(dhi, (pd.DataFrame,pd.Series))
-        
-        if is_pandas:
-            poa_sky_diffuse = poa_sky_diffuse.values    
+    poa_isotropic = dhi * (1 + tools.cosd(surface_tilt)) * 0.5
 
-        diffuse_components = {
-            'poa_sky_diffuse': poa_sky_diffuse,
-            'poa_isotropic': poa_sky_diffuse,
-            'poa_circumsolar': 0,
-            'poa_horizon': 0,
-            }
-
-        poa_sky_diffuse = pd.DataFrame(poa_sky_diffuse)        
-        if is_pandas:
-            # keeps original index
-            poa_sky_diffuse.index = dhi.index
-        
-        return diffuse_components
+    # if pd.DataFrame or Series handles first output as numerical
+    is_pandas = isinstance(dhi, (pd.DataFrame,pd.Series))
+    if is_pandas:
+        poa_isotropic = poa_isotropic.values
     
+    poa_sky_diffuse = poa_isotropic
+
+    if not return_components:
+        poa_sky_diffuse = {'total': poa_sky_diffuse}
     else:
-        return poa_sky_diffuse
+        poa_sky_diffuse = {'isotropic': poa_isotropic,
+                           'circumsolar': 0,
+                           'horizon': 0}
+
+    poa_sky_diffuse = pd.DataFrame(poa_sky_diffuse)
+
+    # if dhi was pandas, out_df inherits index
+    if is_pandas:
+        poa_sky_diffuse.index = dhi.index
+
+    return poa_sky_diffuse
 
 
 def klucher(surface_tilt, surface_azimuth, dhi, ghi, solar_zenith,
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -653,7 +653,7 @@ def isotropic(surface_tilt, dhi, return_components=False):
 
     if return_components:
         diffuse_components = dict()
-        
+
         # original formatting (to be deprecated in v0.14.0)
         diffuse_components['sky_diffuse'] = poa_sky_diffuse # total
         diffuse_components['isotropic'] = poa_sky_diffuse
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -649,29 +649,29 @@ def isotropic(surface_tilt, dhi, return_components=False):
        Energy vol. 201. pp. 8-12
        :doi:`10.1016/j.solener.2020.02.067`
     '''
-    poa_sky_diffuse = dhi * (1 + tools.cosd(surface_tilt)) * 0.5
 
+    poa_sky_diffuse = dhi * (1 + tools.cosd(surface_tilt)) * 0.5    
+        
     if return_components:
-        diffuse_components = dict()
+        is_pandas = isinstance(dhi, (pd.DataFrame,pd.Series))
+        
+        if is_pandas:
+            poa_sky_diffuse = poa_sky_diffuse.values    
 
-        # original formatting (to be deprecated in v0.14.0)
-        diffuse_components['sky_diffuse'] = poa_sky_diffuse # total
-        diffuse_components['isotropic'] = poa_sky_diffuse
-        diffuse_components['circumsolar'] = 0
-        diffuse_components['horizon'] = 0
+        diffuse_components = {
+            'poa_sky_diffuse': poa_sky_diffuse,
+            'poa_isotropic': poa_sky_diffuse,
+            'poa_circumsolar': 0,
+            'poa_horizon': 0,
+            }
 
-        # new formatting
-        diffuse_components['poa_sky_diffuse'] = poa_sky_diffuse
-        diffuse_components['poa_isotropic'] = poa_sky_diffuse
-        diffuse_components['poa_circumsolar'] = 0
-        diffuse_components['poa_horizon'] = 0
-
-        if isinstance(poa_sky_diffuse, pd.Series):
-            # follows `perez` worfklow
-            # shouldn't it include an argument `index=dhi.index`?
-            diffuse_components = pd.DataFrame(diffuse_components)
-
+        poa_sky_diffuse = pd.DataFrame(poa_sky_diffuse)        
+        if is_pandas:
+            # keeps original index
+            poa_sky_diffuse.index = dhi.index
+        
         return diffuse_components
+    
     else:
         return poa_sky_diffuse
 
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -615,7 +615,7 @@ def isotropic(surface_tilt, dhi, return_components=False):
 
     return_components : bool, default False
         Flag used to decide whether to return the calculated diffuse components
-        or not. If `False`, ``sky_diffuse`` is returned. If `True`,
+        or not. If `False`, ``poa_sky_diffuse`` is returned. If `True`,
         ``diffuse_components`` is returned.
 
     Returns
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -624,12 +624,18 @@ def isotropic(surface_tilt, dhi, return_components=False):
         The sky diffuse component of the solar radiation on a tilted
         surface.
 
-    diffuse_components : OrderedDict (array input) or DataFrame (Series input)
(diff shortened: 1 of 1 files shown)


ramaroesilva:
For now, I only updated `irradiance.isotropic` as a testbed for the proposed changes.

Made use of the content from `perez` and `haydavies`docstrings, which were not fully consistent. Also, added the `poa_` prefix to the outputs as suggested by @markcampanelli [here](https://github.com/pvlib/pvlib-python/issues/1553#issuecomment-1931254530)

If this seems good to everyone, I can replicate this for all functions (adding the `return_components` and making documentation more consistent across fuctions, whenever each is needed).

pvlib/irradiance.py
```diff
@@ -630,9 +643,20 @@ def isotropic(surface_tilt, dhi):
        Energy vol. 201. pp. 8-12
        :doi:`10.1016/j.solener.2020.02.067`
     '''
-    sky_diffuse = dhi * (1 + tools.cosd(surface_tilt)) * 0.5
+    poa_sky_diffuse = dhi * (1 + tools.cosd(surface_tilt)) * 0.5
 
-    return sky_diffuse
+    if return_components:
+        diffuse_components = OrderedDict()
```
kandersolar: Minor note: no need to use OrderedDict in new code (https://github.com/pvlib/pvlib-python/issues/1684)

pvlib/irradiance.py
```diff
@@ -613,10 +613,23 @@ def isotropic(surface_tilt, dhi):
     dhi : numeric
         Diffuse horizontal irradiance. [Wm⁻²] DHI must be >=0.
 
+    return_components : bool, default False
+        Flag used to decide whether to return the calculated diffuse components
+        or not. If `False`, ``sky_diffuse`` is returned. If `True`,
```
kandersolar:
```suggestion
        or not. If `False`, ``poa_sky_diffuse`` is returned. If `True`,
```
Assuming this is in reference to the names in the `Returns` section

pvlib/irradiance.py
```diff
@@ -613,10 +613,23 @@ def isotropic(surface_tilt, dhi):
...
+    poa_sky_diffuse : numeric
+        The sky diffuse component of the solar radiation on a tilted
+        surface.
+
+    diffuse_components : OrderedDict (array input) or DataFrame (Series input)
+        Keys/columns are:
+            * sky_diffuse: Total sky diffuse
+            * isotropic
+            * circumsolar
+            * horizon
```
kandersolar: Update with `poa_` prefix

kandersolar (commented):
Thanks @ramaroesilva!  I agree it makes sense to trial it on one function first.  Let's wait to do the others until #2529 has some more discussion.

In the meantime, some minor comments below.

ramaroesilva:
Once more people give feedback on the documentation structuring of this example, I will continue developing the other transposition functions taking into account the `OrderedDict` comment. All `poa_` related changes will move to a separate PR.

pvlib/irradiance.py
```diff
@@ -630,9 +643,20 @@ def isotropic(surface_tilt, dhi):
...
+        diffuse_components = OrderedDict()
+        diffuse_components['poa_sky_diffuse'] = poa_sky_diffuse
+
+        # Calculate the different components
+        diffuse_components['poa_isotropic'] = poa_sky_diffuse
+        diffuse_components['poa_circumsolar'] = 0
+        diffuse_components['poa_horizon'] = 0
+
+        return diffuse_components
+    else:
```
cwhanse, line 654: I would prefer to return both `poa_sky_diffuse` and `diffuse_components`, if `return_componets` is True.

ramaroesilva: Sounds good to me. This "one or the other" is what is implemented in `perez` at the moment.

For clarity, do you mean returning two dictionaries, which makes the number of outputs depend on `return_components`, or having still one dictionary buth with an additional `sky_diffuse` key?

cwhanse: I'm thinking a tuple with two dicts, when `return_components` is True.

ramaroesilva: @cwhanse I just noticed that in my latest commit I included in the `diffuse_components` dict the `sky_diffuse` as one of the keys. Would this - having a single-dict with components and their sum - make sense?

If instead we have two dictionaries as originally proposed, the one having the `sky_diffuse` will have a single key... How would we call the variable and the key without being repetitive? `total_diffuse["sky"]`?

cwhanse: Apologies, I misspoke. Return a tuple `(sky_diffuse, components)` where `sky_diffuse` is the current `numeric` (Series, array, float) and `components` is a dict.

cwhanse: Edited: returning a tuple requires the user to change from

```
sky_diffuse = isotropic(...)
```

to

```
sky_diffuse, _ = isotropic(...)
```

I was (wrongly) thinking that only the first element of the tuple would be assigned to `sky_diffuse`. Not thinking clearly.

So now I'm not so sure that returning a tuple is a good idea. Still, it seems better than requiring the user to change to

```
out = isotropic(...)
sky_diffuse = out['sky_diffuse']
````

Comment (ramaroesilva) [reply pvlib/irradiance.py:658]: Thanks for clarifying and thinking things through. This makes me think that since any of the solutions changes the user handling of functions, for now I should replicate the `perez`modus operandi (targeting v0.13.1) and address your point separately for v0.14 (in line with what was discussed in #2529).

If agreed, you can raise an issue like @kandersolar did in #2529.

Comment (ramaroesilva) [reply pvlib/irradiance.py:654]: Thanks for clarifying and thinking things through. 

Going back to the start:
> I would prefer to return both `poa_sky_diffuse` and `diffuse_components`, if `return_componets` is True.

I just remembered that in my latest commit the `sky_diffuse` is included within `diffuse_components`, following the workflow used in `perez` and `haydavies`. So what we are discussing here is in fact if we should change the current modus operandi of transposition models with `return_components`.

While what you propose  is v0.13.1-friendly for the models not yet returning `return_components`, it would break previous code for those that already do. Having this said, I would suggest moving this specific topic to a separate issue and raise some discussion aiming for v0.14.

What do you think? If agreed, can you can raise a new issue like @kandersolar did in #2529?

Comment (adriesse) [reply pvlib/irradiance.py:654]: > I just remembered that in my latest commit the `sky_diffuse` is included within `diffuse_components`, following the workflow used in `perez` and `haydavies`. So what we are discussing here is in fact if we should change the current modus operandi of transposition models with `return_components`.

All the irradiance component sets I looked at include both components and sums of components.  I think the cleanest way would be (would have been) to include only the lowest level and leave the summation up to the user.

File: pvlib/irradiance.py [hunk 5]
```diff
@@ -630,9 +649,31 @@ def isotropic(surface_tilt, dhi):
...
+        # new formatting
+        diffuse_components['poa_sky_diffuse'] = poa_sky_diffuse
+        diffuse_components['poa_isotropic'] = poa_sky_diffuse
+        diffuse_components['poa_circumsolar'] = 0
+        diffuse_components['poa_horizon'] = 0
+
+        if isinstance(poa_sky_diffuse, pd.Series):
+            # follows `perez` worfklow
+            # shouldn't it include an argument `index=dhi.index`?
+            diffuse_components = pd.DataFrame(diffuse_components)
```
ramaroesilva: Any comments about my question here? This code is used in `perez`.

cwhanse: Yes, should preserve the index from `dhi`

ramaroesilva: Would adding this index bit to the diffuse-component-including `perez` and `haydavies` functions be considered a breaking change?

cwhanse: I don't think that's a breaking change. pvlib strives that a function returns the same data type as the inputs. That's one reason we test functions with different input types. If those functions are returning e.g. Array for Series input, I'd regard that as a bug that needs fixing.

In `perez` the index wasn't explicitly set because the upstream calculations resulted in `sky_diffuse` being a Series. I think it's better to be explicit, just in case the calculation uses some numpy function that casts Series to Array.

ramaroesilva:
Sorry for letting this get stale, work got hectic.

Would like to follow up on @kandersolar initiative who is already including the `poa_` prefix in the outputs in #2627 and try to push this PR (after expanding it to other transposition functions) for v0.14. Now we have the freedom of being allowed to introduce breaking changes.

Previous points made:
- @cwhanse would prefer including both `poa_sky_diffuse` and `diffuse_components` if return_components is True.
- @adriesse suggests  to include only the lowest level and leave the summation up to the user.

Within a modelchain, in theory, the optical losses would use the components, whereas the module temperature most often uses the non-effective GTI. If only the components are kept, I guess that a summation would be needed somewhere within the modelchain.

Keeping both the sum and the components seems easier, otherwise we would need to have the code adapt to the user choice regarding `return_components`.

ramaroesilva:
If we end up deciding to keep both sum and components, I was thinking if a single dict/DataFrame variable should be produced whether `return_components` is `True` or `False` (i.e., adding more keys/columns to the same base structure).

Pros:
- keeps same function calling in both cases (addressing one of @cwhanse points in this discussion)
- avoids duplicating indices in case of DataFrame

Cons:
- would need to adjust functions in `modelchain` (and `get_total_irradiance` maybe) that were receiving either the sum and components
- would be sending unnecessary data when other functions getting this input ask only for sum/components (increased overhead?)

Could also call the variable as `poa_sky_diffuse` and have the sum allocate to a `total` key.

markcampanelli:
@ramaroesilva Here's my idea, which probably breaks too many things. Immutability opens up caching the sum, for a significant speedup.

```python
"""
Example of a frozen dataclass for PAO diffuse components using cache for components sum.
"""

from dataclasses import dataclass
import functools
import typing

import numpy.typing


@dataclass(frozen=True)
class PoaDiffuseComponents:
    """Contanier for POA diffuse irradiance components."""

    poa_isotropic: numpy.typing.NDArray[numpy.floating[typing.Any]]
    poa_circumsolar: numpy.typing.NDArray[numpy.floating[typing.Any]]
    poa_horizon: numpy.typing.NDArray[numpy.floating[typing.Any]]

    @functools.cached_property
    def poa_sky_diffuse(self) -> numpy.typing.NDArray[numpy.floating[typing.Any]]:
        """Return sum of all POA diffuse components."""

        return self.poa_isotropic + self.poa_circumsolar + self.poa_horizon

    @property
    def isotropic(self) -> numpy.typing.NDArray[numpy.floating[typing.Any]]:
        """Return deprecated POA isotropic diffuse-component."""

        return self.poa_isotropic

    @property
    def circumsolar(self) -> numpy.typing.NDArray[numpy.floating[typing.Any]]:
        """Return deprecated POA circumsolar-diffuse component."""

        return self.poa_circumsolar

    @property
    def horizon(self) -> numpy.typing.NDArray[numpy.floating[typing.Any]]:
        """Return deprecated POA horizon-diffuse component."""

        return self.poa_horizon

    @property
    def sky_diffuse(self) -> numpy.typing.NDArray[numpy.floating[typing.Any]]:
        """Return deprecated sum of all POA diffuse components."""

        return self.poa_sky_diffuse


if __name__ == "__main__":

    import time

    rng = numpy.random.default_rng()

    poa_diffuse_components = PoaDiffuseComponents(
        poa_isotropic=rng.uniform(low=0.0, high=1100.0, size=(8760,)),
        poa_circumsolar=rng.uniform(low=0.0, high=1100.0, size=(8760,)),
        poa_horizon=rng.uniform(low=0.0, high=1100.0, size=(8760,)),
    )

    start_time_ns = time.perf_counter_ns()
    poa_diffuse_components.poa_sky_diffuse
    end_time_ns = time.perf_counter_ns()
    print(f"poa_sky_diffuse uncached time: {end_time_ns - start_time_ns} ns")

    start_time_ns = time.perf_counter_ns()
    poa_diffuse_components.poa_sky_diffuse
    end_time_ns = time.perf_counter_ns()
    print(f"poa_sky_diffuse cached time: {end_time_ns - start_time_ns} ns")
```

Example output:
```console
poa_sky_diffuse uncached time: 24667 ns
poa_sky_diffuse cached time: 250 ns
```

kandersolar:
> I was thinking if a single dict/DataFrame variable should be produced whether return_components is True or False (i.e., adding more keys/columns to the same base structure).

For the sake of argument, if a dict/DataFrame is being returned no matter what, I'd suggest we always return the components so that we can simplify the API by getting rid of the `return_components` parameter.  The extra memory would be minor, I expect.  

Some relevant discussion here: https://github.com/pvlib/pvlib-python/issues/434

I'm not sure what I think, but I lean towards keeping the current behavior to minimize disruption.

adriesse:
> * @adriesse suggests  to include only the lowest level and leave the summation up to the user.

Thanks for keeping track of my suggestion.  I think it would be cleaner that way, and summation could be done trivially without knowing or using any of the column names.

ramaroesilva:
@kandersolar, I think I wrapped up a nice template with isotropic.

Profitted from the discussion in #565 led by @cwhanse and made the outputs always being a DataFrame, with columns depending on `return_components`:
- if `True`, three columns: `circumsolar`, `isotropic`, `horizon`.
- if `False`, one column: `total` (following @adriesse suggestion)

Thought of the `total` designation to avoid repeating name of variable and with `poa_` already in the variable name, I thought I could dismiss it in the columns.

ramaroesilva:
Already implemented the same structure for `haydavies` and `reindl` (yet to be pushed) and will move next to recheck the `perez` ones.

However, I antecipate `klucher` being a bit tricky since it considers multiplicative terms rather than additive. If we expand the expression there will be four terms: isotropic, circumsolar, horizon, and a last term representing a sort of interaction between circumsolar and horizon... So, how to make this into three valueso only? Splitting the interaction term half-half? Seems a bit arbitrary...

imagem

ramaroesilva:
@kandersolar @cwhanse @adriesse, now that the v0.14.0 boat has sailed, I reflected a bit on the "no breaking changes" constraint that is posed for v0.14.1.

in this case, having an added `return_components` defaulting to `False` should be enough, no? `False` mimics v0.14.0, `True` adopts new option.

for v0.15.0, if everyone agrees, would like to have all transposition functions outputing pd.DataFrame, with:
- a single "poa_sky_total" (or alternative naming) column, if `return_components` is `False`.
- columns "poa_sky_isotropic", "poa_horizon", "poa_circumsolar" (or alternative namings), if `return_components` is `True`.

ramaroesilva:
still blocked with this point. besides the already tagged ones, mr @AdamRJensen any take on this?

> Already implemented the same structure for `haydavies` and `reindl` (yet to be pushed) and will move next to recheck the `perez` ones.
> 
> However, I antecipate `klucher` being a bit tricky since it considers multiplicative terms rather than additive. If we expand the expression there will be four terms: isotropic, circumsolar, horizon, and a last term representing a sort of interaction between circumsolar and horizon... So, how to make this into three valueso only? Splitting the interaction term half-half? Seems a bit arbitrary...
> imagem

kandersolar:
> @kandersolar @cwhanse @adriesse, now that the v0.14.0 boat has sailed, I reflected a bit on the "no breaking changes" constraint that is posed for v0.14.1.
> 
> in this case, having an added `return_components` defaulting to `False` should be enough, no? `False` mimics v0.14.0, `True` adopts new option.

Yep! That is exactly the idea I tried to explain in https://github.com/pvlib/pvlib-python/issues/2529#issuecomment-3160849211.

adriesse:
> still blocked with this point. 

I think I would suggest doing a parameter variation study first to get a feel for how the interaction term behaves. It should be small.  Splitting in half does seem a bit arbitrary, but maybe you can find a logical  argument to include it with one or the other.

cwhanse:
> However, I antecipate klucher being a bit tricky

I would just exclude `klucher` by ignoring `return_components` and possibly warning the user. pvlib claims to provide implementations that are faithful to the underlying references, so I don't feel that it is appropriate to guess what the authors would have done if requested to return components.

adriesse:
> > However, I anticipate klucher being a bit tricky
> 
> I would just exclude `klucher` by ignoring `return_components` and possibly warning the user. pvlib claims to provide implementations that are faithful to the underlying references, so I don't feel that it is appropriate to guess what the authors would have done if requested to return components.

That may be acceptable way to go for the implementation, but it does put the Klucher model at a disadvantage for PV simulation.  I'd be curious to see how this thing behaves either way.

adriesse:
More generally--now that I think about it--should we be hard-coding these three components? We know there are models with fewer, and there could be models with more.

ramaroesilva:
> More generally--now that I think about it--should we be hard-coding these three components? We know there are models with fewer, and there could be models with more.

My first argument would be that we can always adapt and add more terms to all models, making them be 0 when not applicable (that's what we are doing right now). But this may be naive, assuming that only simple terms exist like up to now. But if more convoluted, interdependent terms start popping up like for klucher...

kandersolar:
Coming back to the design for `isotropic`, a few thoughts:

- I am -1 to returning a single-key dict/dataframe for `return_components=False`.  Everywhere else in pvlib returns a 1-D array/Series when the output is 1-D.  Forcing users to extract out that 1-D array from a trivially 2-D container is unnecessary (and a breaking change).
- Per #2529, the return keys should have the `poa_` prefix.
- I think all the pandas stuff (`is_pandas`, matching `dhi.index`) can be simplified.

How about this?
```python
    poa_isotropic = dhi * (1 + tools.cosd(surface_tilt)) * 0.5

    if return_components:
        diffuse_components = {'poa_isotropic': poa_isotropic,
                              'poa_circumsolar': 0,
                              'poa_horizon': 0}
        if isinstance(poa_isotropic, pd.Series):
            diffuse_components = pd.DataFrame(diffuse_components)

        return diffuse_components

    return poa_isotropic
```

ramaroesilva:
Sounds good to me. Thanks for the insight, being a recent contributor I notice that sometimes my suggestions are a bit hasty/incoherent with the larger picture.

adriesse:
> Sounds good to me. Thanks for the insight, being a recent contributor I notice that sometimes my suggestions are a bit hasty/incoherent with the larger picture.

It's good to inject other ideas!!

adriesse:
> > still blocked with this point.
> 
> I think I would suggest doing a parameter variation study first to get a feel for how the interaction term behaves. It should be small. Splitting in half does seem a bit arbitrary, but maybe you can find a logical argument to include it with one or the other.

Actually it is not hard to do a proportional split for the two anisotropic factors and make them additive. 

```
(1+h) * (1+c) -> (1+a)

a = h + c + hc
  = h + c + (h+c)/(h+c) * hc
  = (h + h/(h+c) * hc) + (c + c/(h+c) * hc)
  = h' + c'

(1+h) * (1+c) -> (1+h'+c')
```

kandersolar:
> I think I would suggest doing a parameter variation study first to get a feel for how the interaction term behaves. It should be small. Splitting in half does seem a bit arbitrary, but maybe you can find a logical argument to include it with one or the other.

For Klucher, I am tempted to support Anton's proportional split, but ultimately I think I'm with Cliff that we shouldn't go beyond what the reference provides.

> More generally--now that I think about it--should we be hard-coding these three components? We know there are models with fewer, and there could be models with more.

It's a good point (though maybe better suited for #1553).  My 2 cents: unusual models that don't fit into the "sum of three components" structure shouldn't prevent us from making our implementation of the popular models (primarily, perez/haydavies/isotropic) more useful.

@ramaroesilva let's focus just on `isotropic` in this PR.  Once that's done, then we can pursue more functions in a follow-up.

AdamRJensen:
Duplicate of #2787

PR #2547, opened 2025-09-19
Drop Python 3.9

kandersolar:
Python 3.9 is end of life starting [October 2025](https://devguide.python.org/versions/).  Might as well drop it now.

I also took the opportunity to advance our minimum numpy, scipy, pandas versions to the lowest that are available for Python 3.10+.


kandersolar:
@echedey-ls do you have any idea why the failing py 3.10-min job is trying to install `numpy==1.19.3` instead of `1.21.2`?  It makes no sense to me.   https://github.com/pvlib/pvlib-python/actions/runs/17867208211/job/50812293647?pr=2547#step:4:213

echedey-ls:
Definitely black magic. Citing the logs:

```bash
Installing pip packages: h5py==3.0.0, numpy==1.21.2, pandas==1.3.3, scipy==1.7.2, pytest-rerunfailures, pytest-remotedata, requests-mock
    error: subprocess-exited-with-error
    
    × pip subprocess to install backend dependencies did not run successfully.
    │ exit code: 1
    ╰─> [956 lines of output]
        Collecting numpy==1.19.3
          Downloading numpy-1.19.3.zip (7.3 MB)
```

In https://github.com/mamba-org/setup-micromamba
There is this link in the readme, to redirect users that want to know caching better (both for downloading and environment): https://github.com/mamba-org/provision-with-micromamba/pull/38#discussion_r808837618
In the end, a user reports his issue is the same as this one: https://github.com/mamba-org/setup-micromamba/issues/50 (short conversation, must read all through it).

I know we have:
```yml
		uses: mamba-org/setup-micromamba@v2
        with:
          cache-downloads: false
          cache-environment: false
```

However, I guess it is caching things, what a coincidence that it's the last numpy version... I won't be able to trace the problem down, so I suggest we wait for 8 days till the cache gets deleted, or just merge as is. I believe in this diff more than that action docs. We could also try to use bare python for the minimal CI - that would require translating the ci files into requirements.txt syntax thou.

>  It makes no sense to me.

Neither to me. Can we delete `.github`? I beg.

kandersolar:
Ah ha, it was an [indirect pin](https://github.com/h5py/h5py/blob/3.0.0/setup.py#L33) via `h5py==3.0.0`.  Increasing to a newer h5py version fixed it.  Thanks for taking a look!

echedey-ls:
Well done and thanks to you Kevin, I thought newer pip reported these incoherencies...

kandersolar:
Merging since this is just a routine update.

PR #2552, opened 2025-09-23
Fix quote characters in pytest GH Actions

kandersolar:
I accidentally used double quotation mark characters in #2547.  It seems that single quotation mark characters are required for the job to run.  My bad!


kandersolar:
The fix worked; tests ran as expected.  Merging.

PR #2559, opened 2025-09-29
Simplify/DRY some infinite_sheds/bifacial.utils calculations

kandersolar:
- ~[ ] Closes #xxxx~
 - [x] Tests added

Projected solar zenith angle (`tan_phi`) is calculated several times in the `infinite_sheds` workflow.  This PR calculates it once and passes it around, simplifying the relevant utility functions and their signatures.

For context, this is a step towards future enhancements I intend to make to `pvlib.bifacial`, which would benefit from working with `tan_phi` instead of the base solar position and module angles.

### DIFF
--- a/pvlib/bifacial/infinite_sheds.py
+++ b/pvlib/bifacial/infinite_sheds.py
@@ -6,7 +6,7 @@ import numpy as np
 import pandas as pd
 from pvlib.tools import cosd, sind, tand
 from pvlib.bifacial import utils
-from pvlib.irradiance import beam_component, aoi, haydavies
+from pvlib.irradiance import aoi, haydavies, poa_components
 
 
 def _poa_ground_shadows(ghi, dhi, albedo, f_gnd_beam, vf_gnd_sky):
@@ -121,8 +121,7 @@ def _poa_ground_pv(poa_ground, gcr, surface_tilt):
     return poa_ground * vf_integ
 
 
-def _shaded_fraction(solar_zenith, solar_azimuth, surface_tilt,
-                     surface_azimuth, gcr):
+def _shaded_fraction(surface_tilt, tan_phi, aoi, gcr):
     """
     Calculate fraction (from the bottom) of row slant height that is shaded
     from direct irradiance by the row in front toward the sun.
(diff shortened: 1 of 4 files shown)


echedey-ls:
You've probably got the email, but docs are failing with:

<details><summary>This failing example</summary>
<p>

```bash
Extension error:
Here is a summary of the problems encountered when running the examples:
Unexpected failing examples (1):
    ../../examples/agrivoltaics/plot_agrivoltaics_ground_irradiance.py failed leaving traceback:
      File "/home/docs/checkouts/readthedocs.org/user_builds/pvlib-python/checkouts/2559/docs/examples/agrivoltaics/plot_agrivoltaics_ground_irradiance.py", line 145, in <module>
        unshaded_ground_fraction = pvlib.bifacial.utils._unshaded_ground_fraction(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    TypeError: _unshaded_ground_fraction() got an unexpected keyword argument 'surface_azimuth'
```

My opinion is that it's bad practice to use internals in public-facing examples. But a fix, using internals or not, is okay with me.

</p>
</details> 

Also, I think this deserves a whatsnew 👀 
Nice improvement.

kandersolar:
> My opinion is that it's bad practice to use internals in public-facing examples. 

Agreed.  We should consider making that ground shaded fraction function public.

kandersolar:
Closing as it's nicer to use `phi` instead of `tan_phi`.  Will pursue a similar refactor in a later PR.

mikofski:
You’re gonna hate this, but you could [memoize](https://en.wikipedia.org/wiki/Memoization) `tan_phi` in a class and reevaluate lazily as needed when inputs update. Crazy right. So over the top. But maybe…?

echedey-ls:
> You’re gonna hate this, but you could [memoize](https://en.wikipedia.org/wiki/Memoization) `tan_phi` in a class and reevaluate lazily as needed when inputs update. Crazy right. So over the top. But maybe…?

Easily hatable TBF 😆 

The easiest option relies on python's `lru_cache` wrapper. At least this one, and I'm convinced many others too, use a hashmap (plain map in python) where `input variables hash`->`cached return value`. Numpy arrays can't be hashed, as they are mutable (plus CPython's default hash implementation is way too basic to work in this case). Also, if you want to bypass this with your suggestion (or other approach, like a wrapper), you would need to check the values of an array matches the values of a stored copy of the array in the cache. So you would end up consuming another significant amount of time.

I'm onboard with Kevin on this approach hehe. Thou that was a good concern @mikofski :)

PR #2563, opened 2025-10-01
Add Batzelis 2017 simple nonlinear PV model

### LINKED ISSUE #2562: Datasheet-only nonlinear model for Pmp/Imp/Vmp/Isc/Voc
I came across this rather nice model; it's firmly rooted in the SDE but uses only datasheet inputs and requires no tedious iteration to calculate the MPP.  Here's a quick comparison of its error (treating the CEC SDM, with consistent inputs, as "truth"), for two modules (c-Si in first row, CdTe in second).  Dotted lines in the first column show the same for PVWatts, for reference.

Image

The comparison for those two modules makes the model seem quite promising to me.

Easy to implement; one simple `lambertw` call, one `np.log`, and the rest just elementary operations.

Reference:
> E. I. Batzelis, "Simple PV Performance Equations Theoretically Well Founded on the Single-Diode Model," in IEEE Journal of Photovoltaics, vol. 7, no. 5, pp. 1400-1409, Sept. 2017, https://doi.org/10.1109/JPHOTOV.2017.2711431

kandersolar:
- [x] Closes #2562 
 - [ ] Tests added

### DIFF
--- a/pvlib/pvarray.py
+++ b/pvlib/pvarray.py
@@ -10,7 +10,7 @@ Supporting functions and parameter fitting functions may also be found here.
 
 import numpy as np
 from scipy.optimize import curve_fit
-from scipy.special import exp10
+from scipy.special import exp10, lambertw
 
 
 def pvefficiency_adr(effective_irradiance, temp_cell,
@@ -394,3 +394,92 @@ def huld(effective_irradiance, temp_mod, pdc0, k=None, cell_type=None,
         k[5] * tprime**2
     )
     return pdc
+
+
+def batzelis(effective_irradiance, temp_cell,
+             isc0, voc0, imp0, vmp0, alpha_sc, beta_voc):
+    """
+    Compute maximum power point, open circuit, and short circuit
+    values using the Batzelis method.
+
+    Batzelis's method [1]_ is a fast method of computing the maximum
+    power current and voltage.  The calculations are rooted in the
+    single-diode equation, but only typical datasheet information
+    is required.
+
+    Parameters
+    ----------
+    effective_irradiance : numeric, non-negative
+        Effective irradiance incident on the PV module. [Wm⁻²]
+    temp_cell : numeric
+        PV module operating temperature. [°C]
+    isc0 : float
+        Short-circuit current at STC. [A]
+    voc0 : float
+        Open-circuit voltage at STC. [V]
+    imp0 : float
+        Maximum power point current at STC. [A]
+    vmp0 : float
+        Maximum power point voltage at STC. [V]
+    alpha_sc : float
+        Short-circuit current temperature coefficient at STC. [1/K]
+    beta_voc : float
+        Open-circuit voltage temperature coefficient at STC. [1/K]    
+
+    Returns
+    -------
+    dict
+        The returned dict-like object always contains the keys/columns:
+
+        * i_sc - short circuit current in amperes.
+        * v_oc - open circuit voltage in volts.
+        * i_mp - current at maximum power point in amperes.
+        * v_mp - voltage at maximum power point in volts.
+        * p_mp - power at maximum power point in watts.
+
+    Notes
+    -----
+    The ``alpha_sc`` and ``beta_voc`` temperature coefficient parameters
+    must be given as normalized values.
+
+    References
+    ----------
+    .. [1] E. I. Batzelis, "Simple PV Performance Equations Theoretically Well
+       Founded on the Single-Diode Model," Journal of Photovoltaics vol. 7, 
+       no. 5, pp. 1400-1409, Sep 2017, :doi:`10.1109/JPHOTOV.2017.2711431`
+    """
+
+    t0 = 298.15
+    delT = temp_cell - (t0 - 273.15)
+    lamT = (temp_cell + 273.15) / t0
+    g = effective_irradiance / 1000
+    lnG = np.log(g)
+
+    # Eq 9-10
+    del0 = (1 - beta_voc * t0) / (50.1 - alpha_sc * t0)
+    w0 = lambertw(np.exp(1/del0 + 1)).real
+
+    # Eqs 27-28
+    alpha_imp = alpha_sc + (beta_voc - 1/t0) / (w0 - 1)
+    beta_vmp = (voc0 / vmp0) * (
+        beta_voc / (1 + del0) +
+        (del0 * (w0 - 1) - 1/(1 + del0)) / t0
+    )
+
+    # Eq 26
+    eps0 = (del0 / (1 + del0)) * (voc0 / vmp0)
+    eps1 = del0 * (w0 - 1) * (voc0 / vmp0) - 1
+
+    # Eqs 22-25
+    isc = g * isc0 * (1 + alpha_sc * delT)
+    voc = voc0 * (1 + del0 * lamT * lnG + beta_voc * delT)
+    imp = g * imp0 * (1 + alpha_imp * delT)
+    vmp = vmp0 * (1 + eps0 * lamT * lnG + eps1 * (1 - g) + beta_vmp * delT)
+
+    return {
+        'p_mp': vmp * imp,
+        'i_mp': imp,
+        'v_mp': vmp,
+        'i_sc': isc,
+        'v_oc': voc,
+    }


kandersolar:
This PR currently adds three new functions.  As they are related to each other I thought it would be helpful to review them together, but I can split the PR into smaller pieces to make it more digestible if needed.

To summarize the three functions:
- `pvlib.ivtools.sdm.fit_desoto_batzelis`: estimate De Soto SDM parameters from datasheet values
- `pvlib.singlediode.batzelis_keypoints`: estimate Pmp/Imp/Vmp/Isc/Voc from SDE parameters
- `pvlib.pvarray.batzelis`: estimate Pmp/Imp/Vmp/Isc/Voc from datasheet values

The third, effectively, combines the first two with `calcparams_desoto`, but with some algebraic manipulations to make the computation more compact.  I think there is also a minor difference in the `calcparams_desoto` calculations.  All three come from the paper linked in #2562.

Questions for reviewers:
- Currently `alpha_sc` has units [1/K] in some places and [A/K] in others.  Better to diverge from the reference (which uses [1/K]) and use [A/K] everywhere for consistency with the rest of pvlib?
- Parameter names likely also should be changed for consistency.  `i_sc` instead of `isc0`, etc?
- Are the current modules the best places for these new functions?

pvlib/ivtools/sdm/desoto.py
```diff
@@ -399,3 +400,64 @@ def _fit_desoto_sandia_diode(ee, voc, vth, tc, specs, const):
...
+        * ``I_L_ref`` [A]
+        * ``I_o_ref`` [A]
+        * ``R_sh_ref`` [Ohm]
+        * ``R_s`` [Ohm]
+
+    References
+    ----------
+    .. [1] E. I. Batzelis, "Simple PV Performance Equations Theoretically Well
+       Founded on the Single-Diode Model," Journal of Photovoltaics vol. 7,
+       no. 5, pp. 1400-1409, Sep 2017, :doi:`10.1109/JPHOTOV.2017.2711431`
```
cwhanse, line 444:
```suggestion
       no. 5, pp. 1400-1409, Sep 2017, :doi:`10.1109/JPHOTOV.2017.2711431`
    .. [2] E. I. Batzelis and S. A. Papathanassiou, "A method for the analytical
       extraction of the single-diode PV model parameters, IEEE Trans.
       Sustain. Energy, vol. 7, no. 2, pp. 504-512, Apr 2016.
       :doi:`10.1109/TSTE.2015.2503435`
```

pvlib/ivtools/sdm/desoto.py
```diff
@@ -399,3 +400,64 @@ def _fit_desoto_sandia_diode(ee, voc, vth, tc, specs, const):
...
     res = sm.RLM(y, new_x).fit()
     return np.array(res.params)[1]
+
+
+def fit_desoto_batzelis(isc0, voc0, imp0, vmp0, alpha_sc, beta_voc):
+    """
+    Determine De Soto single-diode model parameters from datasheet values
+    using Batzelis's method.
+
+    This method is described in Section II.C of [1]_.
```
cwhanse:
```suggestion
    This method is described in Section II.C of [1]_ and fully documented in [2]_.
```

pvlib/ivtools/sdm/desoto.py
```diff
@@ -399,3 +400,64 @@ def _fit_desoto_sandia_diode(ee, voc, vth, tc, specs, const):
...
+
+    Parameters
```
cwhanse, line 413: I would prefer `alpha_sc` and `beta_voc` to be required in units of A/K and V/K, respectively. The parameters `alpha_sc` and `beta_voc` in sdm.fit_desoto, sdm.fit_cec_sam and other function (notably, pvsystem.calcparams_desoto) use those units.

Do we want the STC parameters to have the same names as in pvsystem.sapm? I'm not a fan of the SAPM names, which derive from some old spreadsheet's column headings. But they are the same quantities.

pvlib/singlediode.py
```diff
@@ -861,3 +862,85 @@ def _pwr_optfcn(df, loc):
...
+    # Eqs 3-4
+    isc = Iph * Rsh / (Rs + Rsh)
+    with np.errstate(divide='ignore'):  # zero Iph
+        voc = a * np.log(Iph / Is)
+
+    # Eqs 5-8
+    w = np.real(lambertw(np.e * Iph / Is))
+    # vmp = (1 + Rs/Rsh) * a * (w - 1) - Rs * Iph * (1 - 1/w)  # not needed
+    with np.errstate(divide='ignore', invalid='ignore'):  # zero Iph -> zero w
+        imp = Iph * (1 - 1/w) - a * (w - 1) / Rsh
```
cwhanse, line 1011: Suggest we implement in terms of Gsh = 1/Rsh. The case of infinite Rsh is handled by other pvlib functions to avoid division by zero.

tests/ivtools/sdm/test_desoto.py
```diff
@@ -92,3 +92,27 @@ def test_fit_desoto_sandia(cec_params_cansol_cs5p_220p):
     assert_allclose(result['dEgdT'], -0.0002677)
     assert_allclose(result['EgRef'], 1.3112547292120638)
     assert_allclose(result['cells_in_series'], specs['cells_in_series'])
+
+
+def test_fit_desoto_batzelis():
+    params = {'isc0': 15.98, 'voc0': 50.26, 'imp0': 15.27, 'vmp0': 42.57,
+              'alpha_sc': 0.00046, 'beta_voc': -0.0024}
+    expected = {  # calculated with the function itself
```
cwhanse, line 100: Calculated with the `fit_desoto_batzelis` function?

markcampanelli: @kandersolar Would you be willing to provide a link to the spec sheet for the module here?

kandersolar: https://jinkosolarcdn.shwebspace.com/uploads/68c7b593/JKM650-670N-66QL6-BDV-F1-EN.pdf

Not chosen for any particular reason--just the first one to come up when I searched for a Jinko datasheet.

pvlib/singlediode.py
kandersolar: What would `Gsh` help with here?  The `np.errstate` business is for when `w` is zero because of `Iph` being zero at night.

cwhanse: You're right about this line. What about the calculation of `isc`? There, `Rsh = np.inf` will produce `isc = np.nan`. The correct value would be `isc = Iph  = Iph / (1 + Gsh * Rs)`

cwhanse:
A question about the method: does `batzelis` produce the same results as `fit_desoto_batzelis` + `batzelis_keypoints`? It is not clear to me that should be the case.

pvlib/ivtools/sdm/desoto.py
kandersolar: > units of A/K and V/K

done

> Do we want the STC parameters to have the same names as in pvsystem.sapm?  I'm not a fan of the SAPM names, 

Me neither.  `fit_desoto` uses just the usual `v_mp` etc.  This PR now uses `v_mp` etc. too, although those are obviously not ideal either.

kandersolar:
> A question about the method: does `batzelis` produce the same results as `fit_desoto_batzelis` + `batzelis_keypoints`? It is not clear to me that should be the case.

Close but not exactly the same, in my testing.  The latter has to be `fit_desoto_batzelis` + `calcparams_desoto` + `batzelis_keypoints`, and I think there is at least a difference in the `calcparams_desoto` step (regarding `dEgdT`).  I think there are differences elsewhere too, but I can't point to any specifics.

pvlib/singlediode.py
```diff
@@ -949,3 +950,85 @@ def _pwr_optfcn(df, loc):
                                  df['resistance_shunt'], df['nNsVth'])
 
     return current * df[loc]
+
+
+def batzelis_keypoints(photocurrent, saturation_current, resistance_series,
```
cwhanse:
```suggestion
def batzelis(photocurrent, saturation_current, resistance_series,
```

This function has the same signature as `pvsystem.singlediode` so I don't think it helps to have "keypoints" in the name. Do you think the name needs to be other than `batzelis` to differentiate from `pvsystem.batzelis`? Maybe `batzelis_2017` here.

If the emphasis was only on MPP, then it could be `_2014` as that content is in his 2014 paper (10.1109/TEC.2014.2356017). The equations for Isc and Voc are not in that 2014 paper, though.

tests/test_singlediode.py
```diff
@@ -617,9 +618,59 @@ def test_bishop88_init_cond(method):
...
+    params3 = {k: pd.Series(v) for k, v in params2.items()}
+    output3 = batzelis_keypoints(**params3)
+    assert isinstance(output3, pd.DataFrame)
+    for key in exact_values:
+        exp = pd.Series([exact_values[key]] * 2)
+        np.testing.assert_allclose(output3[key], exp, rtol=rtol)
+
+
+def test_batzelis_keypoints_night():
+    # SDMs produce photocurrent=0 and resistance_shunt=inf at night
```
cwhanse:
```suggestion
    # The De Soto SDM produces photocurrent=0 and resistance_shunt=inf at 0 W/m2 irradiance
```

cwhanse (approved):

adriesse:
>     * Currently `alpha_sc` has units [1/K] in some places and [A/K] in others.  Better to diverge from the reference (which uses [1/K]) and use [A/K] everywhere for consistency with the rest of pvlib?

I think this is the right decision for pvlib because of precedent, although personally I much prefer the normalized values.

>     * Parameter names likely also should be changed for consistency.  `i_sc` instead of `isc0`, etc?

+1

>     * Are the current modules the best places for these new functions?

+1

pvlib/singlediode.py
```diff
@@ -949,3 +950,85 @@ def _pwr_optfcn(df, loc):
                                  df['resistance_shunt'], df['nNsVth'])
 
     return current * df[loc]
+
+
+def batzelis(photocurrent, saturation_current, resistance_series,
+             resistance_shunt, nNsVth):
```
markcampanelli, line 956: @kandersolar Was adding this SDE solver to the `method` options in `pvsystem.singlediode` discussed?

cwhanse: I don't think we discussed it. I'm not in favor of adding it as a `method` option to `pvsystem.singlediode` because `singlediode.batzelis` solves an approximation of the single diode equation, rather than the SDE itself.

PR #2567, opened 2025-10-03
Use scipy's find_minimum in lambertw for MPP

### LINKED ISSUE #2497: Use scipy's new `optimize.elementwise` functions

cwhanse:
- [x] Closes #2497
 - [ ] Tests added

Switches to the old golden section if `find_minimum` is not available, or if `find_minimum` fails to converge.

### DIFF
--- a/pvlib/singlediode.py
+++ b/pvlib/singlediode.py
@@ -913,8 +913,28 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
             v_oc = 0.
 
     # Find the voltage, v_mp, where the power is maximized.
-    # Start the golden section search at v_oc * 1.14
-    p_mp, v_mp = _golden_sect_DataFrame(params, 0., v_oc * 1.14, _pwr_optfcn)
+    # use scipy.elementwise if available
+    use_gs = False
+    try:
+        from scipy.optimize.elementwise import find_minimum
+        init = (0., 0.8*v_oc, 1.01*v_oc)
+        res = find_minimum(_vmp_opt, init,
+                           args=(params['photocurrent'],
+                                 params['saturation_current'],
+                                 params['resistance_series'],
+                                 params['resistance_shunt'],
+                                 params['nNsVth'],))
+        if res.success.all():
+            v_mp = res.x
+            p_mp = -1.*res.f_x
+        else:
+            use_gs = True
+    except ModuleNotFoundError:
+        use_gs = True
+
+    if use_gs:
+        # gracefully switch to old golden section method
+        p_mp, v_mp = _golden_sect_DataFrame(params, 0., v_oc * 1.14, _pwr_optfcn)
 
     # Find Imp using Lambert W
     i_mp = _lambertw_i_from_v(v_mp, **params)
@@ -936,6 +956,15 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
         out += (ivcurve_i, ivcurve_v)
 
     return out
+
+
+def _vmp_opt(v, iph, io, rs, rsh, nNsVth):
+    '''
+    Function to find power from ``i_from_v``.
+    '''
+    current = _lambertw_i_from_v(v, iph, io, rs, rsh, nNsVth)
+
+    return -v * current
 
 
 def _pwr_optfcn(df, loc):


cwhanse:
What do you think of this approach?

pvlib/singlediode.py
```diff
@@ -913,8 +913,28 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
...
+                           args=(params['photocurrent'],
+                                 params['saturation_current'],
+                                 params['resistance_series'],
+                                 params['resistance_shunt'],
+                                 params['nNsVth'],))
+        if res.success.all():
+            v_mp = res.x
+            p_mp = -1.*res.f_x
+        else:
+            use_gs = True
```
kandersolar: I am inclined to not include this safety net; if we plan to get rid of the GS function eventually, might as well let any unforeseen issues with the scipy approach bubble to the surface now so that we can address them sooner rather than later.

Or is there reason to expect the scipy function to fail sometimes?

pvlib/singlediode.py
```diff
@@ -913,8 +913,28 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
             v_oc = 0.
 
     # Find the voltage, v_mp, where the power is maximized.
-    # Start the golden section search at v_oc * 1.14
-    p_mp, v_mp = _golden_sect_DataFrame(params, 0., v_oc * 1.14, _pwr_optfcn)
```
kandersolar, line 917: Just curious... I wonder if anyone knows why `1.14` is the magic number.  I see it goes back to the [early early days](https://github.com/pvlib/pvlib-python/blob/e74644ba99d16113fb976493b5006ffdf7fb1671/pvlib/pvl_singlediode.py#L143) of pvlib-python.

pvlib/singlediode.py
cwhanse: I can't think of a reason for failure other than bad inputs. Because `res.success` is a vector, we could pass over to the GS only those index values which failed, for whatever reason. Or skip the `success` check altogether and just push `v_mp` forward. I worry a little that will forward unconverged, but still numeric, values.

pvlib/singlediode.py
cwhanse: That was `1.14 * Voc_ref`, the STC value of Voc. The factor of 1.14 was meant to push the right endpoint out when temperature is lower than STC.

In pvlib, v_oc reflects temperature and irradiance. I included `1.01 * v_oc` in case the algorithm needs to start with a negative value of power at the right endpoint, since it is minimizing negative power. Now that I write this, I don't think that's necessary - it is provable that power has a unique maximum between `v=0` and `v=v_oc`.

kandersolar: I was wondering why it is `1.14 * Voc_ref` rather than (for example) `1.15 * Voc_ref` or `1.13 * Voc_ref` or `1.2 * Voc_ref`.  Maybe computed from some suitable low temperature and a representative temperature coefficient?  Just idle curiosity.

pvlib/singlediode.py
```diff
@@ -913,8 +913,23 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
...
+                                 params['nNsVth'],))
+        v_mp = res.x
+        p_mp = -1.*res.f_x
+    except ModuleNotFoundError:
+        # switch to old golden section method
+        p_mp, v_mp = _golden_sect_DataFrame(params, 0., v_oc * 1.14,
+                                            _pwr_optfcn)
 
     # Find Imp using Lambert W
     i_mp = _lambertw_i_from_v(v_mp, **params)
```
kandersolar: It's not really related to this PR, but can't we avoid an unnecessary calculation here and just do `i_mp = p_mp / v_mp`?

docs/sphinx/source/whatsnew/v0.13.2.rst
```diff
@@ -27,7 +27,9 @@ Enhancements
   :py:func:`~pvlib.singlediode.bishop88_mpp`,
   :py:func:`~pvlib.singlediode.bishop88_v_from_i`, and
   :py:func:`~pvlib.singlediode.bishop88_i_from_v`. (:issue:`2497`, :pull:`2498`)
-
+* Add capability for :py:func:`~pvlib.singlediode._lambertw` to use scipy
+  find_minimum to get the maximum power point instead of pvlib's golden
+  mean search. (:issue:`2497`, :pull:`2567`)
```
kandersolar:
```suggestion
* Accelerate :py:func:`~pvlib.pvsystem.singlediode` when scipy>=1.15 is installed. (:issue:`2497`, :pull:`2567`)
```

pvlib/singlediode.py
```diff
@@ -913,8 +913,23 @@ def _lambertw(photocurrent, saturation_current, resistance_series,
...
+        from scipy.optimize.elementwise import find_minimum
+        init = (0., 0.8*v_oc, v_oc)
+        res = find_minimum(_vmp_opt, init,
+                           args=(params['photocurrent'],
+                                 params['saturation_current'],
+                                 params['resistance_series'],
+                                 params['resistance_shunt'],
+                                 params['nNsVth'],))
+        v_mp = res.x
+        p_mp = -1.*res.f_x
```
kandersolar, line 928: Test failure is because the test expects `i_mp=0` for `photocurrent=0`, but currently it's getting back `i_mp=nan` on some platforms.  Maybe because `photocurrent=0` violates the requirement of `init`:

>    The abscissae of a standard scalar minimization bracket. A bracket is
>    valid if arrays ``x1, x2, x3 = init`` satisfy ``x1 < x2 < x3`` and
>    ``func(x1) >= func(x2) <= func(x3)``, where one of the inequalities
>    is strict.

We could detect `photocurrent==0` and skip the optimizer for those cases?

pvlib/singlediode.py
cwhanse: Also unrelated: since we removed the `ivcurve_pnts` kwarg from `singlediode` (several versions ago), should we also delete that capability from `_lambertw`?

kandersolar:
@cwhanse FYI it looks like this branch got mixed up with the CEC/PVsyst parameter conversion code.

cwhanse:
Ugh. Starting over.

PR #2568, opened 2025-10-03
add pvwatts_dc_marion

williamhobbs:
- [x] Closes #2566
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing/index.html)
 - [ ] Tests added
 - [ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.
 - [ ] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [x] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [ ] Pull request is nearly complete and ready for detailed review.
 - [ ] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

Still has some work to do, but I thought I would share what I've started to get feedback. Tagging @kandersolar, since we briefly discussed this.

PR #2569, opened 2025-10-07
add marion's adjustment to pvwatts_dc

### LINKED ISSUE #2566: Nonlinear adjustment to pvwattsv5 dc model
It would be nice to have the Marion 2008 [1] nonlinear adjustment to the PVWatts V5 DC power model (`pvlib.pvsystem.pvwatts_dc`) available in pvlib. This adjustment/improvement is already described on the Sandia PVPMC website [2] (*with a typo from the original source included in the figure*). 

I think a new function, e.g., `pvlib.pvsystem.pvwatts_dc_marion`, that is similar to `pvwatts_dc` with an additional input parameter of an irradiance correction factor, `k`, would be the best way to implement this.

An alternative would be a separate "adjustment" function that could be applied to the output of `pvwatts_dc`.

**Additional context**
Related to https://github.com/pvlib/pvlib-python/issues/2562 and this discussion, https://groups.google.com/g/pvlib-python/c/-3UuvBtWByY.

I had an offline exchange with @janinefreeman and @cpaulgilman to see if there was a specific reason that this adjustment was never implemented in PVWatts, and they could not find a record of any reason. 

[1] B. Marion, "Comparison of Predictive Models for Photovoltaic Module Performance," IEEE PVSC 2008, https://docs.nrel.gov/docs/fy08osti/42511.pdf

[2] https://pvpmc.sandia.gov/modeling-guide/2-dc-module-iv/point-value-models/pvwatts/improvements-to-pvwatts/

williamhobbs:
- [x] Closes #2566
 - [ ] Tests added

Still some work to do . Replaces #2568.

### DIFF
--- a/pvlib/pvsystem.py
+++ b/pvlib/pvsystem.py
@@ -2878,14 +2878,36 @@ def scale_voltage_current_power(data, voltage=1, current=1):
 
 @renamed_kwarg_warning(
     "0.13.0", "g_poa_effective", "effective_irradiance")
-def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
+def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.,
+               k=0.0, capped_adjustment=False):
     r"""
-    Implements NREL's PVWatts DC power model. The PVWatts DC model [1]_ is:
+    Implements NREL's PVWatts (Version 5) DC power model. The PVWatts Version
+    5 DC model [1]_ is:
 
     .. math::
 
         P_{dc} = \frac{G_{poa eff}}{1000} P_{dc0} ( 1 + \gamma_{pdc} (T_{cell} - T_{ref}))
 
+    This model has also been referred to as the power temperature coefficient
+    model.
+
+    This function accepts an optional irradiance adjustment factor, `k`, based
+    on based on [2]_. This applies a piece-wise adjustment to power based on
+    irradiance, where `k` is the reduction in actual power at 200 W/m^2
+    relative to ideal power calculated linearly from standard test conditions,
+    normalized to nameplate power at standard test conditions.
+
+    .. math::
+
+        k=\frac{0.2P_{dc0}-P_{200}}{P_{dc0}}
+
+    For example, a 500 W module that produces 95 W at 200 W/m^2 (a 5% relative
+    reduction in efficiency) would have a value of `k` = 0.01.
+
+    This adjustment increases relative efficiency for irradiance above 1000
+    W/m^2, which may not be desired. An optional input, `capped_adjustment`,
+    modifies the adjustment from [2]_ to only apply below 1000 W/m^2.
+   
     Note that ``pdc0`` is also used as a symbol in
     :py:func:`pvlib.inverter.pvwatts`. ``pdc0`` in this function refers to the DC
     power of the modules at reference conditions. ``pdc0`` in
@@ -2909,6 +2931,11 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
     temp_ref: numeric, default 25.0
         Cell reference temperature. PVWatts defines it to be 25 C and
         is included here for flexibility. [C]
+    k: numeric, default 0.005
+        Irradiance correction factor, defined in [2]_. [unitless]
+    modified_marion: Boolean, default False
+        Optional modification to [2]_, where no adjustment is applied above
+        1000 W/m^2.
 
     Returns
     -------
@@ -2920,10 +2947,33 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
     .. [1] A. P. Dobos, "PVWatts Version 5 Manual"
            http://pvwatts.nrel.gov/downloads/pvwattsv5.pdf
            (2014).
+    .. [2] B. Marion, "Comparison of Predictive Models for
+           Photovoltaic Module Performance," 
+           https://doi.org/10.1109/PVSC.2008.4922586,
+           https://docs.nrel.gov/docs/fy08osti/42511.pdf
+           (2008).
     """  # noqa: E501
 
     pdc = (effective_irradiance * 0.001 * pdc0 *
            (1 + gamma_pdc * (temp_cell - temp_ref)))
+
+    # apply Marion's correction if k is anything but zero
+    if k != 0:
+        err_1 = (k * (1 - (1 - effective_irradiance / 200)**4) /
+                 (effective_irradiance / 1000))
+        err_2 = (k * (1000 - effective_irradiance) / (1000 - 200))
+
+        pdc_marion = np.where(effective_irradiance <= 200,
+                              pdc * (1 - err_1),
+                              pdc * (1 - err_2))
+
+        # "cap" Marion's correction at 1000 W/m^2
+        if capped_adjustment is True:
+            pdc_marion = np.where(effective_irradiance >= 1000,
+                                  pdc,
+                                  pdc_marion)
+
+        pdc = pdc_marion
 
     return pdc
 


pvlib/pvsystem.py
```diff
@@ -2878,14 +2878,36 @@ def scale_voltage_current_power(data, voltage=1, current=1):
 
 @renamed_kwarg_warning(
     "0.13.0", "g_poa_effective", "effective_irradiance")
-def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
+def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.,
+               k=0.0, capped_adjustment=False):
```
cwhanse: Prefer `cap_adjustment`

I'd also set `k=None` here.

pvlib/pvsystem.py
```diff
@@ -2878,14 +2878,36 @@ def scale_voltage_current_power(data, voltage=1, current=1):
...
+    relative to ideal power calculated linearly from standard test conditions,
+    normalized to nameplate power at standard test conditions.
+
+    .. math::
+
+        k=\frac{0.2P_{dc0}-P_{200}}{P_{dc0}}
+
+    For example, a 500 W module that produces 95 W at 200 W/m^2 (a 5% relative
+    reduction in efficiency) would have a value of `k` = 0.01.
+
```
cwhanse: Please include Equation from above with `k`, I'd place it here.

pvlib/pvsystem.py
```diff
@@ -2909,6 +2931,11 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
     temp_ref: numeric, default 25.0
         Cell reference temperature. PVWatts defines it to be 25 C and
         is included here for flexibility. [C]
+    k: numeric, default 0.005
+        Irradiance correction factor, defined in [2]_. [unitless]
+    modified_marion: Boolean, default False
```
```suggestion
    cap_adjustment: Boolean, default False
```

pvlib/pvsystem.py
```diff
@@ -2909,6 +2931,11 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
     temp_ref: numeric, default 25.0
         Cell reference temperature. PVWatts defines it to be 25 C and
         is included here for flexibility. [C]
+    k: numeric, default 0.005
```
```suggestion
    k: numeric, optional
```

pvlib/pvsystem.py
```diff
@@ -2909,6 +2931,11 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+        Irradiance correction factor, defined in [2]_. [unitless]
+    modified_marion: Boolean, default False
+        Optional modification to [2]_, where no adjustment is applied above
+        1000 W/m^2.
```
```suggestion
        If True, apply the optional adjustment at and below 1000 W/m^2.
```

pvlib/pvsystem.py
```diff
@@ -2920,11 +2947,34 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+           https://doi.org/10.1109/PVSC.2008.4922586,
+           https://docs.nrel.gov/docs/fy08osti/42511.pdf
+           (2008).
     """  # noqa: E501
 
     pdc = (effective_irradiance * 0.001 * pdc0 *
            (1 + gamma_pdc * (temp_cell - temp_ref)))
 
+    # apply Marion's correction if k is anything but zero
+    if k != 0:
```
```suggestion
    if k is not None:
```

pvlib/pvsystem.py
```diff
@@ -2878,14 +2878,36 @@ def scale_voltage_current_power(data, voltage=1, current=1):
...
         P_{dc} = \frac{G_{poa eff}}{1000} P_{dc0} ( 1 + \gamma_{pdc} (T_{cell} - T_{ref}))
 
+    This model has also been referred to as the power temperature coefficient
+    model.
+
+    This function accepts an optional irradiance adjustment factor, `k`, based
+    on based on [2]_. This applies a piece-wise adjustment to power based on
+    irradiance, where `k` is the reduction in actual power at 200 W/m^2
+    relative to ideal power calculated linearly from standard test conditions,
+    normalized to nameplate power at standard test conditions.
```
```suggestion
    This function accepts an optional irradiance adjustment factor, `k`, based
    on based on [2]_. This applies a piece-wise adjustment to power based on
    irradiance, where `k` is the reduction in actual power at 200 W/m^2
    relative to power calculated at 200 W/m^2 as 0.2*`pdc0`.
```

cwhanse (commented):
I like this better than a separate function.

pvlib/pvsystem.py
williamhobbs: done

pvlib/pvsystem.py
williamhobbs: See my latest change. Is that what you had in mind?

cwhanse: Yeah, that comment was not clear. I **was** thinking we should copy the equation for `P_DC` here, and show how `k` modifies the power. But the equation for the adjustment is not simple (the piecewise part), so I retract that thought. If someone wants to know how the adjustment works, we've provided code and the reference.

pvlib/pvsystem.py
```diff
@@ -2920,11 +2944,34 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
     .. [1] A. P. Dobos, "PVWatts Version 5 Manual"
            http://pvwatts.nrel.gov/downloads/pvwattsv5.pdf
            (2014).
+    .. [2] B. Marion, "Comparison of Predictive Models for
+           Photovoltaic Module Performance,"
+           https://doi.org/10.1109/PVSC.2008.4922586,
```
RDaxini: let's use the doi sphinx role
```suggestion
           :doi:`10.1109/PVSC.2008.4922586`,
```

pvlib/pvsystem.py
```diff
@@ -2878,14 +2878,34 @@ def scale_voltage_current_power(data, voltage=1, current=1):
 
 @renamed_kwarg_warning(
     "0.13.0", "g_poa_effective", "effective_irradiance")
-def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
+def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.,
+               k=None, cap_adjustment=False):
     r"""
-    Implements NREL's PVWatts DC power model. The PVWatts DC model [1]_ is:
+    Implements NREL's PVWatts (Version 5) DC power model. The PVWatts Version
+    5 DC model [1]_ is:
```
RDaxini: I suggest (an imperative) single sentence for the one line description (single sentence helps with the sphinx autosummary for functions) 
Also, let's follow new(-ish?) pvlib style: single-line description -> parameter description -> notes section.
So I suggest moving everything "The PVWatts Version..." onwards to _after_ the parameter description. [Example](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.spectrum.average_photon_energy.html#pvlib-spectrum-average-photon-energy)

```suggestion
    Implement NREL's PVWatts (Version 5) DC power model. The PVWatts Version
    5 DC model [1]_ is:
```

pvlib/pvsystem.py
```diff
@@ -2878,14 +2878,34 @@ def scale_voltage_current_power(data, voltage=1, current=1):
...
     .. math::
 
         P_{dc} = \frac{G_{poa eff}}{1000} P_{dc0} ( 1 + \gamma_{pdc} (T_{cell} - T_{ref}))
 
+    This model has also been referred to as the power temperature coefficient
+    model.
+
+    This function accepts an optional irradiance adjustment factor, `k`, based
+    on [2]_. This applies a piece-wise adjustment to power based on irradiance,
+    where `k` is the reduction in actual power at 200 W/m^2 relative to power
```
RDaxini: nitpick
```suggestion
    where `k` is the reduction in actual power at 200 Wm⁻² relative to power
```

pvlib/pvsystem.py
```diff
@@ -2878,14 +2878,34 @@ def scale_voltage_current_power(data, voltage=1, current=1):
...
+    produces 95 W at 200 W/m^2 (a 5% relative reduction in efficiency) would
+    have a value of `k` = 0.01.
+
+    .. math::
+
+        k=\frac{0.2P_{dc0}-P_{200}}{P_{dc0}}
+
+    This adjustment increases relative efficiency for irradiance above 1000
+    W/m^2, which may not be desired. An optional input, `capped_adjustment`,
+    modifies the adjustment from [2]_ to only apply below 1000 W/m^2.
```
RDaxini: nitpick
```suggestion
    This adjustment increases relative efficiency for irradiance above 1000
    Wm⁻², which may not be desired. An optional input, `capped_adjustment`,
    modifies the adjustment from [2]_ to only apply below 1000 Wm⁻².
```

RDaxini (commented):
Nice work @williamhobbs. Just a few simple comments/suggestions on the docs from me. Only had a quick look at the python implementation, LGTM at a glance but I could have a deeper look later if required

pvlib/pvsystem.py
williamhobbs: Done.

pvlib/pvsystem.py
```diff
@@ -2909,22 +2900,81 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+                              pdc * (1 - err_2))
+
+        # "cap" Marion's correction at 1000 W/m^2
+        if cap_adjustment is True:
+            pdc_marion = np.where(effective_irradiance >= 1000,
+                                  pdc,
+                                  pdc_marion)
+
+        pdc = pdc_marion
+
```
RDaxini: @williamhobbs have you started / do you intend to write tests? The codecov check fails since this section is not covered by tests. Happy to help with that if you need.

williamhobbs: Haven't started, but figured I would need to. Do you have suggestions on tests to add? I have pretty limited experience with the pvlib testing structure/best practices.

cwhanse: I'd hand-calculate half a dozen points, using `k`. Test for correct output with input of three types: float, array, Series. Then test with `cap_adjustment=True`.

RDaxini: Be creative, test whatever comes to mind! Generally speaking, test a range of intended and extreme conditions.

Intended:
For the `if` statements, check whether the indented block is executed if the condition is met (for example, if `k` is not None, use some example values to check whether the correction is applied).
Whether the block is executed correctly should also be checked--- if `k` is not None, then check example irradiance values >200 and <=200 (is the intended behaviour executed correctly in these cases?)

Extreme:
What if the user enters non-physical values, how should these be handled and are they handled in this way? e.g. negative or NaN irradiance values

Someone else might be able to offer a clearer/more succinct explanation 😅

RDaxini: > I'd hand-calculate half a dozen points, using `k`. Test for correct output with input of three types: float, array, Series. Then test with `cap_adjustment=True`.

Good point, check whether different data types are handled appropriately too

cwhanse: @williamhobbs you do not have to test with `k=None`, that is covered by existing tests. Any new tests would be better in their own function, e.g., `test_pvwatts_dc_with_k`.

pvlib/pvsystem.py
```diff
@@ -2909,22 +2900,81 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+    produces 95 W at 200 Wm-2 (a 5% relative reduction in efficiency) would
+    have a value of `k` = 0.01.
+
+    .. math::
+
+        k=\frac{0.2P_{dc0}-P_{200}}{P_{dc0}}
+
+    This adjustment increases relative efficiency for irradiance above 1000
+    Wm⁻², which may not be desired. An optional input, `capped_adjustment`,
+    modifies the adjustment from [2]_ to only apply below 1000 Wm⁻².
```
wholmgren: any concerns about deviating from a reference here? I'm ok with it but imagine it could be a point of contention. perhaps some more documentation clarity would help

pvlib/pvsystem.py
```diff
@@ -2909,22 +2900,81 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
     pdc = (effective_irradiance * 0.001 * pdc0 *
            (1 + gamma_pdc * (temp_cell - temp_ref)))
 
+    # apply Marion's correction if k is anything but zero
+    if k is not None:
+        err_1 = (k * (1 - (1 - effective_irradiance / 200)**4) /
+                 (effective_irradiance / 1000))
+        err_2 = (k * (1000 - effective_irradiance) / (1000 - 200))
+
+        pdc_marion = np.where(effective_irradiance <= 200,
```
wholmgren: np.where will break the paradigm of "return the same object type that was input" since it always returns an array. Options:
1. keep np.where, cast output to match input
2. switch to slicing, assume array input
3. switch to slicing, promote scalars to arrays for compatibility

pvlib/pvsystem.py
```diff
@@ -2909,22 +2900,81 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+                              pdc * (1 - err_1),
+                              pdc * (1 - err_2))
+
+        # "cap" Marion's correction at 1000 W/m^2
+        if cap_adjustment is True:
```
```suggestion
        if cap_adjustment:
```

pvlib/pvsystem.py
```diff
@@ -2909,22 +2900,81 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+           :doi:`10.1109/PVSC.2008.4922586`,
+           https://docs.nrel.gov/docs/fy08osti/42511.pdf
+           (2008).
     """  # noqa: E501
 
     pdc = (effective_irradiance * 0.001 * pdc0 *
            (1 + gamma_pdc * (temp_cell - temp_ref)))
 
+    # apply Marion's correction if k is anything but zero
+    if k is not None:
```
wholmgren, line 2974: needing to use the `is not None` paradigm is a small reason to prefer a separate function

cwhanse:
> We've been talking about changing this function name to `pvwatts_dc_v5` in #1350. I think these new parameters will look quite out of place if we do go through with that. Which makes me think a separate function might be preferable.

I'm not dismissing your point. But to reply, PVWatts v8 (or whatever it is called now) is a moving target that does not have much documentation. I think, if we do anything with the newer PVWatts in pvlib (once it stablizes and is documented somehow), we could name that new function "pvwatts_v8" and leave the existing functions names without the version suffix.

pvlib/pvsystem.py
RDaxini: > @williamhobbs you do not have to test with `k=None`, that is covered by existing tests.

Correct, my bad. I was trying to make a general point but overlooked that in the example.
See the [codecov report](https://app.codecov.io/gh/pvlib/pvlib-python/pull/2569?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=pvlib)

> Any new tests would be better in their own function, e.g., `test_pvwatts_dc_with_k`.

Add to [test_pvsystem.py](https://github.com/pvlib/pvlib-python/blob/main/tests/test_pvsystem.py) (examples also visible there)

pvlib/pvsystem.py
williamhobbs: I also wondered about deviating from the reference. It just gives the option to apply the Marion 2008 correction up to 1000 Wm-2, and stay with standard pvwatts above that, so seems like a minor deviation. 

And I can definitely add more documentation. I tend to get to long-winded and was trying to fight that. I’m open to suggestions, but will also work on some additions.

pvlib/pvsystem.py
williamhobbs: I see what you mean. I tried a few things, but can’t seem to figure out how to get your proposed solutions to work. Any pointers or examples?

pvlib/pvsystem.py
```diff
@@ -2933,9 +2933,12 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.,
...
 
-    This adjustment increases relative efficiency for irradiance above 1000
-    Wm⁻², which may not be desired. An optional input, `capped_adjustment`,
-    modifies the adjustment from [2]_ to only apply below 1000 Wm⁻².
+    For positive `k` values, and `k` is typically positive, this adjustment
+    increases relative efficiency when irradiance is above 1000 Wm⁻². This may
+    not be desired, as modules with nonlinear irradiance response often have
+    peak efficiency near 1000 Wm⁻², and it is either flat or declining at
+    higher irradiance. An optional parameter, `cap_adjustment`, can address
+    this by modifying the adjustment from [2]_ to only apply below 1000 Wm⁻².
```
williamhobbs: @wholmgren, does this help with clarifying the deviation?

pvlib/pvsystem.py
williamhobbs: @wholmgren, I think this addresses your comment about returning the same object type

~https://github.com/pvlib/pvlib-python/pull/2569/commits/9f756bf24a233391e90f1e1cf09caf100cfd64b7~ (Edit: not that one - see the latest commit...)

pvlib/pvsystem.py
williamhobbs: @cwhanse and @RDaxini, I added some tests. I'm not sure if it's what you had in mind, so let me know if I should remove or add anything.

RDaxini: @williamhobbs these look good to me! I see tests covering different dtypes, various combinations of `k` and `cap_adjustment`, etc. Get a second opinion from @cwhanse in case I am missing something.

One other case to add would be the **scalar** case when `cap_adjustment=True` **and** `effective_irradiance>=1000`
That should cover the `else` branch highlighted as missing in the codecov report. Something simple like:

```
def test_pvwatts_dc_cap_adjustment_scalar_above_1000():
    irrad = 1200
    temp_cell = 25
    k = 0.01
    cap_adjustment = True

    out = pvsystem.pvwatts_dc(irrad, temp_cell, 100, -0.003, 25, k, cap_adjustment)
    expected = 120.0

    assert_allclose(out, expected)
```

So in this case, the unadjusted `pdc` value is returned. Right?

pvlib/pvsystem.py
```diff
@@ -2878,19 +2878,10 @@ def scale_voltage_current_power(data, voltage=1, current=1):
...
-    .. math::
-
-        P_{dc} = \frac{G_{poa eff}}{1000} P_{dc0} ( 1 + \gamma_{pdc} (T_{cell} - T_{ref}))
-
-    Note that ``pdc0`` is also used as a symbol in
-    :py:func:`pvlib.inverter.pvwatts`. ``pdc0`` in this function refers to the DC
-    power of the modules at reference conditions. ``pdc0`` in
-    :py:func:`pvlib.inverter.pvwatts` refers to the DC power input limit of
-    the inverter.
+    Implement NREL's PVWatts (Version 5) DC power model.
```
RDaxini:
```suggestion
    Implement NREL's PVWatts (Version 5) DC power model.
```

pvlib/pvsystem.py
```diff
@@ -2909,22 +2900,105 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
...
+        P_{dc} = \frac{G_{poa eff}}{1000} P_{dc0} ( 1 + \gamma_{pdc} (T_{cell} - T_{ref}))
+
+    This model has also been referred to as the power temperature coefficient
+    model.
+
+    This function accepts an optional irradiance adjustment factor, `k`, based
+    on [2]_. This applies a piece-wise adjustment to power based on irradiance,
+    where `k` is the reduction in actual power at 200 Wm⁻² relative to power
+    calculated at 200 Wm-2 as 0.2*`pdc0`. For example, a 500 W module that
+    produces 95 W at 200 Wm-2 (a 5% relative reduction in efficiency) would
```
RDaxini:
```suggestion
    produces 95 W at 200 Wm⁻² (a 5% relative reduction in efficiency) would
```

pvlib/pvsystem.py
```diff
@@ -2909,22 +2900,105 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
     temp_ref: numeric, default 25.0
         Cell reference temperature. PVWatts defines it to be 25 C and
         is included here for flexibility. [C]
+    k: numeric, optional
+        Irradiance correction factor, defined in [2]_. Typically positive.
+        [unitless]
+    cap_adjustment: Boolean, default False
+        If True, apply the optional adjustment at and below 1000 W/m^2.
```
RDaxini: pretty units plz
```suggestion
        If True, apply the optional adjustment at and below 1000 Wm⁻²
```

pvlib/pvsystem.py
```diff
@@ -2909,22 +2900,105 @@ def pvwatts_dc(effective_irradiance, temp_cell, pdc0, gamma_pdc, temp_ref=25.):
     temp_ref: numeric, default 25.0
         Cell reference temperature. PVWatts defines it to be 25 C and
         is included here for flexibility. [C]
```
RDaxini: Idk about you Will, but I prefer my temps in degrees celsius over coulombs ;P
```suggestion
        Cell reference temperature. PVWatts defines it to be 25 °C and
        is included here for flexibility. [°C]
```

tests/test_pvsystem.py
```diff
@@ -2181,6 +2181,46 @@ def test_pvwatts_dc_series():
...
+    out = []
+    expected = [9.0625, 120.25, 0, 123.75, 9.0625, 120.0, 0, 120.0]
+    for cap_adjustment in cap_adjustments:
+        for k in ks:
+            for irrad in irrad_trans:
+                out.append(pvsystem.pvwatts_dc(irrad, temp_cell, 100, -0.003,
+                                               25, k, cap_adjustment))
+    assert_allclose(out, expected)
+
+
```
RDaxini, line 2219: See https://github.com/pvlib/pvlib-python/pull/2569#discussion_r2440438584; I think this test should help. The rest look good already, nice job!
```suggestion

def test_pvwatts_dc_cap_adjustment_scalar_above_1000():
    irrad = 1200
    temp_cell = 25
    k = 0.01
    cap_adjustment = True

    out = pvsystem.pvwatts_dc(irrad, temp_cell, 100, -0.003, 25, k, cap_adjustment)
    expected = 120.0

    assert_allclose(out, expected)
```

williamhobbs: See my changes to `test_pvwatts_dc_with_k_and_cap_adjustment()`. It now includes irradiance > 1000 and `cap_adjustment=True`. I think I made those changes before seeing your comments here. 

Does that cover it?

williamhobbs:
A general comment: after other Will H. pointed out the issue with returning the same object type that was input, my code has gotten a lot harder to follow, with lots of nested ifs, slicing, etc. If anyone has suggestions for cleaning it up, I'm very interested. A few specific thoughts/questions:
- I'm using `if hasattr(effective_irradiance, '__len__')` to see if the inputs are a series or array.
  - Is there a better way to do this?
  - This is checked 3 times (before 3 separate if steps) - should I check it once and then run those three steps? See example below. (_edit: I went ahead and made this change. Looks much better to me._) (_2nd edit: implemented new changes suggested by Kevin, not shown in this comment._)
- I calculate `pdc_marion` the same as `pdc` for PVWattsv5, then apply the error adjustments. This was originally easy for me to make sense of, but now it looks a lot messier... 

Example of reorganizing the ifs. Is this better? (_edit: I went ahead and made this change. Looks much better to me._) (_2nd edit: implemented new changes suggested by Kevin, not shown in this comment._)

<details>

<summary>click to expand and see code</summary>

```python
    # apply Marion's correction if k is anything but zero
    if k is not None:
        err_1 = k * (1 - (1 - effective_irradiance / 200)**4)
        err_2 = k * (1000 - effective_irradiance) / (1000 - 200)

        # if input is Series or array
        if hasattr(effective_irradiance, '__len__'):
            # precalculate pdc before error adjustments
            pdc_marion = (effective_irradiance * 0.001 * pdc0 *
                (1 + gamma_pdc * (temp_cell - temp_ref)))
            
            # apply error adjustments
            pdc_marion[effective_irradiance <= 200] = (
                pdc[effective_irradiance <= 200] -
                (pdc0 * err_1[effective_irradiance <= 200]))
            pdc_marion[effective_irradiance > 200] = (
                pdc[effective_irradiance > 200] -
                (pdc0 * err_2[effective_irradiance > 200]))
            
            # "cap" Marion's correction at 1000 W/m^2
            if cap_adjustment:
                pdc_marion[effective_irradiance >= 1000] = (
                    pdc[effective_irradiance >= 1000])
            
            # set negative power to zero
            pdc_marion[pdc_marion < 0] = 0
        
        # else (input is scalar)
        else:
            # apply error adjustments
            if effective_irradiance <= 200:
                pdc_marion = pdc - (pdc0 * err_1)
            elif effective_irradiance > 200:
                pdc_marion = pdc - (pdc0 * err_2)

            # "cap" Marion's correction at 1000 W/m^2
            if cap_adjustment:
                if effective_irradiance >= 1000:
                    pdc_marion = pdc
            
            # set negative power to zero
            if pdc_marion < 0:
                pdc_marion = 0

        pdc = pdc_marion
```

</details>

kandersolar:
> A general comment: after other Will H. pointed out the issue with returning the same object type that was input, my code has gotten a lot harder to follow, with lots of nested ifs, slicing, etc. If anyone has suggestions for cleaning it up, I'm very interested.

Having different computation paths for different input types results in quite a lot of code, and duplicating the math also introduces the risk of accidentally performing different computations for different input types.

What we do elsewhere in pvlib is to have only one computation path that operates on numpy arrays (the lingua franca) and then fix up the types at the very end.  Something like this (please check that the computations are correct; this is just for illustration):

```python
def pvwatts_dc(...):
    pdc = (effective_irradiance * 0.001 * pdc0 *
           (1 + gamma_pdc * (temp_cell - temp_ref)))

    # apply Marion's correction if k is anything but zero
    if k is not None:

        # preserve input types
        index = pdc.index if isinstance(pdc, pd.Series) else None
        is_scalar = np.isscalar(pdc)
        
        # calculate error adjustments
        err_1 = k * (1 - (1 - effective_irradiance / 200)**4)
        err_2 = k * (1000 - effective_irradiance) / (1000 - 200)
        err = np.where(effective_irradiance <= 200, err_1, err_2)
        if cap_adjustment:
            err = np.where(effective_irradiance >= 1000, 0, err)

        pdc = pdc - pdc0 * err

        if index is not None:
            pdc = pd.Series(pdc, index=index)
        elif is_scalar:
            pdc = float(pdc)

    return pdc
```

williamhobbs:
> one computation path that operates on numpy arrays

Thanks, @kandersolar! This is the exact help I needed.

pvlib/pvsystem.py
wholmgren: sorry for slow response on this one. thanks @williamhobbs for persevering and @kandersolar for the assist

pvlib/pvsystem.py
wholmgren: There is precedent for optional deviations like this in other parts of pvlib. No one has taken the opportunity to comment or thumbs up so I will mark as resolved!

tests/test_pvsystem.py
```diff
@@ -2181,6 +2181,58 @@ def test_pvwatts_dc_series():
...
+    cap_adjustments = [False, True]
+    out = []
+    expected = [9.0625, 120.25, 0, 123.75, 9.0625, 120.0, 0, 120.0]
+    for cap_adjustment in cap_adjustments:
+        for k in ks:
+            for irrad in irrad_trans:
+                out.append(pvsystem.pvwatts_dc(irrad, temp_cell, 100, -0.003,
+                                               25, k, cap_adjustment))
+    assert_allclose(out, expected)
+
```
wholmgren, line 2218: I think this test can be greatly simplified... What exactly is the intention here? Testing with both k and cap adjustment is supplied? Why do we need two `k` values for that? I think the default cap_adjustment case is already covered in other tests so not sure why it's needed here.

tests/test_pvsystem.py
```diff
@@ -2181,6 +2181,58 @@ def test_pvwatts_dc_series():
     assert_series_equal(expected, out)
 
 
+def test_pvwatts_dc_scalars_with_k():
+    expected = 8.9125
+    out = pvsystem.pvwatts_dc(100, 30, 100, -0.003, 25, 0.01)
```
wholmgren: generally preferable to call keyword arguments as keyword arguments and only supply the one that deviates from default. So in this case you'd have `pvsystem.pvwatts_dc(100, 30, 100, -0.003, k=0.01)`

wholmgren: And similar changes below

tests/test_pvsystem.py
wholmgren: maybe I am missing something in https://github.com/pvlib/pvlib-python/pull/2569/files#r2440448406?

tests/test_pvsystem.py
williamhobbs: Gotcha. I think I got everything.

tests/test_pvsystem.py
williamhobbs: I was going for a range of irradiance with and without and `cap_adjustment`, and a range of `k` values including a `k` and `irrad` combination that would have returned negative power without the `pdc = np.where(pdc < 0, 0, pdc)` line. 

Let me know what you think about https://github.com/pvlib/pvlib-python/pull/2569/commits/86eb06ea92eb68490f0afd317338b249e401eb6b.

pvlib/pvsystem.py
```diff
@@ -2878,53 +2878,117 @@ def scale_voltage_current_power(data, voltage=1, current=1):
...
     temp_ref: numeric, default 25.0
-        Cell reference temperature. PVWatts defines it to be 25 C and
-        is included here for flexibility. [C]
+        Cell reference temperature. PVWatts defines it to be 25 °C and
+        is included here for flexibility. [°C]
+    k: numeric, optional
+        Irradiance correction factor, defined in [2]_. Typically positive.
+        [unitless]
+    cap_adjustment: Boolean, default False
+        If True, apply the optional adjustment at and below 1000 Wm⁻²
```
```suggestion
        If True, only apply the optional adjustment at and below 1000 Wm⁻²
```

pvlib/pvsystem.py
```diff
@@ -2878,53 +2878,117 @@ def scale_voltage_current_power(data, voltage=1, current=1):
...
+
+    This model has also been referred to as the power temperature coefficient
+    model.
+
+    This function accepts an optional irradiance adjustment factor, `k`, based
+    on [2]_. This applies a piece-wise adjustment to power based on irradiance,
+    where `k` is the reduction in actual power at 200 Wm⁻² relative to power
+    calculated at 200 Wm-2 as 0.2*`pdc0`. For example, a 500 W module that
+    produces 95 W at 200 Wm⁻² (a 5% relative reduction in efficiency) would
+    have a value of `k` = 0.01.
```
```suggestion
    An optional adjustment can be applied to :math:`P_{dc}` as described in [2]_.
    The adjustment accounts for the variation in module efficiency with
    irradiance. The piece-wise adjustment to power is parameterized by `k`,
    where `k` is the reduction in actual power at 200 Wm⁻² relative to power
    calculated at 200 Wm⁻² as 0.2*`pdc0`. For example, a module that
    is rated at 500 W at STC but produces 95 W at 200 Wm⁻² 
    (a 5% relative reduction in efficiency) would
    have a value of `k` = 0.01.
```

pvlib/pvsystem.py
```diff
@@ -2878,53 +2878,117 @@ def scale_voltage_current_power(data, voltage=1, current=1):
...
+
+    .. math::
+
+        k=\frac{0.2P_{dc0}-P_{200}}{P_{dc0}}
+
+    For positive `k` values, and `k` is typically positive, this adjustment
+    increases relative efficiency when irradiance is above 1000 Wm⁻². This may
```
```suggestion
    would also increase relative efficiency when irradiance is above 1000 Wm⁻².
    This may
```

cwhanse (approved):

PR #2571, opened 2025-10-10
Use chandrupatla to find MPP in lambertw method

### LINKED ISSUE #2497: Use scipy's new `optimize.elementwise` functions

cwhanse:
- [X] Closes #2497 
 - ~~[ ] Tests added~~
 - ~~[ ] Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.~~
 - [X] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [X] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

Uses `'chandrupatla'` when scipy>=1.15. Existing tests are good enough.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.13.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.13.2.rst
@@ -28,7 +28,7 @@ Enhancements
   :py:func:`~pvlib.singlediode.bishop88_v_from_i`, and
   :py:func:`~pvlib.singlediode.bishop88_i_from_v`. (:issue:`2497`, :pull:`2498`)
 * Accelerate :py:func:`~pvlib.pvsystem.singlediode` when scipy>=1.15 is
-  installed. (:issue:`2497`, :pull:`XXXX`)
+  installed. (:issue:`2497`, :pull:`2571`)
 
 
 Documentation


cwhanse:
Restart of #2567

cwhanse:
Pursuing the test failures, one condition tested is `photocurrent=0.` For that input, `i_mp=nan` is returned when the scipy method is used (https://github.com/pvlib/pvlib-python/pull/2567#discussion_r2411690225).

`nan` is returned because the convergence fails on the first iteration, due to violation of the bracket condition. Quote from the scipy documentation for `find_minimum`:

>  requires argument init to provide a three-point minimization bracket: x1 < x2 < x3 such that func(x1) >= func(x2) <= func(x3), where one of the inequalities is strict.

Here, `init = (0., 0.8*v_oc, v_oc)` and `func=_vmp_opt` in the PR.

Evaluating at `photocurrent=0.` and `v_oc=init`, I get `_vmp_opt(init, 0., ...) = array([-0.00000000e+00,  2.18952885e-48,  2.73691106e-48])`. Note that the negative power is actually positive. The func values at this bracket fail the check.

Digging deeper, `_lambertw_i_from_v(init, 0., ...) = array([0., -4.1359e-25, -4.1359e-25])` the negative currents are the root of the problem.

We could set small negative current to zero in `lambertw_i_from_v` But the MPP with `chandralupta` would still fail since `photocurrent=0.` since `func(init)` would be `array([-0., -0., -0.])` and one of the inequalities has to be strict.

I think intercepting small photocurrent and skipping the MPP (or entire IV curve) calculation is better.

Your thoughts?

kandersolar:
> I think intercepting small photocurrent and skipping the MPP (or entire IV curve) calculation is better.

I think this is probably fine, although the indexing is going to make the code clunky.

A possible alternative: instead of `init = (0., 0.8*v_oc, v_oc)`, how about `init = (-1, 0.8*v_oc, v_oc)`?  For `photocurrent=0`, the MPP is at `v=i=0`, and if we're not satisfying the bracket condition for the right pair, how about moving the left side out so that it's satisfied by the left pair?

cwhanse:
> `init = (-1, 0.8*v_oc, v_oc)`?

This is the way. And reverting from `i_mp = p_mp / v_mp` which introduces a divide by 0. and a nan thus a test failure...

cwhanse:
Weird test failure seemingly unrelated to these code changes.

pvlib/pvsystem.py
```diff
@@ -2534,28 +2534,35 @@ def singlediode(photocurrent, saturation_current, resistance_series,
     explicit function of :math:`V=f(I)` and :math:`I=f(V)` as shown in [2]_.
 
     If the method is ``'newton'`` then the root-finding Newton-Raphson method
-    is used. It should be safe for well behaved IV-curves, but the ``'brentq'``
-    method is recommended for reliability.
+    is used. It should be safe for well-behaved IV curves, otherwise the
+    ``'chandralupta``` or ``'brentq'`` methods are recommended for reliability.
```
kandersolar:
```suggestion
    ``'chandrupatla``` or ``'brentq'`` methods are recommended for reliability.
```

pvlib/pvsystem.py
```diff
@@ -2534,28 +2534,35 @@ def singlediode(photocurrent, saturation_current, resistance_series,
...
 
     If the method is ``'brentq'`` then Brent's bisection search method is used
     that guarantees convergence by bounding the voltage between zero and
-    open-circuit.
+    open-circuit. ``'brentq'`` is generally slower than the other options.
+
+    If the method is ``'chandralupta'`` then Chandralupta's method is used
```
kandersolar:
```suggestion
    If the method is ``'chandrupatla'`` then Chandrupatla's method is used
```

pvlib/pvsystem.py
```diff
@@ -2565,8 +2572,9 @@ def singlediode(photocurrent, saturation_current, resistance_series,
         out = _singlediode._lambertw(*args)
         points = out[:7]
     else:
-        # Calculate points on the IV curve using either 'newton' or 'brentq'
-        # methods. Voltages are determined by first solving the single diode
+        # Calculate points on the IV curve using Bishop's algorithm and solving
+        # with 'newton', 'brentq' or 'chandralupta' method.
```
kandersolar:
```suggestion
        # with 'newton', 'brentq' or 'chandrupatla' method.
```

kandersolar:
> Weird test failure seemingly unrelated to these code changes.

I think it is related.  The failing test has negative `v_mp` values (e.g. -5e-16).  Presumably that wasn't possible when the MPP search was bounded by zero.

Would truncating negative `v_mp` to zero after the MPP search make sense?  Not sure what is best here.

cwhanse:
Thanks, I was missing how the test failure related to the changes in this PR. That fixture evaluates the CEC model via the model chain.

The failure exposed additional issues however:

1. The test executes a modelchain which, for input `weather` and for `inverter_model='adr'` produces this output with numpy 2.2 and scipy 1.5.3:

```
weather
                           ghi  dni  dhi
2016-01-01 12:00:00-07:00  500  800  100
2016-01-01 18:00:00-07:00    0    0    0

mc.results.dc[['v_mp', 'p_mp']]
                                   v_mp          p_mp
2016-01-01 12:00:00-07:00  4.062850e+01  1.679352e+02
2016-01-01 18:00:00-07:00 -5.205590e-16  2.691227e-41

mc.results.ac
2016-01-01 12:00:00-07:00     NaN
2016-01-01 18:00:00-07:00     NaN

```

The test fails because of the NaN at the second time step in the AC series:
```
    assert mc.results.ac.iloc[1] < 1
```
Apparently, prior to numpy 2.X, the assert was OK (the ubuntu conda min test passes). For the life of me I can't see why this test wasn't failing on the last few PRs that also use numpy 2.X.

The NaNs are introduced in the AC series in `inverter.adr` at this line: https://github.com/pvlib/pvlib-python/blob/28ea577563e0e51f15dc9e82dce5773fb1c251f6/pvlib/inverter.py#L321

Upstream, there are numeric values in `power_ac`.

Why is a NaN also in the first line, where the DC values look reasonable? Because the test system is malformed. The test system pairs a single 220W module with a 3kW inverter.

I think the fixes here is to repair the test: replace the above assert so that it handles NaN, and correct this fixture so that a reasonable AC power is output for the first time step: https://github.com/pvlib/pvlib-python/blob/28ea577563e0e51f15dc9e82dce5773fb1c251f6/tests/test_modelchain.py#L129

cwhanse:
Depends on #2577 to resolve test failure

kandersolar:
I don't think that explanation is quite right.  I expect it's not the numpy version that matters, but rather scipy's, since that's what determines which of the two optimization branches (chadrupatla vs golden section) is taken.  Here is what I think is happening:

1. With the suggested change to `init = (-1, 0.8*v_oc, v_oc)`, the chandrupatla optimization can produce negative `v_mp`.  For the second timestamp in this test, where irradiance is zero, it is producing `v_mp=-5.205590e-16`.  This is different from the result of the golden section search, which still uses the original `init = (0, 0.8*v_oc, v_oc)` and produces `v_mp=0`.
2. `inverter.adr` sees that both of these voltages (`-5.205590e-16` and `0`) are outside the valid voltage range and are thus `invalid`, setting the output to NaN.  However, it has a special case for `v_dc=0` that turns it back to a non-NaN value:

https://github.com/pvlib/pvlib-python/blob/bc67019578db0795f8487d7947a8ca3708e9d9e4/pvlib/inverter.py#L317-L324

3. All test configurations EXCEPT `-min` use the latest scipy, resulting in the negative `v_mp` and eventually the NaN ac power.  On the `-min` configuration, `chandrupatla` is not available, so it uses the golden section search and eventually computes the non-NaN value as described above.

I don't think #2577 will change this story (although I agree it makes the test more coherent for the first, daytime, timestamp).

What to do about it?  Some ideas...
1. Check the output of `find_minimum`, and set any values within some `atol` of zero to zero.
2. Flip around my earlier suggestion (https://github.com/pvlib/pvlib-python/pull/2571#issuecomment-3406232685): keep the left side of the bracket at zero, but extend the right side a little bit?  For example `init = (0, 0.8*v_oc, v_oc+1)`.  
   * This might not solve the immediate problem, since then it might just produce small _positive_ `v_mp` instead of small negative `v_mp`...
3. Intercept `photocurrent==0` and skip the optimization, as you suggested earlier.  Not "clean" code, but solves the problem and is likely the fastest, assuming a typical simulation where half of the inputs are at night anyway.  And it might prevent headaches someday if the fiddling in options 1/2 stops working for some future version of scipy's `find_minimum`.
4. Decide that the negative `v_mp` is fine, and either the failing test or `inverter.adr` need to change to accommodate it. 

I am starting to (reluctantly) lean towards option 3, although I don't hate option 4.

cwhanse:
I don't hate #4 either, and prefer that to #3. I agree that #2577 doesn't fix the failing test for the 2nd timestamp, but it corrects the value at the first timestamp so that the test can examine that value instead.

cwhanse:
Codecov failure because coverage is done in an environment with scipy>=1.15, so the old golden section MPP function is not hit.

docs/sphinx/source/whatsnew/v0.13.2.rst
```diff
@@ -27,7 +27,8 @@ Enhancements
   :py:func:`~pvlib.singlediode.bishop88_mpp`,
   :py:func:`~pvlib.singlediode.bishop88_v_from_i`, and
   :py:func:`~pvlib.singlediode.bishop88_i_from_v`. (:issue:`2497`, :pull:`2498`)
-
+* Accelerate :py:func:`~pvlib.pvsystem.singlediode` when scipy>=1.15 is
+  installed. (:issue:`2497`, :pull:`XXXX`)
```
kandersolar:
```suggestion
  installed. (:issue:`2497`, :pull:`2571`)
```

adriesse:
> 4\. Decide that the negative `v_mp` is fine, and either the failing test or `inverter.adr` need to change to accommodate it.
> > 
> I am starting to (reluctantly) lean towards option 3, although I don't hate option 4.

I am not entirely up to speed on this PR, but w.r.t. the negative `v_mp` I would say it depends on whether it is just a numerical artifact or is physically plausible.  Assuming the former, why not use `np.clip()` on the result?

As for the `adr` inverter that is now in the spotlight, I had another look at it.  Actually, it would make a lot of sense to go into night mode when the dc voltage is below `v_lim_lower `.  Much better than a check for precise zero values.  Further to that point, the minimum value should perhaps be using `nanmin()` (unless the database has some zero values where the specs are unknown).

cwhanse:
> I am not entirely up to speed on this PR, but w.r.t. the negative `v_mp` I would say it depends on whether it is just a numerical artifact or is physically plausible. Assuming the former, why not use `np.clip()` on the result?

The negative voltage is a numerical artifact, when input DC is negative or very small. This PR improves how the maximum power point is found by `pvsystem.singlediode` so I did not think it appropriate to edit `inverter.adr` to get a `ModelChain` test to pass.

> As for the `adr` inverter that is now in the spotlight, I had another look at it. Actually, it would make a lot of sense to go into night mode when the dc voltage is below `v_lim_lower `. Much better than a check for precise zero values. Further to that point, the minimum value should perhaps be using `nanmin()` (unless the database has some zero values where the specs are unknown).

This all sounds reasonable to me. I think should the improvements deserve a new issue and PR.

adriesse:
> I think should the improvements deserve a new issue and PR.

Agreed.  I'm not in a position to be able to do that PR at the moment though.

PR #2573, opened 2025-10-17
Add function for accessing ERA5

### LINKED ISSUE #2420: ERA5 single-point time series database
ERA5 has recently launched a [time series-specific database ](https://cds.climate.copernicus.eu/datasets/reanalysis-era5-single-levels-timeseries?) for ERA5 which brings down request times from +10 minutes to a few seconds.

Note, that the service is currently listed as experimental. Here's the info from the ECMWF's website.

> Please be aware that this is an experimental catalogue entry which uses an alternative source for the ERA5 data designed for fast access long-timeseries for a single point, and it is not recommended for use operational systems.
> The entry may be temporarily disabled or completely deprecated at any point and it does not come with the same level of operational support as the [parent entry](https://cds.climate.copernicus.eu/datasets/reanalysis-era5-single-levels). Additionally, the formats and structure of the data files are subject to change.

LinkedIn announcement:

I plan on making a PR in the following week, but we may delay merging until the service becomes permanent.

kandersolar:
- [x] Closes #2420
 - [x] Tests added

As described in #2420, the ECMWF CDS now provides a simple and fast API for retrieving timeseries ERA5 data.  No need for any new dependencies (not even `cdsapi`), although the nature of the API requires a sequence of requests, and one poll-and-wait step, to complete.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.13.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.13.2.rst
@@ -21,6 +21,13 @@ Bug fixes
 
 Enhancements
 ~~~~~~~~~~~~
+* Add :py:func:`~pvlib.ivtools.sdm.fit_desoto_batzelis`, a function to estimate
+  parameters for the De Soto single-diode model from datasheet values. (:pull:`2563`)
+* Add :py:func:`~pvlib.singlediode.batzelis`, a function to estimate
+  maximum power, open circuit, and short circuit points using parameters for
+  the single-diode equation. (:pull:`2563`)
+* Add :py:func:`~pvlib.pvarray.batzelis`, a function to estimate maximum power
+  open circuit, and short circuit points from datasheet values. (:pull:`2563`)
 * Add ``method='chandrupatla'`` (faster than ``brentq`` and slower than ``newton``,
   but convergence is guaranteed) as an option for
   :py:func:`pvlib.pvsystem.singlediode`,
@@ -30,6 +37,8 @@ Enhancements
   :py:func:`~pvlib.singlediode.bishop88_mpp`,
   :py:func:`~pvlib.singlediode.bishop88_v_from_i`, and
   :py:func:`~pvlib.singlediode.bishop88_i_from_v`. (:issue:`2497`, :pull:`2498`)
+* Accelerate :py:func:`~pvlib.pvsystem.singlediode` when scipy>=1.15 is
+  installed. (:issue:`2497`, :pull:`2571`)
 * Add :py:func:`~pvlib.iotools.get_era5`, a function for accessing
   ERA5 reanalysis data. (:pull:`2573`)
 
@@ -57,4 +66,4 @@ Maintenance
 
 Contributors
 ~~~~~~~~~~~~
-
+* Cliff Hansen (:ghuser:`cwhanse`)
--- a/docs/sphinx/source/reference/pv_modeling/sdm.rst
+++ b/docs/sphinx/source/reference/pv_modeling/sdm.rst
@@ -17,6 +17,7 @@ Functions relevant for single diode models.
    pvsystem.v_from_i
    pvsystem.max_power_point
    ivtools.sdm.pvsyst_temperature_coeff
+   singlediode.batzelis
 
 Low-level functions for solving the single diode equation.
 
@@ -37,3 +38,4 @@ Functions for fitting diode models
     ivtools.sde.fit_sandia_simple
     ivtools.sdm.fit_cec_sam
     ivtools.sdm.fit_desoto
+    ivtools.sdm.fit_desoto_batzelis
--- a/docs/sphinx/source/reference/pv_modeling/system_models.rst
+++ b/docs/sphinx/source/reference/pv_modeling/system_models.rst
@@ -55,3 +55,11 @@ PVGIS model
    :toctree: ../generated/
 
     pvarray.huld
+
+Other
+^^^^^
+
+.. autosummary::
+   :toctree: ../generated/
+
+    pvarray.batzelis
--- a/pvlib/ivtools/sdm/__init__.py
+++ b/pvlib/ivtools/sdm/__init__.py
@@ -10,7 +10,8 @@ from pvlib.ivtools.sdm.cec import (  # noqa: F401
 
 from pvlib.ivtools.sdm.desoto import (  # noqa: F401
     fit_desoto,
-    fit_desoto_sandia
+    fit_desoto_batzelis,
+    fit_desoto_sandia,
 )
 
 from pvlib.ivtools.sdm.pvsyst import (  # noqa: F401
--- a/pvlib/ivtools/sdm/desoto.py
+++ b/pvlib/ivtools/sdm/desoto.py
@@ -2,6 +2,7 @@ import numpy as np
 
 from scipy import constants
 from scipy import optimize
+from scipy.special import lambertw
 
 from pvlib.ivtools.utils import rectify_iv_curve
 from pvlib.ivtools.sde import _fit_sandia_cocontent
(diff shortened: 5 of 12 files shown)


AdamRJensen:
Discussion on available parameter though the ERA5 time series API can be found [here](https://forum.ecmwf.int/t/new-dataset-published-in-cds-era5-hourly-time-series-data-on-single-levels-from-1940-to-present/11919/13).

AdamRJensen:
Right now you can't pass pvlib variable names as strings:

This doesn't work:
```python
variables = 'ghi'

data, meta = pvlib.iotools.get_era5(
    latitude, longitude, start, end, variables, api_key, timeout=60)
```

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,195 @@
...
+    .. [3] https://confluence.ecmwf.int/pages/viewpage.action?pageId=505390919
+    """  # noqa: E501
+    start = pd.to_datetime(start).strftime("%Y-%m-%d")
+    end = pd.to_datetime(end).strftime("%Y-%m-%d")
+
+    headers = {'PRIVATE-TOKEN': api_key}
+
+    # allow variables to be specified with pvlib names
+    reverse_map = {v: k for k, v in VARIABLE_MAP.items()}
+    variables = [reverse_map.get(k, k) for k in variables]
```
AdamRJensen, line 127:
```suggestion
    # allow single variables to be specified as string
    variables = [variables] if isinstance(variables, str) else variables
    # allow variables to be specified with pvlib names
    reverse_map = {v: k for k, v in VARIABLE_MAP.items()}
    variables = [reverse_map.get(k, k) for k in variables]
```

Right now you can't pass pvlib variable names as strings. For example, this doesn't work:

```python
variables = 'ghi'

data, meta = pvlib.iotools.get_era5(
    latitude, longitude, start, end, variables, api_key, timeout=60)
```

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,195 @@
...
+    'sp': 'pressure',
+    'ssrd': 'ghi',
+    'tp': 'precipitation',
+
+    # long names
+    '2m_dewpoint_temperature': 'temp_dew',
+    '2m_temperature': 'temp_air',
+    'surface_pressure': 'pressure',
+    'surface_solar_radiation_downwards': 'ghi',
+    'total_precipitation': 'precipitation',
```
AdamRJensen, line 21: I suggest adding all relevant ERA5 names here, even if they aren't yet supported by the time series end point (could add a comment mentioned this). This way we would avoid introducing breaking changes when new parameters are added:

- 10m v-component of wind?
- 10m u-component of wind?
- Clear-sky direct solar radiation at surface
- TOA incident solar radiation
- Total sky direct solar radiation at surface

docs/sphinx/source/whatsnew/v0.13.2.rst
```diff
@@ -27,7 +27,8 @@ Enhancements
   :py:func:`~pvlib.singlediode.bishop88_mpp`,
   :py:func:`~pvlib.singlediode.bishop88_v_from_i`, and
   :py:func:`~pvlib.singlediode.bishop88_i_from_v`. (:issue:`2497`, :pull:`2498`)
-
+* Add :py:func:`~pvlib.iotools.get_era5`, a function for accessing
+  ERA-5 reanalysis data. (:pull:`2573`)
```
AdamRJensen:
```suggestion
  ERA5 reanalysis data. (:pull:`2573`)
```

pvlib/iotools/ecmwf.py
```diff
@@ -0,0 +1,195 @@
...
+        metadata['latitude'] = df['latitude'].values[0]
+        metadata['longitude'] = df['longitude'].values[0]
+
+    df.index = pd.to_datetime(df['valid_time']).dt.tz_localize('UTC')
+    df = df.drop(columns=['valid_time', 'latitude', 'longitude'])
+
+    if map_variables:
+        # convert units and rename
+        for shortname in df.columns:
+            converter = UNITS[shortname]
```
AdamRJensen:
```suggestion
        for shortname in df.columns:
            converter = UNITS.get(shortname, same)
```
This seems more robust in case new variables are added to the API end point in the future. I would then remove the "pass" entries in the ``UNITS`` dict.

pvlib/iotools/ecmwf.py
```diff
@@ -0,0 +1,195 @@
...
+        First day of the requested period.
+    end : datetime like or str
+        Last day of the requested period.
+    variables : list of str
+        List of variable names to retrieve.  See [1]_ for options.
+    api_key : str
+        ECMWF CDS API key.
+    map_variables : bool, default True
+        When true, renames columns of the DataFrame to pvlib variable names
+        where applicable. See variable :const:`VARIABLE_MAP`.
```
AdamRJensen:
```suggestion
        When true, renames columns of the DataFrame to pvlib variable names
        where applicable. Also converts units of some variables. See variable
        :const:`VARIABLE_MAP` and :const:`UNITS`.
```

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,195 @@
...
+        Metadata.
+
+    References
+    ----------
+    .. [1] https://cds.climate.copernicus.eu/
+    .. [2] https://cds.climate.copernicus.eu/datasets/reanalysis-era5-single-levels-timeseries?tab=overview
+    .. [3] https://confluence.ecmwf.int/pages/viewpage.action?pageId=505390919
+    """  # noqa: E501
+    start = pd.to_datetime(start).strftime("%Y-%m-%d")
+    end = pd.to_datetime(end).strftime("%Y-%m-%d")
```
AdamRJensen: Should these be localized to UTC first?

pvlib/iotools/ecmwf.py
```diff
@@ -0,0 +1,195 @@
...
+                'https://cds.climate.copernicus.eu/requests?tab=all'
+            )
+            raise Exception(msg)
+
+        num_polls += 1
+        if num_polls * poll_interval > timeout:
+            raise Exception(
+                'Request timed out. Try increasing the timeout parameter or '
+                'reducing the request size.'
+            )
```
AdamRJensen: Could we make this a timeout exception?

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,195 @@
...
+            raise Exception(msg)
+
+        num_polls += 1
+        if num_polls * poll_interval > timeout:
+            raise Exception(
+                'Request timed out. Try increasing the timeout parameter or '
+                'reducing the request size.'
+            )
+
+        time.sleep(1)
```
AdamRJensen, line 172: Could consider reducing this to 0.5

AdamRJensen (approved):
This is a phenomenal contribution 🥳 

I do suggest renaming the module to "era5" instead of "ecmwf" as I foresee we might add other datasets such as the IFS forecasts, where it would be preferred to have the code in different files.

pvlib/iotools/era5.py
kandersolar: > allow single variables to be specified as string

I'm not totally opposed to this, but this usage (single variables as string) isn't supported elsewhere in iotools, is it?  I suggest we think about whether we want to allow this across iotools before adding it here.

pvlib/iotools/era5.py
kandersolar: If they do add new names in the future and we have to add them to our map, I'd be happy to call it a "bugfix" rather than a breaking change.

But if we want to do it now, can we predict the names with confidence?  i.e., are the short and long names the same between the time series end point and the main ERA5 dataset?

kandersolar: And, the U/V wind components do not have a pvlib analog anyway

pvlib/iotools/era5.py
AdamRJensen: I had misread the docstring as saying "list OR str" but in fact it was "list of str". So this can be ignored.

pvlib/iotools/era5.py
AdamRJensen: The API long names seem consistent as I would expect - it is after all the same dataset just different chunking. ~~There's a database of the parameter names and the relations between long and short names: https://codes.ecmwf.int/grib/param-db - short names doesn't seem to match~~. *Edit: short names might be different*

image

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,196 @@
...
+    't2m': k_to_c,
+    'msl': same,
+    'sst': k_to_c,
+    'skt': k_to_c,
+    'sp': same,
+    'ssrd': j_to_w,
+    'strd': j_to_w,
+    'tp': m_to_cm,
+}
+
```
wholmgren: I'd prefix all of this with `_` to discourage others from using it, even within pvlib.

pvlib/iotools/era5.py
```diff
@@ -0,0 +1,196 @@
...
+    latitude : float
+        In decimal degrees, north is positive (ISO 19115).
+    longitude: float
+        In decimal degrees, east is positive (ISO 19115).
+    start : datetime like or str
+        First day of the requested period.
+    end : datetime like or str
+        Last day of the requested period.
+    variables : list of str
+        List of variable names to retrieve.  See [1]_ for options.
```
adriesse:
```suggestion
        List of variable names to retrieve, for example ``['ghi', 
        'temp_air']``. See [1]_ for additional options.
```

PR #2612, opened 2025-12-06
Fix noct_sam() 

### LINKED ISSUE #1462: NOCT_SAM may not handle effective irradiance correctly
**Describe the problem**
The noct_sam temperature model has an `effective_irradiance` input, which would normally have a value less than `poa_global`, mainly due to reflection losses.  This is implemented internally as an adjustment to the `transmittance_absorptance` parameter.  However, this  adjustment produces some questionable results.

**Code fragment to demonstrate the problem**
```
from pvlib.temperature import noct_sam

gpoa = 800
geff = gpoa * 0.75

wind_speed_10m = 1.0 / 0.51

NOCT = 41.6

# open circuit and max power module conditions for reuse
module_oc =dict(noct=NOCT, module_efficiency=0.00)
module_mp =dict(noct=NOCT, module_efficiency=0.20)

weather0 = dict(poa_global=gpoa, 
                temp_air=20, 
                wind_speed=wind_speed_10m,
                )
print()
print('mod temp under gpoa')
print('oc:', noct_sam(**weather0, **module_oc))
print('mp:', noct_sam(**weather0, **module_mp))

weather1 = dict(poa_global=geff, 
                temp_air=20, 
                wind_speed=wind_speed_10m,
                )
print()
print('mod temp under geff (what I would expect to be correct)')
print('oc:', noct_sam(**weather1, **module_oc))
print('mp:', noct_sam(**weather1, **module_mp))

# use the effective_irradiance input
weather2 = dict(poa_global=gpoa, 
                temp_air=20, 
                wind_speed=wind_speed_10m,
                effective_irradiance=geff,
                )
print()
print('mod temp under gpoa using separate geff input (different/incorrect)')
print('oc:', noct_sam(**weather2, **module_oc))
print('mp:', noct_sam(**weather2, **module_mp))

# replicate the tau_alpha adjustment that is inside noct_sam
weather3 = dict(poa_global=gpoa, 
                temp_air=20, 
                wind_speed=wind_speed_10m,
                transmittance_absorptance=0.9 * geff / gpoa,
                )
print()
print('mod temp under gpoa with tau_alpha adjusted externally (same problem)')
print('oc:', noct_sam(**weather3, **module_oc))
print('mp:', noct_sam(**weather3, **module_mp))

```
**Code fragment output**
```
mod temp under gpoa
oc: 41.6
mp: 36.8

mod temp under geff (what I would expect to be correct)
oc: 36.2
mp: 32.6

mod temp under gpoa using separate geff input (different/incorrect)
oc: 41.6
mp: 35.2

mod temp under gpoa with tau_alpha adjusted externally (same problem)
oc: 41.6
mp: 35.2
```
**Comments on output**
Note that under open circuit conditions the module temperature remains at 41.6 in the third and fourth cases despite the reduction in irradiance, which is not logical. The biggest surprise comes when you set `geff` to small values, like 20% of `gpoa` because the module temperature then goes below ambient.

**Additional context**
The `noct_sam` function in pvlib appears to correspond very closely to the code in SAM itself, so this problem would exist there too.  But it is much easier to demonstrate it here.

**Solution**
Reflected irradiance does not contribute to heating or electricity generation, so a single irradiance input for this thermal model is enough.

Chirag3841:
@cwhanse 

 - [x] Closes #1462
- [x] I am familiar with the contributing guidelines
- [x] Tests added  
- [ ] Updates entries in docs reference  
- [x] Adds description to what's new  
- [x] New code is fully documented . 

The problem was that 'noct_sam' did not correctly account for the
'effective_irradiance' parameter. The original implementation always used
'poa_global' when computing the initial temperature rise, which caused:

- Open-circuit modules (efficiency = 0) to ignore 'effective_irradiance'
- Very low 'effective_irradiance' values to produce unrealistic cooling
- Results inconsistent with SAM under certain option combinations

This PR updates 'noct_sam' so that:

- 'effective_irradiance' is used consistently when provided
- Negative or very small irradiance values are clipped to zero
- Computation now matches SAM’s documented behaviour
- I updated the noct_sam function to correctly incorporate effective irradiance into the temperature model.

New tests were added to validate:

- Temperature decreases with reduced effective irradiance
- Open-circuit modules still respond to effective irradiance changes
- Temperature never falls below ambient for extremely small irradiance

These tests pass under the main branch, confirming correct behaviour.

### DIFF
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -20,33 +20,22 @@ from pvlib.tools import _build_kwargs
 
 from pvlib._deprecation import deprecated
 
-# Keys used to detect input data and assign values to the appropriate
-# ModelChain attributes.
+# keys that are used to detect input data and assign data to appropriate
+# ModelChain attribute
+# for ModelChain.weather
+WEATHER_KEYS = ('ghi', 'dhi', 'dni', 'wind_speed', 'temp_air',
+                'precipitable_water')
 
-# Weather-related input columns for ModelChain.weather
-WEATHER_KEYS = (
-    'ghi',               # Global Horizontal Irradiance (W/m^2)
-    'dhi',               # Diffuse Horizontal Irradiance (W/m^2)
-    'dni',               # Direct Normal Irradiance (W/m^2)
-    'wind_speed',        # Wind speed (m/s)
-    'temp_air',          # Ambient air temperature (°C)
-    'precipitable_water' # Column precipitable water (cm)
-)
+# for ModelChain.total_irrad
+POA_KEYS = ('poa_global', 'poa_direct', 'poa_diffuse')
 
-# Plane-of-array irradiance input columns for ModelChain.total_irrad
-POA_KEYS = (
-    'poa_global',   # Total plane-of-array irradiance (W/m^2)
-    'poa_direct',   # Direct POA irradiance (W/m^2)
-    'poa_diffuse'   # Diffuse POA irradiance (W/m^2)
-)
+# Optional keys to communicate temperature data. If provided,
+# 'cell_temperature' overrides ModelChain.temperature_model and sets
+# ModelChain.cell_temperature to the data. If 'module_temperature' is provided,
+# overrides ModelChain.temperature_model with
+# pvlib.temperature.sapm_cell_from_module
+TEMPERATURE_KEYS = ('module_temperature', 'cell_temperature')
 
-# Temperature-related optional input columns for ModelChain
-TEMPERATURE_KEYS = (
-    'module_temperature', # Back-surface module temperature (°C)
-    'cell_temperature',   # Direct cell temperature input (°C)
-)
-
-# All supported input keys combined
 DATA_KEYS = WEATHER_KEYS + POA_KEYS + TEMPERATURE_KEYS
 
 # these dictionaries contain the default configuration for following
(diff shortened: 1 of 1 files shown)


Chirag3841:
@cwhanse 
Kindly review the updated implementation.

cwhanse:
@Chirag3841 please merge the current main, this PR is missing the changes made in #2605

cwhanse:
I don't think we want to patch `noct_sam` this way. My thought is that we replace `poa_global` with `effective_irradiance` in the first arg position, drop `effective_irradiance=None`, and do away with the POA ratio internally. That matches the SAM calculation. The SAM function also handles calculation of effective irradiance, which isn't the job of this function in pvlib.

Replacing the argument requires a deprecation. @Chirag3841 if the deprecation machinery is a bit much, OK to close this PR and thanks for attempting the fix.

Chirag3841:
@cwhanse 
I tried the approach where effective_irradiance=None defaults to poa_global and emits a deprecation warning otherwise.
However, this caused ambiguity in the function signature and resulted in multiple failures across ModelChain, PVSystem, and tests due to mixed positional and keyword arguments.
I now understand that keeping both poa_global and effective_irradiance in the same function is not a good approach.
I will revert this logic and instead keep noct_sam aligned with SAM by using effective_irradiance as the first argument only, and handle backward compatibility via a separate deprecated wrapper function.

PR #2618, opened 2025-12-10
Fix Fuentes int input truncation by forcing float dtype

### LINKED ISSUE #2608: Integer/float bug in `pvlib.temperature.fuentes`
**Describe the bug**
In the temperature.py module [here](https://github.com/pvlib/pvlib-python/blob/6c93b716c5762c2e2c2334926e3e3b832b324350/pvlib/temperature.py#L886), pvlib is using `np.zeros_like` which crucially returns the same type than the input. 

In the unit tests, pvlib uses int instead of float values [here](https://github.com/pvlib/pvlib-python/blob/6c93b716c5762c2e2c2334926e3e3b832b324350/tests/test_temperature.py#L268). This leads to NumPy truncating the first float result produced (`321.19 -> 321`). 

The final subtraction `tmod_array - 273.15` then converts to a float, giving `321 - 273.15 = 47.85`

That means, the results of the Fuentes functions will differ, depending on whether the input is an int or a float.

**To Reproduce**
Run the `test_fuentes_timezone` unit test with ints and then again with floats

**Expected behavior**
Ints and floats should produce the same results if they have the same value (i.e., 100 vs 100.0)

**Screenshots**
Image

kumaradityaapril:
Closes #2608 

## Summary

This PR fixes a dtype-dependent bug in `pvlib.temperature.fuentes` where integer `poa_global` inputs could lead to truncated module temperatures and slightly different results compared to float inputs with the same numeric values.

## Changes

- In `pvlib.temperature.fuentes`:
  - Changed `tmod_array = np.zeros_like(poa_global)` to `tmod_array = np.zeros_like(poa_global, dtype=float)` so internal temperature calculations always use a float dtype.

- In `tests/test_temperature.py`:
  - Added `test_fuentes_int_float_consistency`, which calls `fuentes` with both integer and float `poa_global` series (same values, datetime index) and asserts that the results match within floating-point tolerance.

## Rationale

Previously, when `poa_global` was an integer Series, `tmod_array` was also integer, so intermediate temperatures such as `321.19 K` were truncated to `321` when stored. Converting to Celsius (`tmod_array - 273.15`) produced a different value than for float inputs.

This change ensures that physically identical inputs (`100` vs `100.0`) produce identical Fuentes outputs.

## Testing

- `pytest tests/test_temperature.py::test_fuentes -v`
- `pytest tests/test_temperature.py::test_fuentes_int_float_consistency -v`

### DIFF
--- a/pvlib/temperature.py
+++ b/pvlib/temperature.py
@@ -884,7 +884,7 @@ def fuentes(poa_global, temp_air, wind_speed, noct_installed, module_height=5,
     windmod_array = wind_speed * (module_height/wind_height)**0.2 + 1e-4
 
     tmod0 = 293.15
-    tmod_array = np.zeros_like(poa_global)
+    tmod_array = np.zeros_like(poa_global, dtype = float)
 
     iterator = zip(tamb_array, sun_array, windmod_array, tsky_array,
                    timedelta_hours)
--- a/tests/test_temperature.py
+++ b/tests/test_temperature.py
@@ -260,6 +260,21 @@ def test_fuentes(filename, inoct):
     assert night_difference.max() < 6
     assert night_difference.min() > 0
 
+def test_fuentes_int_float_consistency():
+    """Test that int and float inputs give identical results after dtype fix."""
+    index = pd.date_range('2020-01-01', periods=3, freq='h')
+    
+    poa_int = pd.Series([800, 800, 800], index=index, dtype=int)
+    poa_float = pd.Series([800.0, 800.0, 800.0], index=index)
+    temp_air = pd.Series([25.0, 25.0, 25.0], index=index)
+    wind_speed = pd.Series([1.0, 1.0, 1.0], index=index)
+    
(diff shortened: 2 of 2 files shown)


cwhanse:
@kumaradityaapril please address the formatting issues identified by Flake8.

There are two tests that use the `fuentes` model that now fail, almost certainly because the expected values were computed with the `fuentes` function that was doing integer truncation. I think it's OK to update the expected values and use as many digits as needed to get the tests to pass with the default `rtol` and `atol` values, so we don't have to relax those tolerances.

kumaradityaapril:
Thanks for the review. I’ll address the Flake8 formatting issues and update the Fuentes-related tests.

For the tests that use the Fuentes model, I’ll recompute the expected values using the corrected fuentes implementation and update the hard-coded expectations (with sufficient precision) so that they pass with the default rtol and atol settings, without relaxing tolerances.

cwhanse:
Closing due to lack of progress.

PR #2627, opened 2025-12-16
Add `poa_` prefix to returned names when `return_components=True` in transposition functions

### LINKED ISSUE #2529: Add "poa_" prefix to `return_components=True` transposition model outputs
> Would the powers that be be willing to make the component name’s consistent across all model functions? For example, perez returns sky_diffuse, while poa_components returns poa_sky_diffuse. It also seems that poa_circumsolar, etc. would be more consistent and descriptive for those diffuse components. (A counter argument might be than the underlying plane need not necessarily be for a solar PV array.)

 _Originally posted by @markcampanelli in [#1553](https://github.com/pvlib/pvlib-python/issues/1553#issuecomment-1931254530)_

I think the argument makes sense.  Even further: `pvlib.bifacial.infinite_sheds` returns keys with the `poa_` prefix.  

The downside is that there's not a good way to deprecate DataFrame column names (ref https://github.com/pvlib/pvlib-python/issues/1403#issuecomment-1057405773), so it would be a hard change.  It's unfortunate, but I think I'm +1 to making the change.

kandersolar:
- [x] Closes #2529
 - [x] Tests added

Note that `pvfactors_timeseries` (actually solarfactors under the hood) uses the old names, so the pvfactors tests are going to fail until solarfactors is updated and a new version released.

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.13.2.rst
+++ b/docs/sphinx/source/whatsnew/v0.13.2.rst
@@ -9,7 +9,7 @@ Breaking Changes
 * Following the removal of the NSRDB PSM3 API, the :func:`!pvlib.iotools.get_psm3`,
   :func:`!pvlib.iotools.read_psm3`, and :func:`!pvlib.iotools.parse_psm3`
   functions are removed. (:issue:`2581`, :pull:`2582`)
-* Rename column names to be prefixed with ``"poa_"`` when ``return_components=True``
+* Rename output column names to be prefixed with ``"poa_"`` when ``return_components=True``
   in :py:func:`~pvlib.irradiance.haydavies`, :py:func:`~pvlib.irradiance.perez`,
   and :py:func:`~pvlib.irradiance.perez_driesse`. (:issue:`2529`, :pull:`2627`)
    


ramaroesilva:
@kandersolar nice to see this moving! Will this be done also for `isotropic`, `klucher`, `reindl`, and `king`? I also noticed that the last two have "poa_" in the docstring but not on the actual variable within the code.

Also, I guess I should address #1553 (which is being handled here #2527) on top of this right? Once you comment on my first question (and it is addressed, if I'm right) I can later review this PR, so that its merging is sped up.

kandersolar:
> Will this be done also for isotropic, klucher, reindl, and king?

No, the idea here was to make only the breaking changes (changing returned dict key/dataframe column names), which only affect the functions that already have the `return_components` parameter.  To be clear, these two sets of changes are related, but can be done independently in parallel:
1) The breaking change of renaming the component column names for `haydavies`, `perez`, and `perez_driesse` (this PR)
2) The enhancement of adding `return_components` to the other functions (starting with #2527)

Better to keep them separate because smaller, focused PRs are easier to review, and because it means we can be sure to get the breaking changes into the next release (a .0 release) even if the enhancements aren't ready in time.

> I also noticed that the last two have "poa_" in the docstring but not on the actual variable within the code.

The names of actual variables in the code are not user-facing, so strictly speaking it doesn't matter if they match the docstring (or anything else).  If we want to change them to have `poa_` for code clarity then that's fine, but that's orthogonal to achieving #1553.

> Also, I guess I should address https://github.com/pvlib/pvlib-python/issues/1553 (which is being handled here https://github.com/pvlib/pvlib-python/pull/2527) on top of this right? Once you comment on my first question (and it is addressed, if I'm right) I can later review this PR, so that its merging is sped up.

Hopefully the above addresses this :)  These two changes can be worked on independently.  I do think this PR achieves the intended scope and is ready for review though.

tests/test_irradiance.py
```diff
@@ -210,12 +210,14 @@ def test_haydavies(irrad_data, ephem_data, dni_et):
 
 
 def test_haydavies_components(irrad_data, ephem_data, dni_et):
+    keys = ['poa_sky_diffuse', 'poa_isotropic', 'poa_circumsolar',
+                'poa_horizon']
```
kandersolar:
```suggestion
    keys = ['poa_sky_diffuse', 'poa_isotropic', 'poa_circumsolar',
            'poa_horizon']
```

docs/sphinx/source/whatsnew/v0.13.2.rst
```diff
@@ -9,7 +9,10 @@ Breaking Changes
 * Following the removal of the NSRDB PSM3 API, the :func:`!pvlib.iotools.get_psm3`,
   :func:`!pvlib.iotools.read_psm3`, and :func:`!pvlib.iotools.parse_psm3`
   functions are removed. (:issue:`2581`, :pull:`2582`)
-
+* Rename column names to be prefixed with ``"poa_"`` when ``return_components=True``
```
cwhanse:
```suggestion
* Rename output column names to be prefixed with ``"poa_"`` when ``return_components=True``
```

cwhanse (approved):

docs/sphinx/source/whatsnew/v0.13.2.rst
cwhanse: Looks like this would cause a breaking issue in `solar_factors` e.g. [here](https://github.com/pvlib/solarfactors/blob/ff6c4d6342e1ca7bacee13bb637e19f209286a49/pvfactors/irradiance/utils.py#L88).

cwhanse:
I thought the discussion in #2529 indicated that this change would happen in v0.14, since its breaking?

kandersolar:
The imminent release, which was once called v0.13.2, will be v0.14.0 instead.  The whatsnew will be renamed in the release PR, and I need to clean up the milestones too.

kandersolar:
CI is green after I retriggered everything following making a new solarfactors release ([v1.6.1](https://github.com/pvlib/solarfactors/releases/tag/v1.6.1)) with compatibility for this change.

PR #2646, opened 2025-12-28
Fix ZeroDivisionError in simple_efficiency when load_loss=0

### LINKED ISSUE #2645: simple_efficiency raises ZeroDivisionError when load_loss=0
### Describe the bug

`pvlib.transformer.simple_efficiency` raises a `ZeroDivisionError` when
`load_loss = 0`.

A zero `load_loss` is a physically valid case (e.g., an ideal transformer
with only no-load losses), but the current implementation always solves
a quadratic equation and divides by `2 * load_loss`, which leads to a
division by zero.

---

### To Reproduce
```
from pvlib.transformer import simple_efficiency

simple_efficiency(
    input_power=1000,
    no_load_loss=0.01,
    load_loss=0,
    transformer_rating=1000,
)
```

### Expected behavior

When `load_loss = 0`, the loss model becomes linear rather than quadratic.
The output power should be computed as:
`P_out = P_in - L_no_load * P_nom`
and the function should return a finite value instead of raising an
exception.

---

### Proposed solution

Handle the special case where `load_loss == 0` explicitly before solving
the quadratic equation.

When `load_loss = 0`, the loss model reduces to a linear relationship:
`P_out = P_in - L_no_load * P_nom`

A conditional check can be added at the beginning of the function to
return the linear solution and avoid division by zero, for example:
```
if load_loss == 0:
    return input_power - no_load_loss * transformer_rating
```
This preserves existing behavior for `load_loss > 0` while correctly
handling a physically valid edge case.

aman-coder03:
- [x] Closes #2645
 - [x] Tests added

fixes a `ZeroDivisionError` in `pvlib.transformer.simple_efficiency`
when `load_loss = 0`.

when `load_loss` is zero, the transformer loss model reduces to a linear
relationship rather than a quadratic one. this PR handles that case
explicitly to avoid division by zero while preserving existing behavior
for `load_loss > 0`.

Fixes #2645

### DIFF
--- a/pvlib/transformer.py
+++ b/pvlib/transformer.py
@@ -111,6 +111,8 @@ def simple_efficiency(
     b = 1
     c = no_load_loss - input_power_normalized
 
+    # alternative form of the quadratic equation to avoid
+    # divide-by-zero when a == 0
     disc = (b*b - 4*a*c)**0.5
     output_power_normalized = 2*c / (-b - disc)
 


pvlib/transformer.py
```diff
@@ -105,13 +105,17 @@ def simple_efficiency(
        pg. 101.
     '''  # noqa: E501
 
+    if load_loss == 0:
+        return input_power - no_load_loss * transformer_rating
+
```
cwhanse:
```suggestion
```

I don't think this is necessary with the alternate form of the quadratic equation.

aman-coder03: i have removed it

cwhanse: Could you add a test with `load_loss=0`?

aman-coder03: sure!!

aman-coder03: thanks!! @cwhanse 
i have added a test covering the `load_loss=0` that checks the linear-limit behavior `P_out = P_in − L_no_load * P_nom`
please let me know if you would prefer this test to also cover array like inputs or assert behavior at additional edge cases (negative input power)

cwhanse: Let's keep the scope to addressing the division by zero issue. Test looks good.

Please open a new issue to extend testing to `np.array` and `pd.Series` types.

aman-coder03: thanks @cwhanse 
sure i will open issues separately for each `np.array` and `pd.Series`

echedey-ls: > thanks @cwhanse sure i will open issues separately for each `np.array` and `pd.Series`

@aman-coder03 

Just one issue for both of them types is enough.

aman-coder03: done @echedey-ls 
thanks for your reply!!

tests/test_transformer.py
```diff
@@ -58,3 +58,24 @@ def test_simple_efficiency_known_values():
                                       *args),
         rating,
     )
+
+
+def test_simple_efficiency_zero_load_loss():
+
```
echedey-ls:
```suggestion
```

echedey-ls (commented):
LGTM. Add a whatsnew entry in the bugfixes section and this is good IMO.

@aman-coder03, just so you know, a cleaner and modular approach to write the tests would be to use `@pytest.mark.parametrize`, so you could avoid boiler plate repeated in the test at the top. Let me know if you wanna give it a try, it would definitely look better and introduce you to some of the perks of the testing framework.

aman-coder03:
@echedey-ls i added a `whatsnew` entry under Bug fixes.
Thanks for the suggestion on parametrized tests, happy to refactor those next.

docs/sphinx/source/whatsnew/v0.13.2.rst
```diff
@@ -16,6 +16,10 @@ Deprecations
 
 
 Bug fixes
+* Fixed a division-by-zero error in
```
cwhanse:
```suggestion
* Fix a division-by-zero condition in
```

echedey-ls:
Feel free to merge this test at the top: https://github.com/aman-coder03/pvlib-python/blob/d45d6ea29165380d076481decc555af6b5a366bf/tests/test_transformer.py#L7C1-L41C1
 and the one you have created: https://github.com/aman-coder03/pvlib-python/blob/d45d6ea29165380d076481decc555af6b5a366bf/tests/test_transformer.py#L63C1-L81C38
into the top one.

cwhanse:
@echedey-ls I think it's OK to re-implement the tests is a separate PR, to avoid scope creep in a PR that addresses a division-by-zero condition.

echedey-ls:
@cwhanse I agree with that.

aman-coder03:
thanks both for the guidance @echedey-ls @cwhanse 
i have moved the whatsnew entry to `v0.13.3` as suggested and removed it from v0.13.2.
i will keep the test in this PR limited to the division-by-zero case only, and open a separate PR later if we want to refactor or consolidate the transformer tests

cwhanse:
@aman-coder03 I'm really sorry, could you move this back to v0.13.2? Just move the whatsnew note. We had planned to release v0.13.2 in December and I forgot that we put that off.

aman-coder03:
@cwhanse no issues!
i will do that!

aman-coder03:
These failures appear to be from deprecation tests that are designed to fail
once the removal version (v0.14) is reached.

Since this PR only addresses the transformer division-by-zero issue,
I didn’t modify any irradiance deprecation tests.

echedey-ls:
> These failures appear to be from deprecation tests that are designed to fail once the removal version (v0.14) is reached.

Yes, our bad for not enforcing those changes, I will open an issue. Thanks for spotting that!

aman-coder03:
@echedey-ls can i work on the issue you were saying to open?

echedey-ls:
> @echedey-ls can i work on the issue you were saying to open?

Thanks for the follow up @aman-coder03. I lost my draft without opening it. Please, do NOT work on it until maintainers agree on what action to take.

I will open the issue first.

aman-coder03:
sure @echedey-ls

docs/sphinx/source/whatsnew/v0.14.1.rst
```diff
@@ -14,7 +14,9 @@ Deprecations
 
 Bug fixes
 ~~~~~~~~~
-
+* Fix a division-by-zero condition in
+  :py:func:`~pvlib.transformer.simple_efficiency` when ``load_loss = 0``.
+  (:issue:`2645`)
```
echedey-ls:
```suggestion
  (:issue:`2645`, :pull:`2646`)
```

docs/sphinx/source/whatsnew/v0.14.1.rst
```diff
@@ -14,7 +14,9 @@ Deprecations
 
 Bug fixes
 ~~~~~~~~~
-
+* Fix a division-by-zero condition in
+  :py:func:`~pvlib.transformer.simple_efficiency` when ``load_loss = 0``.
```
echedey-ls:
```suggestion
  :py:func:`pvlib.transformer.simple_efficiency` when ``load_loss = 0``.
```
I suggest that whatsnew uses full qualified name, since reading `simple_efficiency` does not hint that it is regarding a transformer (in a mainly-PV project).

pvlib/transformer.py
```diff
@@ -111,7 +111,8 @@ def simple_efficiency(
     b = 1
     c = no_load_loss - input_power_normalized
 
-    output_power_normalized = (-b + (b**2 - 4*a*c)**0.5) / (2 * a)
+    disc = (b*b - 4*a*c)**0.5
```
kandersolar, line 116:
```suggestion
    # alternative form of the quadratic equation to avoid
    # divide-by-zero when a == 0
    disc = (b*b - 4*a*c)**0.5
```

aman-coder03:
@kandersolar thankyou for your reply
i have committed your suggestion!

pvlib/transformer.py
kandersolar: @aman-coder03 these two comment lines still need to be added

aman-coder03:
```
input_power_normalized = input_power / transformer_rating

    a = load_loss
    b = 1
    c = no_load_loss - input_power_normalized

    # alternative form of the quadratic equation to avoid
    # divide-by-zero when a == 0
    disc = (b*b - 4*a*c)**0.5
    
    output_power = disc * transformer_rating
    return output_power
```
are you saying to implement like this?

kandersolar: No I am not, since that math would not be correct.  My suggestion was just to add the two comment lines, not to change the calculations.  I will add them, do not worry about it.

aman-coder03: sorry, i thought you were suggesting the whole block to be changed, thankss I got it now! thanks for your implementation

aman-coder03:
thankyou to everyone involved in this discussion!! looking forward to contribute more from my side

kandersolar:
Thanks @aman-coder03!

PR #2700, opened 2026-02-24
Fix latlon_to_xy longitude scaling bug and update scaling tests (Fixes #2614)

### LINKED ISSUE #2614: `scaling.py` Numpy function arg bug
**Describe the bug**
https://github.com/pvlib/pvlib-python/blob/cc75255d7df6ce939db7ee6ebaf81e79741c2e88/pvlib/scaling.py#L199 has a bug where longitude gets passed as a second arg to `np.array`. The second arg for `np.array` is the dtype however.

```python
pos = coordinates * np.array(m_per_deg_lat, m_per_deg_lon)
```

What will happen with the current implementation is that `m_per_deg_lat` will be silently used as a multiplier for both axes.

```python
# resulting broadcast
lat * m_per_deg_lat
lon * m_per_deg_lat   # <- this is wrong. it should be using the lon instead
```

**Expected behavior**
Correct implementation is as follows:
```python
pos = coordinates * np.array([m_per_deg_lat, m_per_deg_lon])
```

This could also affect the unit tests, but I didn't check.

muien5080:
Fix incorrect longitude scaling in latlon_to_xy where the NumPy array construction previously misapplied scaling factors, causing longitude to be multiplied by the latitude conversion factor.

Fixes #2614

Changes:
Correct scaling vector construction:
np.array([m_per_deg_lat, m_per_deg_lon])
Ensure longitude scaling depends on mean latitude
Update expected coordinate values in test_scaling.py
Update expected variability reduction (VR) values affected by corrected geometry

Validation:
test_scaling.py passes
Full test suite passes locally:
    1415 passed
    111 skipped
    2x failed
No new warnings introduced

Checklist:
Tests added or updated
Documentation update (not required)
Code passes flake8
All tests pass locally
Linked to relevant issue

### DIFF
--- a/pvlib/scaling.py
+++ b/pvlib/scaling.py
@@ -196,7 +196,7 @@ def latlon_to_xy(coordinates):
     m_per_deg_lon = r_earth * np.cos(np.pi/180 * meanlat) * np.pi/180
 
     # Conversion
-    pos = coordinates * np.array(m_per_deg_lat, m_per_deg_lon)
+    pos = coordinates * np.array([m_per_deg_lat, m_per_deg_lon])
 
     # reshape as (x,y) pairs to return
     try:
--- a/tests/test_scaling.py
+++ b/tests/test_scaling.py
@@ -51,8 +51,8 @@ def time_500ms(clear_sky_index):
 
 @pytest.fixture
 def positions():
-    # Sample positions based on the previous lat/lon (calculated manually)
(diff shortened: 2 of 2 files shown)


cwhanse:
@muien5080 please put the PR template back in your leading comment block, so we maintainers can use that checklist. Thanks.

muien5080:
Template restored.
Thanks.

AdamRJensen:
Closing this PR as the contributor did not follow instructions.

PR #2702, opened 2026-02-28
calculate surface angles using unit normal

### LINKED ISSUE #1976: Ambiguous descriptions of axis_azimuth and axis_tilt in pvlib.tracking.singleaxis() docs
Some details around _axis_azimuth_ and _axis_tilt_ in the docs (https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.tracking.singleaxis.html) seem ambiguous to me. 

Examples:

1. For a north-south aligned tracker axis, sloping 5 degrees down to the south, what are the inputs for _axis_azimuth_ and _axis_tilt_? I assume _axis_azimuth_ = 180, _axis_tilt_ = 5.
2. For the same north-south aligned tracker axis, but sloping 5 degrees down to the **_north_**, what are the inputs? I assume _axis_azimuth_ = 0, _axis_tilt_ = 5, since _axis_tilt_ must be >= 0.

Additionally, the description in the docs includes:
> For example, if tracker _axis_azimuth_ is 180 (oriented south) and _axis_tilt_ is zero...

But the default for _axis_azimuth_ is 0, which could lead to confusion around inputs and output sign convention. 

More clear docs.

**Additional context**
Changing definitions slightly and allowing for negative _axis_tilt_ would make more sense to me, but maybe that causes other problems. [_edit: now I don't think this is a good idea - see the full discussion below_] [**2nd edit: looks like code allows for this even though docs do not. Also, my opinion is easily swayed...**]

For example \# 2 above (sloping down to the North), allowing for _axis_azimuth_ = 180, _axis_tilt_ = -5 makes more since to me. You could unambigously define _axis_azimuth_ as "the South-most direction of the axis of rotation" and _axis_tilt_ as "the tilt of the axis of rotation, with positive values indicating downward slope towards the South-most direction, and negative values indicating downward slope towards the North-most direction...".

I'm not sure how to handle the odd case of a tracker axis aligned perfectly East-West.

And for systems in the Southern Hemisphere, I don't think anything would need to change.

cwhanse:
- [x] Closes #1976
 - [x] Tests added

Changes the calculation of surface angles for tracking.singleaxis to allow for negative axis_tilt.

Changes the output of singleaxis in the case of surface_tilt=0, to consistently return surface_azimuth=axis_azimuth.


pvlib/tracking.py
kandersolar: Remove this reference?

pvlib/tracking.py
```diff
@@ -245,16 +290,18 @@ def calc_surface_orientation(tracker_theta, axis_tilt=0, axis_azimuth=0):
...
-        # handle surface_tilt=0 case:
-        azimuth_delta = np.where(sind(surface_tilt) != 0, azimuth_delta, 90)
-        surface_azimuth = (axis_azimuth + azimuth_delta) % 360
+    # unit normal to rotated tracker surface
+    unit_normal = _unit_normal(axis_azimuth, axis_tilt, tracker_theta)
+
+    # deviate from [1] to allow for negative tilt.
+    # project unit_normal to x-y plane to calculate azimuth
+    surface_azimuth = np.degrees(
+        np.arctan2(unit_normal[:, 0], unit_normal[:, 1]))
```
adriesse, line 299: I would include this line in the function that calculates the unit normal so that it is self-contained w.r.t. its conventions.

pvlib/tracking.py
```diff
@@ -210,6 +210,50 @@ def singleaxis(apparent_zenith, solar_azimuth,
     return out
 
 
+def _unit_normal(axis_azimuth, axis_tilt, theta):
+    """
+    Unit normal to rotated tracker surface, in global E-N-Up coordinates,
+    given by R*(0, 0, 1)^T, where:
+
+        R = Rz(-axis_azimuth) Rx(-axis_tilt) Ry(theta) *
```
echedey-ls:
```suggestion
        R = Rz(-axis_azimuth) Rx(-axis_tilt) Ry(theta)
```

pvlib/tracking.py
```diff
@@ -210,6 +210,50 @@ def singleaxis(apparent_zenith, solar_azimuth,
...
+
+    Rz is a rotation by -axis_azimuth about the z-axis (axis_azimuth
+    is negated to convert from an azimuth angle to a rotation angle). Rx is a
+    rotation by -axis_tilt about the x-axis, where axis_tilt is negated
+    because pvlib's convention is that the positive y-axis is tilted
+    downwards. Ry is a rotation by theta
+    about the y-axis. theta is negated so that a negative.
```
echedey-ls: I think this sentence was not ended correctly, right?

echedey-ls (commented):
I've compared `_unit_normal` with an implementation I did based on scipy rotation objects, and they match. The 3d axis basis is the same E-N-Up, left-hand rotations.

I just left [this comment](https://github.com/pvlib/pvlib-python/issues/1976#issuecomment-4179647476) in the issue, I hope not to delay a fix on negative tilt handling.

pvlib/tracking.py
cwhanse: Are you suggesting that `_unit_normal` calculate and return `surface_azimuth`? I'm unclear what you are suggesting.

adriesse: That was the thought, yes.  Angles in, angle out is independent of the method.  I guess the function name would change too.

pvlib/tracking.py
```diff
@@ -43,8 +43,10 @@ def singleaxis(apparent_zenith, solar_azimuth,
...
     axis_tilt : float, default 0
         The tilt of the axis of rotation (i.e, the y-axis defined by
-        ``axis_azimuth``) with respect to horizontal.
-        ``axis_tilt`` must be >= 0 and <= 90. [degrees]
+        ``axis_azimuth``) with respect to horizontal (degrees). Positive
+        ``axis_tilt`` is *downward* in the direction of ``axis_azimuth``. For
+        example, for a tracker with ``axis_azimuth``=180 and ``axis_tilt``=10,
+        the north end is higher than the south end of the axis.
 
     axis_azimuth : float, default 0
```
adriesse, line 51: All the examples refer to azimuth 180, so why is the default zero?

pvlib/tracking.py
```diff
@@ -84,7 +84,7 @@ def singleaxis(apparent_zenith, solar_azimuth,
         intersection between the slope containing the tracker axes and a plane
```
adriesse, line 84: I haven't looked at these functions much before, and it took me a while to realize that "slope" usually refers to the terrain or ground itself, or in this case  surface floating above the ground, rather than an attribute of a line or surface.  I the documentation could be improved by changing "slope" to "sloped surface" or "sloped ground"or "sloped terrain" or similar depending on the context.  

If this gets a few thumbs up I can make specific suggestions.

pvlib/tracking.py
cwhanse: Thanks for the suggestion. I guess I'm preferring `_unit_normal` as the function because it may find other uses. We could package the few lines here into a helper `_calc_surface_azimuth` but I don't see that as much of an improvement.

pvlib/tracking.py
cwhanse: I don't think there's a particular reason for the default of 0. It's the least common orientation in the northern hemisphere but perhaps its most common in the southern hemisphere. 0 has been the default for a long time, so maybe inertia wins? Examples are different question, though.

pvlib/tracking.py
cwhanse: I agree that this description can be improved. "slope" is not specific. The precise definition should describe an intersection of two planes (because that's how a line is defined, and lines define angles). I think "slope" is intended to mean the plane that contains the tracker axes. @kandersolar

kandersolar: >  I think "slope" is intended to mean the plane that contains the tracker axes.

Right.  +1 to improved wording, although maybe that clarification should be separate from this PR.

pvlib/tracking.py
adriesse: The function name seems very general and I initially suspected it could indeed have other intended uses; but the calculations seem very specialized, which raised some doubt. What other uses do you have in mind?

cwhanse: I am assuming you are referring to the `_unit_normal` function. In this instance it is specific to a surface mounted on a single-axis tracker. I am thinking that having the unit normal for that surface enables other computing angles in a 3D scene, which are much easier to write as dot products than using trig functions. pvlib doesn't do much of that, currently. This function could be used in place of the current AOI calculation, but I don't see any advantage to do so.

adriesse: Yes indeed, having the unit normal for the single axis tracker surface could be used for other subsequent calculations, such as calculating ```surface_tilt``` or AOI. But the ```_unit_normal``` function itself does three tracker-specific rotations in a predetermined sequence, so I don't see how that can be repurposed.

If you did use ```unit_normal``` to calculate both ```surface_tilt``` and ```surface_azimuth```, then that would make a better case for current the division of labor between function and helper, and maybe also serve as a better example for future uses of the vector approach.

pvlib/tracking.py
```diff
@@ -210,6 +211,49 @@ def singleaxis(apparent_zenith, solar_azimuth,
...
+    Parameters
+    ----------
+    axis_azimuth : scalar
+    axis_tilt    : scalar
+    theta        : scalar or array-like
+
+    Returns
+    -------
+    ndarray
+        Shape (3,) if theta scalar
```
kandersolar:
```suggestion
        Shape (1, 3) if theta scalar
```

```python
In [4]: pvlib.tracking._unit_normal(180, 10, 45).shape
Out[4]: (1, 3)
```

pvlib/tracking.py
kandersolar: The PR as it currently stands addresses the linked issue, so I am inclined to go ahead with it.  It can be refactored in the future as needed for some additional purpose.

pvlib/tracking.py
kandersolar, line 287: Not sure this reference is very helpful anymore.  Delete?

tests/test_tracking.py
```diff
@@ -195,6 +195,54 @@ def test_backtrack():
     assert_frame_equal(expect, tracker_data)
 
 
+def test__unit_normal():
+    # with scalar input
+    unorm = tracking._unit_normal(180., 45., 45.)
+    np.allclose(unorm, np.array([-np.sqrt(2)/2, -0.5, -0.5]))
```
kandersolar:
```suggestion
    np.testing.assert_allclose(unorm, np.array([-np.sqrt(2)/2, -0.5, -0.5]))
```

Needs an assertion of some kind

pvlib/tracking.py
```diff
@@ -43,8 +43,10 @@ def singleaxis(apparent_zenith, solar_azimuth,
 
     axis_tilt : float, default 0
         The tilt of the axis of rotation (i.e, the y-axis defined by
-        ``axis_azimuth``) with respect to horizontal.
-        ``axis_tilt`` must be >= 0 and <= 90. [degrees]
```
kandersolar, line 47: What's new entry for allowing negative axis tilt?

tests/test_tracking.py
kandersolar, line 1: Can we add a test along these lines?  (assuming I did not miss that it's redundant with another one you've added)

```python
params = dict(apparent_zenith=45, solar_azimuth=270)

tr_pos = pvlib.tracking.singleaxis(axis_tilt=10, axis_azimuth=0, **params)
tr_neg = pvlib.tracking.singleaxis(axis_tilt=-10, axis_azimuth=180, **params)

tr_neg['tracker_theta'] *= -1  # expect tracker_theta to be negated

for key in tr_pos:
    np.testing.assert_allclose(tr_pos[key], tr_neg[key])
```

This test would fail with the current pvlib, but should pass here.

pvlib/tracking.py
cwhanse: It's still the source for the surface_tilt equation, I'd keep it for that reason.

docs/sphinx/source/whatsnew/v0.15.1.rst
```diff
@@ -38,6 +38,9 @@ Enhancements
 * Accelerate the internals of :py:func:`~pvlib.solarpostion.ephemeris`. (:pull:`2626`)
 * Accelerate the intervals of :py:func:`~pvlib.pvsystem.singlediode` when
   `method='lambertw'`. (:pull:`2732`, :pull:`2723`)
+* :py:func:`~pvlib.tracking.singlediode` accepts negative values for
```
kandersolar:
```suggestion
* :py:func:`~pvlib.tracking.singleaxis` accepts negative values for 
```

kandersolar (approved):
LGTM.  Thanks @cwhanse

kandersolar:
Test failure is not related and seems like the kind of thing that might go away on its own.

markcampanelli:
Thanks @cwhanse for this, and to the reviewers.

PR #2718, opened 2026-03-12
Calculate IV curve for series-connected devices 

cwhanse:
- [ ] Closes #xxxx
 - [x] Tests added

Uses a simplified reverse bias model (vertical asymptote at breakdown voltage), rather than the differentiable model in the bishop88 functions. The simplified model is used for speed and reliability but comes at the cost of some precision for devices with nearly equal short-circuit currents.


cwhanse:
This is ready for some feedback. Some questions in my mind:

- I'm not sure about the interface. The new function (intended to be public) takes in single diode equation parameters (output from one of the pvsystem.calcparams_* functions) and returns the series IV curve. It does not anticipate "adding" IV curves represented by arrays of current and voltage.
- I intend to add a maximum power point to that curve; that is not yet done. As it is now, this function acts as an alternative for pvsystem.i_from_v (compute an IV curve). The IV curve is returned (rather than Isc, Imp, Vmp, Voc as are returned by pvsystem. singlediode) because this function is intended to feed the next stage: compute the IV curve for parallel strings.
- I've put this is a new module called `mismatch`. If that's too general (there's also spectral mismatch) this would be electrical mismatch.

RDaxini (commented):
I am happy to take a deeper dive into the code later, likely next week if that's not too late. I can share some thoughts in response to at least two these questions first though.

> * I'm not sure about the interface. The new function (intended to be public) takes in single diode equation parameters (output from one of the pvsystem.calcparams_* functions) and returns the series IV curve. It does not anticipate "adding" IV curves represented by arrays of current and voltage.

Fair point... I don't have anything insightful to add here at the moment but will try to circle back to it.

> * I intend to add a maximum power point to that curve; that is not yet done. As it is now, this function acts as an alternative for pvsystem.i_from_v (compute an IV curve). The IV curve is returned (rather than Isc, Imp, Vmp, Voc as are returned by pvsystem. singlediode) because this function is intended to feed the next stage: compute the IV curve for parallel strings.

I'm not sure I follow... `pvsystem.i_from_v` is a pointwise solver but I thought that this is a curve constructor, so they seem to be completely different? I might be misunderstanding something here from not having gone into the code yet.

> * I've put this is a new module called `mismatch`. If that's too general (there's also spectral mismatch) this would be electrical mismatch.

-1 to this naming. The generality is one downside but I think the collision with `pvlib.spectrum.mismatch` is a bigger issue. `from pvlib import mismatch` vs `from pvlib.spectrum import mismatch` could be problematic. We might want to consider renaming the spectral mismatch module with a `spectral` clarifier, but that could be discussed separately. `electrical_mismatch` suggestion solves this but, another consideration: mismatch is a loss, so is it definitely the right thing to name this module after? How about `pvlib.circuit` or `pvlib.ivcurve`? These sound to me more linked to what the module does. Not 100% opposed to `electrical_mismatch` though and certainly open other suggestions. The two I mentioned are just the first two that came to mind.

cwhanse:
Having worked in the MPP calculation, I'm inclined to refactor this code to more closely parallel `pvsystem.singlediode` and `pvsystem.i_from_v` and `pvsystem.v_from_i`.  The purposes are common:

- for `pvsystem.singlediode`, solve for Isc, Imp, Voc, Vmp, etc. given arrays of diode equation parameters. Returned values are arrays. It is anticipated that the arrays are time series, resulting from application of `pvsystem.calcparams_xxxxx` to weather time series.
- for `pvsystem.i_from_v` (or `v_from_i`), calculate points (could be many) given arrays of diode equation parameters. Returned values are arrays of voltage (current) corresponding to the input current (voltage). Again, it is anticipated that the array's first dimension is time.

Here, the corresponding functions would be conceptually similar. I'll use `mismatch` for the module name to describe the idea, although I tend to agree with @RDaxini that's not a good choice.

- `mismatch.singlediode` takes in arrays of diode equation parameters (output from pvsystem.calcparams_xxxx`) and returns arrays of Isc, Voc, etc. The difference from `pvsystem.singlediode` is the dimensionality; the photocurrent parameter (for example) should be 2D and have dimensions (number of devices in series, number of times).  Output is 1d: Isc is the string Isc, and has dimension (number of times).
- `mismatch.v_from_i` takes in an array of string current (1d) and arrays of device parameters (2d), and returns an array of string voltage (1d).

I hope this is a more clear picture.

kandersolar:
@cwhanse I think fully wrapping my head around this will take a few attempts.

In the meantime, I ask the perennial annoying question: are there any plans to publish a reference for this approach?  It looks like the bulk of it is a combination of well-established methods, but some parts (e.g. the simple reverse bias breakdown model and the default number of I-V curve points) leave me wanting to read some kind of analysis of how these choices compare with alternatives.  There may be some minor novelty as well, for example identifying the neighborhood of the global MPP using the discretized I-V curve and then refining with the usual optimization approach.

cwhanse:
> are there any plans to publish a reference for this approach?

Not in a journal, as I think the novelty (here, and in the follow-on PR that will support combining parallel IV curves) is low. I could write a technical report for archival purposes.

The simple reverse bias treatment is a surrender to persistent convergence failures when attempting this calculation using `bishop_i_from_v`'s Newton option. The `brentq` option would not have the convergence issues but is not vectorized which I think is necessary here for performance reasons.

I'm OK if this kicks down the road.

I have a corroborating calculation to add to the test suite, from PVLighthouse.

Issue #1867, opened 2023-09-20
`bifacial.utils.vf_ground_sky_2d` interprets `max_rows` asymmetrically


kandersolar:
**Describe the bug**
`vf_ground_sky_2d` considers one fewer sky wedge on the right side of the central wedge than it does on the left side.

**To Reproduce**
I added a line `print(next_edge)` after here: https://github.com/pvlib/pvlib-python/blob/7608eda81d8302d66be6a686f85f9c3eb5cbe955/pvlib/bifacial/utils.py#L170

Then I ran `_ = vf_ground_sky_2d(surface_tilt=0, gcr=0.5, x=0.5, pitch=1, height=1, max_rows=3)` and got the printed text: `[[[0.01598559 0.04556841 0.18086881 0.48507125 0.18086881 0.04556841]]]`.

Each value in that output describes the VF of a visible sky wedge.  The inputs describe a symmetrical system, so the VFs should be symmetrical too, and they are, except one of them is missing... 0.485 for the wedge directly overhead, with three wedges to the left but only two to the right.

**Expected behavior**
I want the number of sky wedges considered in each direction to be equal.

Maybe this is a fencepost problem, and really what I want is effectively a `max_wedges` parameter instead of `max_rows`?

pasquierjb:
k goes from `-max_rows` to `max_rows+1` so the symmetry is for x=0 which is indeed problematic if you intergrate betwen 0 and 1 for a small `max_rows`. However as `max_rows` is set to be high enough to neglect the horizon VF, not sure it will change a lot the results.

![download (7)](https://github.com/pvlib/pvlib-python/assets/12225807/d2697886-47a4-4d94-b0d2-69f213e6c48e)

Sar-thak-3:
Hi @kandersolar, is this issue still open? I would love to contribute to this issue!

Chirag3841:
@kandersolar I want to contribute to this issue if anyone had not taken that .

AdamRJensen:
> [@kandersolar](https://github.com/kandersolar) I want to contribute to this issue if anyone had not taken that .

Hi 👋 @Chirag3841 Thanks for your interest in contributing to pvlib. Please note that making meaningful code contributions to pvlib requires some familiarity with photovoltaics or solar irradiance. If you have this background, we would be happy to help you along.

kandersolar:
> really what I want is effectively a max_wedges parameter instead of max_rows?

[...]

Issue #1861, opened 2023-09-15
Update default transposition model


AdamRJensen:
Having the default transposition model be isotropic seems like a very poor choice. Isotropic models are known to have very large biases and should only be used in very specific use cases (generally not be new users).

I did a brief survey of commercial software and found that the vast majority only offer the Hay-Davirs and Perez model, with the 1990 Perez model seeming to be favored as the default model.

It would seem to make sense to have the same default model used in ``pvlib.modelchain.ModelChain``.

https://github.com/pvlib/pvlib-python/blob/cb4c1355712255c27ec5211a3cd9d93f12518613/pvlib/irradiance.py#L305-L310

AdamRJensen:
I suppose this would seen as a "breaking change" and would have to wait until 0.11.0 if we choose to do this.

kandersolar:
It would be great if `get_total_irradiance` could still be run with the current minimal set of inputs, even if the default model is changed to something more sophisticated.  #1225 got us halfway there by automatically calculating airmass for Perez, but left ET DNI alone since there is not always a datetime index available to calculate it.  Maybe we could examine the inputs to see if a datetime index is available and only raise the error if not?

AdamRJensen:
> It would be great if `get_total_irradiance` could still be run with the current minimal set of inputs, even if the default model is changed to something more sophisticated. ... Maybe we could examine the inputs to see if a datetime index is available and only raise the error if not?

This seems like a good approach to me. I suppose it should be incorporated here:
https://github.com/pvlib/pvlib-python/blob/46851d91bb70e9b5521feb7ec119b057ec3d7f70/pvlib/irradiance.py#L454-L455

 What would be the best way of assessing if the index is datetime? Something like this:

    isinstance(d.index, pd.DatetimeIndex)

kurt-rhee:
I would like to throw my vote in the hat for Perez as the default transposition model, about 90% of predictions run in PlantPredict use the Perez model (this is without regard to the source of the meteorological data).  This is an imperfect measure of true usage industry wide, but I think it is a nice data point.

kandersolar:
> I suppose this would seen as a "breaking change" and would have to wait until 0.11.0 if we choose to do this.

Seems like we can and should do a deprecation period for this first.  That can be implemented any time, so I've removed the v0.11.0 milestone from this issue.

PR #2740, opened 2026-04-21
Implement ANTS-2D bifacial irradiance model

kandersolar:
- ~[ ] Closes #xxxx~
 - [x] Tests added

@AdamRJensen, @cwhanse, and I have a paper describing a new bifacial irradiance model called ANTS-2D.  It is similar to pvlib's `infinite_sheds` model, but extended to allow:
- a discretized module surface (for, e.g., cell-level irradiance values)
- a discretized ground surface, with variable albedo
- sloped terrain, in the fashion of pvlib's existing functionality in that area
- the Perez transposition model, in addition to Hay-Davies and isotropic
- computation of ground-level irradiance (for agriPV)
- fast computation by using analytical integrated view factors instead of burdensome numerical integrals

Details available open-access here: https://doi.org/10.1109/JPHOTOV.2026.3677506

This PR is rather large.  To summarize:
- Add `g0` and `g1` parameters to the view factor functions in `pvlib.bifacial.utils`.  These are analogous to `x0` and `x1` in `vf_row_sky_2d_integ` and extend the functions to subset the ground surface.
- Change `vf_ground_sky_2d_integ` to use Hottel's crossed-string rule instead of burdensome numerical integration.  This makes the `npoints` and `vectorize` parameters unnecessary.
- Change some signatures in `pvlib.bifacial.utils` to be cleaner with the new calculations.
- Minor edits in `pvlib.bifacial.infinite_sheds` to accommodate the `utils` changes
- Create `pvlib.bifacial.ant2d`, which houses the model itself and uses the new  `utils` functionality.

Let me know if it would help reviewers to split it up and review separate PRs, starting with `utils`.


pvlib/bifacial/utils.py
```diff
@@ -174,55 +239,132 @@ def vf_ground_sky_2d(rotation, gcr, x, pitch, height, max_rows=10):
     return vf
 
 
-def vf_ground_sky_2d_integ(surface_tilt, gcr, height, pitch, max_rows=10,
-                           npoints=100, vectorize=False):
+@renamed_kwarg_warning("0.15.2", "surface_tilt", "tracker_rotation")
+def vf_ground_sky_2d_integ(tracker_rotation, gcr, height, pitch, g0=0, g1=1,
```
echedey-ls:
```suggestion
def vf_ground_sky_2d_integ(tracker_rotation, gcr, height, pitch, *, g0=0, g1=1,
```
The rationale is that previous versions assume this call signature

`vf_ground_sky_2d_integ(0, 0.5, 2, 4, 20, 200)`, where `max_rows=20` and `npoints=200`

This wouldn't fail in the new version [1], where values `g0` and `g1` would take 20 and 200 respectively.

[1] At this branch
```
>>> import pvlib
>>> pvlib.bifacial.utils.vf_ground_sky_2d_integ(0, 0.5, 2, 4, 20, 200)
array(1.14662412e-05)
```

[2] On main
```
array([0.49984351])
```

pvlib/bifacial/utils.py
```diff
@@ -351,7 +511,8 @@ def vf_row_ground_2d(surface_tilt, gcr, x):
     return 0.5 * (1 - (1/gcr * cosd(surface_tilt) + x)/p)
 
 
-def vf_row_ground_2d_integ(surface_tilt, gcr, x0=0, x1=1):
+def vf_row_ground_2d_integ(surface_tilt, gcr, height=None, pitch=None,
+                           x0=0, x1=1, g0=0, g1=1, max_rows=20):
```
echedey-ls: I'd say this one is also subject to the problem described in previous comment

echedey-ls (commented):
Some first impressions; also, you may want to link to it in other bifacial "See also" sections.

docs/sphinx/source/whatsnew/v0.15.2.rst
```diff
@@ -18,6 +27,13 @@ Bug fixes
 
 Enhancements
 ~~~~~~~~~~~~
+* Implement the ANTS-2D bifacial irradiance model in
+  :py:func:`pvlib.bifacial.ants2d.get_irradiance`. (:pull:`2740`)
+* Add parameters ``g0`` and ``g1`` to allow segmented ground surfaces in
+  :py:func:`~pvlib.bifacial.utils.vf_ground_sky_2d_integ` and
+  :py:func:`~pvlib.bifacial.utils.vf_row_ground_2d_integ`. (:pull:`2740`)
+* Accelerate :py:func:`~pvlib.bifacial.utils.vf_ground_sky_2d_integ` by one or
```
cwhanse: I think a comment here about "this makes `bifacial.infinite_sheds` faster would be appreciated.

pvlib/bifacial/utils.py
```diff
@@ -37,36 +38,42 @@ def _solar_projection_tangent(solar_zenith, solar_azimuth, surface_azimuth):
...
-    solar_zenith : numeric
-        Solar zenith angle. [degree].
-    solar_azimuth : numeric
-        Solar azimuth. [degree].
+    tracker_rotation : numeric
+        Tracker rotation angle as a right-handed rotation around
+        the same axis as ``phi``. [degree]
+    phi : numeric
+        Projected solar zenith angle, defined around the same axis as
+        ``tracker_rotation``. [degree].
```
cwhanse: I don't understand this definition. A vector can be projected onto a plane or another vector. An angle is formed by two vectors. "around" suggests an orientation of a vector. 

`tan_phi` was defined in the docstrings for `_solar_project_tangent` as

    Tangent of the angle between the zenith vector and the sun vector
    projected to the plane defined by the zenith vector and surface_azimuth.

So I would expect `phi` to be the angle of the sun vector projected to a plane formed by....

Since `phi` isn't clear, `tracker_rotation` becomes unclear also.

pvlib/bifacial/utils.py
```diff
@@ -37,36 +38,42 @@ def _solar_projection_tangent(solar_zenith, solar_azimuth, surface_azimuth):
...
     gcr : float
         Ground coverage ratio, which is the ratio of row slant length to row
         spacing (pitch). [unitless]
-    max_zenith : numeric, default 87
+    height : float, optional
+        Height of the center point of the row above the ground; must be in the
```
cwhanse: Center point means the center of the row slant length? If that's correct I'd make it

        Height above ground of the center point of the row slant length; must be in the

pvlib/bifacial/utils.py
```diff
@@ -37,36 +38,42 @@ def _solar_projection_tangent(solar_zenith, solar_azimuth, surface_azimuth):
...
+        same units as ``pitch``.  Required if ``g0`` is not zero or ``g1`` is
+        not one.
+    pitch : float, optional
+        Distance between two rows; must be in the same units as ``height``.
+        Required if ``g0`` is not zero or ``g1`` is not one.
+    g0 : numeric, default 0
+        Position on the ground surface, as a fraction of the row-to-row
```
cwhanse:
```suggestion
        Left position on the ground surface, as a fraction of the row-to-row
```

`g0` is the left position on the ground, I think.

pvlib/bifacial/utils.py
```diff
@@ -37,36 +38,42 @@ def _solar_projection_tangent(solar_zenith, solar_azimuth, surface_azimuth):
...
+        spacing. ``g0=0`` corresponds to ground underneath the middle of the
+        left row. ``g0`` should be less than ``g1``. [unitless]
+    g1 : numeric, default 1
+        Position on the ground surface, as a fraction of the row-to-row
```
cwhanse:
```suggestion
        Right position on the ground surface, as a fraction of the row-to-row
```

pvlib/bifacial/utils.py
```diff
@@ -37,36 +38,42 @@ def _solar_projection_tangent(solar_zenith, solar_azimuth, surface_azimuth):
...
+    height : float, optional
+        Height of the center point of the row above the ground; must be in the
+        same units as ``pitch``.  Required if ``g0`` is not zero or ``g1`` is
+        not one.
+    pitch : float, optional
+        Distance between two rows; must be in the same units as ``height``.
+        Required if ``g0`` is not zero or ``g1`` is not one.
+    g0 : numeric, default 0
+        Position on the ground surface, as a fraction of the row-to-row
+        spacing. ``g0=0`` corresponds to ground underneath the middle of the
```
cwhanse:
```suggestion
        spacing. ``g0=0`` corresponds to ground underneath the center of the
```

pvlib/bifacial/utils.py
```diff
@@ -37,36 +38,42 @@ def _solar_projection_tangent(solar_zenith, solar_azimuth, surface_azimuth):
...
+        left row. ``g0`` should be less than ``g1``. [unitless]
+    g1 : numeric, default 1
+        Position on the ground surface, as a fraction of the row-to-row
+        spacing. ``g1=1`` corresponds to ground underneath the middle of the
```
cwhanse:
```suggestion
        spacing. ``g1=1`` corresponds to ground underneath the center of the
```

pvlib/bifacial/utils.py
```diff
@@ -37,36 +38,42 @@ def _solar_projection_tangent(solar_zenith, solar_azimuth, surface_azimuth):
...
+        right row. ``g1`` should be greater than ``g0``. [unitless]
+    max_rows : int, default 10
+        Maximum number of rows to consider on either side of the current
+        row. [unitless]
```
cwhanse:
```suggestion
        Maximum number of rows to consider on either side of the left and right
        rows. [unitless]
```

Related to the `[max_rows, max_rows+1]` comment below.

pvlib/bifacial/utils.py
```diff
@@ -83,13 +90,71 @@ def _unshaded_ground_fraction(surface_tilt, surface_azimuth, solar_zenith,
...
+    tracker_rotation = np.atleast_1d(tracker_rotation)[np.newaxis, np.newaxis,
+                                                       :]
+    phi = np.atleast_1d(phi)[np.newaxis, np.newaxis, :]
+
+    g0 = np.atleast_1d(g0)[np.newaxis, :, np.newaxis]
+    g1 = np.atleast_1d(g1)[np.newaxis, :, np.newaxis]
+
+    # TODO seems like this should be np.arange(-max_rows, max_rows+1)?
+    # see GH #1867
+    k = np.arange(-max_rows, max_rows)[:, np.newaxis, np.newaxis]
```
cwhanse: I would do `(-max_rows, max_rows + 1)` or is this a breaking change needing deprecation?

pvlib/bifacial/utils.py
```diff
@@ -37,36 +38,42 @@ def _solar_projection_tangent(solar_zenith, solar_azimuth, surface_azimuth):
...
     gcr : float
         Ground coverage ratio, which is the ratio of row slant length to row
         spacing (pitch). [unitless]
-    max_zenith : numeric, default 87
+    height : float, optional
+        Height of the center point of the row above the ground; must be in the
+        same units as ``pitch``.  Required if ``g0`` is not zero or ``g1`` is
+        not one.
+    pitch : float, optional
+        Distance between two rows; must be in the same units as ``height``.
```
cwhanse:
```suggestion
        Distance between centers of two rows; must be in the same units as ``height``.
```

"centers" is arbritrary. I think we want to avoid the confusion that the distance is the gap between the right edge of left row, and the left edge of right row.

cwhanse (commented):
I'm OK reviewing the whole PR, if we can do it in a sequence of smaller review bites.

pvlib/bifacial/utils.py
kandersolar: Making these keyword-only arguments would be breaking, which makes me reluctant to take it on here.  @echedey-ls how about the alternative of moving `g0` and `g1` to the end of the signature so that existing usage is not affected?

pvlib/bifacial/utils.py
kandersolar: I edited to `Projected solar zenith angle, defined as a right-handed rotation around the same axis as ``tracker_rotation`` `, paralleling the description of `tracker_rotation`.  I would not object to taking the approach of the original `tan_phi` though.

pvlib/bifacial/utils.py
kandersolar: I went through and fixed all these to what I think makes sense.  Which `+` constant is appropriate depends on the function.  I also went ahead and fixed #1867, calling it a bugfix.

Edit: not sure what I've done is the right way forward. Will discuss in #1867.

pvlib/bifacial/utils.py
cwhanse: The angle phi is a rotation? Or is the projection of vectors that form phi somehow a rotation? I'm still not seeing this clearly.

pvlib/bifacial/utils.py
echedey-ls: That should work, yes, assuming nobody suppresses the warnings by setting those old params to `None`. Cause whenever they get deleted, then that function call would fail. I can only think of an approach that allows code to never fail, but that requires a deprecation period for positional params `max_rows` and `npoints` prior to merging this PR. [1, 2]

Your suggestion seems to be the most agile, the use-case I propose to be an absolute edge-case, and I prefer to merge this PR in next minor release rather than to make it wait for two minor versions.

[1] Decorator for deprecating positional-allowed params https://github.com/pyvista/pyvista/blob/main/pyvista/_deprecate_positional_args.py
[2] Example usage of [1] https://github.com/pyvista/pyvista/blob/21dd07fb3444356c61aacaf83510ba71cebd9780/pyvista/plotting/helpers.py#L65

pvlib/bifacial/utils.py
echedey-ls: How about
```
phi : numeric
        Projected solar zenith angle, defined as a right-handed rotation on
        the plane perpendicular to the rotation axis, with respect to horizontal. [degree].
```
?

Both `phi` and `tracker_rotation` are currently defined based one on another, which may make it extra difficult to see. I would add `rotation axis` to the definition of `tracker_rotation` too. And clarify later (notes section?, the same params definitions?, IDK) to interpret fixed-tilt as an specific case of SAT.

cwhanse: > Projected solar zenith angle, defined as a right-handed rotation on
        the plane perpendicular to the rotation axis, with respect to horizontal. [degree].

This still describes projecting an angle the way I read it. Angles aren't projected; vectors are. I think that the sun vector is projected, and the angle phi is formed by the projection of the sun vector to the plane described above, and the line of intersection of that plane with the horizontal plane.

I don't think there's a better name for the angle than `phi` and calling it `projected solar zenith angle` confuses the description.

What about

```
phi : numeric
    The angle between the sun vector projected to a plane defined by the zenith and surface_azimuth, and surface_azimuth.
```

I tried to draw a picture. 

proj_solar_zenith

echedey-ls: Oh, I now see your point. I agree with you. Probably it was a typo in `plane defined by the zenith and surface_azimuth, and surface_azimuth.`, but I'm convinced it is only a plane that contains `zenith` and `surface_azimuth` when `axis_tilt=0`. Otherwise the plane is not vertical. I lean towards the definition as the plane normal to the rotation axis (+ clarify fixed tilt as a special case of SAT, or even just link to ```:py:func:`pvlib.shading.projected_solar_zenith_angle` ``` @kandersolar ).

I suggest to add projected solar zenith to the nomenclature and agree on an informative acronym/param name for it, like `psza` that I've read when discussing other PRs. Retrospecting a little bit into #1904, maybe I should have left PSZA calculation up to the user, to reuse the value in other algorithms like this one...

kandersolar: I believe the original thinking for this name (which was some years ago now) was to call to mind the following parallel:
- solar zenith angle is the angle between (1) the solar vector and (2) the global z-axis.
- phi is the angle between (1) the solar vector projected onto the (possibly tilted) tracker x-z plane and (2) the tracker z-axis.

So, in the projected 2-D world we use for tracking, row-to-row shading etc, phi is in some sense the analog of the solar zenith angle.  But now I wish we had called this quantity something else, because of this good point:

> Angles aren't projected; vectors are.

Another proposal:
```
phi : numeric
    The angle between (1) the sun vector projected onto the plane perpendicular to
    the PV rows and (2) the vector normal to the plane containing the PV rows.
    See :py:func:`pvlib.shading.projected_solar_zenith_angle`.
```

cwhanse: "plane perpendicular to the PV rows" meaning the plane whose normal is the tracking axis?

kandersolar: Yes, avoiding using the term "tracking axis" to sidestep the tracker/fixed tilt issue.

cwhanse: Proposed phi definition is clear to me

pvlib/bifacial/utils.py
```diff
@@ -174,55 +239,132 @@ def vf_ground_sky_2d(rotation, gcr, x, pitch, height, max_rows=10):
...
     """
     Integrated view factor to the sky from the ground underneath
     interior rows of the array.
 
     Parameters
     ----------
-    surface_tilt : numeric
-        Surface tilt angle in degrees from horizontal, e.g., surface facing up
-        = 0, surface facing horizon = 90. [degree]
+    tracker_rotation : numeric
```
cwhanse, line 251: I'm not persuaded that `tracker_rotation` and `surface_tilt` are interchangeable. When `axis_tilt` is not 0 the rotation angle is not the same as the angle from horizontal. In this case, the 2D geometry being considered is in the plane perpendicular to the tracker rotation axis, not perpendicular to the ground. Does that matter?

Also, with `tracker_rotation` as the input, this public function relies on context outside the scope of the function: `axis_tilt` and `axis_aximuth` are needed to know what `tracker_rotation` is. If we need the input to be `tracker_rotation` then maybe consider making this function private.

pvlib/bifacial/utils.py
```diff
@@ -174,55 +237,130 @@ def vf_ground_sky_2d(rotation, gcr, x, pitch, height, max_rows=10):
...
+           This parameter has no effect; integrated view factors are now
+           calculated exactly instead of with discretized approximations.
+           This parameter will be removed in the future.
+
+    vectorize : bool, optional
+
+        .. deprecated:: 0.15.2
+
+           This parameter has no effect; calculations are now vectorized
+           with no memory usage penality.
```
cwhanse:
```suggestion
           with no memory usage penalty.
```

pvlib/bifacial/utils.py
```diff
@@ -174,55 +237,130 @@ def vf_ground_sky_2d(rotation, gcr, x, pitch, height, max_rows=10):
...
+           This parameter will be removed in the future.
 
     Returns
     -------
     fgnd_sky : numeric
         Integration of view factor over the length between adjacent, interior
         rows.  Shape matches that of ``surface_tilt``. [unitless]
```
cwhanse: Replace `surface_tilt` or not, pending outcome of comment about `tracker_rotation`

pvlib/bifacial/utils.py
```diff
@@ -382,12 +557,139 @@ def vf_row_ground_2d_integ(surface_tilt, gcr, x0=0, x1=1):
...
+
+    return ac, ad, bc, bd
+
+
+def _dist(p1, p2):
+    return ((p1[0] - p2[0])**2 + (p1[1] - p2[1])**2)**0.5
+
+
+def _angle(p1, p2):
+    return np.arctan2(p2[1] - p1[1], p2[0] - p1[0])
```
cwhanse, line 711:
```suggestion
    # polar angle of the ray from p1 to p2
    return np.arctan2(p2[1] - p1[1], p2[0] - p1[0])
```

pvlib/bifacial/utils.py
kandersolar: Implemented in fa94412

williamhobbs:
Is there a reason that `pvlib.irradiance.perez_driesse` is not an option? That would be useful to include if it's compatible.

pvlib/bifacial/utils.py
kandersolar: > I'm not persuaded that `tracker_rotation` and `surface_tilt` are interchangeable. When `axis_tilt` is not 0 the rotation angle is not the same as the angle from horizontal. In this case, the 2D geometry being considered is in the plane perpendicular to the tracker rotation axis, not perpendicular to the ground. Does that matter?

I think `tracker_rotation` is correct here:
- Adding `g0` and `g1` (rather than implicitly hardcoding them as 0/1) means that we can't assume symmetry around the ground midpoint anymore.  That means that we need to distinguish east- and west-facing rotations of the same tilt, so a signed quantity like `tracker_rotation` is needed.
- The VF must be calculated from the ground's point of view.  When `axis_tilt` is nonzero, that means the ground surface is tilted too (the model assumes that the tracking axis is parallel to the ground surface).  In the reference frame aligned with the ground, `tracker_rotation` is the required `surface_tilt` analog.

> Also, with `tracker_rotation` as the input, this public function relies on context outside the scope of the function: `axis_tilt` and `axis_azimuth` are needed to know what `tracker_rotation` is. If we need the input to be `tracker_rotation` then maybe consider making this function private.

Yeah, I guess it is a little awkward.  Making the function private seems a shame, since it's useful in other contexts.  Would noting the geometric link between `tracker_rotation`, `g0`, and `g1` be a sufficient alternative?  `g0` and `g1` are documented using left and right, so maybe something like:

```
    tracker_rotation : numeric
        Tracker rotation angle.  Positive rotations indicate raising the
        row's right edge.  [degree]
```

By the way, we are talking about `vf_ground_sky_2d_integ`, but the existing `vf_ground_sky_2d` signature has the same issue (tracker rotation input, but no axis tilt/azimuth): https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.bifacial.utils.vf_ground_sky_2d.html

kandersolar:
> Is there a reason that `pvlib.irradiance.perez_driesse` is not an option? That would be useful to include if it's compatible.

The only reason is that I did not think to include it in the reference paper, so including it here would be going beyond the reference.  If no reviewers object to that in this case, I'd be fine adding it in (and maybe documenting it as an extension).

williamhobbs:
>I'd be fine adding it in (and maybe documenting it as an extension).

That sounds great to me!

pvlib/bifacial/utils.py
cwhanse: " (the model assumes that the tracking axis is parallel to the ground surface)"

This is a key assumption and constraint. I think users may assume that `axis_tilt` allows for a tracker to have its rotation axis elevated at one end relative to the ground surface.

cwhanse: With this constraint, I agree that `tracker_rotation` and `surface_tilt` are interchangeable.

kandersolar: [d3679fe](https://github.com/pvlib/pvlib-python/pull/2740/commits/d3679fe54a465dd868ea177875a9428b40db3962) adds notes on this point.

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,603 @@
...
+        Tracker rotation angle as a right-handed rotation around
+        the axis defined by ``axis_tilt`` and ``axis_azimuth``.  For example,
+        with ``axis_tilt=0`` and ``axis_azimuth=180``, ``tracker_theta > 0``
+        results in ``surface_azimuth`` to the West while ``tracker_theta < 0``
+        results in ``surface_azimuth`` to the East. [degree]
+    phi : numeric
+        Projected solar zenith angle. [degrees]
+    gcr : numeric
+        Ground coverage ratio, which is the ratio of row slant length to row
+        spacing (pitch). [unitless]
```
cwhanse, line 39: Do we need to clarify here, that "row spacing" is a distance along the slanted ground, rather than the horizontal distance between row centers. I think that internal to `ants2d` gcr is the slanted distance, but for infinite sheds and other contexts, gcr is the horizontal distance.

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,603 @@
...
+from pvlib.tools import cosd, sind, tand, acosd
+from pvlib.bifacial import utils
+from pvlib.irradiance import aoi_projection, haydavies, perez, perez_driesse
+from pvlib.shading import projected_solar_zenith_angle, shaded_fraction1d
+from pvlib.tracking import calc_surface_orientation
+
+
+def _shaded_fraction(tracker_rotation, phi, gcr, x0=0, x1=1):
+    """
+    Calculate fraction (from the bottom) of row slant height that is shaded
```
cwhanse: If I follow the code in this function correctly, it is returning the fraction of the interval (x0, x1) that is shaded. I think that needs to be clear here at the function description.

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,603 @@
...
+    ground_irradiance : numeric
+        Irradiance incident on the ground surface, partitioned according
+        to ``x0`` and ``x1``. Sum of direct and diffuse components. [Wm⁻²]
+    albedo : numeric
+        Surface albedo. If a scalar, it is applied to all ground segments and
+        timestamps.  Otherwise, must be specified as an array with shape
+        (n_ground_segments, n_timestamps). [unitless]
+    x0 : numeric, default 0
+        Position on the row's slant length, as a fraction of the slant length.
+        ``x0=0`` corresponds to the left side of the row.
```
cwhanse:
```suggestion
        ``x0=0`` corresponds to the left edge of the row.
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,603 @@
...
+        ``x0`` should be less than ``x1``.  If specified as array, it
+        must have the same length as ``x1``. [unitless]
+    x1 : numeric, default 1
+        Position on the row's slant length, as a fraction of the slant length.
+        ``x1=1`` corresponds to the right side of the row.
```
cwhanse:
```suggestion
        ``x1=1`` corresponds to the right edge of the row.
```

pvlib/bifacial/utils.py
```diff
@@ -367,27 +535,178 @@ def vf_row_ground_2d_integ(surface_tilt, gcr, x0=0, x1=1):
...
     x1 : numeric, default 1.
         Position on the row's slant length, as a fraction of the slant length.
-        x1 should be greater than x0. [unitless]
+        ``x1`` should be greater than ``x0``. [unitless]
+    max_rows : int, default 20
+        Maximum number of rows to consider in front and behind the current row.
+    g0 : numeric, default 0
+        Position on the ground surface, as a fraction of the row-to-row
+        spacing. ``g0=0`` corresponds to ground underneath the middle of the
+        left row. ``g0`` should be less than ``g1``. [unitless]
```
cwhanse: I'm not clear what "left" and "right" row mean. Do they refer to the rows immediately to the left and right of the row being considered (where x0 and x1 live)? Or is the row being considered the "right" row?

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,603 @@
...
+        ``x1`` should be greater than ``x0``. If specified as array, it
+        must have the same length as ``x0``. [unitless]
+    g0 : numeric
+        Position on the ground surface, as a fraction of the row-to-row
+        spacing. ``g0=0`` corresponds to ground underneath the middle of the
+        left row. ``g0`` should be less than ``g1``. If specified as array, it
```
cwhanse, line 144: Comment on "left" and "right" made on `utils.vf_row_ground_2d_integ`

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,603 @@
...
+        segments the ground surface is divided into.  Otherwise, it must be
+        a list of pairs ``(g0, g1)`` where ``g0`` and ``g1`` are fractions of
+        the pitch and ``g0 < g1``.  The pairs must be non-overlapping and must
+        cover the entire ground surface.  ``albedo`` can be specified for
+        each segment.
+    axis_tilt : numeric, default 0
+        Tilt of the axis of rotation with respect to horizontal. [degree]
+    cross_axis_slope : numeric, default 0
+        The angle, relative to horizontal, of the line formed by the
+        intersection between the slope containing the tracker axes and a plane
```
cwhanse:
```suggestion
        intersection between the plane containing the tracker axes and a plane
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,603 @@
...
+        timestamps.  Otherwise, must be specified as an array with shape
+        (``n_ground_segments``, ``len(tracker_rotation)``). [unitless]
+    model : str, default 'perez'
+        Irradiance model - can be one of 'isotropic', 'haydavies', 'perez',
+        or 'perez_driesse'.
+    dni_extra : numeric, optional
+        Extraterrestrial direct normal irradiance. Required when
+        ``model='haydavies'``, ``model='perez'``, or ``model='perez_driese'``.
+        [Wm⁻²]
+    airmass : numeric, optional
```
cwhanse, line 373: Seems a bit awkward to default `model='perez'` which requires `airmass`, then describe `airmass` as optional with no default. Maybe default `model='isotropic'` which (I think) would be consistent with `infinite_sheds`.

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,603 @@
...
+                        cross_axis_slope):
+    slope_azimuth = axis_azimuth + np.degrees(
+        np.arctan2(sind(cross_axis_slope),
+                   cosd(cross_axis_slope) * sind(axis_tilt))
+    )
+    slope_tilt = acosd(cosd(axis_tilt) * cosd(cross_axis_slope))
+
+    height = height * cosd(slope_tilt)
+    pitch = pitch / cosd(cross_axis_slope)
+    gcr = gcr * cosd(cross_axis_slope)
```
cwhanse, line 284: This unsettles me. From here forward, `gcr` is not a horizontal distance between row centers. Are we confident that doesn't upset later functions that use `gcr`, e.g., `utils.vf_ground_sky_2d_integ`?

pvlib/bifacial/utils.py
kandersolar: Neither, really.  In these two parameter descriptions, "left" and "right" are referring to the members of any pair of adjacent rows, and `g0` and `g1` are fractions of the ground surface between them.  The (g0, g1) interval is repeated across all inter-row ground surfaces. 

The intent was to describe the origin and positive direction for `g0` and `g1`.  Clarifications welcome!

cwhanse (commented):
Review of file `ants2d.py`. OK with file `infinite_sheds.py`.

pvlib/bifacial/ants2d.py
kandersolar: It's not a horizontal distance in the user's reference frame, but it is horizontal in the rotated ANTS-2D frame, which is where the relevant calculations happen.  All downstream uses of `gcr` (and the other parameters transformed by this function) assume the array to be horizontal, and this function transforms all the relevant parameters accordingly.

pvlib/bifacial/ants2d.py
kandersolar: I'd rather not default to `isotropic`, in part because the paper focuses mostly on perez and also because of #1861.

What if we keep `perez` as the default and compute `airmass` and `dni_extra` automatically if needed?  It would be another extension of the paper, but a reasonable one IMHO.

cwhanse: I'm OK with this suggestion.

pvlib/bifacial/ants2d.py
cwhanse: I follow now. And it's immutable so no danger that a user's `gcr` is modified.

pvlib/bifacial/utils.py
```diff
@@ -367,27 +535,178 @@ def vf_row_ground_2d_integ(surface_tilt, gcr, x0=0, x1=1):
...
+    # dimensions: k/max_rows, ground segment, row segment, time
+
+    # cheat a little to prevent numerical issues with surface_tilt==180, -180
+    surface_tilt = np.where(surface_tilt == 180, 179.9999, surface_tilt)
+    surface_tilt = np.where(surface_tilt == -180, -179.9999, surface_tilt)
+
+    surface_tilt = \
+        np.atleast_1d(surface_tilt)[np.newaxis, np.newaxis, np.newaxis, :]
+
+    x0 = np.atleast_1d(x0)[np.newaxis, np.newaxis, :, np.newaxis]
```
cwhanse, line 600: Please add a comment describing the new dimensions. The request is to aid future maintainers.

pvlib/bifacial/utils.py
```diff
@@ -351,7 +518,8 @@ def vf_row_ground_2d(surface_tilt, gcr, x):
     return 0.5 * (1 - (1/gcr * cosd(surface_tilt) + x)/p)
 
 
-def vf_row_ground_2d_integ(surface_tilt, gcr, x0=0, x1=1):
+def vf_row_ground_2d_integ(surface_tilt, gcr, height=None, pitch=None,
+                           x0=0, x1=1, max_rows=20, g0=0, g1=1):
     r'''
     Calculate the average view factor to the ground from a segment of the row
     surface between x0 and x1.
```
cwhanse: Suggest

Calculate the view factor to the ground segment (g0, g1) from the row surface segment (x0, x1).

pvlib/bifacial/utils.py
cwhanse: I think the comment appears 9 lines earlier, I overlooked it.

pvlib/bifacial/utils.py
cwhanse:
```suggestion
        Position on the ground surface, as a fraction of the spacing between
        adjacent rows. ``g0=0`` corresponds to ground underneath the middle of the
        left row. ``g0`` should be less than ``g1``. [unitless]
```

pvlib/bifacial/utils.py
```diff
@@ -367,27 +535,178 @@ def vf_row_ground_2d_integ(surface_tilt, gcr, x0=0, x1=1):
...
+        ``x1`` should be greater than ``x0``. [unitless]
+    max_rows : int, default 20
+        Maximum number of rows to consider in front and behind the current row.
+    g0 : numeric, default 0
+        Position on the ground surface, as a fraction of the row-to-row
+        spacing. ``g0=0`` corresponds to ground underneath the middle of the
+        left row. ``g0`` should be less than ``g1``. [unitless]
+    g1 : numeric, default 1
+        Position on the ground surface, as a fraction of the row-to-row
+        spacing. ``g1=0`` corresponds to ground underneath the middle of the
```
cwhanse:
```suggestion
        Position on the ground surface, as a fraction of the spacing between
        adjacent rows. ``g1=0`` corresponds to ground underneath the middle of the
```

pvlib/bifacial/ants2d.py
cwhanse: Edited: "...internal to `ants2d` gcr is the slanted distance..."

tests/bifacial/test_ants2d.py
```diff
@@ -0,0 +1,656 @@
...
+    fs = ants2d._shaded_fraction(tracker_rotation, phi, gcr)
+    assert_allclose(fs, expected)
+    fs = ants2d._shaded_fraction(-tracker_rotation, -phi, gcr)
+    assert_allclose(fs, expected)
+
+    # sun too high for shade
+    assert 0 == ants2d._shaded_fraction(10, 20, 0.5)
+    assert 0 == ants2d._shaded_fraction(-10, -20, 0.5)
+
+    # sun behind the modules (AOI > 90)
```
cwhanse: Very minor nit: "sun behind the modules" I think I would comment this case as "module facing away from the sun" since the function being tested returns "fraction...shaded from direct irradiance by the row in front toward the sun." It is implied that `_shaded_fraction` is specific to one face of a module, which (in this case) is face pointed away from the sun.

cwhanse (approved):
Thorough tests!

williamhobbs:
I think the documentation for `ants2d.get_irradiance` needs clarification on how to model fixed-tilt systems. I.e., how to convert tilt angle and azimuth into `tracker_rotation` and `axis_azimuth` (and similar for `axis_tilt` and `cross_axis_slope`).

williamhobbs:
In `pvlib.irradiance.get_total_irradiance`, the input for `model` for Perez-Driesse is `perez-driesse`, but here it is `perez_driesse`. I suggest changing it to `perez-driesse` for consistency, unless there is a reason to do it differently here.

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,625 @@
...
+        Diffuse horizontal irradiance. [Wm⁻²]
+    dni : numeric
+        Direct normal irradiance. [Wm⁻²]
+    albedo : numeric
+        Surface albedo. If a scalar, it is applied to all ground segments and
+        timestamps.  Otherwise, must be specified as an array with shape
+        (``n_ground_segments``, ``len(tracker_rotation)``). [unitless]
+    model : str, default 'perez'
+        Irradiance model - can be one of 'isotropic', 'haydavies', 'perez',
+        or 'perez_driesse'.
```
williamhobbs:
```suggestion
        or 'perez-driesse'.
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,625 @@
...
+    dni_extra : numeric, optional
+        Extraterrestrial direct normal irradiance. Required when
+        ``model='haydavies'``, ``model='perez'``, or ``model='perez_driese'``.
```
williamhobbs:
```suggestion
        ``model='haydavies'``, ``model='perez'``, or ``model='perez-driesse'``.
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,625 @@
...
+        Computed from the datetime index of the inputs if not provided. [Wm⁻²]
+    airmass : numeric, optional
+        Relative airmass. Required when ``model='perez'`` or
+        ``model='perez_driese'``. Computed from ``solar_zenith`` if needed.
```
williamhobbs:
```suggestion
        ``model='perez-driesse'``. Computed from ``solar_zenith`` if needed.
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,625 @@
...
+    }
+    return output
+
+
+def _apply_sky_diffuse_model(dni, dhi, model, solar_zenith, solar_azimuth,
+                             dni_extra, airmass):
+
+    # Section II.C.2
+
+    if model in ['haydavies', 'perez', 'perez_driesse']:
```
williamhobbs:
```suggestion
    if model in ['haydavies', 'perez', 'perez-driesse']:
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,625 @@
...
+        # determine circumsolar irradiance, add it to DNI
+
+        if model == 'haydavies':
+            if dni_extra is None:
+                raise ValueError(f'Must supply dni_extra for {model} model')
+            diffuse_model_func = haydavies
+            extra_kwargs = {}
+
+        elif model in ['perez', 'perez_driesse']:
```
williamhobbs:
```suggestion
        elif model in ['perez', 'perez-driesse']:
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,625 @@
...
+        The following quantities are included:
+
+        - ``ground_direct``: direct irradiance incident on the ground surface.
+          [Wm⁻²]
+        - ``ground_diffuse``: diffuse irradiance incident on the ground
+          surface. [Wm⁻²]
+
+    Notes
+    -----
+    - ``model='perez_driesse'`` was not evaluated in [1]_ and is included
```
williamhobbs:
```suggestion
    - ``model='perez-driesse'`` was not evaluated in [1]_ and is included
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,625 @@
...
+        # get the index of the first pandas input, if there is one
+        pd_index = next(
+            x.index for x in maybe_array_inputs if isinstance(x, pd.Series)
+        )
+    except StopIteration:
+        pd_index = None  # no pandas inputs
+
+    # preparation steps
+
+    if airmass is None and model in ['perez', 'perez_driesse']:
```
williamhobbs:
```suggestion
    if airmass is None and model in ['perez', 'perez-driesse']:
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,625 @@
...
+        airmass = get_relative_airmass(solar_zenith)
+
+    if (dni_extra is None and
+            model in ['perez', 'perez_driesse', 'haydavies'] and
```
williamhobbs:
```suggestion
            model in ['perez', 'perez-driesse', 'haydavies'] and
```

pvlib/bifacial/ants2d.py
williamhobbs: note that there was also a type here

pvlib/bifacial/ants2d.py
williamhobbs: note that there was also a type here

williamhobbs:
Could it be worthwhile to return *unshaded* irradiance, e.g., `poa_front_direct_unshaded`? It would help with some types of shade loss models [1, 2] and with general loss accounting (e.g., if someone wanted to make waterfall/Sankey charts). 

I haven't fully thought through trade-offs and alternatives yet, but I thought it was worth bringing up. 

[1] Anderson, Kevin. 2020. Maximizing Yield with Improved Single-Axis Backtracking on Cross-Axis Slopes: Preprint. Golden, CO: National Renewable Energy Laboratory. NREL/CP-5K00-76023. https://www.nlr.gov/docs/fy20osti/76023.pdf

[2] W. Hobbs, K. Anderson, M. Mikofski, and M. Ghiz, “An approach to modeling linear and non-linear self-shading losses with PVlib,” in PV Perform. Model. Collaborative Workshop, Sandia Nat. Lab., May 2024. [Online]. Available: https://www.sandia.gov/app/uploads/sites/243/dlm_uploads/2024/05/2024_PVPMC_hobbs_pvlib_self-shade.pdf

kandersolar:
> Could it be worthwhile to return unshaded irradiance, e.g., poa_front_direct_unshaded? [...]

I think this would be easy enough for the direct component, but sky diffuse would need at least a little bit extra to disentangle r2r shading from the reduction due to surface tilt.  Currently, it uses a single VF that accounts for both effects.

The alternative would, I think, be to just use `get_total_irradiance` with the same inputs.  Not as convenient for the user, but maybe good enough?

williamhobbs:
Hmm. I think that unshaded direct is all I would need for shade loss modeling. So having that would be convenient.

But I see what you mean about diffuse. And maybe returning unshaded direct but not unshaded diffuse could confuse users?

Anyone else have thoughts?

docs/sphinx/source/user_guide/modeling_topics/bifacial.rst
```diff
@@ -114,6 +114,19 @@ as additional ``dni`` for transposition and shading purposes.
 This model is influenced by the 2D model published by Marion, *et al.* in [2].
 
 
+ANTS-2D
```
AdamRJensen, line 118: Should add a link to the paper and the main function.

Also, there should be added an extra bullet at the top of the section introducing ANTS-2D as the third option in pvlib.

pvlib/bifacial/infinite_sheds.py
```diff
@@ -235,14 +236,25 @@ def get_irradiance_poa(surface_tilt, surface_azimuth, solar_zenith,
         Incidence angle modifier, the fraction of direct irradiance incident
         on the surface that is not reflected away. [unitless]
 
-    npoints : int, default 100
+    npoints : int, optional
         Number of discretization points for calculating integrated view
         factors.
 
-    vectorize : bool, default False
+        .. deprecated:: 0.15.2
```
AdamRJensen: Update version. Also below.

pvlib/bifacial/utils.py
```diff
@@ -37,38 +38,46 @@ def _solar_projection_tangent(solar_zenith, solar_azimuth, surface_azimuth):
...
         Maximum zenith angle. For solar_zenith > max_zenith, unshaded ground
         fraction is set to 0. [degree]
+    g0 : numeric, default 0
+        Left position on the ground surface, as a fraction of the row-to-row
+        spacing. ``g0=0`` corresponds to ground underneath the center of the
+        left row. ``g0`` should be less than ``g1``. [unitless]
+    g1 : numeric, default 1
+        Right position on the ground surface, as a fraction of the row-to-row
+        spacing. ``g1=1`` corresponds to ground underneath the center of the
+        right row. ``g1`` should be greater than ``g0``. [unitless]
```
AdamRJensen:
```suggestion
        left tracker post. ``g0`` should be less than ``g1``. [unitless]
    g1 : numeric, default 1
        Right position on the ground surface, as a fraction of the row-to-row
        spacing. ``g1=1`` corresponds to ground underneath the center of the
        right tracker post. ``g1`` should be greater than ``g0``. [unitless]
```

pvlib/bifacial/utils.py
```diff
@@ -298,19 +442,41 @@ def vf_row_sky_2d_integ(surface_tilt, gcr, x0=0, x1=1):
         Ratio of the row slant length to the row spacing (pitch). [unitless]
     x0 : numeric, default 0
         Position on the row's slant length, as a fraction of the slant length.
-        x0=0 corresponds to the bottom of the row. x0 should be less than x1.
-        [unitless]
+        ``x0=0`` corresponds to the bottom of the row. ``x0`` should be less
+        than ``x1``. [unitless]
```
AdamRJensen:
```suggestion
        ``x0=0`` corresponds to the lower edge of the row. ``x0`` should
        be less than ``x1``. [unitless]
```

pvlib/bifacial/utils.py
```diff
@@ -351,10 +518,11 @@ def vf_row_ground_2d(surface_tilt, gcr, x):
...
 
 
-def vf_row_ground_2d_integ(surface_tilt, gcr, x0=0, x1=1):
+def vf_row_ground_2d_integ(surface_tilt, gcr, height=None, pitch=None,
+                           x0=0, x1=1, max_rows=20, g0=0, g1=1):
     r'''
-    Calculate the average view factor to the ground from a segment of the row
-    surface between x0 and x1.
+    Calculate the view factor to the ground segment (g0, g1) from the row
+    surface segment (x0, x1).
```
AdamRJensen: Backticks around x and g?

pvlib/bifacial/utils.py
```diff
@@ -367,27 +535,177 @@ def vf_row_ground_2d_integ(surface_tilt, gcr, x0=0, x1=1):
         = 0, surface facing horizon = 90. [degree]
     gcr : numeric
         Ratio of the row slant length to the row spacing (pitch). [unitless]
+    height : float, optional
+        Height of the center point of the row above the ground; must be in the
+        same units as ``pitch``.  Required if ``g0`` or ``x0` is not zero or
+        if ``g1`` or ``x1`` is not one.
```
AdamRJensen, line 541: Should the "or" be changed to "and"?

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,633 @@
...
+def get_irradiance(tracker_rotation, axis_azimuth, solar_zenith, solar_azimuth,
+                   gcr, height, pitch, ghi, dhi, dni,
+                   albedo, model='perez', dni_extra=None, airmass=None,
+                   row_segments=1, ground_segments=10, axis_tilt=0,
+                   cross_axis_slope=0, max_rows=None,
+                   return_ground_components=False):
+    """
+    Get front and rear irradiance using the ANTS-2D bifacial irradiance model.
+
+    The ANTS-2D model [1] assumes the PV system comprises parallel,
```
AdamRJensen:
```suggestion
    The ANTS-2D model [1]_ assumes the PV system comprises parallel,
```

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,633 @@
...
+    evenly spaced rows on flat or uniformly sloped ground. Rows can be on fixed
+    racking or single axis trackers. The model calculates irradiance at a
+    location far from the ends of any rows, in effect, assuming that the
+    rows (sheds) are infinitely long, and parallel to the (potentially sloped)
+    ground surface.
+
+    The model accounts for the following effects:
```
AdamRJensen, line 315: I'm missing some mention that the model can output irradiances on discretized surfaces (which could be used for cell-level irradiance modeling).

pvlib/bifacial/ants2d.py
```diff
@@ -0,0 +1,633 @@
...
+        - ``shaded_fraction_back``: fraction of row slant height that is
+          shaded from direct irradiance on the back surface by adjacent
+          rows. [unitless]
+
+    ground_irradiance : dict or DataFrame
+        ``ground_irradiance`` is a DataFrame when inputs are Series
+        and ``n_ground_segments=1``, a dict of scalars when inputs are scalars
+        and ``n_ground_segments=1``, and a dict of ``np.ndarray``
+        otherwise.  Only returned when ``return_ground_components=True``.
+        The following quantities are included:
```
AdamRJensen, line 451:
```suggestion
        Only returned when ``return_ground_components=True``.
        ``ground_irradiance`` is a DataFrame when inputs are Series
        and ``n_ground_segments=1``, a dict of scalars when inputs are scalars
        and ``n_ground_segments=1``, and a dict of ``np.ndarray``
        otherwise. 
        The following quantities are included:
```

I suggest the first line mentions that this is only returned in special cases.

AdamRJensen (approved):

AdamRJensen:
In the future, the gallery example [plot_agrivoltaics_ground_irradiance.py](https://github.com/pvlib/pvlib-python/pull/2740/changes#diff-6952d3bc2134822a17d39102db7f67dc5879230271d566f2e9d116673f4aeec1) should be modified to use ANTS-2D.

pvlib/bifacial/utils.py
kandersolar: "or" is correct here; you can only ignore height if treating the entire ground surface as one segment.  Changing either endpoint makes height required.

kandersolar:
> Hmm. I think that unshaded direct is all I would need for shade loss modeling. So having that would be convenient.

@AdamRJensen points out that get_total_irradiance can't handle sloped terrain and nonuniform albedo, so nevermind about what I said earlier about it maybe being a suitable replacement (at least, in the general case).

However, I suppose many cases could get by with just using the irradiance for the top cell as the "unshaded" irradiance, so I will leave this for a future improvement.

PR #2741, opened 2026-04-23
Replace interp1d #2394

### LINKED ISSUE #2394: `scipy.interpolate.interp1d` is discouraged
SciPy now discourages the use of `interp1d`.  From https://docs.scipy.org/doc/scipy-1.15.2/reference/generated/scipy.interpolate.interp1d.html :

> This class is considered legacy and will no longer receive updates. While we currently have no plans to remove it, we recommend that new code uses more modern alternatives instead. For a guide to the intended replacements for [interp1d](https://docs.scipy.org/doc/scipy-1.15.2/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d) see [1-D interpolation](https://docs.scipy.org/doc/scipy-1.15.2/tutorial/interpolate/1D.html#tutorial-interpolate-1dsection).

We use `interp1d` in a few places:

```bash
$ git grep -n interp1d
docs/examples/shading/plot_partial_module_shading_simple.py:41:from scipy.interpolate import interp1d
docs/examples/shading/plot_partial_module_shading_simple.py:181:    """convenience wrapper around scipy.interpolate.interp1d"""
docs/examples/shading/plot_partial_module_shading_simple.py:182:    f_interp = interp1d(np.flipud(df['i']), np.flipud(df['v']), kind='linear',
pvlib/iam.py:443:        See scipy.interpolate.interp1d for more options.
pvlib/iam.py:473:    from scipy.interpolate import interp1d
pvlib/iam.py:486:    interpolator = interp1d(theta_ref, iam_ref, kind=method,
pvlib/spectrum/response.py:9:from scipy.interpolate import interp1d
pvlib/spectrum/response.py:70:    interpolator = interp1d(SR_DATA[0], SR_DATA[1],
```

I suppose there is no rush to change, but updating these to use whatever scipy recommends at that link might be a good first PR for some pvlib user interested in becoming a contributor.

jason-rpkt:
- [X] Closes #2394 
 - [X] I attest that all AI-generated material has been vetted for accuracy and is in compliance with the pvlib license
 - [X] Tests added

The existing tests pass. I have added 2 in test_iam.py under test_iam_interp.


cwhanse:
Reason for importing `CubicSpline` instead of using `make_interp_spline(..., k=3)`? The latter would clean up the case processing, I think.

jason-rpkt:
no particular reason. Just that I had seen `CubicSpline` first in https://docs.scipy.org/doc/scipy-1.17.0/tutorial/interpolate/1D.html#tutorial-interpolate-1dsection and in response.py it was hardcoded as cubic. 

I have replaced `CubicSpline` by `make_interp_spline` with k=3. Now only importing `make_interp_spline`.

pvlib/spectrum/response.py
```diff
@@ -66,16 +66,15 @@ def get_example_spectral_response(wavelength=None):
     if wavelength is None:
         resolution = 5.0
         wavelength = np.arange(280, 1200 + resolution, resolution)
+    x = SR_DATA[0]
+    y = SR_DATA[1]
+    spline = make_interp_spline(
+        x, y, k=3)
```
kandersolar:
```suggestion
    spline = make_interp_spline(x, y, k=3)
```

No reason to not fit this on one line, right?

pvlib/iam.py
```diff
@@ -470,7 +470,7 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
     '''
     # Contributed by Anton Driesse (@adriesse), PV Performance Labs. July, 2019
 
-    from scipy.interpolate import interp1d
+    from scipy.interpolate import make_interp_spline
```
kandersolar: This line being here inside the function (instead of at the top of the file as usual) is a relic of the days when scipy was an optional dependency.  Optional for this PR, but moving the import to the top would be an improvement.

pvlib/iam.py
```diff
@@ -483,10 +483,31 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
         raise ValueError("Negative value(s) found in 'iam_ref'. "
                          "This is not physically possible.")
 
-    interpolator = interp1d(theta_ref, iam_ref, kind=method,
-                            fill_value='extrapolate')
-    aoi_input = aoi
+    theta_ref = np.asarray(theta_ref)
+    iam_ref = np.asarray(iam_ref)
```
kandersolar: Why is `np.asarray` needed here?

pvlib/iam.py
```diff
@@ -483,10 +483,31 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
...
+            return spline(x)
 
+    elif method == "cubic":
+        spline = make_interp_spline(theta_ref, iam_ref, k=3)
+
+        def interpolator(x):
+            return spline(x)
+
+    else:
+        raise ValueError(f"Invalid interpolation method '{method}'.")
```
kandersolar: This is, technically, a breaking change, since the `interp1d` way also supported `'nearest'`, `'nearest-up'`, `'zero'`, `'slinear'`, `'previous'`, and `'next'`.  I doubt these got much use, if any.  Any thoughts on how to handle that?

pvlib/iam.py
```diff
@@ -483,10 +483,31 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
...
+        spline = make_interp_spline(theta_ref, iam_ref, k=2)
+
+        def interpolator(x):
+            return spline(x)
 
+    elif method == "cubic":
+        spline = make_interp_spline(theta_ref, iam_ref, k=3)
+
+        def interpolator(x):
+            return spline(x)
```
kandersolar: This can surely be simplified as @cwhanse mentioned in https://github.com/pvlib/pvlib-python/pull/2741#issuecomment-4305763777

pvlib/iam.py
```diff
@@ -440,7 +440,7 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
     method : str, default 'linear'
         Specifies the interpolation method.
         Useful options are: 'linear', 'quadratic', 'cubic'.
-        See scipy.interpolate.interp1d for more options.
+        See scipy.interpolate for more options.
```
kandersolar: This line may need to be edited, depending on https://github.com/pvlib/pvlib-python/pull/2741/changes#r3137773820

pvlib/spectrum/response.py
jason-rpkt: made into one line

pvlib/iam.py
jason-rpkt: moved import to top

pvlib/iam.py
jason-rpkt: simplified

pvlib/iam.py
jason-rpkt: I'm not sure I know how to update to support all of those. 
Is it OK to update the error message?

pvlib/iam.py
jason-rpkt: Do you mean to use `np.asanyarray` to keep consistency with `np.asanyarray(aoi)` below? Or to remove entirely? 

`np.asarray` is to convert to numpy array but I can remove if you prefer.

cwhanse: I think the issue is that the docstrings states `numeric` for input parameters. pvlib means `numeric` as one of float, array or Series. Testing is all in one function `test_iam_interp` which inputs either list, float or array (but not Series).

My opinion: rewrite the test to remove inputting lists, which obviates `np.asarray` inside `iam.interp`. And add a test that inputs Series.

jason-rpkt: Hi both, thanks for your comments. Not sure if I got this right, but I have rewritten the test to remove inputting lists. (And removed `np.asarray `in `iam.interp` ) 
I have also added a check for list inputs at the beginning of `iam.interp `
Let me know if this is OK. 

--Looking at it again, it seems that keeping `np.asarray` might be "cleaner" than the latest version.

pvlib/iam.py
```diff
@@ -483,10 +485,19 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
...
+        interpolator = make_interp_spline(theta_ref, iam_ref, k=1)
+
+    elif method == "quadratic":
+        interpolator = make_interp_spline(theta_ref, iam_ref, k=2)
 
+    elif method == "cubic":
+        interpolator = make_interp_spline(theta_ref, iam_ref, k=3)
+
+    else:
+        raise ValueError(f"Invalid interpolation method '{method}'.")
```
```suggestion
        raise ValueError(f"Interpolation method '{method}' is not supported in pvlib-python.")
```

pvlib/iam.py
```diff
@@ -469,9 +470,10 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
     pvlib.iam.sapm
     '''
     # Contributed by Anton Driesse (@adriesse), PV Performance Labs. July, 2019
-
-    from scipy.interpolate import interp1d
-
+    if isinstance(theta_ref, list):
```
cwhanse: We normally don't check the type of input. If a user inputs a non-allowed type e.g. a list, `make_interp_spline` will emit an error. That's OK.

pvlib/iam.py
cwhanse: Since `interp1d` is legacy but not intended for removal, it is still available and we could fall back:

```
elif method in {'nearest', 'nearest-up', 'zero', 'slinear', 'previous', 'next'}:
    interpolator = interp1d(theta_ref, iam_ref, kind=method,
                               fill_value='extrapolate')
```

kandersolar: I think that's reasonable, as long as we deprecate these weird `method`s too.   If someone really wants these special interpolations, they can do it themselves and provide a pre-calculated input.

pvlib/iam.py
```diff
@@ -469,9 +470,10 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
...
+        raise TypeError("theta_ref cannot be a list")
+    if isinstance(iam_ref, list):
+        raise TypeError("iam_ref cannot be a list")
     # Scipy doesn't give the clearest feedback, so check number of points here.
     MIN_REF_VALS = {'linear': 2, 'quadratic': 3, 'cubic': 4, 1: 2, 2: 3, 3: 4}
```
adriesse, line 474: You could add a dictionary with names and k values, or something like that, to use below.

pvlib/iam.py
adriesse: None of those options really make any sense here.  I would vote for dropping them immediately.

docs/examples/shading/plot_partial_module_shading_simple.py
```diff
@@ -178,9 +178,9 @@ def plot_curves(dfs, labels, title):
 
 
 def interpolate(df, i):
-    """convenience wrapper around scipy.interpolate.interp1d"""
-    f_interp = interp1d(np.flipud(df['i']), np.flipud(df['v']), kind='linear',
-                        fill_value='extrapolate')
+    """convenience wrapper around scipy.interpolate"""
+    f_interp = make_interp_spline(np.flipud(df['i']), np.flipud(df['v']), k=1)
+
```
adriesse: This is one where performance is likely important.  Maybe check if np.interp is faster?

pvlib/iam.py
```diff
@@ -483,10 +481,21 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
...
-    aoi_input = aoi
+    if method == "linear":
+        interpolator = make_interp_spline(theta_ref, iam_ref, k=1)
+
+    elif method == "quadratic":
+        interpolator = make_interp_spline(theta_ref, iam_ref, k=2)
 
+    elif method == "cubic":
+        interpolator = make_interp_spline(theta_ref, iam_ref, k=3)
+
```
```suggestion
    kvals = {'linear': 1, 'quadratic': 2, 'cubic': 3}
    if method in kvals:
        interpolator = make_interp_spline(theta_ref, iam_ref, k=kvals[method])

```

pvlib/iam.py
```diff
@@ -483,10 +481,16 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
...
                          "This is not physically possible.")
 
-    interpolator = interp1d(theta_ref, iam_ref, kind=method,
-                            fill_value='extrapolate')
-    aoi_input = aoi
+    kvals = {'linear': 1, 'quadratic': 2, 'cubic': 3}
+    if method in kvals:
+        interpolator = make_interp_spline(theta_ref, iam_ref, k=kvals[method])
 
+    else:
```
cwhanse, line 497:
```suggestion
    elif method in {'nearest', 'nearest-up', 'zero', 'slinear', 'previous', 'next'}:
        msg = (
            f"Interpolation method {method} is deprecated in pvlib"
        )
        warn_deprecated(since="0.15.3", removal="0.16.0", addendum=msg)
        interpolator = interp1d(theta_ref, iam_ref, kind=method, fill_value='extrapolate')
    else:
```

Out of an abundance of caution and perhaps formality, let's deprecate the old methods instead of removing immediately.

@jason-rpkt please fix this if it doesn't run. You'll need to import `warn_deprecated` from pvlib._deprecation. We should add a test that checks that the deprecation and value errors are raised, for code coverage purposes.

pvlib/iam.py
```diff
@@ -440,7 +442,7 @@ def interp(aoi, theta_ref, iam_ref, method='linear', normalize=True):
     method : str, default 'linear'
         Specifies the interpolation method.
         Useful options are: 'linear', 'quadratic', 'cubic'.
-        See scipy.interpolate.interp1d for more options.
+        See scipy.interpolate for more options.
```
```suggestion
```

cwhanse (approved):

pvlib/iam.py
cwhanse: PR needs an entry in the v0.15.3 whatsnew file.

docs/sphinx/source/whatsnew/v0.15.3.rst
```diff
@@ -10,7 +10,7 @@ Breaking Changes
 
 Deprecations
 ~~~~~~~~~~~~
-
+* Deprecate scipy.interpolate.interp1d. (:issue:`2394`, :pull:`2741`)
```
```suggestion
* Deprecate non-polynomial interpolation options in :py:func:`pvlib.iam.interp` (:issue:`2394`, :pull:`2741`)
```

cwhanse:
thanks for your persistence @jason-rpkt

PR #2749, opened 2026-05-05
Fix calc_surface_orientation broadcasting for >1-D tracker_theta

### LINKED ISSUE #2747: calc_surface_orientation broadcasting error for 2-D inputs in 0.15.1 (regression vs 0.15.0)
**Describe the bug**

`pvlib.tracking.calc_surface_orientation` regressed in 0.15.1: passing 2-D arrays for `tracker_theta` (and broadcasting-compatible 2-D `axis_tilt` / `axis_azimuth`, e.g. shape `(timestamps, sites)`) raises a broadcasting `ValueError`. The same call works in 0.15.0 and earlier.

The cause is the new `_unit_normal` helper introduced in #2702, which builds its return value with `np.column_stack((x, y, z))`. `np.column_stack` only stacks 1-D inputs into columns; for 2-D inputs of shape `(T, N)` it concatenates along axis=1, producing a `(T, 3*N)` array instead of the intended `(T, N, 3)`. The subsequent `unit_normal[:, 0]` / `unit_normal[:, 1]` then collapse to shape `(T,)`, and the downstream

```python surface_azimuth = np.where(surface_tilt == 0., axis_azimuth - 90., surface_azimuth)```

fails because the condition and true branch are still `(T, N)` while the false branch is now `(T,)`.

**To Reproduce**

```python
import numpy as np
import pvlib

T, N = 4655, 105
tracker_theta = np.zeros((T, N))
axis_tilt = np.zeros((T, N))
axis_azimuth = np.full((T, N), 180.0)

pvlib.tracking.calc_surface_orientation(tracker_theta, axis_tilt, axis_azimuth)
# pvlib 0.15.0: returns dict of (T, N) arrays — works
# pvlib 0.15.1: ValueError: operands could not be broadcast together with shapes (4655,105) (4655,105) (4655,)
```

Bisected against pip-installed releases on Python 3.11 / numpy 2.4.4 / pandas 3.0.2:

| pvlib | result |
|---|---|
| 0.11.2 → 0.15.0 | works |
| 0.15.1 | broken |

**Expected behavior**

Return `surface_tilt` and `surface_azimuth` arrays with the same shape as the broadcast inputs (the 0.15.0 behavior).

**Versions**

- ``pvlib.__version__``: 0.15.1 (broken), 0.15.0 (works)
- ``pandas.__version__``: 3.0.2
- ``numpy.__version__``: 2.4.4
- python: 3.11.9
- platform: Windows

**Additional context**

Regression introduced in #2702 (milestone v0.15.1). Suggested one-line fix in
`_unit_normal`:

```python
# before
return np.column_stack((x, y, z))
# after
return np.stack((x, y, z), axis=-1)
```

`np.stack(..., axis=-1)` preserves the leading dimensions for inputs of any rank and is equivalent to the old behavior for 1-D inputs, so it should be a drop-in replacement.

Hit in production while batch-processing tracker geometry across (timestamps × bays); pinned to `pvlib<0.15.1` as a workaround.

ArthurOnnoTerabase:
The unit-normal-based implementation introduced in v0.15.1 (#2702) used np.column_stack inside _unit_normal, which only handles 1-D inputs correctly.  For 2-D (or higher rank) tracker_theta the column-stacked array had the wrong shape, and the subsequent np.where in calc_surface_orientation raised a ValueError.

Switch to np.stack(..., axis=-1) and use ellipsis indexing in calc_surface_orientation so the function preserves the input rank for arbitrary-shape tracker_theta.  np.atleast_1d preserves the historical (1, 3) return shape of _unit_normal for scalar input.

Also adds a parametrised regression test that exercises 2-D and 3-D tracker_theta inputs.

Closes #2747

 - [X] Closes #2747 
 - [X] I attest that all AI-generated material has been vetted for accuracy and is in compliance with the pvlib license
 - [X] Tests added
 - [X] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).

### DIFF
--- a/tests/test_tracking.py
+++ b/tests/test_tracking.py
@@ -564,8 +564,7 @@ def test_calc_surface_orientation_2d(shape):
     # Regression test for GH#2747: calc_surface_orientation must accept
     # tracker_theta of arbitrary rank, not just 1-D.  Compare the >1-D result
     # to the 1-D result computed on the flattened input.
-    rng = np.random.default_rng(0)
-    rotations_flat = rng.uniform(-90, 90, size=int(np.prod(shape)))
+    rotations_flat = np.linspace(-90, 90, int(np.prod(shape)))
     rotations_nd = rotations_flat.reshape(shape)
 
     out_1d = tracking.calc_surface_orientation(rotations_flat,
--- a/tests/test_tracking.py
+++ b/tests/test_tracking.py
@@ -557,3 +557,32 @@ def test_calc_surface_orientation_special():
         # in a modulo-360 sense.
         np.testing.assert_allclose(np.round(out['surface_azimuth'], 4) % 360,
                                    expected_azimuths, rtol=1e-5, atol=1e-5)
+
+
+@pytest.mark.parametrize('shape', [(3, 5), (1, 7), (4, 1), (2, 3, 4)])
+def test_calc_surface_orientation_2d(shape):
+    # Regression test for GH#2747: calc_surface_orientation must accept
+    # tracker_theta of arbitrary rank, not just 1-D.  Compare the >1-D result
+    # to the 1-D result computed on the flattened input.
+    rng = np.random.default_rng(0)
+    rotations_flat = rng.uniform(-90, 90, size=int(np.prod(shape)))
+    rotations_nd = rotations_flat.reshape(shape)
+
+    out_1d = tracking.calc_surface_orientation(rotations_flat,
+                                               axis_tilt=20,
+                                               axis_azimuth=180)
+    out_nd = tracking.calc_surface_orientation(rotations_nd,
+                                               axis_tilt=20,
+                                               axis_azimuth=180)
+
+    assert out_nd['surface_tilt'].shape == shape
+    assert out_nd['surface_azimuth'].shape == shape
+    np.testing.assert_allclose(out_nd['surface_tilt'].reshape(-1),
+                               out_1d['surface_tilt'])
+    np.testing.assert_allclose(out_nd['surface_azimuth'].reshape(-1),
+                               out_1d['surface_azimuth'])
+
+    # _unit_normal must preserve the input rank, appending a trailing axis
+    # of length 3.
+    unorm = tracking._unit_normal(180., 20., rotations_nd)
+    assert unorm.shape == shape + (3,)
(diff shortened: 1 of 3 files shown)


cwhanse (approved):
Thanks @ArthurOnnoTerabase

tests/test_tracking.py
```diff
@@ -557,3 +557,32 @@ def test_calc_surface_orientation_special():
...
                                    expected_azimuths, rtol=1e-5, atol=1e-5)
+
+
+@pytest.mark.parametrize('shape', [(3, 5), (1, 7), (4, 1), (2, 3, 4)])
+def test_calc_surface_orientation_2d(shape):
+    # Regression test for GH#2747: calc_surface_orientation must accept
+    # tracker_theta of arbitrary rank, not just 1-D.  Compare the >1-D result
+    # to the 1-D result computed on the flattened input.
+    rng = np.random.default_rng(0)
+    rotations_flat = rng.uniform(-90, 90, size=int(np.prod(shape)))
```
kandersolar: This will be the first and only (I think) random number generation in pvlib's test suite.  I lean towards not using random numbers for (1) fear of them silently changing on some new numpy and (2) opaque serendipitous effects.  Of course the values themselves should not matter, but in principle I think ensuring stability is a good idea.

Could we get rid of the RNG and use `rotations_flat = np.linspace(-90, 90, size=int(np.prod(shape)))` instead?

kandersolar (commented):
Great PR @ArthurOnnoTerabase, thanks!  One request below.

tests/test_tracking.py
ArthurOnnoTerabase: Fair point. Done! Update was tested locally before pushing the new commit, worked as expected (the test still reproduces the broadcasting `ValueError `on the unfixed code).

Indeed, I asked Cursor to check and it could not find any other random number generator in the suite. Shall the _Contributing_ documentation include a short mention about refraining from using random generators?

cwhanse: I don't think that note is needed.

kandersolar (approved):
Thanks @ArthurOnnoTerabase!

Issue #1409, opened 2022-02-18
DC model unexpectedly fills NaNs with 0


harakiricode:
I would like to report a behaviour in the dc models of pvlib, which was very unintuitive to me. Since it has been coded very explicitly though, I assume the behaviour has been considered.

I am on version 0.6.1 of pvlib, but the issue occurs on the current version, too. Using, for example the desoto dc model, NaNs are filled with 0 after calling scale_voltage_to_current_power. ([v0.6.1](https://github.com/pvlib/pvlib-python/blob/v0.6.1/pvlib/modelchain.py#L438l) and [current master](https://github.com/pvlib/pvlib-python/blob/master/pvlib/modelchain.py#L778)). In my specific case, this means, that after having passed NaNs to the model chain in my weather dataframe, I receive valid numbers in the model chain result. I would have either expected the afflicted time steps to be NaN, a warning or an error.

Is this behaviour by the dc model - and as a result by the entire model chain - intended?

cwhanse:
You raise a fair point. That has been the behavior since the original `ModelChain.singlediode` method in v0.4.0. I can't say exactly why; my guess is so that `ModelChain.result.dc.sum()` returns a valid value.

wholmgren:
I agree that behavior is very bad and inconsistent with the general pvlib policy of "nan in, nan out". I don't remember, but if I had to guess, I may have added it as a very lazy way of handling missing nighttime values.

A pull request would be welcome.

matsuobasho:
@cwhanse I'm trying to find a more challenging issue to tackle that will be most useful to users (so looked at the 'help wanted' ones).  I'm not using pvlib every day so don't have such a good gauge.  This is an older one, would like some guidance whether this will be useful to resolve now.  If not, can you point to another one that's active and is a pain point right now?  Perhaps this one https://github.com/pvlib/pvlib-python/issues/2245

cwhanse:
@matsuobasho either #1409 or #2245 would be welcome. I think @2245 is easier. Here, we recorded the issue but didn't investigate further. my fear is numerous little edits needed to tests.

#1711 would also be nice to close.

PR #2763, opened 2026-05-26
fix: remove zero-fill in singlediode DC model to preserve NaNs (Close…

Mohamedhassan268:
This PR addresses Issue #1409. I removed the .fillna(0) behavior inside ModelChain._singlediode

### DIFF
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -665,9 +665,6 @@ class ModelChain:
             self.results.dc,
             unwrap=False
         )
-        self.results.dc = tuple(dc.fillna(0) for dc in self.results.dc)
-        # If the system has one Array, unwrap the single return value
-        # to preserve the original behavior of ModelChain
         if self.system.num_arrays == 1:
             self.results.diode_params = self.results.diode_params[0]
             self.results.dc = self.results.dc[0]


pvlib/modelchain.py
```diff
@@ -665,9 +665,6 @@ def _make_diode_params(photocurrent, saturation_current,
             self.results.dc,
             unwrap=False
         )
-        self.results.dc = tuple(dc.fillna(0) for dc in self.results.dc)
-        # If the system has one Array, unwrap the single return value
-        # to preserve the original behavior of ModelChain
```
cwhanse, line 670: Please undo this delete - the comment is about the following `if` statement.

cwhanse (commented):
@Mohamedhassan268 I doubt there is an existing test in test_modelchain.py that includes `nan` in the weather input. I am concerned that removing the `fillna(0)` will uncover other problems.

Can you make a new test for `test_modelchain.py` that checks for proper handling of `nan` in `weather` input? Just so we can see what may go wrong?

In the end I think we'll want to add `nan` to the `weather` fixture. However, that's not a simple task because test_modelchain.py has a lot of tests that use the `weather` fixture, and I suspect that some tests rely on the index length of 2 to match `weather` with other data.

kandersolar:
Closing for lack of response.

Issue #2750, opened 2026-05-06
GSoC 2026: Standardizing Diffuse Irradiance Components and Extending ModelChain Optical Loss Modeling


cbcrespo:
This issue tracks my work for [Google Summer of Code 2026](https://summerofcode.withgoogle.com/) on improving diffuse irradiance handling and optical loss modeling in pvlib. This project will be carried out under the supervision of @AdamRJensen and @ramaroesilva, and has two main goals:

1.	Standardizing the outputs of diffuse transposition models
2.	Extending `ModelChain` to support component-specific optical loss calculations, via component-specific incident angle modifiers (IAM)

If you're interested, you can follow regular updates and discussions on my [blog](https://cbcrespo.github.io/). Official project description on the GSoC website [here](https://summerofcode.withgoogle.com/programs/2026/projects/vM390Eg2).

# Motivation

Currently, diffuse transposition models in `pvlib.irradiance` provide inconsistent outputs. Some models (e.g., Perez) can return individual components (circumsolar, isotropic, horizon), while others only return total diffuse irradiance. This was previously identified and discussed in issue #1553.

This inconsistency limits the `ModelChain` workflow, which:

-	Applies a single angle of incidence (AOI) model
-	Uses one IAM value for total irradiance
-	Cannot distinguish between irradiance components. Since optical losses depend strongly on incidence angle, this reduces physical accuracy.

# Proposed Changes
## Phase 1: Standardize diffuse component outputs
-	Introduce a consistent `return_components` parameter across all diffuse transposition models. PR #2527 has begun addressing this by modifying `irradiance.isotropic` to support this parameter, as previously implemented in `irradiance.perez`.
-	Ensure all models return: 
    - Circumsolar
    - Isotropic
    - Horizon
    - Total diffuse irradiance
-	For models that lack explicit components, return zeros for missing components
-	Follow the structure used in `pvlib.irradiance.perez`

## Phase 2: Extend ModelChain for component-specific IAM
- Consider plane-of-array (POA) components instead of global irradiance
- Add use of component-specific IAM when calculating effective $G_{POA}$
- Allow users to pass a dictionary to `aoi_model`, specifying IAM model for each component
- Maintain full backward compatibility

My name is Carolina Crespo, I am a PhD student at the University of Lisbon working on firm power in Iberia using PyPSA-Eur. I am a regular user of pvlib, which I teach (although at a superficial level) to 1st-year Bachelors students.

[...]

PR #2775, opened 2026-06-09
Add support for return_components to Reindl

cbcrespo:
- [ ] ~Closes #xxxx~
 - [x] Tests added

This PR is part of my [GSoC 2026 project](https://github.com/pvlib/pvlib-python/issues/2750). One of the goals is to add support for `return_components` to all diffuse transposition models in `pvlib`, which currently is supported by some models (e.g. `perez`) but not others. This parameter allows the user to choose whether they want only the total diffuse irradiance, or are interested in the split between different diffuse components (sky diffuse, circumsolar, horizon brightening).

This first PR starts by implementing `return_components` in `reindl`.

Reindl expands the Hay and Davies model, which considers only the isotropic and circumsolar components, and is defined as:
imagem
with A Rb being the circumsolar component and the rest being the isotropic component.

Reindl expands this by adding a horizon brightening component, but this new component is not additive but rather multiplied by the existing isotropic component:
imagem

In my implementation, I essentially considered that the isotropic term was:
imagem
while the horizon brightening term was:
imagem

I updated the function docstrings accordingly. I haven't yet added a test. I saw that e.g. `haydavies` has two separate tests, for `return_components=False` and `return_components=True`. However, from what I gathered, the test values came from the matlab implementation of `pvlib`, and therefore it seems kind of silly to create a test using values derived from the python implementation to test the python implementation. Curious as to what are the community's thoughts on this.


pvlib/irradiance.py
```diff
@@ -973,16 +991,32 @@ def reindl(surface_tilt, surface_azimuth, dhi, dni, ghi, dni_extra,
...
 
-    return sky_diffuse
+    term1 = (1 - AI) * SVF
+    term2 = AI * Rb
+    term3 = (1 - AI) * SVF * h
+
+    sky_diffuse = dhi * (term1 + term2 + term3)
+
+    if return_components:
+        diffuse_components = OrderedDict()
```
kandersolar: I know this would be consistent with existing functions, but at some point we will switch those to normal dictionaries (see #1684).  I suggest getting ahead of that change here and using a normal dictionary.

pvlib/irradiance.py
```diff
@@ -912,10 +912,27 @@ def reindl(surface_tilt, surface_azimuth, dhi, dni, ghi, dni_extra,
...
+    numeric, OrderedDict, or DataFrame
+        Return type controlled by ``return_components`` argument.
+        If ``return_components=False``, `sky_diffuse` is returned.
+        If ``return_components=True``, `diffuse_components` is returned.
+
+    sky_diffuse : numeric
+        The sky diffuse component of the solar radiation on a tilted
+        surface.
+
+    diffuse_components : OrderedDict (array input) or DataFrame (Series input)
```
kandersolar: If switching the return from OrderedDict to dict, change here as well

pvlib/irradiance.py
```diff
@@ -912,10 +912,27 @@ def reindl(surface_tilt, surface_azimuth, dhi, dni, ghi, dni_extra,
...
 
+    return_components : bool, default False
+        Flag used to decide whether to return the calculated diffuse components
+        or not.
+
     Returns
     -------
-    poa_sky_diffuse : numeric
-        The sky diffuse component of the solar radiation. [Wm⁻²]
+    numeric, OrderedDict, or DataFrame
```
kandersolar: And here

kandersolar (commented):
Nice PR @cbcrespo!  I think your interpretation of the horizon component is reasonable, although since it is not stated in the reference, I think we need to document it ourselves.  A sentence in the `Notes` section of the docstring would be appropriate, I think.

> I haven't yet added a test. I saw that e.g. haydavies has two separate tests, for return_components=False and return_components=True. However, from what I gathered, the test values came from the matlab implementation of pvlib, and therefore it seems kind of silly to create a test using values derived from the python implementation to test the python implementation. Curious as to what are the community's thoughts on this.

Yes, ideally the test values would come from an independent implementation.  If you wanted, you could reimplement the model in Excel (or whatever you like) and calculate test values that way.  In this case, I personally would just use the python implementation to compute the test values, for two reasons:
1) the math is not very complicated, so we can be fairly confident that there is no bug in the code
2) the test's intent is only partially to check that the correct numeric values are returned; equally important is that the test verifies that `return_components=True` produces a dict with the correct structure. 

Also, a few comments below regarding whether the returned object should be `OrderedDict` or `dict`.

ramaroesilva:
> I think your interpretation of the horizon component is reasonable, although since it is not stated in the reference, I think we need to document it ourselves. A sentence in the `Notes` section of the docstring would be appropriate, I think.

@kandersolar just for clarification, this interpretation can be seen as factual rather than reasonable since, while it is not explicit in the paper, it is remaining term after subtracting the hay-davies models which predeces it (and to which reindl only adds an horizon term).

@cbcrespo explanation of the interpretation is a bit convoluted because in the beginning we we're trying too much to make a physical interpretation of the term and we wanted to open up a bite the discussion. but then, when talking to Adam, we noticed that indeed this is simply the extra term when we multiply the multiplicative terms.

Anyway, something can be added to the Notes if that is deemed relevant, but it should be really short imo.

cbcrespo:
Thank you for your feedback @kandersolar! I have now added a small sentence to the notes explaining the decomposition, as well as corrected an existent sentence which wrongly stated that "the model described here generates ONLY the diffuse radiation from the sky and circumsolar".

I've also changed OrderedDict to Dict and will continue doing so as I work on the other diffuse transposition models.

In addition, I've now added a `test_reindl_components` based on the existing `test_haydavies_components`, and I believe everything is now in order.

cbcrespo:
The Read the Docs build fail is unrelated to my PR - see #2795.

docs/sphinx/source/whatsnew/v0.15.3.rst
```diff
@@ -18,6 +18,8 @@ Bug fixes
 
 Enhancements
 ~~~~~~~~~~~~
+* Add support for diffuse irradiance components to :py:func:`pvlib.irradiance.reindl`
+  when ``return_components=True``. (:issue:`2750`, :pull:`2775`)
```
AdamRJensen:
```suggestion
* Add support for returning the individual diffuse irradiance components
  from :py:func:`pvlib.irradiance.reindl` when ``return_components=True``.
  (:issue:`2750`, :pull:`2775`)
```

pvlib/irradiance.py
```diff
@@ -916,10 +916,27 @@ def reindl(surface_tilt, surface_azimuth, dhi, dni, ghi, dni_extra,
...
+        If ``return_components=False``, ``sky_diffuse`` is returned.
+        If ``return_components=True``, ``diffuse_components`` is returned.
+
+    sky_diffuse : numeric
+        The sky diffuse component of the solar radiation on a tilted
+        surface.
+
+    diffuse_components : Dict (array input) or DataFrame (Series input)
+        Keys/columns are:
+            * poa_sky_diffuse: Total sky diffuse
```
AdamRJensen: I would leave out this combined value since it is not a component. What does the other functions do?

cbcrespo: The functions with existing `return_components` support (`perez`, `perez_driesse`, `haydavies`) also return the total sky diffuse irradiance.

AdamRJensen: I guess since all the other ones are returning ``poa_sky_diffuse`` it is more consistent to also do it here.

AdamRJensen:
@cwhanse Would you have time to review this PR?

cwhanse:
> @cwhanse Would you have time to review this PR?

EOL on vacation so yes, I do have time to review.

pvlib/irradiance.py
```diff
@@ -980,16 +1001,31 @@ def reindl(surface_tilt, surface_azimuth, dhi, dni, ghi, dni_extra,
...
+
+    sky_diffuse = dhi * (term1 + term2 + term3)
+
+    if return_components:
+        diffuse_components = {
+            'poa_sky_diffuse': sky_diffuse,
+            'poa_isotropic': dhi * term1,
+            'poa_circumsolar': dhi * term2,
+            'poa_horizon': dhi * term3
+        }
```
adriesse, line 1020: This code pattern has redundant multiplications, as @markcampanelli pointed out, so consider reorganizing it slightly.

ramaroesilva: so, I guess having `term3 = term1 * h` and:
- if `sky_diffuse` is not included as a `dict` key: add an `else` to the `if return_components` and move the sum computation there
- if it is included, compute the components before creating the `dict` and make `sky_diffuse` be their sum

AdamRJensen: > so, I guess having `term3 = term1 * h` and:
> 
> * if `sky_diffuse` is not included as a `dict` key: add an `else` to the `if return_components` and move the sum computation there
> * if it is included, compute the components before creating the `dict` and make `sky_diffuse` be their sum

Yes to ``term3 = term1 *h``.

Let's keep ``sky_diffuse`` as a key for now.

docs/sphinx/source/whatsnew/v0.15.3.rst
```diff
@@ -18,6 +18,9 @@ Bug fixes
 
 Enhancements
 ~~~~~~~~~~~~
+* Add support for returning the individual diffuse irradiance components
+  from :py:func:`pvlib.irradiance.reindl` when ``return_components=True``.
```
```suggestion
* Add  ``return_components`` kwarg to :py:func:`pvlib.irradiance.reindl` to
  support returning the components of sky diffuse irradiance.
```

Diffuse irradiance includes ground reflections which isn't considered in the sky diffuse models.

cwhanse (commented):
I'll hold the review here. I've noted an issue that will cross all of these sky diffuse PRs: we don't have consistent definitions of these terms in the current docstrings:

- sky_diffuse
- poa_sky_diffuse
- poa_isotropic (usually missing)
- poa_circumsolar (usually missing)

Each PR is editing the definition in its function so we're not likely to get consistency in this way.

I think we should agree on definitions and apply them throughout.

pvlib/irradiance.py
ramaroesilva: so maybe after all is done, we do a separate PR on removing the combined, if the consensus is to not return it. I would say we have time until the next release to do it.

ramaroesilva:
> I'll hold the review here. I've noted an issue that will cross all of these sky diffuse PRs: we don't have consistent definitions of these terms in the current docstrings:
> 
>     * sky_diffuse
> 
>     * poa_sky_diffuse
> 
>     * poa_isotropic (usually missing)
> 
>     * poa_circumsolar (usually missing)
> 
> 
> Each PR is editing the definition in its function so we're not likely to get consistency in this way.
> 
> I think we should agree on definitions and apply them throughout.

there seemed to be a consensus to only return the components that a model covers (i.e., no key returned, rather than one with an all-zero timeseries).

that being said, it seems that lingering `sky_diffuse` should have a `poa_` prefix, but I was under the impression that @kandersolar had handled that in #2627.

cwhanse (approved):

pvlib/irradiance.py
```diff
@@ -911,10 +911,30 @@ def reindl(surface_tilt, surface_azimuth, dhi, dni, ghi, dni_extra,
...
+    diffuse_components : dict (array input) or DataFrame (Series input)
+        Keys/columns are:
+            * poa_sky_diffuse: The sky diffuse component of irradiance on a
+            tilted plane. [Wm⁻²]
+            * poa_isotropic: The portion of sky diffuse irradiance on a tilted
+            plane from the isotropic sky dome. [Wm⁻²]
+            * poa_circumsolar: The portion of sky diffuse irradiance on a
+            tilted plane from the circumsolar region. [Wm⁻²]
+            * poa_horizon: The portion of sky diffuse irradiance on a tilted
+            plane from the horizon. [Wm⁻²]
```
AdamRJensen: Not rendering correctly

AdamRJensen:
@pvlib/pvlib-core I plan on merging this PR in a few days when the last minor fixes has been addressed.

PR #2776, opened 2026-06-09
Fix wrong reference on irradiance.reindl

### LINKED ISSUE #2765: Likely wrong reference on irradiance.reindl
In [irradiance.reindl](https://pvlib-python.readthedocs.io/en/latest/reference/generated/pvlib.irradiance.reindl.html), two references are cited for this function. I believe that [1] is a wrong reference, since it corresponds to a decomposition paper. However, from a paragraph within that same paper (see printscreen below), I'm almost sure the original idea was to cite Reindl's thesis, which addresses both decomposition and transposition.

**To Reproduce**
1. Go to [[1] in irradiance.reindl](https://pvlib-python.readthedocs.io/en/latest/reference/generated/pvlib.irradiance.reindl.html#id5)

**Expected behavior**
Replace [1] url with https://web.archive.org/web/20240709013622/https://minds.wisconsin.edu/bitstream/handle/1793/47852/0001.pdf

**Screenshots**
From Reindl's decomposition paper:

Image

Image

cbcrespo:
- [x] Closes #2765 

Fixes the wrong reference in the [irradiance.reindl](https://pvlib-python.readthedocs.io/en/latest/reference/generated/pvlib.irradiance.reindl.html) docstrings, previously identified in #2765.

### DIFF
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -948,17 +948,18 @@ def reindl(surface_tilt, surface_azimuth, dhi, dni, ghi, dni_extra,
 
     References
     ----------
-    .. [1] Reindl, D. T., 1988. Estimating diffuse radiation on horizontal
-       surfaces and total radiation on tilted surfaces. M.Sc. thesis,
-       University of Wisconsin-Madison. Available from:
+    .. [1] D. T. Reindl, "Estimating diffuse radiation on horizontal surfaces
+       and total radiation on tilted surfaces," M.S. thesis, University of
+       Wisconsin-Madison, Madison, WI, USA, 1988. Available:
        https://web.archive.org/web/20240709013622/https://minds.wisconsin.edu/bitstream/handle/1793/47852/0001.pdf
-       (Last accessed: 09/06/26)
-    .. [2] Reindl, D. T., Beckmann, W. A., Duffie, J. A., 1990. Evaluation of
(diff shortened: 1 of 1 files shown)


ramaroesilva:
Never reviewed a PR before, but seems good to me :-)

AdamRJensen:
@cbcrespo congrats with opening your first PR!

You'll notice that the "Python Flake8 Linter" is marked as red and with an x. If you click this, you can see the details which the linter is unhappy about (it's about good practice formatting).

While pvlib is not very consistent, we made a choice a while ago to use the IEEE citation format. You can read more about this [here](https://pvlib-python.readthedocs.io/en/stable/contributing/style_guide.html). Could you update the existing citations to match the IEEE format?

cwhanse:
@cbrespo 

For future PRs, please include the checkboxes in the template that's automatically added in the PR description. We use those to manage the review work. You can strikethrough those that don't apply using "~~" on both ends, like this:

 - ~~[ ] doesn't matter~~

cbcrespo:
Thanks for the tips @AdamRJensen and @cwhanse.

I think I've fixed the problem now, and I've also changed the reference style to IEEE.

cwhanse (approved):

kandersolar (approved):
Thanks @cbcrespo!

PR #2780, opened 2026-06-10
Add dataset option to get_era5

AdamRJensen:
 - [x] Tests added

This PR adds a new parameter to the ``get_era5`` function, such that users can change the dataset to be ERA5-land instead of ERA5.

### DIFF
--- a/pvlib/iotools/era5.py
+++ b/pvlib/iotools/era5.py
@@ -69,7 +69,7 @@ def get_era5(latitude, longitude, start, end, variables, api_key,
     This API [2]_ provides a subset of parameters of the full ERA5 datasets,
     see [3]_ for available variables. A comparison of ERA5 and ERA5-land is
     available in [4]_.
-    
+
     Parameters
     ----------
     latitude : float
--- a/pvlib/iotools/era5.py
+++ b/pvlib/iotools/era5.py
@@ -66,9 +66,10 @@ def get_era5(latitude, longitude, start, end, variables, api_key,
 
     A CDS API key is needed to access this API.  Register for one at [1]_.
 
-    This API [2]_ provides a subset of the full ERA5 dataset.  See [3]_ for
-    the available variables.  Data are available on a 0.25° x 0.25° grid.
-
+    This API [2]_ provides a subset of parameters of the full ERA5 datasets,
+    see [3]_ for available variables. A comparison of ERA5 and ERA5-land is
+    available in [4]_.
+    
     Parameters
     ----------
     latitude : float
@@ -116,6 +117,7 @@ def get_era5(latitude, longitude, start, end, variables, api_key,
     .. [1] https://cds.climate.copernicus.eu/
     .. [2] https://cds.climate.copernicus.eu/datasets/reanalysis-era5-single-levels-timeseries?tab=overview
     .. [3] https://confluence.ecmwf.int/pages/viewpage.action?pageId=505390919
+    .. [4] https://confluence.ecmwf.int/display/CKB/The+family+of+ERA5+datasets
     """  # noqa: E501
 
     def _to_utc_dt_notz(dt):
--- a/pvlib/iotools/era5.py
+++ b/pvlib/iotools/era5.py
@@ -58,6 +58,7 @@ UNITS = {
 
 
 def get_era5(latitude, longitude, start, end, variables, api_key,
+             dataset="reanalysis-era5-single-levels-timeseries",
              map_variables=True, timeout=60,
              url='https://cds.climate.copernicus.eu/api/retrieve/v1/'):
     """
@@ -84,6 +85,10 @@ def get_era5(latitude, longitude, start, end, variables, api_key,
         See [1]_ for additional options.
     api_key : str
         ECMWF CDS API key.
+    dataset : str, default "reanalysis-era5-single-levels-timeseries"
+        The dataset to query.  May be either
+        "reanalysis-era5-single-levels-timeseries" or
+        "reanalysis-era5-land-timeseries".
     map_variables : bool, default True
         When true, renames columns of the DataFrame to pvlib variable names
         where applicable. Also converts units of some variables. See variable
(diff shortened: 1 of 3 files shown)


pvlib/iotools/era5.py
```diff
@@ -84,6 +85,10 @@ def get_era5(latitude, longitude, start, end, variables, api_key,
         See [1]_ for additional options.
     api_key : str
         ECMWF CDS API key.
+    dataset : str, default "reanalysis-era5-single-levels-timeseries"
+        The dataset to query.  May be either
+        "reanalysis-era5-single-levels-timeseries" or
+        "reanalysis-era5-land-timeseries".
```
IoannisSifnaios: It might be nice to give some more info on these, in order to be able to choose which one fits best for one's application. Or at least maybe point to this link that explains them? https://confluence.ecmwf.int/display/CKB/The+family+of+ERA5+datasets

pvlib/iotools/era5.py
IoannisSifnaios: Note that the 0.25 resolution is for the "reanalysis-era5-single-levels-timeseries" while the "reanalysis-era5-land-timeseries" has a 0.1 resolution

AdamRJensen: I removed the info and added the link from your comment above.

pvlib/iotools/era5.py
```diff
@@ -84,6 +86,10 @@ def get_era5(latitude, longitude, start, end, variables, api_key,
...
```
echedey-ls:
```suggestion
        ``"reanalysis-era5-single-levels-timeseries"`` or
        ``"reanalysis-era5-land-timeseries"``. See [4]_ for details.
```
`details`, `differences`, `capabilities of each endpoint`, whatever you come up with.

pvlib/iotools/era5.py
```diff
@@ -84,6 +86,10 @@ def get_era5(latitude, longitude, start, end, variables, api_key,
         See [1]_ for additional options.
     api_key : str
         ECMWF CDS API key.
+    dataset : str, default "reanalysis-era5-single-levels-timeseries"
```
echedey-ls:
```suggestion
    dataset : str, default ``"reanalysis-era5-single-levels-timeseries"``
```

pvlib/iotools/era5.py
```diff
@@ -137,7 +144,7 @@ def _to_utc_dt_notz(dt):
             "data_format": "csv"
         }
     }
-    slug = "processes/reanalysis-era5-single-levels-timeseries/execution"
+    slug = f"processes/{dataset}/execution"
```
echedey-ls, line 147: This means that if somebody copies the `dataset` parameter with a typo (e.g., an space) the error is harder to spot. I suggest to add a guardrail right before (`if dataset not in {"...", "..."}`. In addition to that, the new function signature is not backwards-compatible in those calls that provided positional parameters up to `map_variables` (inclusive). The guardrail would provide a meaningful message in this case, where `dataset=True/False`.

AdamRJensen: Hmm, I like the idea, but overall I'm against it. My reason is that if a new compatible dataset is added, then users would not be able to use the function. For example, I was not aware that the ERA5-Land had become an option.

Indeed, it is not backward compatible, this is why I am lobbying for making most input parameters keyword only.

echedey-ls: I admit to not suggesting that to avoid being pedantic 😆

I don't think it's the right time for this release, but I remember PyVista has [at least in v0.46.something] a deprecation decorator that warns against the use of positional arguments (while it also allowed passing some). Just in case it's helpful in the future if iotools API was to be standardized.

pvlib/iotools/era5.py
```diff
@@ -58,15 +58,17 @@ def _m_to_cm(m):
...
     """
     Retrieve ERA5 reanalysis data from the ECMWF's Copernicus Data Store.
 
     A CDS API key is needed to access this API.  Register for one at [1]_.
 
-    This API [2]_ provides a subset of the full ERA5 dataset.  See [3]_ for
-    the available variables.  Data are available on a 0.25° x 0.25° grid.
+    This API [2]_ provides a subset of parameters of the full ERA5 datasets,
+    see [3]_ for available variables. A comparison of ERA5 and ERA5-land is
+    available in [4]_.
```
adriesse, line 71: What do you think about adding a Notes section to briefly explain the problem with coastal grid points?

AdamRJensen: That would be fine, but I don't have the bandwidth right. Feel free to add a suggestion

kandersolar:
I think some more changes will be needed here.  Unfortunately it seems that the land dataset API returns separate CSV files for each request variable, instead of bundling them all into one like the single-level API does.

Since `get_era5` assumes there is only one CSV present, the function as it currently stands in this PR returns only one variable when specifying the land dataset.

AdamRJensen:
> I think some more changes will be needed here. Unfortunately it seems that the land dataset API returns separate CSV files for each request variable, instead of bundling them all into one like the single-level API does.
> 
> Since `get_era5` assumes there is only one CSV present, the function as it currently stands in this PR returns only one variable when specifying the land dataset.

This is sadly true. I did find it documented in the ERA5-land hourly time-series [documentation](https://confluence.ecmwf.int/pages/viewpage.action?pageId=536218894).

> Parameters are stored in groups to ensure a proper access performance. The number of NetCDF files downloaded will be equal to total number of groups of all the parameters selected. Users can check the groups in the Parameters listings section [below](https://confluence.ecmwf.int/pages/viewpage.action?pageId=536218894#ERA5Landhourlytimeseriesdataonsinglelevelsfrom1950topresent:ProductUserGuide(PUG)-Parameterslistings).

I have verified that the current implementation can indeed return two parameters when the parameters are in the same group. This of course is not sufficient. I'll look into it later.

Thanks for spotting this @kandersolar!

Issue #2795, opened 2026-06-24
[Docs] ReadTheDocs builds failing


echedey-ls:
**Describe the bug**
RTD build of PRs #2787 and #2793 are failing. 

I'm attaching a comment I had prepared to submit to RTD issues. But to sum it up, seems like the only timezones failing are those whose type is `link` (of any type), as per wikipedia table: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

I haven't been able to understand whether "link" type is an euphemism for "deprecated", or a reason for OSes to not bundle them. I'll defer opening an issue to RTD in case anybody of you @pvlib/pvlib-maintainer are more knowledgeable than me in this topic - which is almost 100% the case.

Down below is my nearly-completed comment that I was about to upload to RTD when I found some of the timezones did in fact work. Maybe the solution is just to go ahead with canonical timezones, but what I don't quite understand is that Ubuntu changelog does not have any mention to timezone database changes. And I haven't dug Ubuntu's source code to see exactly which database version they ship, with or without link type timezones.

`vvv Comment as written to upload to RTD vvv`

## Details

* Read the Docs project URL: https://pvlib-python.readthedocs.io/en/stable/
* Build URL (if applicable): https://app.readthedocs.org/projects/pvlib-python/builds/33291002/
* Read the Docs username (if applicable): NA

## Expected Result

Documentation to build successfully.

## Actual Result

See log (trimmed):

```bash
../../examples/solar-tracking/plot_discontinuous_tracking.py failed leaving traceback:
    Traceback (most recent call last):
      File "/home/docs/checkouts/readthedocs.org/user_builds/pvlib-python/checkouts/2787/docs/examples/solar-tracking/plot_discontinuous_tracking.py", line 53, in <module>
        times = pd.date_range('2019-06-01', '2019-06-02', freq='1min', tz='US/Eastern')
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/docs/checkouts/readthedocs.org/user_builds/pvlib-python/envs/2787/lib/python3.12/site-packages/pandas/core/indexes/datetimes.py", line 1491, in date_range
        dtarr = DatetimeArray._generate_range(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2803, opened 2026-07-03
Do not merge: debugging ReadTheDocs timezone database issue

kandersolar:
Investigating #2795.  Not intended to be merged.

First step: roll back to the previous OS (really, docker image), and see if the problem goes away.  If it does, that will narrow down the issue to the new RTD release @echedey-ls found.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -9,7 +9,7 @@ sphinx:
     configuration: docs/sphinx/source/conf.py
 
 build:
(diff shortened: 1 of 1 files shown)


echedey-ls:
Closing as fixed by #2809 (latest build of `main` docs https://app.readthedocs.org/projects/pvlib-python/builds/33479482/) , thanks for taking the time to debug a little further @kandersolar :))

Issue #2801, opened 2026-07-02
Keep or exclude total diffuse irradiance in the output when `return_components=True`


cbcrespo:
As part of my GSoC 2026 project (#2750) I am working on diffuse irradiance models in `pvlib` to expand support for the `return_components` parameter, which is currently only implemented in some functions. When `return_components=True`, these functions return a dictionary containing the individual diffuse irradiance components (isotropic, circumsolar, horizon), but also their sum (total diffuse irradiance).

Several people (@AdamRJensen, @adriesse) have raised questions on whether it makes sense to keep the total or if it should be removed. Including the total is convenient and avoids requiring users to recompute it themselves if necessary. Excluding the total avoids redundant outputs and makes the meaning of `return_components` more explicit: return the components only.

The purpose of this issue is therefore to open a more focused discussion on this point and hopefully reach a consensus.

For reference, `irradiance.get_total_irradiance` and `irradiance.poa_components` currently also return several values that could be considered redundant: `poa_global`, `poa_direct`, `poa_diffuse`, `poa_sky_diffuse`, and `poa_ground_diffuse`. But in this case, there is no parameter controlling which values are returned.

markcampanelli:
I'm going to make a bold opening move here with goal to "keep it simple stupid (KISS)" and suggest--
- Remove entirely the `return_components` input argument in all relevant functions in `irradiance.py`.
- Always return a dictionary with `numpy.ndarray` fields (possibly this is a 0-dimensional `numpy.ndarray` for the "scalar" case)--
  -  A `poa_sky_diffuse_total ` field.
  - A `poa_sky_diffuse_components` field, which is--
    -  `None` if this function doesn't provide any components, OR
    -   A sub-dictionary with the named POA diffuse components, when calculated by the function.

NOTE: Further investigate here to ensure "ease of use" by expected downstream consumers such as SAPM- vs. Marion-based effective irradiance computations for diffuse-component IAMs. Also, I mentioned elsewhere to consider we should try to indicate when, say, the circumsolar or horizon components are considered "concentrated", because that may affect how downstream consumers compute with the provided values. Including this info here seems to me like a stretch, but considering the downstream consumers NOW might shed some more light on if this would be worth the additional complexity.

[...]

Issue #2539, opened 2025-08-21
Consider updating pvlib.tracking.singleaxis() to return zeros (or similar) for nighttime instead of NaNs


williamhobbs:
**Is your feature request related to a problem? Please describe.**
`pvlib.tracking.singleaxis` returns NaNs for intervals where the sun is below the horizon. This can result in NaNs propogating to lots of other pvlib outputs and can cause confusion, other issues, etc. See https://groups.google.com/g/pvlib-python/c/ZeVmTC7eddY as an example.

It appears to happen here: 

https://github.com/pvlib/pvlib-python/blob/6dfeaf8ee97db6a7a8f3c8109eed8b5d285c4e41/pvlib/tracking.py#L155

**Describe the solution you'd like**
Perhaps change the default to be zero for _surface_tilt_ and _tracker_theta_ (and whatever is most appropriate for _aoi_ and _surface_azimuth_). Or have an optional input for `pvlib.tracking.singleaxis` to specify whether it returns NaNs or numbers at night. 

**Describe alternatives you've considered**
Applying `.fillna(0)` (or similar) to the output of `pvlib.tracking.singleaxis` is what I typically do. 

**Additional context**
It looks like this zero vs nan situation was briefly discussed in https://github.com/pvlib/pvlib-python/issues/569. It seems it would be worth further discussion. **Is there a significant reason that NaN is the output at night?**

`fillna(0)` is used in the gallery [here](https://pvlib-python.readthedocs.io/en/stable/gallery/solar-tracking/plot_single_axis_tracking.html) and [here](https://pvlib-python.readthedocs.io/en/stable/gallery/solar-tracking/plot_single_axis_tracking_on_sloped_terrain.html), but not [here](https://pvlib-python.readthedocs.io/en/stable/gallery/solar-tracking/plot_discontinuous_tracking.html). 

Tagging @cwhanse.

adriesse:
It depends on the variable. In this case NaNs make sense because the algorithm cannot calculate the optimal angle under these conditions, and I would think those NaNs are best processed downstream.

williamhobbs:
[...]

PR #2808, opened 2026-07-05
DRAFT: Handle dni/dhi=0/0 in irradiance.perez without NaNs

markcampanelli:
- [x] Inspired by #2801
 - [x] Tests added

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.15.3.rst
+++ b/docs/sphinx/source/whatsnew/v0.15.3.rst
@@ -14,7 +14,7 @@ Deprecations
 
 Bug fixes
 ~~~~~~~~~
-* Fix :py:func:`pvlib.irradiance.perez` to not return NaN when both DHI and
+* Fix :py:func:`pvlib.irradiance.perez` to return 0 when both DHI and
   DNI are zero. (:issue:`2801`, :pull:`2808`)
 * Fix :py:func:`pvlib.irradiance.perez` to not return zero (instead NaN) when
   airmass is NaN and either DHI or DNI are NaN. (:issue:`2801`, :pull:`2808`)
--- a/docs/sphinx/source/whatsnew/v0.15.3.rst
+++ b/docs/sphinx/source/whatsnew/v0.15.3.rst
@@ -14,7 +14,10 @@ Deprecations
 
 Bug fixes
 ~~~~~~~~~
-
+* Fix :py:func:`pvlib.irradiance.perez` to not return NaN when both DHI and
+  DNI are zero. (:issue:`2801`, :pull:`2808`)
+* Fix :py:func:`pvlib.irradiance.perez` to not return zero (instead NaN) when
+  airmass is NaN and either DHI or DNI are NaN. (:issue:`2801`, :pull:`2808`)
 
 Enhancements
 ~~~~~~~~~~~~
@@ -48,3 +51,4 @@ Contributors
 * Eesh Saxena (:ghuser:`eeshsaxena`)
 * Karl Hill (:ghuser:`karlhillx`)
 * Yonry Zhu (:ghuser:`yonryzhu`)
+* Mark Campanelli (:ghuser:`markcampanelli`)
--- a/docs/sphinx/source/whatsnew/v0.15.3.rst
+++ b/docs/sphinx/source/whatsnew/v0.15.3.rst
@@ -16,7 +16,7 @@ Bug fixes
 ~~~~~~~~~
 * Fix :py:func:`pvlib.irradiance.perez` to return 0 when both DHI and
   DNI are zero. (:issue:`2801`, :pull:`2808`)
-* Fix :py:func:`pvlib.irradiance.perez` to not return zero (instead NaN) when
+* Fix :py:func:`pvlib.irradiance.perez` to return NaN when
   airmass is NaN and either DHI or DNI are NaN. (:issue:`2801`, :pull:`2808`)
 
 Enhancements
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -1049,7 +1049,8 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
     DHI is zero, then DNI is also zero, otherwise a FloatingPointError is
     raised due to a division of a nonzero (and not NaN) value by zero. It is
     also expected that extraterrestrial irradiance is positive. If airmass
-    is NaN, then the total and all components are zero.
+    is NaN, then the total and all components are zero if they should not 
+    otherwise be NaN.
 
     Warning
     -------
--- a/pvlib/irradiance.py
+++ b/pvlib/irradiance.py
@@ -1049,8 +1049,8 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
     DHI is zero, then DNI is also zero, otherwise a FloatingPointError is
     raised due to a division of a nonzero (and not NaN) value by zero. It is
     also expected that extraterrestrial irradiance is positive. If airmass
-    is NaN, then the total and all components are zero if they should not 
-    otherwise be NaN.
+    is NaN, then the total and all components are zero if they should not
+    otherwise be NaN because DHI or DNI was NaN.
 
     Warning
     -------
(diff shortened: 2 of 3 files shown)


pvlib/irradiance.py
markcampanelli, line 1186: @pvlib/maintainers I question if this is wise. Perhaps if a user passes a NaN for DNI or DHI (presumably resulting in a NaN for `eps`), then the function should fail?

echedey-ls: I think nan in -> nan out is the philosophy we try to adhere to. Dunno what makes more sense for this specific case.

markcampanelli: K thanks. Sorry I had missed that.

echedey-ls: It ain't something to be sorry about, it a legitimate question (the motto, better safe than sorry).

Discussions related to this point:
- https://github.com/pvlib/pvlib-python/issues/2539
- https://github.com/pvlib/pvlib-python/issues/338

And there was one regarding ModelChain, don't remember exactly where.

markcampanelli: I added many more combinations of NaN DNI and DHI to the new tests.

markcampanelli: @echedey-ls So in my testing, I realized that if airmass is NaN, then this trumps DNI or DHI being NaN, and the result is zero instead of NaN (including the components, when returned). That sounds like incorrect behavior. Would you agree?

markcampanelli, line 1242: @cbcrespo I think this is where one has to be careful about the order of operations when computing the components first and then summing to get the total.

A very edgy case here would be if the components added up to zero, but are themselves not all zero (indeed some components can legitimately be negative in this model). This would likely be exceedingly rare for typical floating point calculations, however.

That said, I think the algorithm semantics would be easier to follow if:
- The components were computed first.
- The sum of components was computed next.
- Finally, the total and components are all set to zero if the sum is (strictly) negative.

Although, if a negative total comes out, should this function raise instead?

(I tend to like things to break obviously rather than fail silently, and a negative total sounds like a model breakage to me, but I could be mis-interpreting the situation.)

markcampanelli: I changed the function's behavior to preserve NaNs inherited from DHI or DNI. New test now passing again.

markcampanelli: On second thought, the "spirit of pvlib-python" might be to set everything to NaN if the total comes out negative, instead of raising.

markcampanelli:
@pvlib/maintainers I have two outstanding issues before moving this out of DRAFT and requesting a full review:
1) I raise a new `FloatingPointError` when dividing a non-zero (and non-NaN) DHI by a zero DNI.
2) I think that if the total sky diffuse POA comes out negative after summing all the components, then we should raise an error OR set this value and all corresponding components to NaN (current behavior sets everything to zero, which I think is significantly misleading).

I think returning NaNs instead of raising is probably closer to what users expect, but I am not sure here. If this is true, then I can change both of the above cases to return NaNs.

docs/sphinx/source/whatsnew/v0.15.3.rst
```diff
@@ -14,7 +14,10 @@ Deprecations
 
 Bug fixes
 ~~~~~~~~~
-
+* Fix :py:func:`pvlib.irradiance.perez` to not return NaN when both DHI and
+  DNI are zero. (:issue:`2801`, :pull:`2808`)
```
```suggestion
* Fix :py:func:`pvlib.irradiance.perez` to return 0 when both DHI and
  DNI are zero. (:issue:`2801`, :pull:`2808`)
```

docs/sphinx/source/whatsnew/v0.15.3.rst
```diff
@@ -14,7 +14,10 @@ Deprecations
...
+* Fix :py:func:`pvlib.irradiance.perez` to not return zero (instead NaN) when
```
```suggestion
* Fix :py:func:`pvlib.irradiance.perez` to return NaN when
```

pvlib/irradiance.py
```diff
@@ -1042,10 +1042,15 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
     Perez models determine the diffuse irradiance from the sky (ground
     reflected irradiance is not included in this algorithm) on a tilted
     surface using the surface tilt angle, surface azimuth angle, diffuse
-    horizontal irradiance, direct normal irradiance, extraterrestrial
-    irradiance, sun zenith angle, sun azimuth angle, and relative (not
-    pressure-corrected) airmass. Optionally a selector may be used to
-    use any of Perez's model coefficient sets.
+    horizontal irradiance (DHI), direct normal irradiance (DNI),
+    extraterrestrial irradiance, sun zenith angle, sun azimuth angle, and
+    relative (not pressure-corrected) airmass. Optionally a selector may be
```
cwhanse, line 1047: Let's remove the list of input quantities. That was ported from Matlab before the current docstring format was settled.

cwhanse (commented):
@markcampanelli can you populate this table? I'm having trouble following the intent of handling the combinations of 0 and/or NaN


| air mass| DHI | DNI | sky diffuse output |
| :--- | :--- | :--- | :--- |
| 0 | 0 | 0 |   |
| 0 | 0 | NaN |  |
| 0 | NaN | 0 |  |
| 0 | NaN | NaN |  |
| NaN | 0 | 0 |  |
| NaN | 0 | NaN |  |
| NaN | NaN | 0 |  |
| NaN | NaN | NaN |  |

markcampanelli:
> @markcampanelli can you populate this table? I'm having trouble following the intent of handling the combinations of 0 and/or NaN

|case | air mass | DHI | DNI | total | circumsolar | isotropic | horizon |
|:--| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| a | 42 | 0 | 0 | 0 | 0 | 0 | 0 |
| b | 42 | 0 | NaN | NaN | NaN | NaN | NaN |
| c | 42 | NaN | 0 | NaN | NaN | NaN | NaN |
| d | 42 | NaN | NaN | NaN | NaN | NaN | NaN |
| e | NaN | 0 | 0 | 0 | 0 | 0 | 0 |
| f | NaN | 0 | NaN | NaN | NaN | NaN | NaN |
| g | NaN | NaN | 0 | NaN | NaN | NaN | NaN |
| h | NaN | NaN | NaN | NaN | NaN | NaN | NaN |
| i | 42 | 100 | 0 | -22.2 -> NaN | -65.5 -> NaN | 37.0 -> NaN | 6.34 -> NaN |
| j | NaN | 100 | 0 | NaN or 0? | NaN or 0?  | NaN or 0?  | NaN or 0?  |
| k | 42 | 0 | 100 | NaN | NaN | NaN | NaN |
| l | NaN | 0 | 100 | NaN or 0? | NaN or 0? | NaN or 0? | NaN or 0? |

I added components columns and four rows to the end of your table template.

The first added row shows a negative sky-diffuse `total` result that is negative, along with examples of components that produce it. The row after this is same input but with NaN airmass.

The last two added rows actually raise a `FloatingPointError` in my current code, but I think, for consistency, this should change to the NaNs shown (or zeros when airmass is NaN?).

After making this table, it is unclear to me if a NaN airmass should supersede a "nonsensical" DHI+DNI combo, for example, DHI=0 and DNI=100. If "NaN airmass wins", then the output is all zeros, if "nonsensical DHI+DNI" wins, then output is all NaNs. Which to choose?

markcampanelli:
@cwhanse If I remove raising the `FloatingPointError` when, say, `dni/dhi = 100/0`, then the `eps` (epsilon) parameter evaluates to `inf` (following standard IEEE 754), and subsequently the total sky diffuse and all components compute to zero due to how `inf` is treated in the subsequent lookup table for `eps` (considered clear sky, I think). TBH: I'm not sure exactly what to do here, but semantically I still think that NaNs make more sense here than zeros or raising an exception.

More broadly, should this `perez` function be first checking the inputs to ensure that `0 <= dni < inf` and `0 <= dhi < inf`, but (possibly) allowing upstream NaNs to "pass through"?

cwhanse:
@markcampanelli as a general rule (but not always), pvlib avoids checking inputs. GIGO applies.

Case e is an exception to the usual nan-in/nan-out philosophy, but it's been that way forever so I'm not inclined to suggest changing it. Case j is a variation of case e, so to be consistent, 0 output.

Case l is a problem that illustrates why the nan-in/nan-out rule has value. If pvlib outputs nan, then we'd do that by removing the overwriting airmass nan with 0 (Cases e and j). Since I'm not inclined to change that behavior, pvlib outputs 0 for case l. Otherwise we write some more complicated logic - ugh.

Case i: I'd prefer to output the negative numbers rather than substituting nan where the values are non-physical (this was likely the motivation for overwriting airmass nan). Outputting nan should be a clue that you've got nan inputs. Outputting negative irradiance is a different signal - you're using the model outside it's intended range.

PR #2822, opened 2026-07-12
Stop ModelChain from filling missing DC/AC values with zero (gh-1409)

### LINKED ISSUE #1409: DC model unexpectedly fills NaNs with 0

adi-IL:
## What does this PR do?

Fixes a long standing behavior where `ModelChain` silently replaced missing (NaN) values in its DC and AC results with zero.

## Background

The single diode path ended with

`self.results.dc = tuple(dc.fillna(0) for dc in self.results.dc)`

and the pvwatts inverter did

`self.results.ac = ac.fillna(0)`

A user who passed genuine gaps (NaN) into the weather therefore received zero power for those time steps instead of NaN, masking the missing data. This is inconsistent with pvlib's documented `NaN in, NaN out` policy. (See gh-1409.)

The masking was originally added so that `ModelChain.results.dc.sum()` returned a value at night. We checked that this is no longer needed: at night the single diode model already returns zero power on its own (zero photocurrent gives zero `p_mp`), so removing the fill leaves nighttime correct while letting true gaps stay NaN.

## The fix

Remove both `fillna(0)` calls. Genuine missing inputs now propagate as NaN through the DC and AC results, and nighttime remains zero power.

## Testing

- Added `test_run_model_preserves_nan_inputs`: feeds a three step weather series (day, NaN gap, night) through the cec/single diode DC model and asserts the gap stays NaN, night is zero, and daytime is positive.
- Added `test_pvwatts_inverter_preserves_nan_inputs`: same idea through the pvwatts inverter.
- Full `tests/test_modelchain.py` suite passes (159 tests).

Closes #1409

### DIFF
--- a/docs/sphinx/source/whatsnew/v0.15.3.rst
+++ b/docs/sphinx/source/whatsnew/v0.15.3.rst
@@ -14,6 +14,12 @@ Deprecations
 
 Bug fixes
 ~~~~~~~~~
+~~~~~~~~
+
+* Fixed ``ModelChain`` no longer replacing missing (NaN) values in the DC
+  and AC results with zero. Genuine gaps in the input now propagate as NaN,
+  following the "NaN in, NaN out" policy, while nighttime correctly remains
+  zero power. (:issue:`1409`)
 
 
 Enhancements
--- a/pvlib/modelchain.py
+++ b/pvlib/modelchain.py
@@ -665,7 +665,10 @@ class ModelChain:
             self.results.dc,
             unwrap=False
         )
-        self.results.dc = tuple(dc.fillna(0) for dc in self.results.dc)
+        # Missing (NaN) values in the inputs are intentionally preserved so
+        # that the result follows the "NaN in, NaN out" policy. At night the
+        # single diode model already returns zero power, so night is not
+        # affected by not filling NaNs here.
         # If the system has one Array, unwrap the single return value
         # to preserve the original behavior of ModelChain
         if self.system.num_arrays == 1:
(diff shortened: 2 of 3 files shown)


cwhanse:
I'm closing this for three reasons, none of which have to do with the improvement that the submitter proposes:

1. This solution has not been discussed and agreed to in the relevant Issue. The submitter's comment appears entirely written by AI.
2. The submitter hasn't included the PR template.
3. I am fatigued with reviewing AI-enabled submissions that usually require more time from a reviewer than the submitter puts into the PR.

Sorry to be blunt. If you'd like to contribute to pvlib (we welcome that), please discuss the proposed solution to #1409 in the relevant issue, along with evidence that your solution addresses the concern.

PR #2824, opened 2026-07-15
Add `marion_diffuse_interpolate`, to be used instead of `marion_diffuse` for vector `surface_tilt` inputs

### LINKED ISSUE #1402: pvlib.iam.marion_integrate uses too much memory for vector inputs
`pvlib.iam.marion_integrate` (which is mostly relevant as a helper for `pvlib.iam.marion_diffuse`) needs quite a bit of memory when passed vector inputs.  An input of length 1000 allocates around 2GB of memory on my machine, so naively passing in a standard 8760 would use roughly 17-18 GB.  Unfortunately I was very much focused on fixed tilt simulations when I wrote pvlib's implementation and never tried it out on large vector inputs, so this problem went unnoticed until @spaneja pointed it out to me. 

I think any vectorized implementation of this algorithm is going to be rather memory-heavy, so I'm skeptical that achieving even a factor of 10 reduction in memory usage is possible here without completely changing the approach (and likely shifting the burden from memory to CPU).  However, here are two low-hanging fruits worth considering:

1. The current implementation has a handful of large 2-D arrays local to the function that only get released when the function returns.  Some of them are only used near the beginning of the function but still take up memory for the entire function duration.  Using the `del` statement to instruct python that those arrays are no longer needed allows python to reclaim that memory immediately and recycle it for subsequent allocations.  This is probably a simplification of what actually happens, but it seems consistent with the below observations.
2. `np.float32` cuts memory usage in half compared with `np.float64` and (probably) doesn't meaningfully change the result.  It's not like `surface_tilt` has more than a few sig figs anyway.

Here is a rough memory and timing comparison (using [memory_profiler](https://github.com/pythonprofilers/memory_profiler), very handy).  `pvlib` is the current implementation; the two `del` variants use a strategic sprinkling of `del` but are otherwise not much different from `pvlib`.  This is for an input of length 1000.  The traces here are memory usage sampled at short intervals across a single function invocation; for example the blue `pvlib` trace shows that the function call took 1.4 seconds to complete and had a peak memory usage slightly higher than 2GB.

image

So using a few `del`s cuts peak memory usage roughly in half. Dropping down to `np.float32` cuts it roughly in half again (and gives a nontrivial speedup too).  It's possible that further improvements can be had with other tricks (e.g. using the `out` parameter that some numpy functions provide) but I've not yet explored them.

**My main question**: are we open to using these two strategies in pvlib?  Despite being built into python itself, `del` still seems unpythonic to me for some reason.  Switching away from `float64` is objectionable to the extent that it's the standard in scientific computing and is therefore baked into the models by assumption.  I think I'm cautiously open to both of the above approaches, iff they are accompanied by good explanatory comments and switching to `float32` can be reasonably shown to not introduce a meaningful difference in output. 

---

Remark: even ignoring this memory bloat, I tend to think that applying `marion_integrate` directly to an 8760 is a bit strange.  In simulations with time series `surface_tilt`s, a better approach IMHO is to calculate the IAM values only for `np.linspace(0, 90, 1)` or similar and use `pvlib.iam.interp` to generate the 8760 IAM series.  If nothing else, we might suggest that in the docs.

cbcrespo:
- [x] Closes #1402
 - [x] Tests added

As noted in #1402, `marion_diffuse` uses too much memory (and takes a long time) when the `surface_tilt` input is a vector, which occurs for trackers. `marion_diffuse` calls `marion_integrate` for each `surface_tilt` and each region (sky, horizon or ground). `marion_integrate` integrates over a solid angle for each of those. @cwhanse, @markcampanelli, @adriesse, @kdebrab and others discussed how this issue could be circumvented via interpolation.

This PR proposes a solution via a new function called `marion_diffuse_tracking`, to be used instead of `marion_diffuse` for trackers.

`marion_diffuse_tracking` creates vectors of 0-90 or 0-180º (for bifacial arrays) with a 0.5º step, `marion_integrate` is called for each angle in that vector and each region only once, and the results are used to create an interpolator function (using `scipy.interpolate.PchipInterpolator`). This interpolator is then used to obtain an IAM value for each value in `surface_tilt`.

Regardless of the length of `surface_tilt`, `marion_integrate` is called a fixed number of times, helping to contain the computational burden.

In addition, the interpolation is cached, meaning that e.g. for a `PVSystem` object with multiple tracker `Array`s which have identical tilt vectors, the interpolator will only have to be created once for each region.

More details on time and memory savings, as well as deviation, [here](https://github.com/pvlib/pvlib-python/issues/1402#issuecomment-4936971909). For one example for a 8760-length vector, `marion_diffuse_tracking` uses 4% of the max memory and 2% of the time of `marion_diffuse`.

### DIFF
--- a/pvlib/iam.py
+++ b/pvlib/iam.py
@@ -645,7 +645,7 @@ def marion_diffuse(model, surface_tilt, **kwargs):
     return iam
 
 
-def marion_diffuse_tracking(model, surface_tilt, **kwargs):
+def marion_diffuse_tracking(model, surface_tilt, resolution=0.5, **kwargs):
     """
     Determine diffuse irradiance incidence angle modifiers using Marion's
     method of integrating over solid angle. This function is designed for
@@ -665,6 +665,10 @@ def marion_diffuse_tracking(model, surface_tilt, **kwargs):
         Surface tilt angles in decimal degrees.
         The tilt angle is defined as degrees from horizontal
         (e.g. surface facing up = 0, surface facing horizon = 90).
+
+    resolution : float, default 0.5
+        The resolution in degrees of the tilt angle vector used for
+        the integration.
 
     **kwargs
         Extra parameters passed to the IAM function.
@@ -713,7 +717,7 @@ def marion_diffuse_tracking(model, surface_tilt, **kwargs):
     iam = {}
     for region in ['sky', 'horizon', 'ground']:
         interpolator = _get_marion_interpolator(iam_function,
-                                                region, full_range)
+                                                region, resolution, full_range)
         iam_values = interpolator(surface_tilt)
         if isinstance(surface_tilt, pd.Series):
             iam_values = pd.Series(iam_values, index=surface_tilt.index)
@@ -725,16 +729,17 @@ def marion_diffuse_tracking(model, surface_tilt, **kwargs):
 
 
 @functools.cache
-def _get_marion_interpolator(iam_function, region, full_range=False):
+def _get_marion_interpolator(iam_function, region, resolution,
+                             full_range=False):
     """
     Cached interpolator for the Marion integration of an IAM function over
     solid angle. Helper function for :py:func:`marion_efficient` function
     to avoid repeated calculations leading to excessive memory use.
     """
     if full_range:
-        tilt = np.arange(0, 180.5, 0.5)
+        tilt = np.arange(0, 180.5, resolution)
     else:
-        tilt = np.arange(0, 90.5, 0.5)
+        tilt = np.arange(0, 90.5, resolution)
     iam = marion_integrate(iam_function, tilt, region)
     return PchipInterpolator(tilt, iam)
 
--- a/pvlib/iam.py
+++ b/pvlib/iam.py
@@ -12,6 +12,7 @@ import numpy as np
 import pandas as pd
 import functools
 from scipy.optimize import minimize
+from scipy.interpolate import PchipInterpolator
 from pvlib.tools import cosd, sind, acosd
 
 # a dict of required parameter names for each IAM model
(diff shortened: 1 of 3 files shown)


ramaroesilva:
kudos to @kandersolar for raising the issue and everyone else who contributed to the conceptualization of this.

was now thinking that neither me and @cbcrespo have never modelled bifacial, and we have no idea how IAM is calculated for that (not even sure if `marion_diffuse` is applicable). would be important that someone who's experienced in this to take a look.

> marion_diffuse_tracking creates vectors of 0-90 or 0-180º with a 0.5º step

likely it's not as simple as this....

pvlib/iam.py
```diff
@@ -644,6 +645,100 @@ def marion_diffuse(model, surface_tilt, **kwargs):
...
+            iam_values = pd.Series(iam_values, index=surface_tilt.index)
+        elif isinstance(surface_tilt, list):
+            iam_values = iam_values.tolist()
+        iam[region] = iam_values
+
+    return iam
+
+
+@functools.cache
+def _get_marion_interpolator(iam_function, region, full_range=False):
```
AdamRJensen: I would make the resolution an input parameter instead of hard coding it.

RDaxini:
> was now thinking that neither me and @cbcrespo have never modelled bifacial, and we have no idea how IAM is calculated for that (not even sure if `marion_diffuse` is applicable). would be important that someone who's experienced in this to take a look.
> 
> > marion_diffuse_tracking creates vectors of 0-90 or 0-180º with a 0.5º step
> 
> likely it's not as simple as this....

I wonder if @shirubana is available to comment on this as bifacial PV is one of her many areas of expertise

kandersolar:
In the style of Marion's method, bifacial IAM can be pretty straightforward (assuming you already have the front-side IAMs in hand).  By symmetry:

- iam_rear_ground = iam_front_sky
- iam_rear_sky = iam_front_ground
- iam_rear_horizon = iam_front_horizon

Flaws of this approach include: slight inconsistency in the horizon IAM, as the horizon band is placed slightly above the horizon (If it was centered on the horizon, it would be perfect, alas), the assumption of no neighboring rows or other objects blocking the view (of course this is also true for the front side in Marion's method, so not an "additional" deficiency here), and the assumption of isotropic ground-reflected irradiance (ignoring the row's own shadow).

However I'm not sure we need to be thinking about bifacial modeling here.  This is just a way to compute many Marion IAMs quickly.  How best to model bifacial IAM is a separate problem which doesn't need to be addressed here IMHO.

ramaroesilva:
Thanks for the input @kandersolar. I think Carolina was worried how this would behave if a user sets up a bifacial system using ModelChain or simply wants to use this function directly as an efficient interpolation gimmick for bifacial. To be honest, I'm not even aware how bifacial is treated within pvlib, and I should check it as it makes it more difficult to imagine/reflect on the matter (from your feedback, it seems each side of the module is handled separately).

Besides this bifaciality, anything else to add? Do you see any issue with interpolating the IAM curve with 0.5 degree resolution (as an arbitrary, high-resolution/theoretically safe delta)?

pvlib/iam.py
```diff
@@ -644,6 +645,105 @@ def marion_diffuse(model, surface_tilt, **kwargs):
     return iam
 
 
+def marion_diffuse_tracking(model, surface_tilt, resolution=0.5, **kwargs):
+    """
+    Determine diffuse irradiance incidence angle modifiers using Marion's
+    method of integrating over solid angle. This function is designed for
+    trackers, where ``surface_tilt`` is a vector, to avoid the computational
+    burden of calling ``marion_integrate`` for each tilt angle.
```
```suggestion
    Determine incidence angle modifiers (IAMs) for diffuse irradiance
    using Marion's method of integrating over solid angles.
    
    This function supports trackers for which ``surface_tilt`` is a vector.
```

pvlib/iam.py
```diff
@@ -644,6 +645,105 @@ def marion_diffuse(model, surface_tilt, **kwargs):
...
+    Instead, the IAM function is integrated once for a range of tilt angles,
+    and then interpolated to the requested tilt angles. For fixed-tilt systems,
```
```suggestion
    The IAM function is integrated once for tilt angles specified by
    ``resolution``. IAM at other angles are determined by interpolation.
    For fixed-tilt systems,
```

pvlib/iam.py
```diff
@@ -644,6 +645,105 @@ def marion_diffuse(model, surface_tilt, **kwargs):
...
+    -------
+    iam : dict
+        IAM values for each type of diffuse irradiance:
+
+        * 'sky': radiation from the sky dome (zenith <= 90)
+        * 'horizon': radiation from the region of the sky near the horizon
+          (89.5 <= zenith <= 90)
+        * 'ground': radiation reflected from the ground (zenith >= 90)
+
+        See [1]_ for a detailed description of each class.
```
```suggestion
        See [1]_ for a detailed description of each type.
```

cwhanse (approved):
Editorial suggestions for the docstring.

pvlib/iam.py
```diff
@@ -644,6 +645,106 @@ def marion_diffuse(model, surface_tilt, **kwargs):
...
+        `'ashrae', 'physical', 'martin_ruiz', 'sapm', 'schlick'`.
+
+    surface_tilt : numeric
+        Surface tilt angles in decimal degrees.
+        The tilt angle is defined as degrees from horizontal
+        (e.g. surface facing up = 0, surface facing horizon = 90).
+
+    resolution : float, default 0.5
+        The resolution in degrees of the tilt angle vector used for
+        the integration.
```
kandersolar:
```suggestion
        the interpolation.
```

I think?

kandersolar:
```suggestion
        The resolution in degrees of the tilt angle vector used for
        the interpolation.
```

Maybe clearer?  The integration itself also has a resolution, and could potentially be confused with this one.

pvlib/iam.py
```diff
@@ -644,6 +645,106 @@ def marion_diffuse(model, surface_tilt, **kwargs):
...
+
+    return iam
+
+
+@functools.cache
+def _get_marion_interpolator(iam_function, region, resolution,
+                             full_range=False):
+    """
+    Cached interpolator for the Marion integration of an IAM function over
+    solid angle. Helper function for :py:func:`marion_efficient` function
```
kandersolar: Update this to whatever the final function name is

pvlib/iam.py
```diff
@@ -644,6 +645,106 @@ def marion_diffuse(model, surface_tilt, **kwargs):
     return iam
 
 
+def marion_diffuse_tracking(model, surface_tilt, resolution=0.5, **kwargs):
+    """
+    Determine incidence angle modifiers (IAMs) for diffuse irradiance
+    using Marion's method of integrating over solid angles.
```
kandersolar:
```suggestion
    Determine incidence angle modifiers (IAMs) for diffuse irradiance
    using interpolation for Marion's method of integrating over solid angles.
```

Just so that it shows a slightly different description here: https://pvlib-python.readthedocs.io/en/stable/reference/pv_modeling/iam.html

pvlib/iam.py
```diff
@@ -644,6 +645,106 @@ def marion_diffuse(model, surface_tilt, **kwargs):
...
+        IAM values for each type of diffuse irradiance:
+
+        * 'sky': radiation from the sky dome (zenith <= 90)
+        * 'horizon': radiation from the region of the sky near the horizon
+          (89.5 <= zenith <= 90)
+        * 'ground': radiation reflected from the ground (zenith >= 90)
+
+        See [1]_ for a detailed description of each type.
+
+    See Also
```
kandersolar, line 715: I suggest listing this new function in the `See also` for `marion_diffuse` also, maybe with a brief note explaining the memory use thing (if sphinx supports that, I'm not sure).

kandersolar (commented):
A proposal: change the function name to `marion_diffuse_interpolate`.  Upside: a better description of what the function does.  Downside: less suggestive of why you might want to use it.  Any reactions?

ramaroesilva:
> A proposal: change the function name to `marion_diffuse_interpolate`. Upside: a better description of what the function does. Downside: less suggestive of why you might want to use it. Any reactions?

Seems a peaceful option to me. We can always explain the why in the function documentation, I'd say it's rather likely that people end up taking a look.

pvlib/iam.py
echedey-ls: In addition to the `See also`, I encourage to use some kind of admonition too, [after the description body](https://github.com/pvlib/pvlib-python/blob/3da60241c5dab603f9075b16e58d2593ba3d79c8/pvlib/iam.py#L594). These two catch my attention:

1. https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-tip
2. https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso

Edit: very good point btw

pvlib/iam.py
cbcrespo: Edited to "The resolution in degrees of the tilt angle vector used to create the interpolator.", I thought this was the clearest option

PR #2826, opened 2026-07-17
Fix `perez` ZeroDivisionError on scalar dhi=0

chuenchen309:
`pvlib.irradiance.perez` crashes with `ZeroDivisionError` on scalar input when `dhi == 0`, even though `dhi == 0` is documented-valid (`dhi : numeric`, "must be >= 0") and occurs at every nighttime timestamp.

```python
from pvlib import irradiance as ir
ir.perez(30, 180, 0.0, 800.0, 1400.0, 40.0, 120.0, 1.5)   # ZeroDivisionError
```

The identical input as an array, as a `np.float64` scalar, and via the sibling `perez_driesse` all return a clean value:

```python
import numpy as np
ir.perez(30, 180, np.array([0.0]), np.array([800.0]), 1400.0, 40.0, 120.0, np.array([1.5]))  # [0.]
ir.perez_driesse(30, 180, 0.0, 800.0, 1400.0, 40.0, 120.0, 1.5)  # 0.0
```

**Cause:** the clearness term divides by `dhi`. The `np.errstate(invalid='ignore')` wrapper lets the `dhi == 0` array path flow through `np.digitize` to a finite result, but a *native Python scalar* `dhi` uses Python's `/`, which raises `ZeroDivisionError` — `np.errstate` doesn't affect native scalar division. The array path was defended against `dhi == 0`; the native-scalar path was not.

**Fix:** compute the division with `np.true_divide` (and add `divide='ignore'`) so a native scalar takes the same inf/nan path the array input already does. The scalar result now matches the array result exactly (`0.0`), `dhi == dni == 0` gives `nan` like the array, and non-zero `dhi` is unchanged.

A regression test is added (`test_perez_scalar_dhi_zero`): red before the fix (`ZeroDivisionError`), green after, asserting the scalar result matches the array result. The 8 existing `perez` tests still pass; `flake8` clean.

- [x] Closes issue — *N/A (no issue; small correctness fix)*
- [x] Tests added
- [ ] Updates entries in `docs/sphinx/source/reference` — *N/A, no API change*
- [x] Adds a whatsnew entry
- [x] New code documented (fix is on an existing documented function; comment explains the `np.true_divide` rationale)

*Disclosure: prepared with AI assistance (Claude Code, Claude Opus 4.8). I reproduced the crash, wrote the failing regression test first, verified the scalar result matches the documented array behavior, and ran the perez suite + flake8 before opening.*


pvlib/irradiance.py
```diff
@@ -1146,8 +1146,12 @@ def perez(surface_tilt, surface_azimuth, dhi, dni, dni_extra,
...
 
     # epsilon is the sky's "clearness"
-    with np.errstate(invalid='ignore'):
-        eps = ((dhi + dni) / dhi + kappa * (z ** 3)) / (1 + kappa * (z ** 3))
+    # np.true_divide so a Python-scalar dhi=0 yields inf/nan like the array
+    # path (handled below via digitize) instead of raising ZeroDivisionError,
+    # which np.errstate cannot suppress for native scalar division.
+    with np.errstate(invalid='ignore', divide='ignore'):
+        eps = (np.true_divide(dhi + dni, dhi) + kappa * (z ** 3)) \
+            / (1 + kappa * (z ** 3))
```
kandersolar:
```suggestion
    # np.divide so a Python-scalar dhi=0 yields inf/nan like the array
    # path (handled below via digitize) instead of raising ZeroDivisionError,
    # which np.errstate cannot suppress for native scalar division.
    with np.errstate(invalid='ignore', divide='ignore'):
        eps = (np.divide(dhi + dni, dhi) + kappa * (z ** 3)) \
            / (1 + kappa * (z ** 3))
```

`true_divide` is just an alias for `divide`, may as well use the simpler name

kandersolar (approved):

Issue #2846, opened 2026-08-19
Add support for diffuse irradiance components and component-specific IAM to `ModelChain`


cbcrespo:
I am currently at the final stage of my GSoC (#2750). After expanding support for diffuse components in transposition functions and adding support for diffuse component-specific IAM in the `Array` and `PVSystem` classes, the remaining step is to integrate this functionality into `ModelChain`.

Here are some points I think are worth discussing:

# Default behavior

The existing behavior applied `fd` to all diffuse irradiance. `fd` was essentially always 1, meaning no optical losses were accounted for for diffuse irradiance.

What should the new default behavior be?

`marion_diffuse` is a good candidate for a default diffuse IAM model, since it is the only model that supports all three diffuse components: sky diffuse, horizon, and ground. (`martin_ruiz_diffuse` and `schlick_diffuse` support sky ground only).

As for `marion_diffuse`'s `model` parameter (i.e. the direct IAM model used by the Marion integration method), I'm considering doing something like the existing `infer_aoi_model` which selects a model based on the available parameters when the user does not explicitly choose one.

# FD

Do we want to keep FD around as an option?

# Handling missing IAM for existing components

Let's say the user chooses the `perez` transposition model, and therefore diffuse irradiance includes the horizon component, and then chooses the `martin_ruiz_diffuse` IAM model, which does not provide an IAM for the horizon component. How should this be handled?

Options:
- Throw an error prompting the user to select a compatible combination of transposition and IAM models
- Apply the sky diffuse IAM to the horizon component
- Apply no optical losses to the horizon component
- Something else?

cwhanse:
My thoughts: keep `fd` as a module parameter, and add a new kwarg to `ModelChain` say `aoi_diffuse_model` default None. If `aoi_diffuse_model` is present, then execute a new step `aoi_diffuse_model` in `ModelChain.run_model` methods, which populates `ModelChainResult.aoi_diffuse_modifier`. Then `modelchain.effective_irradiance_model` uses `aoi_diffuse_modifer` inside this little function: https://github.com/pvlib/pvlib-python/blob/f02515571ab9d7cba2cab9af0c7e1c8155b10323/pvlib/modelchain.py#L1064

I'd default  `aoi_diffuse_model = None` to keep behavior consistent with prior releases.

[...]

PR #2845, opened 2026-08-14
Add support for diffuse IAM in the `Array` and `PVSystem` classes

cbcrespo:
- [x] Closes #2812
 - [x] I am familiar with the [contributing guidelines](https://pvlib-python.readthedocs.io/en/latest/contributing/index.html)
 - [x] I attest that all AI-generated material has been vetted for accuracy and is in compliance with the pvlib license
 - [x] Tests added
 - [ ] ~Updates entries in [`docs/sphinx/source/reference`](https://github.com/pvlib/pvlib-python/blob/main/docs/sphinx/source/reference) for API changes.~
 - [x] Adds description and name entries in the appropriate "what's new" file in [`docs/sphinx/source/whatsnew`](https://github.com/pvlib/pvlib-python/tree/main/docs/sphinx/source/whatsnew) for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` or this Pull Request with `` :pull:`num` ``. Includes contributor name and/or GitHub username (link with `` :ghuser:`user` ``).
 - [x] New code is fully documented. Includes [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, examples, and comments where necessary.
 - [ ] Pull request is nearly complete and ready for detailed review.
 - [x] Maintainer: Appropriate GitHub Labels (including `remote-data`) and Milestone are assigned to the Pull Request and linked Issue.

This PR adds support for diffuse component-specific IAM in the `Array` and `PVSystem` classes, laying the groundwork for its later inclusion in the `ModelChain` class.

PR #2847, opened 2026-08-21
Add support for diffuse irradiance components and component-specific IAM to `ModelChain`

### LINKED ISSUE #2811: Support diffuse irradiance components and component-wise IAM in ModelChain
This issue is part of my [GSoC 2026 project](https://github.com/pvlib/pvlib-python/issues/2750). The first goal was to add support for `return_components` to all diffuse transposition models in `pvlib`, which was previously supported by some models (e.g. perez) but not others. This parameter allows the user to choose whether they want only the total diffuse irradiance, or are interested in the split between different diffuse components (sky diffuse, circumsolar, horizon brightening).

This has now been completed for all diffuse transposition models in `pvlib` except for `klucher`, which does not inherently support different components. This means that we can now implement component-wise IAM calculations in `ModelChain` - which is the second part of my GSoC project. In practice, this means treating circumsolar as direct, and calculating a separate IAM for direct and each of the available diffuse components out of sky isotropic, horizon, and ground.

The end goal is to enable `ModelChain` to calculate effective irradiance using component-wise incidence angle modifiers (IAM), rather than applying a single IAM to direct irradiance and assuming no diffuse optical losses (current behavior).

The first step is to make sure that the `ModelChain` workflow also supports diffuse components, which is currently not the case. After that, component-wise IAM can be introduced. Both of these steps require changes to the three classes involved in the `ModelChain` framework: `Array`, `PVSystem`, and `ModelChain` itself. The `ModelChain` workflow relies on the `PVSystem` class, which in turn relies on the `Array` class.

The goal of this overarching issue is to discuss the overall plan. Other issues will be created for more in-depth discussions of more specific points.

Currently, the plan is as follows:
- `Array` class:
  - `Array.get_irradiance()`: add a boolean `diffuse_components` argument (which gets passed to `irradiance.get_total_irradiance()`)  
  - `Array.get_iam()`: currently only returns IAM for direct irradiance. Support for diffuse component-specific IAM must be added. Issue #2812 discusses how to implement this.
- `PVSystem` class:
  - `PVSystem.get_irradiance()`: add a boolean `diffuse_components` argument (which gets passed to `Array.get_irradiance()`)
  - `PVSystem.get_iam()`: currently only returns IAM for direct irradiance. Support for diffuse component-specific IAM must be added. How this should be done depends on the decision about the architecture of `Array.get_iam`.
- `ModelChain` class:
  - `diffuse_components` parameter should be added on `init`
    - passing `diffuse_components=True` and `transposition_model='klucher'` should throw an error
  - `aoi_model` is a parameter defined on `init`.
    - currently all direct IAM models are supported except for `schlick`. Is there a reason for this?
  - `ModelChain.prepare_inputs()`:
    - calls `PVSystem.get_irradiance()`. Would need a small change to support `diffuse_components`. 
  - `ModelChain.aoi_model` is a property. It has a setter that attributes different functions to it via user choice (defined on init via the `aoi_model` parameter) or via `infer_aoi_model()` if necessary. One possible approach is to rename the current `aoi_model` to `aoi_model_direct` and create a new `aoi_model_diffuse`, but the implementation details of this should probably be discussed in a separate issue.
    - Direct: all existing direct IAM models are supported except for schlick. Is there a reason for this?
    - Diffuse: currently not implemented in `ModelChain`. `martin_ruiz_diffuse`, `marion_diffuse`, and `schlick_diffuse` should all be allowed. One caveat is that `marion_diffuse` is an integration method which uses one of the direct IAM models from ‘ashrae’, ‘physical’, ‘martin_ruiz’, ‘sapm’, and ‘schlick’. This probably requires adding an additional optional `aoi_model_marion` parameter, which is only used if `aoi_model_diffuse` is `marion`.
      - Note: as noted by @markcampanelli on #2750, `marion_diffuse` has a high memory use on trackers, due to integration over a solid angle for a timeseries of `surface_tilt` as opposed to a single value. Over a long timeseries, this can generate redundant calculations (if the same angle is repeated twice). One possible solution is to first do an interpolation of the `marion_diffuse` integral results for each diffuse components and then use that to get the IAM for the timeseries of tilt angles, as suggested by @markcampanelli. I think this deserves a more in-depth discussion in a separate issue.
  - `ModelChain.effective_irradiance_model()`
    - calculates effective irradiance as:
      - `spect_mod * (total_irrad['poa_direct'] * aoi_mod + fd * total_irrad['poa_diffuse'])`
      - fd is 1 by default, meaning no diffuse optical losses.
    - should be changed to something more like:
      - `spect_mod * ((total_irrad['poa_direct'] + total_irrad['poa_circumsolar']) * aoi_mod + iam_isotropic * total_irrad['poa_isotropic'] + iam_horizon * total_irrad['poa_horizon'] + iam_ground * total_irrad['poa_ground'])`
      - but while being mindful of which components are indeed present.
				
Finally, as you can see, this plan requires extensive changes, but all those changes are interdependent. I am aware it is generally not good to create huge PRs with many changes, but in this case, I was wondering how you all would advise me to split the work.

cbcrespo:
- [x] Closes #2811, #2846, #2750
 - [x] Tests added

This PR adds support for diffuse irradiance components and component-specific IAM to the `ModelChain` class, thereby closing the final stage of my GSoC project (#2750) 😄

Note: This PR is currently based on #2845, which has not yet been merged. I expect reviewing will become easier once that PR is merged. Also, I’m very open to revisiting any of the design choices below based on feedback.

This adds a new `iam_diffuse_model` property to `ModelChain`, which can be any of the three existing diffuse IAM functions `marion_diffuse`, `martin_ruiz_diffuse`, or `schlick_diffuse`, as well as `'no_loss'`, which does not apply optical losses to diffuse irradiance. By default, `iam_diffuse_model` is `None`, in which case the previous behavior is preserved: `ModelChain` fetches the `FD` (fraction diffuse) module parameter for each array and applies it to the overall diffuse irradiance. `FD` is essentially always 1, meaning this behavior will generally be equivalent to `no_loss`.

For `marion_diffuse` specifically, an additional parameter `marion_diffuse_model` is necessary (corresponding to the `model` parameter of `iam.marion_diffuse`). It defaults to `None`, but can be inferred from the module parameters via the new `infer_marion_diffuse_model`, which follows a similar logic to the existing `infer_aoi_model`.

A new `iam_diffuse_modifier` is added to the `ModelChainResult` dataclass.

Circumsolar irradiance is treated as direct (i.e., the direct IAM is applied to it). The remaining diffuse components have their respective component-specific IAMs applied.

If a diffuse component is present for which the selected IAM model does not provide an IAM value, no optical losses are applied to that component and a warning is raised. This would happen, for example, with `perez` and `martin_ruiz_diffuse`: the `perez` transposition model provides isotropic, circumsolar, and horizon components, while the `martin_ruiz_diffuse` diffuse IAM model does not provide a horizon IAM.

If the selected transposition model does not provide individual components for sky diffuse irradiance (as is the case for `klucher`), optical losses are applied to the ground diffuse component only. The aggregate sky diffuse irradiance is left unmodified (IAM = 1), and a warning is raised to make this behavior explicit.


ramaroesilva:
> If the selected transposition model does not provide individual components for sky diffuse irradiance (as is the case for `klucher`), optical losses are applied to the ground diffuse component only. The aggregate sky diffuse irradiance is left unmodified (IAM = 1), and a warning is raised to make this behavior explicit.

I understand that this is a clean way to handle things, however I thing it will just make people not use transposition models which don't provide components like `klucher` (it's bit weird to account for ground but not to sky diffuse).

Would it make sense that users in this case are allowed to use `martin_ruiz_diffuse` or `schlick_diffuse` and apply isotropic losses to the total sky diffuse? Personally, I would prefer this than not having any IAM at all for the sky diffuse, even if it meant raising a warning saying that IAM-wise the modelchain will interpret the sky diffuse as isotropic.

PR #2860, opened 2026-09-21
Drop python 3.10; increase required dependency versions

kandersolar:
- ~[ ] Closes #xxxx~
 - ~[ ] I attest that all AI-generated material has been vetted for accuracy and is in compliance with the pvlib license~ no AI
 - ~[ ] Tests added~

Python 3.10 reaches end of life in October (https://devguide.python.org/versions/).  May as well remove it now.  And, dropping support for 3.10 means we can increment our dependencies to exclude old versions that can't be installed on python 3.11+.


kandersolar:
Ready for review.  The test failure is unrelated and looks like the kind of thing that might go away on its own.

cwhanse:
This decorator should go https://github.com/pvlib/pvlib-python/blob/b3d09bf61ce705939d16c50176acf1d582a8ee11/tests/conftest.py#L227 here, or in a different maintenance PR?

cwhanse (approved):

kandersolar:
I think we do not want to make any changes to that decorator.  With this PR the minimum scipy advances to 1.9.2.  I think we should only remove that decorator when it eventually reaches 1.15 and pvlib can assume that chandrupatla is always available.

cwhanse:
Right - 1.9 < 1.15 in semantic versioning

RDaxini:
For my own understanding/knowledge... in the case of no associated bugs/test failures, how are these changes decided? These changes being which specific version of other dependencies are required, and when to implement the update/min version requirements in general.

kandersolar:
Taking scipy as an example, our previous minimum of 1.7.2 is not available for python 3.11 or higher (check [PyPI](https://pypi.org/project/scipy/1.7.2/#files)), meaning after this PR it cannot be installed easily on any python version we support.  It would be difficult for us to test it in CI, and there is likely little benefit to users anyway (since they cannot install it either).  So, it makes sense to increase our minimum version to something that _can_ be installed on a python version we support.  For scipy, the first version to support python 3.11 was 1.9.2, so that's where I set our new minimum.

I note that other large projects take a more aggressive approach to updating their minimum versions.  For example, see [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) and [SPEC 0](https://scientific-python.org/specs/spec-0000/), which suggest supporting only dependency versions released in the last two years.  We don't see why dropping usable versions makes sense (see https://github.com/pvlib/pvlib-python/issues/828#issuecomment-561721447), so we don't follow those guidelines for pvlib.

RDaxini:
@kandersolar thanks!

RDaxini:
Tangentially related:
This PR (especially https://github.com/pvlib/pvlib-python/pull/2860#issuecomment-5771813846) helped me review and notice some inconsistencies with the dependency requirements in the PVDeg repo. We will fix this in v0.8.0. Thanks again!

